PR #51, opened 2016-03-14
trying to fix travis ci

amueller:
let's see if this works.


mfeurer:
Oh, secret API keys are not used when having a pull request from outside the group...

amueller:
yeah so I think we really need a public read only api key

amueller:
merge this so master is fixed?

PR #60, opened 2016-03-15
slight cleanup of the notebooks

amueller:
python3 compat
don't call the connector openml
don't read a non-standard key file)


mfeurer:
Is there any nice way to compare two ipynb's together?

amueller:
no :(

amueller:
here it is particularly bad because previously it was an old version of jupyter notebooks.

mfeurer:
okay, I'll just merge it

PR #61, opened 2016-03-15
add makefile to main repo, add config for coverage

amueller:


mfeurer:
The Makefile says that the readme says why the tests would fail on windows, should that comment be there?

amueller:
fixed

PR #64, opened 2016-03-15
style fixes

amueller:
pep8 fixes, consistent indentation, remove unused inputs, use numpydoc documentation format


openml/entities/task.py
```diff
@@ -1,18 +1,10 @@
-import sys
```
mfeurer: Why does this commit remove this python 2/3 compatibility measure?

PR #73, opened 2016-03-15
add publish method to OpenMLRun

amueller:


amueller:
ok this should work now. separates out all the "run" specific stuff of api connector

PR #93, opened 2016-03-17
ADD download runs

mfeurer:


openml/__init__.py
```diff
@@ -26,4 +26,4 @@
 __version__ = "0.2.1"
 
 __all__ = ['APIConnector', 'OpenMLDataset', 'OpenMLRun', 'OpenMLSplit',
-           'datasets', 'OpenMLTask', 'OpenMLFlow']
+           'datasets', 'OpenMLTask', 'OpenMLFlow', 'runs']
```
amueller: you need to ``from . import runs`` - actually we need to do the same for flows and tasks... I forgot that it seems.

openml/runs/run.py
```diff
@@ -344,3 +351,117 @@ def get_cached_run(api_connector, run_id):
...
+def get_runs_list_by_flow(api_connector, flow_id):
+    return _get_runs_list(api_connector, flow_id, 'flow')
+
+
+def get_runs_list_by_uploader(api_connector, flow_id):
+    return _get_runs_list(api_connector, flow_id, 'uploader')
+
+
+def _get_runs_list(api_connector, id_, by):
+    if isinstance(by, list):
```
amueller: why don't you put this in the ``get_runs_list_by_filters`` function?

openml/runs/run.py
```diff
@@ -344,3 +351,117 @@ def get_cached_run(api_connector, run_id):
...
+
+def get_runs_list_by_task(api_connector, task_id):
+    return _get_runs_list(api_connector, task_id, 'task')
+
+
+def get_runs_list_by_flow(api_connector, flow_id):
+    return _get_runs_list(api_connector, flow_id, 'flow')
+
+
+def get_runs_list_by_uploader(api_connector, flow_id):
```
amueller: should be ``uploader_id`` not ``flow_id``

openml/runs/run.py
```diff
@@ -344,3 +351,117 @@ def get_cached_run(api_connector, run_id):
...
+    return _get_runs_list(api_connector, tag, 'tag')
+
+
+def get_runs_list(api_connector, run_ids):
+    return _get_runs_list(api_connector, run_ids, 'run')
+
+
+def get_runs_list_by_filters(api_connector, id_=None, task=None, flow=None,
+                             uploader=None):
+    id__ = []
```
amueller:
```python
selectors =  zip(["id", "task", "flow", "uploader"], [id_, task, flow, uploader]):
selectors = filter(lambda x: x[1] is not None, selectors)
```
should do pretty much the same as this function

amueller:
cool. there is some weird indentation in some places, and I think the numpy standard is to do
```
Parameters
----------------
```
with a dash and not an =.

PR #119, opened 2016-03-25
MAINT improve coverage: openml.datasets.functions

mfeurer:


mfeurer:
This is a first step to increase the test coverage of the module. Once it's higher I'll start working on the issues with high priority. @amueller do you have time to look at this PR and merge it?

amueller:
please ping me again end of the week. Thinks are crazy here ;)

mfeurer:
Sorry for the delay, things got crazy here, too. Do you have time to review this ATM? I will now continue with dataset.py, task.py and task_functions.py.

openml/datasets/functions.py
```diff
@@ -171,44 +182,44 @@ def check_datasets_active(dids):
...
     active = {}
 
-    dataset_list_idx = 0
     for did in dids:
-        # TODO replace with a more efficient while loop!
-        for idx in range(dataset_list_idx, len(dataset_list)):
-            if did == dataset_list[idx]['did']:
-                active['did'] = bool(dataset_list[idx]['status'])
-        dataset_list_idx = idx
+        for dataset in dataset_list:
```
amueller: err this loop seems pretty bad (I know it was there before).
why not iterate over ``dataset_list`` and build a dictionary from all datasets to bool and then subset it with ``dids``?
(it's a bit weird then to actually provide ``dids`` as we always compute it for all).

amueller: Maybe that tells us that ``list_datasets`` should actually return a dict indexed by ``did``?

openml/datasets/functions.py
```diff
@@ -236,59 +247,75 @@ def get_dataset(did):
...
-    arff_file = _get_dataset_arff(did, description=description)
+    did_cache_dir = _create_dataset_cache_directory(did)
+
+    try:
+        description = _get_dataset_description(did_cache_dir, did)
+        arff_file = _get_dataset_arff(did_cache_dir, description)
+        # TODO not used yet, figure out what to do with them...
+        features = _get_dataset_features(did_cache_dir, did)
+        qualities = _get_dataset_qualities(did_cache_dir, did)
+    except Exception as e:
```
amueller: Be more specific in the exceptions you expect if possible. Not sure though.

openml/flows/flow.py
```diff
@@ -37,9 +37,10 @@ def __init__(self, model, id=None, uploader=None,
         self.tag = tag
         self.model = model
         self.source = "FIXME DEFINE PYTHON FLOW"
+        # TODO add scikit-learn here!
         self.name = (model.__module__ + "." +
                      model.__class__.__name__)
-        self.external_version = 'Tsklearn_' + sklearn.__version__
+        self.external_version = 'sklearn_' + sklearn.__version__
```
amueller: wasn't the T some openml convention? Not sure though.

openml/tasks/functions.py
```diff
@@ -23,21 +23,19 @@ def _get_cached_tasks():
...
 
     return tasks
 
 
 def _get_cached_task(tid):
     for cache_dir in [config.get_cache_directory(), config.get_private_directory()]:
         task_cache_dir = os.path.join(cache_dir, "tasks")
-        task_file = os.path.join(task_cache_dir,
-                                 "tid_%d.xml" % int(tid))
+        task_file = os.path.join(task_cache_dir, str(tid), "task.xml")
```
amueller: so you changed the file names? Or was it broken before? This breaks all the users caches (if anyone uses this) ;)

openml/tasks/task.py
```diff
@@ -29,11 +29,6 @@ def __init__(self, task_id, task_type, data_set_id, target_feature,
         if cost_matrix is not None:
             raise NotImplementedError("Costmatrix")
 
-    def __str__(self):
```
amueller: Why did you remove this? (it should have been ``__repr__`` I think though)

openml/testing.py
```diff
@@ -14,6 +15,19 @@ class TestBase(unittest.TestCase):
     """
 
     def setUp(self):
+        # This cache directory is checked in to git to simulate a populated
+        # cache
+        self.static_cache_dir = None
+        static_cache_dir = os.path.dirname(os.path.abspath(inspect.getfile(self.__class__)))
```
amueller: I think you can just use ``__file__`` to refer to the path of this file.

openml/testing.py
```diff
@@ -14,6 +15,19 @@ class TestBase(unittest.TestCase):
...
+        for i in range(2):
```
amueller: I'm not sure I understand this loop.

tests/datasets/test_datasets.py
```diff
@@ -9,49 +10,65 @@
...
+    def test_get_cached_dataset_description_not_cached(self):
+        openml.config.set_cache_directory(self.static_cache_dir)
+        self.assertRaisesRegexp(OpenMLCacheException, "Dataset description for "
+                                                      "did 3 not cached",
+                                openml.datasets.functions._get_cached_dataset_description,
+                                3)
+
+    def test_get_cached_dataset_arff(self):
+        openml.config.set_cache_directory(self.static_cache_dir)
+        description = openml.datasets.functions._get_cached_dataset_arff(2)
```
amueller: maybe ``did=2`` - I like explicit parameter names.

tests/tasks/test_task.py
```diff
@@ -0,0 +1,91 @@
...
+        patch.return_value = mock.MagicMock()
+        mm = mock.MagicMock()
+        patch.return_value._retrieve_class_labels = mm
+        patch.return_value._retrieve_class_labels.return_value = 'LA'
+        retval = openml.tasks.get_task(1)
+        self.assertEqual(patch.call_count, 1)
+        self.assertIsInstance(retval, openml.OpenMLTask)
+        self.assertEqual(retval.class_labels, 'LA')
+
+    def test_get_X_and_Y(self):
```
amueller: hmm if we want to follow sklearn api, ``Y`` should be lowercase (because it is usually a vector). Not sure if that was my fault.

amueller:
sorry for the late reply. I'm still not done with my book. I didn't check all the tests in detail but it looks good overall.

mfeurer:
No problem. I will try in incorporate these changes soonish and rebase the current work on that.

openml/datasets/functions.py
mfeurer: I'm not sure what kind of exceptions can happen. But for all exceptions, the cache dir might be in an undefined state, thus I want to catch them all.

openml/datasets/functions.py
mfeurer: I'll fix the first comment when I push the next time, and I opened #136 for the second comment.

openml/tasks/functions.py
mfeurer: The old file name assumed a different directory structure, which is outdated. I think we should only worry about this once we have a serious amount of users?

openml/tasks/task.py
mfeurer: I removed this to had all objects consistent. There is an issue about adding __repr__ to all objects to fix this.

openml/testing.py
mfeurer: No, that makes it crash.

openml/testing.py
mfeurer: Me neither, it'll be gone in the next push.

mfeurer:
Okay, I'll incorporated your suggestions. Can you have a look again?

joaquinvanschoren:
Just Reading through this long thread. Is there something server-side that
I should look at?

Cheers,
Joaquin

mfeurer:
Thanks Joaquin, ubt this only adds new test to the client in order to improve test coverage, so no need for changes on the server side.

mfeurer:
@amueller I pulled in the latest changes from the development branch. Could you please check the code?

openml/flows/flow.py
mfeurer: I checked with Joaquin. No, it wasn't a convention.

openml/datasets/functions.py
```diff
@@ -42,7 +48,8 @@ def _list_cached_datasets():
 
             if "dataset.arff" in dataset_directory_content and \
                     "description.xml" in dataset_directory_content:
-                datasets.append(did)
+                if did not in datasets:
```
amueller: how could it already be there? If that was the case, please add a non-regression test checking that after this change, no duplicates appear.

openml/datasets/functions.py
```diff
@@ -176,44 +187,44 @@ def check_datasets_active(dids):
...
-        # TODO replace with a more efficient while loop!
-        for idx in range(dataset_list_idx, len(dataset_list)):
-            if did == dataset_list[idx]['did']:
-                active['did'] = bool(dataset_list[idx]['status'])
-        dataset_list_idx = idx
+        if did not in active:
+            raise ValueError('Could not find dataset %d in OpenML dataset list.'
+                             % did)
+
+    active = {did: active[did] for did in dids}
```
amueller: That's ok, though maybe more elegant:
```python
try:
    active = {did: dataset['status'] == 'active'}
except KeyError:
    raise ValueError(ValueError('Could not find dataset %d in OpenML dataset list.' % did)
return active
```

openml/datasets/functions.py
amueller: alrighty

openml/flows/flow.py
amueller: lol. Why the TODO? that's what the external_version is for, right?

openml/runs/run.py
```diff
@@ -151,6 +155,8 @@ def run_task(task, model):
     train_times = []
 
     rep_no = 0
+    # TODO use different iterator to only provide a single iterator (less
```
amueller: I don't know I understand what that means here.

openml/tasks/functions.py
amueller: fair

openml/tasks/functions.py
```diff
@@ -65,9 +63,18 @@ def get_estimation_procedure_list():
         "estimationprocedure/list")
     procs_dict = xmltodict.parse(xml_string)
     # Minimalistic check if the XML is useful
-    assert procs_dict['oml:estimationprocedures']['@xmlns:oml'] == \
-        'http://openml.org/openml'
-    assert type(procs_dict['oml:estimationprocedures']['oml:estimationprocedure']) == list
+    if 'oml:estimationprocedures' not in procs_dict:
```
amueller: I guess there was some debugging here? This seems pretty heavy handed but it it was useful at some point, maybe it's necessary. I wonder if this can be generalized.

openml/testing.py
amueller: hm... that's pretty odd... but alright.

tests/datasets/test_datasets.py
```diff
@@ -9,49 +10,65 @@
...
-            </oml:authenticate>"""
-        with mock.patch("openml._api_calls._perform_api_call",
-                        return_value=api_mock_return_value) as api_mock:
-            dataset = _get_cached_dataset(2)
-            self.assertIsInstance(dataset, OpenMLDataset)
-            self.assertTrue(api_mock.is_called_once())
+    def test__get_cached_dataset(self, ):
+        openml.config.set_cache_directory(self.static_cache_dir)
+        dataset = _get_cached_dataset(2)
+        self.assertIsInstance(dataset, OpenMLDataset)
```
amueller: hm smoketest... not super strict but whatever...

tests/tasks/test_task.py
```diff
@@ -0,0 +1,91 @@
...
+        # Classification task
+        task = openml.tasks.get_task(1)
+        X, Y = task.get_X_and_y()
```
amueller: this was renamed... shouldn't that fail now?

amueller: never mind I'm blind.

amueller:
I'm merging this now as it seems a good improvement. My comments still apply ;)

openml/datasets/functions.py
mfeurer: It could have been there because of line 31, which iterates over two directories. If it was in the 'official' directory, it should not be added again if it is also in the 'private' directory. Because this is not documented and not tested, I opened issue #151 and I will remove this.

openml/flows/flow.py
mfeurer: Removed in 62b68c680a5321f130b4cead43a2aed87522e845

openml/runs/run.py
mfeurer: I opened up issue #154

openml/tasks/functions.py
mfeurer: Hm, a comment would have been useful here. I don't recall why it's there and just leave it until it produces problems.

tests/datasets/test_datasets.py
mfeurer: what would be the correct way?

PR #138, opened 2016-06-28
add filenames to run upload.

amueller:
this should solve the xml / arff file name issue.


amueller:
@mfeurer good to merge? Not sure if master would pass without it or not, as @janvanrijn changed stuff again ;)

amueller:
Ok master is still failing so this fixes master.

mfeurer:
Since master is rolled back, we can defer that question to later.

PR #139, opened 2016-07-04
Push documentation to different folders

Ayaro:


Ayaro:
I was not able to test on my separate circleci instance due to multiple package install errors: https://circleci.com/gh/Ayaro/openml-python

mfeurer:
Could you please try `pip install -q` instead of `pip install`? There seems to be no other way than disabling the complete output

PR #148, opened 2016-08-30
FIX upload all xml files as files (not as post argument)

mfeurer:


mfeurer:
Coverage decreased because I had to deactivate tests which currently cannot be run. Further commits will fix the decrease in coverage.

openml/runs/run.py
```diff
@@ -79,10 +79,11 @@ def publish(self):
         """
         predictions = arff.dumps(self._generate_arff())
         description_xml = self._create_description_xml()
-        data = {'predictions': ("predictions.csv", predictions),
-                'description': ("description.xml", description_xml)}
+        print(description_xml)
```
amueller: don't think you want that here

tests/flows/test_flow.py
```diff
@@ -17,6 +17,7 @@ def check_flow(flow):
         for flow in flows:
             check_flow(flow)
 
+    @unittest.skip('Not tested until test sentinels are added back.')
```
amueller: what sentinels?

mfeurer: Flow names should be unique, [this commit](https://github.com/openml/openml-python/pull/150/commits/7e352c1f4079e86e76ed9f1217cf22277bf879bd) adds sentinels to make sure during each test creates a flow with a unique name.

PR #150, opened 2016-08-31
Maint/minor improvements

mfeurer:


openml/flows/flow.py
```diff
@@ -51,10 +52,16 @@ def _generate_flow_xml(self):
             Flow represented as XML string.
         """
         model = self.model
+
+        if config._testmode:
+            flow_name = '%s%s' % (config.testsentinel, self.name)
```
amueller: I don't like this. There should be no testing code in the main code. Why do you want this?

openml/flows/flow.py
```diff
@@ -137,6 +144,12 @@ def _check_flow_exists(name, version):
     if not (type(version) is str and len(version) > 0):
         raise ValueError('Argument \'version\' should be a non-empty string')
 
+    if config._testmode:
```
amueller: same thing. This should be handled in testing.

openml/datasets/dataset.py
```diff
@@ -75,7 +76,7 @@ def __init__(self, id=None, name=None, version=None, description=None,
                 logger.debug("Data pickle file already exists.")
             else:
                 try:
-                    data = self._get_arff()
```
amueller: why is ``format`` an attribute? Can we find that out from the file? Shouldn't the file reader handle this?

openml/tasks/functions.py
```diff
@@ -235,7 +236,7 @@ def get_task(task_id):
 
     # TODO look into either adding the class labels to task xml, or other
     # way of reading it.
-    class_labels = dataset._retrieve_class_labels()
+    class_labels = dataset._retrieve_class_labels(task.target_feature)
```
amueller: why is this called ``target_feature``? That's not a feature, is it? ``target_value``? ``target_name``? ``target_column``?

amueller:
The IO stuff is for python2.x, right? Can you put this in it's own PR and if it passes tests we merge?

amueller:
It seems a bit like each of the commits should be it's own PR ;)

amueller:
can you explain the sentinel business?

tests/datasets/test_datasets.py
```diff
@@ -206,6 +206,15 @@ def test_publish_dataset(self):
...
         self.assertEqual(return_code, 200)
 
+    def test__retrieve_class_labels(self):
+        openml.config.set_cache_directory(self.static_cache_dir)
+        labels = openml.datasets.get_dataset(2)._retrieve_class_labels()
+        self.assertEqual(labels, ['1', '2', '3', '4', '5', 'U'])
+        labels = openml.datasets.get_dataset(2)._retrieve_class_labels(
+            target_attribute='product-type')
+        self.assertEqual(labels, ['C', 'H', 'G'])
+        print(labels)
```
amueller: no print please.

amueller:
Do you want to split up the PR? Alternatively, I can cherry pick 8314fac8b478bde672aac3633e42dee9c705133b 7a32fd4c01409992b71405ffad0b7dffdf449889 be03ca604851b8e1383f60fda59b337a871cbe15 9449250fecc16e12ec78d415cab53177e5b65419 into develop, and you can try to reorganize the rest? There is some test coverage, some sparse arff stuff, some sentinel stuff and some target value stuff that I think should be separate PRS.

mfeurer:
Okay, let's do it like that. I assumed the changes are small enough and only fixes, so they could be put together.

amueller:
and why is this not a PR onto master?

mfeurer:
master = current 'stable'
develop = the development branch

according to the [gitflow model](http://nvie.com/posts/a-successful-git-branching-model/)

amueller:
sorry I meant "why is this not a PR onto develop"

mfeurer:
This is now wrong; it wasn't because it actually built upon the already merged changes.

amueller:
I cherry picked the simple changes into develop.

openml/flows/flow.py
mfeurer: To change the name to be unique for each test run. I found this to be more reliable than deleting runs on the test server. Would it help to add a method `get_name` which I mock during test time?

openml/flows/flow.py
mfeurer: Would be handled by the suggestion above.

openml/datasets/dataset.py
mfeurer: Format is returned by the dataset xml. In the long run there will be more formats than arff and sparse_arff, so I would prefer to not try to infer this. Plus, the current arff reader can't do this.

openml/tasks/functions.py
mfeurer: Will fix that.

mfeurer:
Is [this comment](https://github.com/openml/openml-python/pull/150#issuecomment-243802154) still up to date, given your cherry-pick to develop?

openml/tasks/functions.py
mfeurer: This is actually an upstream issue: http://www.openml.org/api_docs/#!/task/get_task_id

Not sure now if I should fix this.

mfeurer:
This PR is now split into #155 #156 #157 #158

PR #156, opened 2016-08-31
FIX allow to load sparse data

mfeurer:
Second split of #150

### DIFF
--- a/openml/datasets/dataset.py
+++ b/openml/datasets/dataset.py
@@ -76,7 +76,7 @@ class OpenMLDataset(object):
                 logger.debug("Data pickle file already exists.")
             else:
                 try:
-                    data = self._get_arff()
+                    data = self._get_arff(self.format)
                 except OSError as e:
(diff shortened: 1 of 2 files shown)


mfeurer:
@amueller ready to be reviewed/merged

amueller:
I'm not entirely sure I understand this. So the user decides how to store the data? Or is the format in the ARFF different? Or does the information come from the dataset info on openml?

mfeurer:
The arff format is different. Ideally, the parser would figure that out itself, but it doesn't. The information comes from the dataset info in openml.

amueller:
Can't we snoop it from the file?

mfeurer:
Would be possible. But this solution is definitely easier and faster.

amueller:
I guess this is fine for now.

amueller:
btw, @raghavrv what happened to your arff reader?

joaquinvanschoren:
Openml tells you if the dataset is arff or sparse arff.

PR #157, opened 2016-08-31
FIX allow to retrieve class labels by the tasks class label flag

mfeurer:
Third split of #150

### DIFF
--- a/openml/tasks/functions.py
+++ b/openml/tasks/functions.py
@@ -236,7 +236,7 @@ def get_task(task_id):
 
     # TODO look into either adding the class labels to task xml, or other
     # way of reading it.
-    class_labels = dataset._retrieve_class_labels()
+    class_labels = dataset._retrieve_class_labels(task.target_feature)
     task.class_labels = class_labels
     return task
 
--- a/openml/tasks/functions.py
+++ b/openml/tasks/functions.py
@@ -236,7 +236,7 @@ def get_task(task_id):
 
     # TODO look into either adding the class labels to task xml, or other
     # way of reading it.
-    class_labels = dataset._retrieve_class_labels(task.target_feature)
+    class_labels = dataset.retrieve_class_labels(task.target_name)
     task.class_labels = class_labels
     return task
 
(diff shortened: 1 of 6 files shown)


mfeurer:
@amueller ready to be reviewed/merged

openml/tasks/functions.py
```diff
@@ -236,7 +236,7 @@ def get_task(task_id):
 
     # TODO look into either adding the class labels to task xml, or other
     # way of reading it.
-    class_labels = dataset._retrieve_class_labels()
+    class_labels = dataset._retrieve_class_labels(task.target_feature)
```
amueller: I don't understand this. What is this for regression or clustering or whatever? Is this classification specific? Is this None otherwise?
Also, as I said in my other comment, ``target_feature`` is a pretty bad name. I think it's called ``target`` in other places but maybe it should be ``target_name``? We should definitely unify this.

tests/datasets/test_datasets.py
```diff
@@ -200,6 +200,15 @@ def test_publish_dataset(self):
...
         self.assertEqual(return_code, 200)
 
+    def test__retrieve_class_labels(self):
+        openml.config.set_cache_directory(self.static_cache_dir)
+        labels = openml.datasets.get_dataset(2)._retrieve_class_labels()
+        self.assertEqual(labels, ['1', '2', '3', '4', '5', 'U'])
+        labels = openml.datasets.get_dataset(2)._retrieve_class_labels(
+            target_attribute='product-type')
+        self.assertEqual(labels, ['C', 'H', 'G'])
+        print(labels)
```
amueller: no print

openml/tasks/functions.py
amueller: also, it's weird that the task needs to access a private function of the dataset. That's not good.

mfeurer: Okay, I'll make the function public. Also, target_name sounds good to me, although the name comes from OpenML.

This is classification specific, and I should add a comment for that. It's actually necessary when creating the arff file with the results, which needs the real class name (not just the index as given by y) to upload the probabilities for each class to the server.

mfeurer:
I'll submit a fix for the issues.

mfeurer:
@amueller, I fixed the issues you mentioned.

amueller:
lgtm. thanks :)

PR #158, opened 2016-08-31
Fix/add sentinels

mfeurer:
This PR adds sentinels to flow names. This is necessary to upload the same flow in different test runs (travis-ci runs three tests in parallel) because of the (now) different name. Also, it does not require the user or us to immediately delete a flow uploaded by test.


openml/flows/flow.py
```diff
@@ -111,6 +113,10 @@ def _ensure_flow_exists(self):
 
         return int(flow_id)
 
+    def _get_name(self):
```
amueller: that's a slightly better solution ;)

openml/flows/flow.py
```diff
@@ -3,6 +3,7 @@
 import sklearn
 
 from .._api_calls import _perform_api_call
+from .. import config
```
amueller: why did you need to add that?

amueller:
Ok so you want the flows to have unique names? Why is that necessary? Aren't flows identified by their ID?

openml/flows/flow.py
mfeurer: For the bad solution I had before, this can be removed.

mfeurer:
A flow is identified by its name and external version. The ID is assigned by the server. Thus, if I upload a new flow with a name like DummyClassifier, there is a high chance it's already there. This is especially the case when several of the same tests run at the same time. Having different names circumvents this.

mfeurer:
I just removed the unused import. Are you okay with this PR?

openml/testing.py
```diff
@@ -39,6 +42,20 @@ def setUp(self):
...
+        pid = os.getpid()
+        md5 = hashlib.md5()
+        md5.update(apikey.encode('utf-8'))
+        md5.update(str(pid).encode('utf-8'))
+        random_number = np.random.randint(1000000000)
+        md5.update(str(random_number).encode('utf-8'))
+        sentinel = md5.hexdigest()
+        # For testing the hash code mustn't be bulletproof
+        self.sentinel = '%sTESTSENTINEL999' % sentinel[:8]
+        openml.config.testsentinel = self.sentinel
```
amueller: why not just use the system time in ns or something? you md5 hash it, so you just want something unique, right?

tests/flows/test_flow.py
```diff
@@ -17,9 +24,9 @@ def check_flow(flow):
         for flow in flows:
             check_flow(flow)
 
-    @unittest.skip('Not tested until test sentinels are added back.')
-    def test_upload_flow(self):
+    @mock.patch.object(openml.OpenMLFlow, '_get_name', autospec=True)
+    def test_upload_flow(self, name_mock):
         flow = openml.OpenMLFlow(model=DummyClassifier(), description="test description")
+        name_mock.return_value = '%s%s' % (self.sentinel, flow.name)
```
amueller: why not just
```python
flow.name = "TEST " + hashlib.md5(str(time.time()).encode("utf-8")).hexdigest()
```
?

amueller:
Ok so the names are unique on the server, and the client generates the name?

Why not just assign some value to ``flow.name`` in the test after creation of the flow before publishing?

PR #188, opened 2016-11-23
use ordered dict for component_reference

amueller:
this should fix travis?

### DIFF
--- a/openml/flows/sklearn_converter.py
+++ b/openml/flows/sklearn_converter.py
@@ -219,7 +219,7 @@ def _serialize_model(model):
             sub_components[k] = rval
             component_reference = OrderedDict()
             component_reference['oml:serialized_object'] = 'component_reference'
-            component_reference['value'] = OrderedDict(key=k, step_name=None))
+            component_reference['value'] = OrderedDict(key=k, step_name=None)
             component_reference = sklearn_to_flow(component_reference)
             parameters[k] = json.dumps(component_reference)
 
--- a/openml/flows/sklearn_converter.py
+++ b/openml/flows/sklearn_converter.py
@@ -217,9 +217,9 @@ def _serialize_model(model):
             # A subcomponent, for example the base model in
             # AdaBoostClassifier
             sub_components[k] = rval
-            component_reference = \
-                OrderedDict('oml:serialized_object': 'component_reference',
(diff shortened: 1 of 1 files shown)


amueller:
@mfeurer do you use some github bot?

mfeurer:
Nope, I don't know where the in progress comes from. Maybe waffle.io is still active. I'll have a look tomorrow.

amueller:
Not sure if the test failures are related... the string comparisons seem very unstable and hard to debug....

amueller:
I still get a random failures in ``FAIL: test_sklearn_to_upload_to_flow (tests.flows.test_flow.TestFlow)``

amueller:
It looks like ``json.dumps`` doesn't respect ``OrderedDict``:
```python
import json
from collections import OrderedDict
json.dumps(OrderedDict(a=1, b=2, c=3))
```
```
> '{"c": 3, "a": 1, "b": 2}'
```

amueller:
hm ok but that's not where the error is actually from. its from ``xml2dict``, I think.

amueller:
Ok, never mind, I don't understand ordered dicts, it seems.
The problem is that ``flow.components['estimator'].parameters['steps']`` contains dicts, not ``OrderedDicts``, and so the order is arbitrary. I have no idea where they come from, though.

amueller:
Actually, this looks to be just a json string of a dict. I don't understand how sending a string to the server and then retrieving the string again messes up the order in the dict. We are actually writing to the test server, right?

amueller:
Ok I have no idea what happens here, maybe you can look into it?

mfeurer:
I just removed waffle.io (which was the github bot). I'll now have a look at what's going wrong here.

mfeurer:
Okay, I've got a fix for this, but I'm not allowed to push to your PR. Could you please give me permissions to do so?

amueller:
It's odd that you can't push to my branch given that there is the "allow edits from maintainers" checkbox checked on the right.
The standard way (or maybe previous standard) to resolve this is to either make a PR to my branch from your fork, or to make a PR to the main repository from your branch. The second might be easier.

mfeurer:
Okay, I added a new branch called `amueller-ordered_dict_fix` to this github repo. Are you taking care of merging it into your branch?

PR #200, opened 2017-03-02
Fix/#187

janvanrijn:
fix issue 187


janvanrijn:
I merged this one, as Joaquins students are eager to use the Pipeline features and travis seems to run 5/6 tests OK. Also not reproducible on my system. 

Any clue why this one test fails? Something random server side? I will make an issue out of this so we can check this soon.

PR #201, opened 2017-03-02
Merge develop into master

janvanrijn:
Master does not run, does not make sense to keep this way. develop is sort of stable. (will receive a next fix this week)


joaquinvanschoren:
This is the wrong way around (merges master into develop)

PR #223, opened 2017-04-11
Issue210

janvanrijn:
- Fixed issue 210 by making a copy of the model for every CV fold (this is important to prevent data leakage). 
- Restructured the way an arff trace is generated (arff content and arff header are extracted separately, this was needed because of new assumptions and in order to keep the code clean after previous point)
- Apparently, I also added the 'upload incompatible run with error message' in this branch. Sorry, but it's it's just two additional lines anyway.

### DIFF
--- a/openml/flows/sklearn_converter.py
+++ b/openml/flows/sklearn_converter.py
@@ -551,27 +551,3 @@ def _deserialize_cross_validator(value, **kwargs):
 
 def _format_external_version(model_package_name, model_package_version_number):
     return '%s==%s' % (model_package_name, model_package_version_number)
-
-
-def get_traceble_model(model):
-    '''
-    Returns whether the model can produce an openml trace. If yes,
-    return that model. Returns false otherwise.
-    Clause holds true for instances of BaseSearchCV and Pipelines that
-    contain exactly one BaseSearchCV (no use to search deeper in the
-    Tree, maybe later)
-    '''
-    if isinstance(model, sklearn.model_selection._search.BaseSearchCV):
-        return True
-    count = 0
-    returnValue = None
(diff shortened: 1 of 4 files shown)


mfeurer:
Unittests are broken.

openml/flows/sklearn_converter.py
```diff
@@ -551,3 +551,27 @@ def _deserialize_cross_validator(value, **kwargs):
...
+    return that model. Returns false otherwise.
+    Clause holds true for instances of BaseSearchCV and Pipelines that
+    contain exactly one BaseSearchCV (no use to search deeper in the
+    Tree, maybe later)
+    '''
+    if isinstance(model, sklearn.model_selection._search.BaseSearchCV):
+        return True
+    count = 0
+    returnValue = None
+    if isinstance(model, sklearn.pipeline.Pipeline):
```
mfeurer: I do not see any usecase for this. Do you have one in mind?

openml/runs/functions.py
```diff
@@ -160,22 +159,27 @@ def _run_task_get_arffcontent(model, task, class_labels):
...
+                model_fold.fit(trainX, trainY)
 
-            if isinstance(model, BaseSearchCV):
-                _add_results_to_arfftrace(arff_tracecontent, fold_no, model, rep_no)
-                model_classes = model.best_estimator_.classes_
-            else:
-                model_classes = model.classes_
+                if get_traceble_model(model_fold):
+                    arff_tracecontent.extend(_extract_arfftrace(model_fold, rep_no, fold_no))
+                    model_classes = model_fold.best_estimator_.classes_
```
mfeurer: In the worst case, the model classes can be different for each fold. Do we cope for that?

PR #241, opened 2017-04-27
Fix169

janvanrijn:


openml/runs/functions.py
```diff
@@ -88,6 +94,24 @@ def run_task(task, model, avoid_duplicate_runs=True, flow_tags=None):
 
     return run
 
+def initialize_model_from_run(run_id):
```
janvanrijn: Agreed, I added tests.

openml/runs/functions.py
```diff
@@ -111,6 +135,44 @@ def _run_exists(task_id, setup_id):
         assert(exception.code == 512)
         return False
 
+def _get_seeded_model(model, seed=None):
+    '''Sets all the non-seeded components of a model with a seed.
```
janvanrijn: I'm not sure. One could argue that that is a restriction of the run_tasks function, as that is the function the user interacts with. Furthermore, that function is responsible for the check.

This function only adds seeds to unseeded models

openml/runs/functions.py
```diff
@@ -111,6 +135,44 @@ def _run_exists(task_id, setup_id):
...
+            a seed
+    '''
+
+    rs = np.random.RandomState(seed)
+    model_params = model.get_params()
+    random_states = {}
+    for param_name in sorted(model_params):
+        if 'random_state' in param_name:
+            currentValue = model_params[param_name]
+            # important to draw the value at this point (and not in the if statement)
```
janvanrijn: Added description. 

I'm not sure if we really need this, but seems nice property to respect.

openml/runs/functions.py
mfeurer: But it can raise an exception:

```
import numpy as np
import sklearn.ensemble

import openml


rf = sklearn.ensemble.RandomForestClassifier(
    random_state=np.random.RandomState(1))
openml.runs.functions._get_seeded_model(rf, 5)
```
But you're right, it should be documented in the `run_tasks()` function.

openml/runs/functions.py
mfeurer: k

openml/runs/functions.py
janvanrijn: Mea culpa, i will add it

openml/setups/setup.py
```diff
@@ -0,0 +1,35 @@
...
+        self.flow_id = flow_id
+        self.parameters = parameters
+
+
+class OpenMLParameter(object):
+    """Parameter object (used in setup).
+
+       Parameters
+       ----------
+       flow_id : int
```
mfeurer: It looks like you copied the docstring from the object above and didn't adapt it.

openml/runs/run.py
```diff
@@ -176,50 +176,38 @@ def _parse_parameters(model, flow):
...
+            _params = []
+            for _param_name in _flow.parameters:
+                _current = OrderedDict()
+                _current['oml:name'] = _param_name
+                _current['oml:value'] = _flow.parameters[_param_name]
+                if _main_call:
+                    _current['oml:component'] = main_id
                 else:
-                    raise ValueError("parameter %s not in flow description of flow %s" %(param,flow.name))
+                    _current['oml:component'] = _param_dict[_flow.name]
```
mfeurer: Why is it once an ID, and once a name?

openml/runs/run.py
```diff
@@ -165,7 +165,7 @@ def _create_description_xml(self):
         return description_xml
 
     @staticmethod
-    def _parse_parameters(model, flow):
+    def _parse_parameters(model, server_flow):
```
mfeurer: Looking at this again I'm actually surprised that this is not called `run_task`, but only when publishing. But maybe this should be its own issue/PR.

tests/test_setups/test_setup_functions.py
```diff
@@ -45,33 +46,47 @@ def test_nonexisting_setup_exists(self):
...
+        # contains all special cases, 0 params, 1 param, n params.
+        # Non scikitlearn flows.
+        setups = [18, 19, 20, 118]
+        num_params = [8, 0, 3, 1]
+
+        for idx in range(len(setups)):
+            current = openml.setups.get_setup(setups[idx])
+            assert current.flow_id > 0
+            if num_params[idx] == 0:
+                assert current.parameters is None
```
mfeurer: Could you please use `self.assert` in the unit tests? It gives nicer outputs.

openml/setups/setup.py
janvanrijn: I loove copy/pasting. fixed it.

openml/runs/run.py
janvanrijn: Can you elaborate a bit? I don really understand which / why

tests/test_setups/test_setup_functions.py
janvanrijn: You mean self.asserts()?
Sure.

janvanrijn:
test_existing_setup_exists makes use of sentinels. That gives us a general problem when serializing a flow (or setup) and comparing it to one on the server, as the flow (setup) serialization is not aware of the (and which) sentinel string was used. For that reason, I slightly generalized one of the functions, such that it does not rely on name mappings for the main flow (boolean flag that indicates a function call on depth 1). This way, we do not have to remove the sentinels and we can in this context still test flows without subflows.

tests/test_setups/test_setup_functions.py
```diff
@@ -87,6 +87,6 @@ def test_setup_get(self):
             current = openml.setups.get_setup(setups[idx])
             assert current.flow_id > 0
             if num_params[idx] == 0:
-                self.asserts(current.parameters is None)
+                self.assertTrue(current.parameters is None)
```
mfeurer: Sorry, I should have been more specific. The correct one here would be `self.assertIsNone`.

tests/test_setups/test_setup_functions.py
```diff
@@ -87,6 +87,6 @@ def test_setup_get(self):
...
             else:
-                self.asserts(len(current.parameters) == num_params[idx])
+                self.assertTrue(len(current.parameters) == num_params[idx])
```
mfeurer: Sorry, I should have been more specific. The correct one here would be `self.assertEqual`.

openml/runs/run.py
mfeurer: Sorry, I meant "that this is not called IN `run_task`"

mfeurer:
Thanks for the explanation of the sentinels. Maybe it would be good to add that to the actual unit test. What I'm wondering right now is if this is save with respect to running the unit tests in parallel (which happens on travis-ci). You assume:
```
# although the flow exists, we can be sure there are no
# setups (yet) as it hasn't been ran
```
it could happen that it was already run. Maybe adding a sentinel to a hyperparameter? Or is there some other way of making a setup unique?

Once we figured out this part, I think we can merge this PR before it becomes too big and create new PRs for the missing functionality.

janvanrijn:
I had already added this in the comment:
```
# because of the sentinel, we can not use flows that contain subflows
```

Also, the other comment was slightly confusing. I changed this to:
```
# although the flow exists (created as of previous statement),
# we can be sure there are no setups (yet) as it was just created
# and hasn't been ran
```

Should be fine now? Let's merge

PR #242, opened 2017-04-27
Reuse the predictions from previous call.

Yatoom:
This seems to pass the unit tests, and will result in a faster execution time.

### DIFF
--- a/openml/runs/functions.py
+++ b/openml/runs/functions.py
@@ -197,7 +197,7 @@ def _run_task_get_arffcontent(model, task, class_labels):
                 modelpredict_starttime = time.process_time()
             
             ProbaY = model_fold.predict_proba(testX)
-            PredY = model_fold.predict(testX)
+            PredY = np.argmax(ProbaY, axis=1)
             if can_measure_runtime:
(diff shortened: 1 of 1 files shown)


mfeurer:
This would basically let the OpenML connector do the tie-breaking instead of the classifier itself, right? Would this work with classification thresholds other than 0.5 in the binary case?

Yatoom:
Ah I see. On second thought, this wasn't a very good idea. Indeed tie-breaking already forms a problem.

```
import numpy as np

X_train = [[1, 1, 1], [3, 3, 3]]
y_train = [1, 3]
X_test = [[2, 2, 2]]
y_test = [2]

knn = KNeighborsClassifier(n_neighbors=2)
knn.fit(X_train, y_train)
print(knn.predict_proba(X_test))                     # --> [[0.5, 0.5]]
print(knn.predict(X_test))                           # --> [1]
print(np.argmax(knn.predict_proba(X_test), axis=1))  # --> [0]

```
But next to that, the Sklearn models use different techniques inside `predict()` and `predict_proba()`.

PR #243, opened 2017-04-28
Allow list-like objects in Sklearn converter.

Yatoom:
Rather than checking if an object is a `list` or `tuple`, I think we could also allow generators, ranges and numpy arrays to be iterated over.

The Sequence class covers lists, tuples, ranges and strings. To avoid iterating over strings, the code block is moved below the block that checks for strings. The `np.ndarray` covers the numpy arrays, and `inspect.isgenerator(o)` covers generator objects.

### DIFF
--- a/openml/flows/sklearn_converter.py
+++ b/openml/flows/sklearn_converter.py
@@ -39,8 +39,8 @@ def sklearn_to_flow(o, parent_model=None):
     if _is_estimator(o):
         # is the main model or a submodel
         rval = _serialize_model(o)
-    elif inspect.isgeneratorfunction(o) or (isinstance(o, Sequence) and not isinstance(o, str)):
-        # this parameter is a generator, list, range, xrange, or tuple
+    elif inspect.isgeneratorfunction(o) or isinstance(o, (Sequence, np.ndarray)) and not isinstance(o, str):
+        # this parameter is a generator, list, range, xrange, tuple or numpy array
         rval = [sklearn_to_flow(element, parent_model) for element in o]
         if isinstance(o, tuple):
             rval = tuple(rval)
--- a/openml/flows/sklearn_converter.py
+++ b/openml/flows/sklearn_converter.py
@@ -1,6 +1,6 @@
 """Convert scikit-learn estimators into an OpenMLFlows and vice versa."""
 
-from collections import OrderedDict
+from collections import OrderedDict, Sequence
 import copy
 from distutils.version import LooseVersion
 import importlib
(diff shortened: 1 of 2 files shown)


mfeurer:
Could you please provide us with a usecase where this is necessary? In general, we try to put strict assumptions on the types we allow into the converter in order to make unit testing and de-serealization easier.

Yatoom:
Here is a simple use-case:
```python
import openml
from sklearn.ensemble import ExtraTreesClassifier
from sklearn.model_selection import RandomizedSearchCV
import numpy as np

task = openml.tasks.get_task(49)

param_grid = {
    'criterion': ["gini", "entropy"],
    'max_features': np.arange(0.05, 1.01, 0.05),
    'min_samples_split': range(2, 21),
    'min_samples_leaf': range(1, 21),
    'bootstrap': [True, False]
}

etc = ExtraTreesClassifier(n_jobs=-1, n_estimators=10, verbose=True)

clf = RandomizedSearchCV(etc, param_distributions=param_grid)

run = openml.runs.run_task(task, clf)
run.publish()
```
I agree that it is not necessary, since you can convert these to lists instead, but it would make the API less restrictive and friendlier to the end-user.

mfeurer:
Thanks for the example. Yes, this makes sense. If you can provide unit tests I'd be happy to merge this PR.

Yatoom:
Hmm, would this be what you want in a unit test?
```python
    def test_list_like_objects(self):
        param_distributions = {
            # List
            'criterion': ["gini", "entropy"],

            # Numpy array
            'max_features': np.arange(0.05, 1.01, 0.05),

            # Range (uses xrange() in python2, and range() in python3)
            'min_samples_split': moves.range(2, 21),

            # Generator
            'min_samples_leaf': [i * 2 for i in moves.range(1, 10)],

            # Tuple
            'bootstrap': (True, False)
        }

        param_distributions_lists = {
            # List
            'criterion': ["gini", "entropy"],

            # Numpy array converted to list
            'max_features': list(np.arange(0.05, 1.01, 0.05)),

            # Range (uses xrange() in python2, and range() in python3) converted to list
            'min_samples_split': list(moves.range(2, 21)),

            # Generator converted to list
            'min_samples_leaf': list([i * 2 for i in moves.range(1, 10)]),

            # Tuple
            'bootstrap': (True, False)
        }

        serialized = sklearn_to_flow(param_distributions)
        deserialized = flow_to_sklearn(serialized)
        self.assertEqual(deserialized, param_distributions_lists)
        self.assertIsNot(deserialized, param_distributions_lists)
```
When serializing and deserializing you end up with lists (except for tuples), so this checks if this conversion was done correctly and no errors occurred.

mfeurer:
Looks good to me. Do you know of any cases where it is supposed to fail/not accept a generator object? If yes, could you add a test for that as well?

Yatoom:
@mfeurer could you please check this again? I have used the above unit test, but this time with an actual generator, and I have added a unit test to check if the normal TypeError is raised when we add an illegal object (i.e. it is not an estimator or cross validator) via the generator.

PGijsbers:
Closed because right now we do not want to support generators. This is for the reason that we can not easily de-serialize it. If the generator only represents a set list of values (such as the `arange`), for now we would prefer the explicit conversion by the end user.

Issue #246, opened 2017-05-04
serialize parameter grid in optimization trace


janvanrijn:
So we can easily re-instantiate if needed

PR #249, opened 2017-05-08
Fix246

janvanrijn:
Please check the PR. I can imagine that one of the Unit tests must be rewritten at a later stage (for speed) but for now it seems like it does the trick.

### DIFF
--- a/openml/runs/functions.py
+++ b/openml/runs/functions.py
@@ -133,6 +133,7 @@ def initialize_model_from_run(run_id):
     run = get_run(run_id)
     return initialize_model(run.setup_id)
 
+
 def initialize_model_from_trace(run_id, repeat, fold, iteration=None):
     '''
     Initialize a model based on the parameters that were set
@@ -142,7 +143,8 @@ def initialize_model_from_trace(run_id, repeat, fold, iteration=None):
     Parameters
     ----------
     run_id : int
-        The Openml run_id. Should contain a trace file
+        The Openml run_id. Should contain a trace file, 
+        otherwise a OpenMLServerException is raised
 
     repeat: int
         The repeat nr (column in trace file)
@@ -152,7 +154,9 @@ def initialize_model_from_trace(run_id, repeat, fold, iteration=None):
 
     iteration: int
         The iteration nr (column in trace file). If None, the
-        best (selected) iteration will be searched (slow)
+        best (selected) iteration will be searched (slow), 
+        according to the selection criteria implemented in
+        OpenMLRunTrace.get_selected_iteration
 
     Returns
     -------
--- a/openml/runs/functions.py
+++ b/openml/runs/functions.py
@@ -17,7 +17,7 @@ from ..setups import setup_exists, initialize_model
 
 from ..exceptions import OpenMLCacheException, OpenMLServerException
 from ..util import URLError, version_complies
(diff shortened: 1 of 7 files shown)


mfeurer:
Adds missing functionality described in #246

openml/runs/functions.py
```diff
@@ -112,6 +122,50 @@ def initialize_model_from_run(run_id):
...
+def initialize_model_from_trace(run_id, repeat, fold, iteration=None):
+    '''
+    Initialized a model based on the parameters that were set
+    by an optimization procedure (i.e., using the exact same
+    parameter settings)
+
+    Parameters
+    ----------
+    run_id : int
+        The Openml run_id. Should contain a trace file
```
mfeurer: What happens if it doesn't?

openml/runs/functions.py
```diff
@@ -112,6 +122,50 @@ def initialize_model_from_run(run_id):
...
+
+    repeat: int
+        The repeat nr (column in trace file)
+
+    fold: int
+        The fold nr (column in trace file)
+
+    iteration: int
+        The iteration nr (column in trace file). If None, the
+        best (selected) iteration will be searched (slow)
```
mfeurer: Currently, multiple models in the trace can be marked as selected. What happens in that case?

openml/runs/trace.py
```diff
@@ -0,0 +1,55 @@
...
+        self.trace_iterations = trace_iterations
+
+    def get_selected_iteration(self, fold, repeat):
+        for (r, f, i) in self.trace_iterations:
+            if r == repeat and f == fold and self.trace_iterations[(r, f, i)].selected == True:
+                return i
+        raise ValueError('Could not find the selected iteration for rep/fold %d/%d' %(repeat,fold))
+
+    def __str__(self):
+        return '[Run id: %d, %d trace iterations]' %(self.run_id, len(self.trace_iterations))
```
mfeurer: I think we should decide on a format for all classes in the OpenML package before adding functions like `__str__` and `__repr__`.

tests/test_runs/test_run_functions.py
```diff
@@ -26,6 +29,47 @@
...
+                # probably because openml eval engine has not executed this run yet
+                time.sleep(10)
+                secCount += 10
+
+        run_prime = openml.runs.run_task(task, model_prime, avoid_duplicate_runs=False)
+        predictions_prime = run_prime._generate_arff_dict()
+
+        self.assertEquals(len(predictions_prime['data']), len(predictions['data']))
+
+        # The original search model does not submit confidence bounds,
```
mfeurer: Do you mean probabilities here?

openml/runs/functions.py
janvanrijn: I added to the comments

openml/runs/functions.py
janvanrijn: added to the comments

openml/runs/trace.py
janvanrijn: Added a comment that this will change. Was actually very useful for debugging, seems suboptimal to remove

tests/test_runs/test_run_functions.py
janvanrijn: No, number of (predicted) instances.

Issue #193, opened 2017-01-31
Allow to pass a flow object to run_task


mfeurer:
Currently, the function `run_task` only accepts a model as an input; which is then used to query the OpenML server for a flow. If the user constructs the flow himself/herself, `run_task` should accept this flow object as well.

joaquinvanschoren:
+1

mfeurer:
The merge of PR #253 enables this.

PR #253, opened 2017-05-11
Allow running a flow on a task

mfeurer:
Side effect: during unit testing, it is possible to add a sentinel to the flow name to not work with flows which are already uploaded to the server.

Adds #193.

Further changes:
* model of a flow is now created when downloading the flow
* parameters of a run are now parsed when the run is executed
* simplifies interface of several functions
* simplified downloading of a flow from a flow id

### DIFF
--- a/openml/flows/flow.py
+++ b/openml/flows/flow.py
@@ -324,10 +324,15 @@ class OpenMLFlow(object):
         arguments['model'] = None
         flow = cls(**arguments)
 
-        if arguments['external_version'].startswith('sklearn'):
+        # try to parse to a model because not everything that can be
+        # deserialized has to come from scikit-learn. If it can't be
+        # serialized, but comes from scikit-learn this is worth an exception
+        try:
             from .sklearn_converter import flow_to_sklearn
             model = flow_to_sklearn(flow)
-        else:
+        except Exception as e:
+            if arguments['external_version'].startswith('sklearn'):
+                raise e
             model = None
         flow.model = model
 
--- a/openml/flows/flow.py
+++ b/openml/flows/flow.py
@@ -322,7 +322,16 @@ class OpenMLFlow(object):
         arguments['tags'] = tags
 
         arguments['model'] = None
-        return cls(**arguments)
+        flow = cls(**arguments)
+
+        if 'sklearn' in arguments['external_version']:
+            from .sklearn_converter import flow_to_sklearn
+            model = flow_to_sklearn(flow)
+        else:
+            model = None
+        flow.model = model
+
+        return flow
 
     def publish(self):
         """Publish flow to OpenML server.
@@ -332,32 +341,38 @@ class OpenMLFlow(object):
         self : OpenMLFlow
 
         """
+        import openml.flows.functions
 
         xml_description = self._to_xml()
 
         file_elements = {'description': xml_description}
         return_value = _perform_api_call("flow/", file_elements=file_elements)
-        self.flow_id = int(xmltodict.parse(return_value)['oml:upload_flow']['oml:id'])
+        flow_id = int(xmltodict.parse(return_value)['oml:upload_flow']['oml:id'])
+        flow = openml.flows.functions.get_flow(flow_id)
         try:
-            _check_flow(self)
+            openml.flows.functions.assert_flows_equal(self, flow)
         except ValueError as e:
             message = e.args[0]
             raise ValueError("Flow was not stored correctly on the server. "
                              "New flow ID is %d. Please check manually and "
                              "remove the flow if necessary! Error is:\n'%s'" %
-                             (self.flow_id, message))
+                             (flow_id, message))
+        _copy_server_fields(flow, self)
         return self
+
+
+def _copy_server_fields(source_flow, target_flow):
+    fields_added_by_the_server = ['flow_id', 'uploader', 'version',
+                                  'upload_date']
+    for field in fields_added_by_the_server:
+        setattr(target_flow, field, getattr(source_flow, field))
+
+    for name, component in source_flow.components.items():
+        assert name in target_flow.components
+        _copy_server_fields(component, target_flow.components[name])
 
 
 def _add_if_nonempty(dic, key, value):
     if value is not None:
         dic[key] = value
 
-
-def _check_flow(flow):
-    # Import is not possible at the top of the file as this would cause an
-    # ImportError due to an import cycle.
-    import openml.flows.functions
-
-    flow_copy = openml.flows.functions.get_flow(flow.flow_id)
-    openml.flows.functions.assert_flows_equal(flow, flow_copy)
--- a/openml/flows/flow.py
+++ b/openml/flows/flow.py
@@ -324,7 +324,7 @@ class OpenMLFlow(object):
         arguments['model'] = None
         flow = cls(**arguments)
 
-        if 'sklearn' in arguments['external_version']:
+        if arguments['external_version'].startswith('sklearn'):
             from .sklearn_converter import flow_to_sklearn
             model = flow_to_sklearn(flow)
         else:
--- a/openml/flows/flow.py
+++ b/openml/flows/flow.py
@@ -349,15 +349,15 @@ class OpenMLFlow(object):
         return_value = _perform_api_call("flow/", file_elements=file_elements)
         flow_id = int(xmltodict.parse(return_value)['oml:upload_flow']['oml:id'])
         flow = openml.flows.functions.get_flow(flow_id)
+        _copy_server_fields(flow, self)
         try:
-            openml.flows.functions.assert_flows_equal(self, flow)
+            openml.flows.functions.assert_flows_equal(self, flow, flow.upload_date)
         except ValueError as e:
             message = e.args[0]
             raise ValueError("Flow was not stored correctly on the server. "
                              "New flow ID is %d. Please check manually and "
                              "remove the flow if necessary! Error is:\n'%s'" %
                              (flow_id, message))
-        _copy_server_fields(flow, self)
         return self
 
 
--- a/openml/flows/flow.py
+++ b/openml/flows/flow.py
@@ -341,6 +341,10 @@ class OpenMLFlow(object):
         self : OpenMLFlow
 
         """
+        # Import at top not possible because of cyclic dependencies. In
+        # particular, flow.py tries to import functions.py in order to call
+        # get_flow(), while functions.py tries to import flow.py in order to
+        # instantiate an OpenMLFlow.
         import openml.flows.functions
 
         xml_description = self._to_xml()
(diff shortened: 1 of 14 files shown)


openml/runs/functions.py
```diff
@@ -25,7 +26,31 @@
...
 
-def run_task(task, model, avoid_duplicate_runs=True, flow_tags=None, seed=None):
+def run_model_on_task(task, model, avoid_duplicate_runs=True, flow_tags=None,
+                      seed=None):
+    flow = sklearn_to_flow(model)
+
+    # returns flow id if the flow exists on the server, False otherwise
+    flow_id = flow_exists(flow.name, flow.external_version)
+
+    if flow_id == False:
```
janvanrijn: Now we are back to publishing a flow before knowing whether it actually valid (before we first ran the task)?

openml/runs/functions.py
```diff
@@ -51,23 +76,18 @@ def run_task(task, model, avoid_duplicate_runs=True, flow_tags=None, seed=None):
...
-    flow_id = flow_exists(flow.name, flow.external_version)
+    flow.model = _get_seeded_model(flow.model, seed=seed)
 
     # skips the run if it already exists and the user opts for this in the config file.
     # also, if the flow is not present on the server, the check is not needed.
-    if avoid_duplicate_runs and flow_id:
-        flow = get_flow(flow_id)
-        setup_id = setup_exists(flow, model)
+    if avoid_duplicate_runs:
+        flow_from_server = get_flow(flow.flow_id)
```
janvanrijn: Why don't we require that the 'run_flow_on_task' runs on a flow from the server?

openml/runs/functions.py
```diff
@@ -79,18 +99,12 @@ def run_task(task, model, avoid_duplicate_runs=True, flow_tags=None, seed=None):
...
 
-    if flow_id == False:
-        # means the flow did not exists. As we could run it, publish it now
-        flow = flow.publish()
-    else:
-        # flow already existed, download it from server
-        # TODO (neccessary? is this a post condition of this function)
-        flow = get_flow(flow_id)
-
     run.flow_id = flow.flow_id
```
janvanrijn: should be set in run constructor

openml/runs/run.py
```diff
@@ -156,56 +152,82 @@ def _create_description_xml(self):
...
 
         """
-        if server_flow.flow_id is None:
-            raise ValueError("The flow parameter needs to be downloaded from server")
+
+        # Depth-first search to check if all components were uploaded to the
+        # server before parsing the parameters
+        stack = list()
+        stack.append(flow)
+        while len(stack) > 0:
```
janvanrijn: can we make a separate function of this (modularity / readability)

janvanrijn: and reusability :)

openml/setups/functions.py
```diff
@@ -26,9 +26,11 @@ def setup_exists(downloaded_flow, sklearn_model):
     '''
 
     # sadly, this api call relies on a run object
-    openml_param_settings = openml.runs.OpenMLRun._parse_parameters(sklearn_model, downloaded_flow)
-    description = xmltodict.unparse(_to_dict(downloaded_flow.flow_id, openml_param_settings), pretty=True)
-    file_elements = {'description': ('description.arff',description)}
+    openml_param_settings = openml.runs.OpenMLRun._parse_parameters(flow)
```
janvanrijn: maybe explicitly state in comments that this function raises an error if the flow does not contain all flow ids?

tests/test_runs/test_run_functions.py
```diff
@@ -87,10 +90,22 @@ def _check_serialized_optimized_run(self, run_id):
 
         return True
 
+    def _perform_run(self, task_id, num_instances, clf,
+                     random_state_value=None, check_setup=True):
+
+        def _remove_random_state(flow):
```
janvanrijn: why remove random state? This seems like part of the behaviour we want to test

openml/flows/flow.py
```diff
@@ -322,7 +322,16 @@ def _from_dict(cls, xml_dict):
         arguments['tags'] = tags
 
         arguments['model'] = None
-        return cls(**arguments)
+        flow = cls(**arguments)
+
+        if 'sklearn' in arguments['external_version']:
```
mfeurer: I will replace this with `startswith('sklearn')`. In the long run, we probably need to build a plugin-like system in which the converters can register themselves for an 'external_version' string.

openml/flows/flow.py
```diff
@@ -332,32 +341,38 @@ def publish(self):
...
         xml_description = self._to_xml()
 
         file_elements = {'description': xml_description}
         return_value = _perform_api_call("flow/", file_elements=file_elements)
-        self.flow_id = int(xmltodict.parse(return_value)['oml:upload_flow']['oml:id'])
+        flow_id = int(xmltodict.parse(return_value)['oml:upload_flow']['oml:id'])
+        flow = openml.flows.functions.get_flow(flow_id)
         try:
-            _check_flow(self)
+            openml.flows.functions.assert_flows_equal(self, flow)
```
mfeurer: It's still on my todo-list to test for this expected error.

openml/flows/functions.py
```diff
@@ -24,9 +24,6 @@ def get_flow(flow_id):
     flow_dict = xmltodict.parse(flow_xml)
     flow = OpenMLFlow._from_dict(flow_dict)
 
-    if 'sklearn' in flow.external_version:
-        flow.model = flow_to_sklearn(flow)
```
mfeurer: This piece of code is removed, I can't add anything here.

openml/runs/functions.py
mfeurer: That's actually an issue I somehow forgot. Will take care of this.

openml/runs/functions.py
mfeurer: We then couldn't change the parameter values of the model in the flow as the run function would always use the parameters from the server.

openml/runs/functions.py
mfeurer: Of course

openml/runs/run.py
mfeurer: I actually thought the very same thing ;) It's already done.

tests/test_runs/test_run_functions.py
mfeurer: It's removed after checking the value to make sure that `assert_flow_equal` works.

mfeurer:
Flows are published after running them (only if they haven't been published before). @janvanrijn this is ready for review again.

openml/flows/flow.py
```diff
@@ -332,32 +341,38 @@ def publish(self):
         self : OpenMLFlow
 
         """
+        import openml.flows.functions
```
janvanrijn: imports at the top! (right?)

mfeurer: This is not possible because of cyclic dependencies. In particular, `flow.py` tries to import `functions.py` in order to call `get_flow()`, while `functions.py` tries to import `flow.py` in order to instantiate an `OpenMLFlow`. I will add a comment.

mfeurer:
Sorry for the failing tests. It seems that I was too sloppy yesterday evening. The tests are passing.

openml/runs/functions.py
```diff
@@ -51,23 +63,19 @@ def run_task(task, model, avoid_duplicate_runs=True, flow_tags=None, seed=None):
...
-    model = _get_seeded_model(model, seed)
-    flow = sklearn_to_flow(model)
 
-    # returns flow id if the flow exists on the server, False otherwise
-    flow_id = flow_exists(flow.name, flow.external_version)
+    flow.model = _get_seeded_model(flow.model, seed=seed)
 
     # skips the run if it already exists and the user opts for this in the config file.
     # also, if the flow is not present on the server, the check is not needed.
+    flow_id = flow_exists(flow.name, flow.external_version)
```
janvanrijn: document somewhere that we need to have the 'avoid_duplicate_runs' to false if we want offline experiments

openml/runs/run.py
```diff
@@ -156,58 +152,86 @@ def _create_description_xml(self):
...
+                if isinstance(_tmp, openml.flows.OpenMLFlow):
+                    continue
+                try:
+                    _tmp = json.dumps(_tmp)
+                except TypeError as e:
+                    # Python3.5 exception message:
+                    # <openml.flows.flow.OpenMLFlow object at 0x7fed87978160> is not JSON serializable
+                    # Python3.6 exception message:
+                    # Object of type 'OpenMLFlow' is not JSON serializable
+                    if 'OpenMLFlow' in e.args[0] and \
```
janvanrijn: document what happens in case of the catch (and why)

janvanrijn: in the catch, please try to define what can reasonably fall into that (because we handle it further down) and raise exception if we have something unexpected

openml/setups/functions.py
```diff
@@ -4,20 +4,23 @@
...
 
 from .setup import OpenMLSetup, OpenMLParameter
+from openml.flows import sklearn_to_flow, flow_exists
 
 
-def setup_exists(downloaded_flow, sklearn_model):
+def setup_exists(flow, model=None):
     '''
-    Checks whether a flow / hyperparameter configuration already exists on the server
+    Checks whether a hyperparameter configuration already exists on the server.
```
janvanrijn: please document why model can be none (i.e., flow.model is set)
please raise exception if flow.model is set and model is set, and these do not agree? (should never happen but still..)

openml/setups/functions.py
```diff
@@ -75,37 +89,39 @@ def initialize_model(setup_id):
...
+    for _param in setup.parameters:
+        _flow_id = setup.parameters[_param].flow_id
+        _param_name = setup.parameters[_param].parameter_name
+        _param_value = setup.parameters[_param].value
+        if _flow_id not in parameters:
+            parameters[_flow_id] = {}
+        parameters[_flow_id][_param_name] = _param_value
 
     def _reconstruct_flow(_flow, _params):
-        # sets the values of flow parameters (and subflows) to
```
janvanrijn: small todo (sorry, should have been me in prev pull request): document what types _flow (flow object?) and _params (totally forgot) are? this would make this function better understandable

openml/runs/run.py
mfeurer: I will add additional checks.

openml/setups/functions.py
mfeurer: What do you mean by models do not agree? You mean names don't match? Or the parameter names don't match? Or the parameter values?

Matching of parameter names is done somewhat in _parse_parameters, but I could make that more strict (I'll think about it...)

mfeurer: Okay, I'll add a more strict check.

PR #265, opened 2017-06-13
Support for Learning Curve Tasks

janvanrijn:
@mfeurer very curious for your opinion

I will walk you through it in the office


openml/runs/functions.py
```diff
@@ -349,82 +361,93 @@ def _run_task_get_arffcontent(model, task, class_labels):
...
+                try:
+                    # for measuring runtime. Only available since Python 3.3
+                    if can_measure_runtime:
+                        modelfit_starttime = time.process_time()
+                    model_fold.fit(trainX, trainY)
+
+                    if can_measure_runtime:
+                        modelfit_duration = (time.process_time() - modelfit_starttime) * 1000
+                        user_defined_measures_sample['usercpu_time_millis_training'][rep_no][fold_no][sample_no] = modelfit_duration
+                        user_defined_measures_fold['usercpu_time_millis_training'][rep_no][fold_no] = modelfit_duration
```
mfeurer: This will overwrite the result for each new sample size, right?

openml/runs/functions.py
```diff
@@ -349,82 +361,93 @@ def _run_task_get_arffcontent(model, task, class_labels):
     X, Y = task.get_X_and_y()
     arff_datacontent = []
     arff_tracecontent = []
-    user_defined_measures = defaultdict(lambda: defaultdict(dict))
+    user_defined_measures_fold = defaultdict(lambda: defaultdict(dict))
+    user_defined_measures_sample = defaultdict(lambda: defaultdict(lambda: defaultdict(dict)))
```
mfeurer: I don't get the difference between the two.

openml/runs/functions.py
```diff
@@ -571,7 +594,8 @@ def _create_run_from_xml(xml):
 
     files = dict()
     evaluations = dict()
-    detailed_evaluations = defaultdict(lambda: defaultdict(dict))
+    fold_evaluations = defaultdict(lambda: defaultdict(dict))
+    sample_evaluations = defaultdict(lambda: defaultdict(lambda: defaultdict(dict)))
```
mfeurer: I still don't understand the difference. If there are no subsamples, the only subsample index would be `0`, thus the result could be saved only in the `sample_evaluations` field, right?

openml/tasks/task.py
```diff
@@ -36,32 +38,28 @@ def get_dataset(self):
...
+        if self.task_type_id == 1:
+        # if 'Supervised Classification'.lower() in self.task_type.lower():
             target_dtype = int
-        elif 'Supervised Regression'.lower() in self.task_type.lower():
+        # elif 'Supervised Regression'.lower() in self.task_type.lower():
+        elif self.task_type_id == 2:
             target_dtype = float
+        # elif ''.lower('Learning Curve') in self.task_type.lower():
+        elif self.task_type_id == 3:
+            target_dtype = int
```
mfeurer: Why is the target of a learning curve automatically an integer?

mfeurer:
Also, could you please merge into develop?

openml/runs/functions.py
janvanrijn: We now have measures per fold (dict with 3 levels of keys) and measures per sample (dict with 4 levels of keys).

openml/runs/functions.py
janvanrijn: Yes. In the case of measures per fold, we are interested in how something ran on the full dataset. Which is enforced by the way that the last loop is always the result on the full dataset

openml/runs/functions.py
janvanrijn: It's all about assumptions of the datastructure. We can always know for sure that a fold_evaluation in a 3-deep dict, and a sample_evaluation is a 4-deep dict (if these have entrees).

openml/tasks/task.py
janvanrijn: Always classification

mfeurer: Maybe you want to make [the specification](https://www.openml.org/tt/3) more clear.

PR #269, opened 2017-06-15
Notebook

joaquinvanschoren:
To avoid issues later, I would like to merge this with develop as well


mfeurer:
@janvanrijn @joaquinvanschoren Should we enable the unit tests for notebooks again? Currently they're disabled because they took ~1min. As the unit tests got more expensive anyway, this one minute doesn't matter that much any more...

janvanrijn:
Having the notebook in the repo gives us the disadvantage that we need to 'maintain' the notebook as well, for that reason enabling the unit test seems like a good idea

PR #271, opened 2017-06-19
Evaluationlist

janvanrijn:
Support for evaluation list fn


mfeurer:
Why is PR #269 part of this PR?

openml/evaluations/functions.py
```diff
@@ -0,0 +1,93 @@
+import xmltodict
+
+from .._api_calls import _perform_api_call
+from ..evaluations import OpenMLEvaluation
+
+def list_evaluations(function, offset=None, size=None, id=None, task=None, setup=None,
+              flow=None, uploader=None, tag=None):
+    """List all run-evaluation pairs matching all of the given filters.
+
+        Perform API call `/evaluation/function{function}/{filters}
```
mfeurer: Shouldn't this be:

    ``/evaluation/function/{function}/{filters}``

?

Also, it would be great to add to the docstring a note where the user can find a list of _functions_.

openml/evaluations/evaluation.py
```diff
@@ -0,0 +1,18 @@
+
+class OpenMLEvaluation(object):
+
+    def __init__(self, run_id, task_id, setup_id, flow_id, flow_name,
```
mfeurer: Could you please add a docstring describing the class. Otherwise one needs to guess what `value` and `array_data` are.

PR #272, opened 2017-06-19
Study

janvanrijn:


mfeurer:
Why are PR #271 and #269 part of this PR?

mfeurer:
I'm struggling to understand what functionality this PR adds and what to use it for. Could you please add a brief explanation of that to this PR?

janvanrijn:
Sure will add that. But would be great if you can accept the pr's
nonetheless before you leave ;)

janvanrijn:
You are right, I was sloppy with documentation. Added documentation to the study and evaluation object, and also to the get study function.

Additionally (this should be documented on openml and not in this api): get study gives a list of items beyond the tag name. According to #339, a study is more comprehensive than just all items that are tagged with "study_x" (not implemented in webinterface, but implemented in api). This function accounts for this and enables users to download all related items from a study (as the listing functions do not suffice for this)

PR #277, opened 2017-06-23
[WIP] Bump version + better version handling

mfeurer:
This PR bumps the version number so new features can only be used with a version number higher than the version number used for the benchmark study. Furthermore, it fixes how the version number is set in the `setup.py` file.


mfeurer:
Yes, this PR automatically installs liac-arff so you don't have to do it manually any more.

PR #278, opened 2017-07-04
initial commit for setup list

janvanrijn:
added a function to obtain setup lists

### DIFF
--- a/openml/flows/sklearn_converter.py
+++ b/openml/flows/sklearn_converter.py
@@ -533,7 +533,7 @@ def _serialize_cross_validator(o):
     for key in args:
         # We need deprecation warnings to always be on in order to
         # catch deprecated param values.
-        # This is set in utils/__init__.py but it gets overwritten
+        # This is set in utils/__init__.py.py.py but it gets overwritten
         # when running under python3 somehow.
         warnings.simplefilter("always", DeprecationWarning)
         try:
--- a/openml/flows/sklearn_converter.py
+++ b/openml/flows/sklearn_converter.py
@@ -533,7 +533,7 @@ def _serialize_cross_validator(o):
     for key in args:
         # We need deprecation warnings to always be on in order to
         # catch deprecated param values.
-        # This is set in utils/__init__.py.py.py but it gets overwritten
+        # This is set in utils/__init__.py but it gets overwritten
         # when running under python3 somehow.
         warnings.simplefilter("always", DeprecationWarning)
         try:
--- a/openml/setups/functions.py
+++ b/openml/setups/functions.py
@@ -74,7 +74,7 @@ def get_setup(setup_id):
     return _create_setup_from_xml(result_dict)
 
 
-def list_setups(flow=None, tag=None, offset=None, size=None):
+def list_setups(flow=None, tag=None, setup=None, offset=None, size=None):
     """List all setups matching all of the given filters.
 
         Perform API call `/setup/list/{filters}
@@ -84,6 +84,8 @@ def list_setups(flow=None, tag=None, offset=None, size=None):
         flow : int, optional
 
         tag : str, optional
+
+        setup : list(int), optional
 
         offset : int, optional
 
@@ -100,6 +102,8 @@ def list_setups(flow=None, tag=None, offset=None, size=None):
         api_call += "/offset/%d" % int(offset)
     if size is not None:
         api_call += "/limit/%d" % int(size)
+    if size is not None:
+        api_call += "/setup/%s" % ','.join([str(int(i)) for i in setup])
     if flow is not None:
         api_call += "/flow/%s" % flow
     if tag is not None:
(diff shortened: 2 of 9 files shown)


openml/flows/sklearn_converter.py
```diff
@@ -533,7 +533,7 @@ def _serialize_cross_validator(o):
     for key in args:
         # We need deprecation warnings to always be on in order to
         # catch deprecated param values.
-        # This is set in utils/__init__.py but it gets overwritten
+        # This is set in utils/__init__.py.py.py but it gets overwritten
```
mfeurer: Why do you add `.py.py`?

openml/setups/functions.py
```diff
@@ -74,6 +74,76 @@ def get_setup(setup_id):
     return _create_setup_from_xml(result_dict)
 
 
+def setup_list(flow=None, tag=None, offset=None, size=None):
```
mfeurer: Wouldn't `list_setups` be a better name?

openml/setups/functions.py
```diff
@@ -74,6 +74,76 @@ def get_setup(setup_id):
...
+    xml_string = openml._api_calls._perform_api_call(api_call)
+
+    setups_dict = xmltodict.parse(xml_string)
+    # Minimalistic check if the XML is useful
+    if 'oml:setups' not in setups_dict:
+        raise ValueError('Error in return XML, does not contain "oml:setups": %s'
+                         % str(setups_dict))
+    elif '@xmlns:oml' not in setups_dict['oml:setups']:
+        raise ValueError('Error in return XML, does not contain '
+                         '"oml:runs"/@xmlns:oml: %s'
```
mfeurer: The error message looks like there is a copy-paste error as it references runs.

openml/setups/functions.py
```diff
@@ -74,6 +74,76 @@ def get_setup(setup_id):
...
+                         % str(setups_dict))
+    elif setups_dict['oml:setups']['@xmlns:oml'] != 'http://openml.org/openml':
+        raise ValueError('Error in return XML, value of  '
+                         '"oml:runs"/@xmlns:oml is not '
```
mfeurer: Same here.

openml/setups/functions.py
```diff
@@ -74,6 +74,76 @@ def get_setup(setup_id):
...
+        setups_list = setups_dict['oml:setups']['oml:setup']
+    elif isinstance(setups_dict['oml:setups']['oml:setup'], dict):
+        setups_list = [setups_dict['oml:setups']['oml:setup']]
+    else:
+        raise TypeError()
+
+    setups = dict()
+    for setup_ in setups_list:
+        # making it a dict to give it the right format
+        current = _create_setup_from_xml({'oml:setup_parameters': setup_})
```
mfeurer: This seems to be inconsistent with the other list functions which only return a describing dictionary. Is this already like this on the API side?

openml/study/functions.py
```diff
@@ -27,9 +27,11 @@ def get_study(study_id):
     creator = result_dict['oml:creator']
     tags = []
     for tag in _multitag_to_list(result_dict, 'oml:tag'):
-        tags.append({'name': tag['oml:name'],
-                     'window_start': tag['oml:window_start'],
-                     'write_access': tag['oml:write_access']})
+        current_tag = {'name': tag['oml:name'],
+                       'write_access': tag['oml:write_access']}
```
mfeurer: Hm, what is `write_access` and `window_start`?

openml/setups/functions.py
janvanrijn: In what sense inconsitent?

openml/study/functions.py
janvanrijn: the tags that are associated with a study determine which entities belong to the study. window start and write_access are both specifying this in more detail

janvanrijn:
Agreed on almost all points , now fixed

openml/setups/functions.py
mfeurer: Inconsistent in the sense that other list functions return a dictionary mapping from the ID to another dictionary describing the dataset/task etc. Here, the list function returns a dict which maps from ID to a setup object. (Check [here](https://github.com/openml/openml-python/blob/master/openml/datasets/functions.py#L138) for the dataset list function.)

mfeurer:
Hm, it seems like there's something messed up with the imports (looks like some form of cyclic import).

openml/setups/functions.py
janvanrijn: I see your point, but in this sense this is more rich, right? 

This listing function returns all information available, whereas other functions show just a summary of the function. There is no point in throwing this away

mfeurer: Okay, makes sense.

openml/study/functions.py
mfeurer: Hm, is this documented somewhere? I never saw this before and still don't really know what this additional information is good for.

PR #279, opened 2017-07-07
Get local evaluations

janvanrijn:
[this PR comes after 'setup list', as there are some changes in setup list that are not in this one]

Allows to obtain evaluation measures from an (unpublished) run by means of sklearn functions

### DIFF
--- a/openml/runs/run.py
+++ b/openml/runs/run.py
@@ -127,7 +127,9 @@ class OpenMLRun(object):
         if self.data_content is not None:
             predictions_arff = self._generate_arff_dict()
         elif 'predictions' in self.output_files:
-            raise ValueError('Not Implemented Yet: Function can currently only be used on locally executed runs (contributor needed!)')
+            predictions_file_url = _file_id_to_url(self.output_files['predictions'], 'predictions.arff')
+            predictions_arff = arff.loads(openml._api_calls._read_url(predictions_file_url))
+            # TODO: make this a stream reader
         else:
             raise ValueError('Run should have been locally executed.')
 
--- a/openml/runs/run.py
+++ b/openml/runs/run.py
@@ -2,6 +2,7 @@ from collections import OrderedDict, defaultdict
 import json
 import sys
 import time
+import numpy as np
 
 import arff
 import xmltodict
@@ -181,7 +182,7 @@ class OpenMLRun(object):
                 y_pred = values_predict[rep][fold][last_sample]
                 y_true = values_correct[rep][fold][last_sample]
                 scores.append(sklearn_fn(y_true, y_pred, **kwargs))
-        return scores
+        return np.array(scores)
 
     def publish(self):
         """Publish a run to the OpenML server.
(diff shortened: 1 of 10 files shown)


openml/runs/run.py
```diff
@@ -106,6 +107,83 @@ def _generate_trace_arff_dict(self):
 
         return arff_dict
 
+    def get_metric_score(self, sklearn_fn, kwargs={}):
+        '''Calculates metric scores based on predicted values. Assumes the
+        run has been executed locally (and contans run_data). Furthermore,
+        it assumes that the 'correct' field has been set (which is
```
mfeurer: What is the `correct` field? Where is it set and what does it imply?

openml/runs/run.py
```diff
@@ -106,6 +107,83 @@ def _generate_trace_arff_dict(self):
 
         return arff_dict
 
+    def get_metric_score(self, sklearn_fn, kwargs={}):
```
mfeurer: What about renaming this to `metric_fn`?

tests/test_runs/test_run_functions.py
```diff
@@ -385,6 +413,40 @@ def test_initialize_cv_from_run(self):
...
+        # invoke OpenML run
+        run = openml.runs.run_model_on_task(task, clf)
+
+        # compare with the scores in user defined measures
+        accuracy_scores_provided = []
+        for rep in run.fold_evaluations['predictive_accuracy'].keys():
+            for fold in run.fold_evaluations['predictive_accuracy'][rep].keys():
+                accuracy_scores_provided.append(run.fold_evaluations['predictive_accuracy'][rep][fold])
+        accuracy_scores = run.get_metric_score(sklearn.metrics.accuracy_score)
+        self.assertEquals(sum(accuracy_scores_provided), sum(accuracy_scores))
```
mfeurer: You could use [numpy-specific test-functions](https://docs.scipy.org/doc/numpy/reference/generated/numpy.testing.assert_array_almost_equal.html#numpy.testing.assert_array_almost_equal) here to make the code more readable.

openml/runs/run.py
```diff
@@ -106,6 +107,83 @@ def _generate_trace_arff_dict(self):
...
+        else:
+            raise ValueError('Run should have been locally executed.')
+
+        def _attribute_list_to_dict(attribute_list):
+            # convenience function
+            res = dict()
+            for idx in range(len(attribute_list)):
+                res[attribute_list[idx][0]] = idx
+            return res
+        attribute_dict = _attribute_list_to_dict(predictions_arff['attributes'])
```
mfeurer: It took my a while to figure out what you're doing here. Could you please add a comment like: `Create a mapping to map from the name of attributes present in the arff prediction file to their index. This is necessary because the number of classes can be different for different tasks.`?

openml/runs/run.py
janvanrijn: done

openml/runs/run.py
janvanrijn: specified in comments

PR #286, opened 2017-07-31
Added a function to automatically loop over paged calls

joaquinvanschoren:


openml/utils.py
```diff
@@ -37,4 +37,33 @@ def extract_xml_tags(xml_tag_name, node, allow_none=True):
             return None
         else:
             raise ValueError("Could not find tag '%s' in node '%s'" %
-                             (xml_tag_name, str(node)))
\ No newline at end of file
+                             (xml_tag_name, str(node)))
+            
+def list_all(listing_call, *args, **filters):
+    """Helper to handle paged listing requests.
+    Example usage: evaluations = list_all(list_evaluations, "predictive_accuracy", task=mytask)
```
mfeurer: Please format like this:

    Example usage: ``evaluations = list_all(list_evaluations, "predictive_accuracy", task=mytask)``

This will make sure that if we render API documentation in the future, this will be properly displayed as code.

PR #290, opened 2017-08-11
fixes bug re custom model selection

janvanrijn:


openml/flows/sklearn_converter.py
```diff
@@ -586,10 +586,13 @@ def check(param_dict, disallow_parameter=False):
         elif isinstance(model, sklearn.model_selection.RandomizedSearchCV):
             param_distributions = model.param_distributions
         else:
+            if hasattr(model, 'param_distributions'):
+                param_distributions = model.param_distributions
+            else:
+                raise AttributeError('Using subclass BaseSearchCV other than {GridSearchCV, RandomizedSearchCV}. Could not find attribute param_distributions. ')
```
mfeurer: I think it should be `Using a subclass of BaseSearchCV other than ...`.

PR #300, opened 2017-08-18
Initializes sklearn object from flow with default hyperparam configuration

janvanrijn:
@mfeurer I am really wondering what you think of the implementation. There is no hurry with this feature


mfeurer:
Hey, I'd be interested in the rational behind this before actually looking at the implementation.

janvanrijn:
Sklearn has for most hyperparameters some pretty good defaults, which can be a topic of research (e.g., answering the question, how much better are we than the default hyperparameter configuration). This function allows to reinitialize an OpenML flow with these defaults. 

The difficulty lies in hyperparameters that do not have any defaults (e.g., the steps of a pipeline). These need to be inferred from the flowname.

mfeurer:
Okay, so instead of creating the flow via a pipeline with the appropriate hyperparameter configuration, you want to download  an existing flow and alter its hyperparameter configuration?

mfeurer:
Really interesting. If you resolve the merge conflict and update the documentation, I think we can merge this.

janvanrijn:
Thanks for checking this. 

I predict that it will work after my most recent commit. 

> update the documentation

The undocumented part of the function is the 'components' part in kwargs. TBH I have no clue what this does again, and what the rationale is (this function was undocumented before). Can you add it?

Issue #174, opened 2016-09-09
Allow evaluation of models which do not support predict_proba


mfeurer:

PGijsbers:
The merge should have solved the issue.

PR #318, opened 2017-10-04
Fix174

janvanrijn:
Attempts to fix issue #174

### DIFF
--- a/tests/test_runs/test_run_functions.py
+++ b/tests/test_runs/test_run_functions.py
@@ -716,12 +716,6 @@ class TestRun(TestBase):
         num_folds = 10
         num_repeats = 1
 
-        clf = SGDClassifier(loss='hinge', random_state=1)
-        self.assertRaisesRegexp(AttributeError,
-                                "probability estimates are not available for loss='hinge'",
-                                openml.runs.functions._run_task_get_arffcontent,
-                                clf, task, class_labels)
-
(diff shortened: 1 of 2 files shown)


tests/test_runs/test_run_functions.py
```diff
@@ -898,3 +901,25 @@ def test_run_on_dataset_with_missing_labels(self):
...
+            ])
+            clf2 = sklearn.pipeline.Pipeline(steps=[
+                ('imputer', sklearn.preprocessing.Imputer()), ('estimator', HardNaiveBayes())
+            ])
+
+            arff_content1, arff_header1, _, _, _ = _run_task_get_arffcontent(clf1, task, task.class_labels)
+            arff_content2, arff_header2, _, _, _ = _run_task_get_arffcontent(clf2, task, task.class_labels)
+
+            # verifies last two arff indices (predict and correct)
+            # TODO: programmatically check wether these are indeed features (predict, correct)
```
mfeurer: I'm not sure what you want to check here, but with the code below you only select one column. From looking at a random predictions arff file, I'm asking myself at the moment why you only want to compare a single column. Can't you just check whether both arff contents are equal?

janvanrijn: We can't, because one will have probabilities per class, the other will have hard predictions as probabilities (i.e., just ones and zeros).

The last two columns are predictions and the correct value, respectively. I changed the slicing.

PR #329, opened 2017-10-09
Versionstring fix

### LINKED ISSUE #328: DEPENDENCY_PATTERN doesn't implement pep 440
The proper definition of version strings is here:
https://www.python.org/dev/peps/pep-0440/#developmental-releases

I have scikit-learn 0.20.dev0 which makes the tests fail real hard (I think because of the trailing zero).

amueller:
Fixes #328.


amueller:
I find it weird to define our own regex. I'm sure it's still wrong. Can't we use a standard library for this? I'm sure setuptools could do that for us.

PR #335, opened 2017-10-09
Fix #334: always create a list of tasks from the tasks API call, even…

### LINKED ISSUE #334: openml.tasks.list_tasks() fails when loading a single task with size=1
When running `oml.tasks.list_tasks()` with `size=1`, a `TypeError` is always raised.

**To reproduce:**
```
task_list = openml.tasks.list_tasks(size=1)
```
yields
```
TypeError                                 Traceback (most recent call last)
<ipython-input-6-d00af9922e60> in <module>()
----> 1 task_list = openml.tasks.list_tasks(size=1) # Get first 5000 tasks
      2 
      3 mytasks = pd.DataFrame.from_dict(task_list, orient='index')
      4 mytasks = mytasks[['tid','did','name','task_type','estimation_procedure','evaluation_measures']]
      5 print("First 5 of %s tasks:" % len(mytasks))

[..]/openml-python/openml/tasks/functions.py in list_tasks(task_type_id, offset, size, tag)
    123         api_call += "/tag/%s" % tag
    124 
--> 125     return _list_tasks(api_call)
    126 
    127 

[..]/openml-python/openml/tasks/functions.py in _list_tasks(api_call)
    150         proc_dict = dict((x['id'], x) for x in procs)
    151         for task_ in tasks_dict['oml:tasks']['oml:task']:
--> 152             tid = int(task_['oml:task_id'])
    153             task = {'tid': tid,
    154                     'ttid': int(task_['oml:task_type_id']),

TypeError: string indices must be integers
```

**Cause:**
The cause of this is that, when `size=1` is passed to `oml.tasks.list_tasks()`, an API call will be made requesting a single task. This will cause the returned XML string to contain a single `oml:task` element, which in turn will cause `xmltodict` to make the tasks children non-iterable.

engelen:
… if there is just a single task

### DIFF
--- a/openml/tasks/functions.py
+++ b/openml/tasks/functions.py
@@ -129,7 +129,7 @@ def _list_tasks(api_call):
     xml_string = _perform_api_call(api_call)
     with open('/tmp/list_tasks.xml', 'w') as fh:
         fh.write(xml_string)
-    tasks_dict = xmltodict.parse(xml_string)
+    tasks_dict = xmltodict.parse(xml_string, force_list=('oml:task'))
     # Minimalistic check if the XML is useful
     if 'oml:tasks' not in tasks_dict:
         raise ValueError('Error in return XML, does not contain "oml:runs": %s'


mfeurer:
Thanks, this makes total sense. Could you please do the following:

* redirect to the development branch
* check whether this could be applied to other listing calls? - maybe we can get rid of some if-statements?

engelen:
Due to some unintended mixup of branches, I've moved this into a new pull request.

PR #338, opened 2017-10-09
Add task rollback if task download fails

### LINKED ISSUE #302: cache mechanism can become corrupt
Today, I had the following problem. When loading a task it kept on showing errors. Further inspection led me to the conclusion that this was due to the fact that the task arrived corrupt in the cache. 

We should try to fix this a little bit better, as this is hard to reproduce and due to the error message almost impossible to debug.

mfeurer:
This PR adds a rollback if task download fails (fixes #302).

### DIFF
--- a/openml/tasks/functions.py
+++ b/openml/tasks/functions.py
@@ -283,11 +283,12 @@ def _create_task_cache_directory(task_id):
     task_cache_dir = os.path.join(
         config.get_cache_directory(), "tasks", str(task_id)
     )
-    try:
-        os.makedirs(task_cache_dir)
-    except (OSError, IOError):
-        # TODO add debug information!
+    if os.path.exists(task_cache_dir) and os.path.isdir(task_cache_dir):
         pass
+    elif os.path.exists(task_cache_dir) and not os.path.isdir(task_cache_dir):
+        raise ValueError('Task cache dir exists but is not a directory!')
+    else:
+        os.makedirs(task_cache_dir)
     return task_cache_dir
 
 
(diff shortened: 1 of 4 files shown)


openml/tasks/functions.py
```diff
@@ -224,36 +225,90 @@ def get_task(task_id):
...
+def _remove_task_cache_dir(tid_cache_dir):
+    """Remove the task cache directory
+
+    This function is NOT thread/multiprocessing safe.
+
+    Parameters
+    ----------
+    """
+    try:
+        os.rmdir(tid_cache_dir)
```
amueller: why two different tries? Why not just shutil?

openml/tasks/functions.py
```diff
@@ -224,36 +225,90 @@ def get_task(task_id):
...
+        Path of the created dataset cache directory.
+    """
+    task_cache_dir = os.path.join(
+        config.get_cache_directory(), "tasks", str(task_id)
+    )
+    try:
+        os.makedirs(task_cache_dir)
+    except (OSError, IOError):
+        # TODO add debug information!
+        pass
```
amueller: why should this pass? That seems like a severe error! (unless it already exists, but we're not testing that).

openml/tasks/functions.py
```diff
@@ -224,36 +225,90 @@ def get_task(task_id):
...
+    tid_cache_dir = _create_task_cache_dir(task_id)
 
     with lockutils.external_lock(
             name='datasets.functions.get_dataset:%d' % task_id,
             lock_path=os.path.join(config.get_cache_directory(), 'locks'),
     ):
         try:
-            with io.open(xml_file, encoding='utf8') as fh:
-                task = _create_task_from_xml(fh.read())
+            task = _get_task_description(task_id)
```
amueller: I don't get why the logic changed so much here...

mfeurer: To make it look like getting a dataset. Things are now in their own little functions.

amueller: hm ok. I can try to go through the details. I mean it looks ok but I haven't looked through the details.

openml/tasks/functions.py
mfeurer: Historic reasons... I changed this, and also changed this for the dataset functions.

openml/tasks/functions.py
mfeurer: You're right. I just changed this, also for the dataset functions.

openml/tasks/functions.py
mfeurer: The tests are passing and I just shuffled some lines, so don't think there's something weird going on here.

PR #340, opened 2017-10-09
Add unit test for loading non-sklearn flows, fixes #218

### LINKED ISSUE #218: Function get_flow to work with other flows
Currently, it seems to immediatelly deserialize a flow. However, this way it fails on downloading flows from other tools. Although the user will not be able to run it, he should be able to inspect it.

mfeurer:
Issue #218 was actually fixed in #253, however, #253 didn't add a unit test. This PR adds the missing unit test.

### DIFF
--- a/openml/runs/functions.py
+++ b/openml/runs/functions.py
@@ -677,10 +677,18 @@ def _create_run_from_xml(xml):
                          'description XML' % run_id)
 
     if 'predictions' not in files:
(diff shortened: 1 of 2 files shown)


mfeurer:
No, they're not. But I don't know them and I don't know what happens in case of them. I only added this specific error message because I realized that it could be improved in that special case.

Issue #197, opened 2017-02-23
get_data rounds off REAL values


joaquinvanschoren:
If you run this:
```
liver = oml.datasets.get_dataset(8) # Download Liver-disorders data
X, y = liver.get_data(target=liver.default_target_attribute);
y
```
You'll see that all targets y are integers. 
The actual target is a REAL number, and gets rounded!

mfeurer:
Try `X, y = liver.get_data(target=liver.default_target_attribute, target_dtype=float)`. In the long run I suggest that you use tasks in order to retrieve the datasets you're looking at. Tasks usually handle this themselves.

joaquinvanschoren:
Thanks! Would be good to add that to the documentation, though?

mfeurer:
How about making this function private, and only exposing this correctly via the task object? People should rarely have to access raw data without a task.

Thoughts @joaquinvanschoren @janvanrijn @amueller

janvanrijn:
Agreed.

2017-10-09 17:44 GMT+02:00 Matthias Feurer <notifications@github.com>:

amueller:
closed by #352

Issue #134, opened 2016-06-23
make task the standard interface to work with, not dataset


amueller:

amueller:
closed by #352

PR #352, opened 2017-10-12
Make task the default concept to work with

mfeurer:
This PR fixes:

* #197 by not automatically casting data on the dataset level
* #134 by rewriting the user guide.

### DIFF
--- a/doc/usage.rst
+++ b/doc/usage.rst
@@ -31,7 +31,6 @@ programmatically after loading the package:
 .. code:: python
 
     >>> import openml
-
     >>> apikey = 'Your API key'
     >>> openml.config.apikey = apikey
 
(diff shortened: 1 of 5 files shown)


openml/datasets/dataset.py
```diff
@@ -242,6 +242,12 @@ def get_data(self, target=None, target_dtype=int, include_row_id=False,
         else:
             if isinstance(target, six.string_types):
                 target = [target]
+            legal_target_types = (int, float)
```
amueller: Is float float64? Why not float32? And why do we require to cast?

doc/usage.rst
```diff
@@ -64,191 +63,41 @@ Config file:
...
-When you have to retrieve several datasets, you can use the convenience function
-:meth:`openml.datasets.get_datasets()`, which downloads all datasets given by
-a list of IDs:
-
-    >>> ids = [12, 14, 16, 18, 20, 22]
-    >>> datasets = openml.datasets.get_datasets(ids)
-    >>> print(datasets[0].name)
-    mfeat-factors
+OpenML contains several key concepts which it needs to make machine learning
+research shareable. A machine learning experiment consists of several runs,
```
amueller: I would say an ML experiment could also be a single run.

doc/usage.rst
```diff
@@ -64,191 +63,41 @@ Config file:
...
+which describe the performance of an algorithm (called a flow in OpenML) on a
+task. Task is the combination of a dataset, a split and an evaluation metric. In
```
amueller: "A task"

doc/usage.rst
```diff
@@ -64,191 +63,41 @@ Config file:
...
+this user guide we will go through listing and exploring existing tasks to
+actually running machine learning algorithms on them. In a further user guide
```
amueller: Maybe say "run is flow + setup + task and produces metric and predictions"? Right now you don't explain "run" right? Maybe make the key concepts bold.

doc/usage.rst
```diff
@@ -64,191 +63,41 @@ Config file:
...
-While datasets provide the most basic information for a machine learning task,
-they do not provide enough information for a reproducible machine learning
-experiment. A task defines how to split the dataset into a train and test set,
-whether to use several disjoint train and test splits (cross-validation) and
-whether this should be repeated several times. Also, the task defines a target
-metric for which a flow should be optimized.
+Tasks are containers, defining how to split the dataset into a train and test
+set, whether to use several disjoint train and test splits (cross-validation)
+and whether this should be repeated several times. Also, the task defines a
+target metric for which a flow should be optimized. You can think of a task as
```
amueller: I would make the "You can" sentence the first sentence. I think more essential is that the task defines which dataset to use, which column (if any) is the target and whether it's a classification, regression, clustering etc task.

doc/usage.rst
```diff
@@ -64,191 +63,41 @@ Config file:
...
+an experimentation protocol, describing how to apply a machine learning model
+to a dataset in a way that it is comparable with the results of others (more
+on how to do that further down).
 
-Just like datasets, tasks are identified by IDs and can be accessed in three
-different ways:
+Tasks are identified by IDs and can be accessed in two different ways:
```
amueller: Can we not filter by tags? Maybe I would say "you can explore tasks on the website or via ``list_tasks``. You can get a single task with ``get_task``". Because these two methods do semantically very different things.

doc/usage.rst
```diff
@@ -293,71 +134,55 @@ Let's find out more about the datasets:
 
 Now we can restrict the tasks to all tasks with the desired resampling strategy:
```
amueller: filtering by CV strategy seems a bit unnatural to me. Can we do it by dataset?

doc/usage.rst
```diff
@@ -293,71 +134,55 @@ Let's find out more about the datasets:
...
+Resampling strategies can be found on the `OpenML Website <http://www.openml.org/search?type=measure&q=estimation%20procedure>`_.
 
-A list of tasks, filtered tags, can be retrieved via:
+We can further filter the list of tasks to only contain datasets with more than
+500 samples, but less than 1000 samples:
 
 .. code:: python
 
-    >>> tasks = openml.tasks.list_tasks(tag='study_1')
+    >>> filtered_tasks = filtered_tasks.query('NumberOfInstances > 500 and NumberOfInstances < 1000')
```
amueller: Or just move this up, this seems more natural then the CV type to me? Or motivate the CV type?

doc/usage.rst
```diff
@@ -390,89 +215,22 @@ Properties of the task are stored as member variables:
...
+~~~~~~~~~~~~~
+Creating runs
+~~~~~~~~~~~~~
 
-Tasks and datasets allow us to download all information to run an experiment
-locally. In order to upload and share results of such an experiment we need
-the concepts of flows and runs.
+In order to upload and share results of running a machine learning algorithm
+on a task, we need to create an :class:`~openml.OpenMLRun`. A run object can
+be created by running a :class:`~openml.OpenMLFlow` or a scikit-learn model on
```
amueller: scikit-learn compatible?

PR #353, opened 2017-10-12
Cast data qualities to float

### LINKED ISSUE #345: cast dataset qualities to appropriate types
dataset qualities are currently all string, though NumberOfClasses etc should be integers.

mfeurer:
Fixes #345.


amueller:
I'd argue the xsd is wrong? Semantically they are not floats, right?

amueller:
I'd say this is better then before but still kinda stupid ;)

mfeurer:
Some qualities are floats, but some aren't. And except for writing down which to cast as int I wouldn't how to do this.

PR #356, opened 2017-10-12
tagging for other stuff

amueller:


openml/tasks/task.py
```diff
@@ -96,6 +96,28 @@ def get_split_dimensions(self):
 
         return self.split.repeats, self.split.folds, self.split.samples
 
+    def push_tag(self, tag):
+        """Annotates this flow with a tag on the server.
```
mfeurer: Should be `Annotates thi task``, right?

openml/tasks/task.py
```diff
@@ -96,6 +96,28 @@ def get_split_dimensions(self):
...
+
+        Parameters
+        ----------
+        tag : str
+            Tag to attach to the flow.
```
mfeurer: Same here.

openml/tasks/task.py
```diff
@@ -96,6 +96,28 @@ def get_split_dimensions(self):
...
+        """
+        data = {'flow_id': self.flow_id, 'tag': tag}
+        _perform_api_call("/flow/tag", data=data)
```
mfeurer: And here.

openml/tasks/task.py
```diff
@@ -96,6 +96,28 @@ def get_split_dimensions(self):
...
 
+    def push_tag(self, tag):
+        """Annotates this flow with a tag on the server.
+
+        Parameters
+        ----------
+        tag : str
+            Tag to attach to the flow.
+        """
+        data = {'flow_id': self.flow_id, 'tag': tag}
```
mfeurer: And here.

openml/tasks/task.py
```diff
@@ -96,6 +96,28 @@ def get_split_dimensions(self):
...
+        _perform_api_call("/flow/tag", data=data)
+
+    def remove_tag(self, tag):
+        """Removes a tag from this flow on the server.
```
mfeurer: And here.

openml/tasks/task.py
```diff
@@ -96,6 +96,28 @@ def get_split_dimensions(self):
...
+
+        Parameters
+        ----------
+        tag : str
+            Tag to attach to the flow.
```
mfeurer: And here.

openml/tasks/task.py
```diff
@@ -96,6 +96,28 @@ def get_split_dimensions(self):
...
+        """
+        data = {'flow_id': self.flow_id, 'tag': tag}
```
mfeurer: And here.

openml/tasks/task.py
```diff
@@ -96,6 +96,28 @@ def get_split_dimensions(self):
...
+        _perform_api_call("/flow/untag", data=data)
```
mfeurer: And here.

tests/test_flows/test_flow.py
```diff
@@ -67,6 +67,19 @@ def test_get_flow(self):
         self.assertEqual(subflow_3.parameters['L'], '-1')
         self.assertEqual(len(subflow_3.components), 0)
 
+    def test_tagging(self):
+        flow = openml.flows.get_flow(4024)
```
mfeurer: That's dangerous. If the server is purged, this flow will no longer exist. Maybe you should just query the server for a flow and take the first.

openml/tasks/task.py
amueller: looks like I missed one search and replace.

tests/test_flows/test_flow.py
amueller: This flow is used in another test, which is why I selected it ;) We can have the setup query for a flow and then use that everywhere, though.

amueller: Though the server was purged and it didn't fail, right?

mfeurer: But that flow is loaded from the production server.

mfeurer: Querying for a flow would be better.

amueller: lol it's actually using the production server in the other test...

amueller:
should be good now

PR #361, opened 2017-10-13
Fix keyerror in quality lookup

mfeurer:


openml/datasets/dataset.py
```diff
@@ -455,13 +455,14 @@ def _data_features_supported(self):
...
 
 
-
 def _check_qualities(qualities):
     if qualities is not None:
         qualities_ = {}
         for xmlquality in qualities:
             name = xmlquality['oml:name']
-            if xmlquality['oml:value'] is None:
+            if 'oml:value' not in xmlquality:
```
amueller: now about
```python
if xmlquality.get('oml:value', None) is None:
    value = float('NaN')
```

openml/datasets/dataset.py
```diff
@@ -455,13 +455,15 @@ def _data_features_supported(self):
...
 def _check_qualities(qualities):
     if qualities is not None:
         qualities_ = {}
         for xmlquality in qualities:
             name = xmlquality['oml:name']
-            if xmlquality['oml:value'] is None:
+
+            if xmlquality.get('oml:value', None) is None:
+                value = float('NaN')
+            elif 'oml:value' not in xmlquality:
```
amueller: you don't need  this line ;) that was kind of the point...

PR #363, opened 2017-10-13
String feature supports, use pandas dataframes

amueller:
This is a draft of adding string feature support via using pandas dataframes. I think using pandas would be nice to handle even float vs categorical etc. But that makes a lot of tests fail, in particular on dtypes.

After this change, columns can have very different dtypes. For example all null columns now have object dtype. Do we want to leave it like that?

### DIFF
--- a/openml/datasets/dataset.py
+++ b/openml/datasets/dataset.py
@@ -196,20 +196,44 @@ class OpenMLDataset(object):
             with io.open(filename, encoding='utf8') as fh:
                 return decode_arff(fh)
 
-    def get_data(self, target=None,
-                 include_row_id=False,
+    def get_data(self, target=None, include_row_id=False,
                  include_ignore_attributes=False,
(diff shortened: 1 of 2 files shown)


mfeurer:
This looks very promising. Can I somehow help here with comments etc?

amueller:
Ugh, I forgot this wasn't done yet. Ping me? Take over? hrm...

mfeurer:
I don't think anyone if Freiburg has capacities to take this over at the moment. I'm afraid it has to wait some time...

joaquinvanschoren:
Hi all. Gael just reminded me that this is still open. Do we have some bandwidth now to look at this?

mfeurer:
@glemaitre was working on this in #548 and wanted to continue on this, too.

mfeurer:
Should be superseded by #548. Please reopen if this is not the case.

PR #365, opened 2017-10-15
Cheatsheet

joaquinvanschoren:
This is the cheat sheet used for the hackathon


mfeurer:
I just added the unit tests for the cheat sheet to make sure it stays up-to-date.

joaquinvanschoren:
I couldn't find a way to build the PDF automatically. There is a `bug` in nbconvert in that it ignores anything done by notebook extensions, and I used a notebook extension to create the 2-column layout. So the only way to create the PDF is to install the notebook extension, load the notebook, end then print and save as PDF. I can remove the PDF but then I'll need to host it elsewhere since people won't go through all these steps.

I can tell you how to do a normal nbconvert, but then you loose the layout (all cells will just be below each other).

amueller:
I wouldn't call that a bug in nbconvert. nbconvert is independent of the notebook...

PGijsbers:
I propose we close this PR.
I had a look and the content of this notebook is by now superseded by the example gallery, which is up-to-date and covers the same topics.

PR #374, opened 2017-11-09
ensure features are represented nicely in jupyter / ipython, which ab…

amueller:
ipython/juptyer uses ``__repr__`` not ``__str__`` for displaying python object. But what we have for ``OpenMLDataFeature`` doesn't really adhere to ``__repr__`` specifications (i.e. the string used to generate the object). So I used ``_repr_pretty_`` which ipython/jupyter prefers over ``__repr__``.
Before:
image
After:
feature_repr


mfeurer:
Does this also nicely print outside of jupyter?

amueller:
it doesn't change anything outside of ipython. If you want to display something, usually you should use ``__str__``, in which case there is already a nice representation.

PR #376, opened 2017-11-16
requirements

janvanrijn:


amueller:
that's for testing only, right?

amueller:
Can't we skip the notebook tests if nbconvert is not installed?

PR #379, opened 2017-11-21
Modularizes the _run_task_get_arffcontent function

janvanrijn:
so we can access individual models of folds, and manipulate them

### DIFF
--- a/openml/runs/functions.py
+++ b/openml/runs/functions.py
@@ -99,7 +99,8 @@ def run_flow_on_task(task, flow, avoid_duplicate_runs=True, flow_tags=None,
     # execute the run
     res = _run_task_get_arffcontent(flow.model, task)
 
-    if not isinstance(flow.flow_id, int):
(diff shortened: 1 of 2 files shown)


openml/runs/functions.py
```diff
@@ -632,21 +720,23 @@ def _create_run_from_xml(xml):
...
-        if isinstance(run['oml:output_data']['oml:file'], dict):
-            # only one result.. probably due to an upload error
-            file_dict = run['oml:output_data']['oml:file']
-            files[file_dict['oml:name']] = int(file_dict['oml:file_id'])
-        elif isinstance(run['oml:output_data']['oml:file'], list):
-            # multiple files, the normal case
-            for file_dict in run['oml:output_data']['oml:file']:
+        output_data = run['oml:output_data']
+        if 'oml:file' in output_data:
+            if isinstance(output_data['oml:file'], dict):
```
mfeurer: You could use the `force_list` argument to the xml parsing here.

Issue #383, opened 2017-11-27
task list bug-> doctest failing


janvanrijn:
In case a task has only one input (clustering) we need to assume xmltodict will return a dict instead of list

janvanrijn:
edit: Only in Python 2.7 that doesn't like the task_.get(var, list()) struct

PR #384, opened 2017-11-27
Fix383

janvanrijn:
Fixes doc tests and error reported in #383

### DIFF
--- a/openml/tasks/functions.py
+++ b/openml/tasks/functions.py
@@ -176,7 +176,11 @@ def _list_tasks(api_call):
                     'status': task_['oml:status']}
 
             # Other task inputs
-            for input in task_.get('oml:input', list()):
+            task_inputs = task_.get('oml:input')
+            if isinstance(task_inputs, dict):
+                task_inputs = [task_inputs]
+
+            for input in task_inputs:
                 if input['@name'] == 'estimation_procedure':
                     task[input['@name']] = proc_dict[int(input['#text'])]['name']
                 else:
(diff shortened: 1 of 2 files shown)


openml/tasks/functions.py
```diff
@@ -176,7 +176,11 @@ def _list_tasks(api_call):
                     'status': task_['oml:status']}
 
             # Other task inputs
-            for input in task_.get('oml:input', list()):
+            task_inputs = task_.get('oml:input')
```
mfeurer: I'd prefer if you use the [force_list](https://github.com/openml/openml-python/blob/develop/openml/tasks/functions.py#L145) argument in order to avoid this if statement.

tests/test_tasks/test_task_functions.py
```diff
@@ -42,6 +42,12 @@ def test__get_estimation_procedure_list(self):
         self.assertIsInstance(estimation_procedures[0], dict)
         self.assertEqual(estimation_procedures[0]['task_type_id'], 1)
 
+
+    def test_list_clustering_task(self):
+        # as shown by #383, clustering tasks can give problems to server
+        openml.config.server = self.production_server
+        openml.tasks.list_tasks(task_type_id=5, size=10)
```
mfeurer: Hm, what's the expected outcome here? That is passes at all?

openml/tasks/functions.py
janvanrijn: Done.

tests/test_tasks/test_task_functions.py
janvanrijn: Added

PR #394, opened 2018-02-11
Setup caching

janvanrijn:
This PR contains three branches:
- Localruns (enables loading runs that are stored locally)
- Adds an independent tagging function
- corrects caching of runs, adds caching of setups


janvanrijn:
@mfeurer Could you have a look this week? Thanks!

openml/utils.py
```diff
@@ -40,6 +42,53 @@ def extract_xml_tags(xml_tag_name, node, allow_none=True):
         else:
             raise ValueError("Could not find tag '%s' in node '%s'" %
                              (xml_tag_name, str(node)))
+
+
+def _tag_entity(entity_type, entity_id, tag, untag=False):
+    """Abstract function that can be used as a partial for tagging entities
```
mfeurer: I do not see where this is used as a `partial` (assuming you mean `functools.partial`). Could you please clarify the docstring?

tests/test_setups/test_setup_functions.py
```diff
@@ -149,4 +149,14 @@ def test_setuplist_offset(self):
 
         all = set(setups.keys()).union(setups2.keys())
 
-        self.assertEqual(len(all), size * 2)
\ No newline at end of file
+        self.assertEqual(len(all), size * 2)
+
+    def test_get_cached_setup(self):
+        openml.config.set_cache_directory(self.static_cache_dir)
+        openml.setups.functions._get_cached_setup(1)
```
mfeurer: It would be great to have a test which checks that there's an exception for trying to access an uncached dataset.

openml/utils.py
```diff
@@ -40,6 +42,53 @@ def extract_xml_tags(xml_tag_name, node, allow_none=True):
...
+        tag : str
+            The tag
+
+        untag : bool
+            Set to true if needed to untag, rather than tag
+
+        Returns
+        -------
+        tags : list
+            List of tags that the entity is still tagged with
```
mfeurer: Maybe write `List of tags that the entity is (still) tagged with` because the number of tags could also increase, right?

tests/test_setups/test_setup_functions.py
janvanrijn: Don't see how this would be different from test_get_uncached_setup() ?

We now test the following cases:
* test_get_setup (downloads a setup and caches it)
* test_get_cached_setup (retrieves a setup from cache)
* test_get_uncached_setup (tries to retrieve an uncached setup, and throws an error)

openml/utils.py
janvanrijn: Agreed

openml/utils.py
janvanrijn: I haven't put these functions in yet. I will change the description, as this function can also be used standalone.

janvanrijn:
Localruns: a run object that is stored locally before uploading (hence it does not contain run id, uploader, upload_date, etc)

PR #399, opened 2018-02-19
fixes #373 + unit test

### LINKED ISSUE #373: supported types in runs/function.py, tuple as supported type missing
# model dependent attributes for trace arff
    for key in model.cv_results_:
        if key.startswith('param_'):
            # supported types should include all types, including bool, int float
            supported_types = (bool, int, float, six.string_types)
            if all(isinstance(i, supported_types) or i is None for i in model.cv_results_[key]):
                type = 'STRING'
            else:
                raise TypeError('Unsupported param type in param grid')

            # we renamed the attribute param to parameter, as this is a required
            # OpenML convention
            attribute = ("parameter_" + key[6:], type)
            trace_attributes.append(attribute)
    return trace_attributes

janvanrijn:


janvanrijn:
@mfeurer you had one concern while addressing this PR, however I cannot find it anywhere documented in emails or review. Do you still remember it? Then I will address it

PR #420, opened 2018-03-14
Improve error message

mfeurer:


openml/datasets/dataset.py
```diff
@@ -148,8 +148,10 @@ def remove_tag(self, tag):
     def __eq__(self, other):
         if type(other) != OpenMLDataset:
             return False
-        elif self.id == other._id or \
```
janvanrijn: Funny that this contained id all the time. This was a bug? dataset_id seems correct

PR #421, opened 2018-03-14
ADD unit test to ensure example listing

mfeurer:


tests/test_study/test_study_examples.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+class TestStudyFunctions(TestBase):
+    _multiprocess_can_split_ = True
+    """Test the example code of Bischl et al. (2018)"""
+
+    def test_Figure1a(self):
+        """Test listing in Figure 1a on a single task and the old OpenML100 study
+
+        import openml
```
janvanrijn: you sure want to copy the code into the comment?

tests/test_study/test_study_examples.py
```diff
@@ -0,0 +1,49 @@
...
+            task = openml.tasks.get_task(task_id)                        # download the OpenML task
+            X, y = task.get_X_and_y()                                    # get the data (not used in this example)
+            openml.config.apikey = 'FILL_IN_OPENML_API_KEY'              # set the OpenML Api Key
+            run = openml.runs.run_model_on_task(task,clf)                # run classifier on splits (requires API key)
+            score = run.get_metric_fn(sklearn.metrics.accuracy_score) # print accuracy score
+            print('Data set: %s; Accuracy: %0.2f' % (task.get_dataset().name,score.mean()))
+            run.publish()                                                # publish the experiment on OpenML (optional)
+            print('URL for run: %s/run/%d' %(openml.config.server,run.run_id))
+        """
+        import openml
```
janvanrijn: Wouldn't it be possible to somehow load this from the appropriate (paper) repo? Or don't you want these dependencies? Or should we add this test actually there?

tests/test_study/test_study_examples.py
mfeurer: That's the original code from the listing which is a bit different from the actual code of the unit test. So yes, I want this code there. I'll add a sentence about this.

tests/test_study/test_study_examples.py
mfeurer: I would like to avoid such dependencies such that users can run unit tests locally without having to install such dependencies which will most likely never make it to pypi.

Issue #358, opened 2017-10-13
implement paging (at least in list_runs)


amueller:
Right now it seems not to be possible to analyze the runs for any of the openml100 (that i tried)
```python
openml.runs.list_runs(task=[146607])
```
> OpenMLServerException: Too many results

I think we should implement paging in the python api

mfeurer:
Here you go: https://github.com/openml/openml-python/blob/master/openml/utils.py#L44

amueller:
cool. I think we should call this in all the ``list_x`` functions.

mfeurer:
@ArlindKadra just added this.

PR #426, opened 2018-03-25
Paging

ArlindKadra:
Incorporated paging in the list_*  functions #358. Refactored the code.

### DIFF
--- a/openml/datasets/functions.py
+++ b/openml/datasets/functions.py
@@ -8,6 +8,7 @@ import shutil
 from oslo_concurrency import lockutils
 import xmltodict
 
+import openml.utils
 from .dataset import OpenMLDataset
 from ..exceptions import OpenMLCacheException, OpenMLServerNoResult
 from .. import config
@@ -136,9 +137,10 @@ def _get_cached_dataset_arff(dataset_id):
         raise OpenMLCacheException("ARFF file for dataset id %d not "
                                    "cached" % dataset_id)
 
-
 def list_datasets(offset=None, size=None, status=None, **kwargs):
-    """Return a list of all dataset which are on OpenML.
+
+    """
+    Return a list of all dataset which are on OpenML.
 
     Parameters
     ----------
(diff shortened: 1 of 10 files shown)


openml/datasets/functions.py
```diff
@@ -169,29 +171,32 @@ def list_datasets(offset=None, size=None, status=None, **kwargs):
         If qualities are calculated for the dataset, some of
         these are also returned.
     """
-    api_call = "data/list"
-    if offset is not None:
-        api_call += "/offset/%d" % int(offset)
 
-    if size is not None:
-        api_call += "/limit/%d" % int(size)
+    return openml.utils.list_all(_list_datasets, offset=offset, size=size, status=status, **kwargs)
```
mfeurer: This looks to be >80 lines. Could you please try to stick to PEP8 as much as possible?

openml/datasets/functions.py
```diff
@@ -169,29 +171,32 @@ def list_datasets(offset=None, size=None, status=None, **kwargs):
...
-
-
-def _list_datasets(api_call):
+def __list_datasets(api_call):
     # TODO add proper error handling here!
     try:
         xml_string = _perform_api_call(api_call)
-    except OpenMLServerNoResult:
-        return dict()
+    except OpenMLServerNoResult as e:
```
mfeurer: Is there a reason why you changed this? I think returning a dict is more natural because it behaves like when listing an empty directory.

mfeurer: Also, it appears to me that this piece of code is repeated in all listing functions. Could you please put that code into the utils module?

openml/datasets/functions.py
ArlindKadra: True, I will keep the line length under 80. Regarding the arguments documentation, I kept the arguments as they were since we did not discuss any changes to them. In this case they can be included in the **kwargs as they are not positional and can all be set in one go. However, in other functions when we have lists, I would say we keep the list arguments separate from the primitives.

openml/datasets/functions.py
ArlindKadra: There is a reason. It breaks the way that the list method in the utils works. As an example, without a limit, the previous implementation would never throw an error and we would be stuck in a loop. 

In my opinion we should not change that, as we are performing checks and also actually creating the dict there. We can however remove the try and except as we are actually catching in the utils method.

ArlindKadra: Basically utils.list_all -> loops _list_* -> forms the api call and passes it to the __list_* function.
With the current flow it is not possible to do that.

mfeurer: Okay, got it, that makes sense. Thanks for removing the try/except, this should make the code a lot easier.

openml/datasets/functions.py
mfeurer: Hm, given the structure of the REST API and the fact that the usage of the method right now is really unclear, it would be good if revert function signature. Also, I think calling it `**filters` instead of `**kwargs` will make the whole thing more similar to the REST API.

PR #427, opened 2018-03-29
MAINT bump required liac-arff version

mfeurer:
New version is more resilient with respect to commas in data. However, it cannot yet read all commas, that's for the future.


mfeurer:
Yep...

Issue #396, opened 2018-02-11
caching (similar functions)


janvanrijn:
I occured to me that there are many functions that actually do the same: caching (runs, datasets, flows, etc)

Would be good to unify these functions and get rid of many maintainability things.

mfeurer:
Great idea! We should have a decorator to add caching to functions which only retrieve a single XML. I'm not sure whether we can actually generalize caching of datasets and tasks (the main objects) in a reasonable way.

PR #432, opened 2018-04-04
Have different cache directories for different servers

mfeurer:


mfeurer:
This now also partially solves #396. However, not as nicely as expected.

tests/test_setups/test_setup_functions.py
```diff
@@ -111,7 +111,7 @@ def test_existing_setup_exists_3(self):
 
     def test_get_setup(self):
         # no setups in default test server
-        openml.config.server = 'https://www.openml.org/api/v1/xml/'
+        openml.config.set_server_url('https://www.openml.org/api/v1/xml/')
```
janvanrijn: why is this still a fn?

openml/testing.py
```diff
@@ -66,8 +68,14 @@ def setUp(self):
 
     def tearDown(self):
         os.chdir(self.cwd)
-        shutil.rmtree(self.workdir)
+        for i in range(10):
```
janvanrijn: make 10 a variable?

ArlindKadra:
@mfeurer @janvanrijn

mfeurer:
Thanks for the fix @ArlindKadra. Could you please describe which change was actually necessary to fix it? Also, could you please explain the reasoning for changing the `static_cache_dir` variable?

Issue #411, opened 2018-03-02
Can't find cache dir on Mac OSX


enzolutions:
Similar to [#120] with an without relative paths

My config file have

```
cachedir = '/Users/enzo/.openml/cache/'
```

```
Traceback (most recent call last):
  File "script-1.py", line 1, in <module>
    import openml
  File "/Users/enzo/.pyenv/versions/miniconda3-4.3.30/lib/python3.6/site-packages/openml-0.6.0-py3.6.egg/openml/__init__.py", line 17, in <module>
    from . import config
  File "/Users/enzo/.pyenv/versions/miniconda3-4.3.30/lib/python3.6/site-packages/openml-0.6.0-py3.6.egg/openml/config.py", line 131, in <module>
    _setup()
  File "/Users/enzo/.pyenv/versions/miniconda3-4.3.30/lib/python3.6/site-packages/openml-0.6.0-py3.6.egg/openml/config.py", line 47, in _setup
    set_cache_directory(cache_dir)
  File "/Users/enzo/.pyenv/versions/miniconda3-4.3.30/lib/python3.6/site-packages/openml-0.6.0-py3.6.egg/openml/config.py", line 78, in set_cache_directory
    os.mkdir(dir_)
FileNotFoundError: [Errno 2] No such file or directory: "'/Users/enzo/.openml/cache/'"
```

mfeurer:
Could you please try if it work if you replace `os.mkdir` with `os.makedirs`?

enzolutions:
I wasn't using anything in my code, to be able to use cache I have to add the following line in my script

```
openml.config.set_cache_directory(os.path.expanduser('~/.openml/cache'))
```

mfeurer:
Thanks for reporting, this indeed seems to be a bug. We'll have a look into this.

mfeurer:
We added the `expanduser` command and this should work now (on the development branch). Thanks again for reporting this issue.

Issue #120, opened 2016-03-31
Can't find cache dir on Mac OSX


joaquinvanschoren:
I get the following error when initiating the connector, although the cache dir does exist.
I'm running this from a Jupyter Notebook, Python 3.

```
---> 10 connector = APIConnector(apikey = key)
 
/Users/joa/anaconda/lib/python3.5/site-packages/openml-0.2.1-py3.5.egg/openml/apiconnector.py in __init__(self, cache_directory, apikey, server, verbosity, configure_logger, private_directory)
    168                      self._private_directory_tasks]:
    169             if not os.path.exists(dir_) and not os.path.isdir(dir_):
--> 170                 os.mkdir(dir_)
    171 
    172     def _parse_config(self):

FileNotFoundError: [Errno 2] No such file or directory: '~/.openml/cache'
```

Is it possible that the home dir is not found with '~'?

joaquinvanschoren:
Fixed this by putting an absolute path in my config file, relative paths do not work. I guess we can close this for now.

PR #441, opened 2018-04-11
Adding support for home directory symbol

ArlindKadra:
Issue #411 was already fixed with #432. Added support for paths which contain the home symbol, based on #120.


ArlindKadra:
@mfeurer you can take a look at the small change.

PR #448, opened 2018-04-20
Fix unit tests

mfeurer:
This PR fixes a few unit tests and adds one which I forgot to check in quite a while ago. Most importantly, the return values to retrieve all runs for a combination of task/setup is now consistent. Furthermore, I think cleaning the test server from lots of runs was really important.

### DIFF
--- a/tests/test_runs/test_run_functions.py
+++ b/tests/test_runs/test_run_functions.py
@@ -527,10 +527,13 @@ class TestRun(TestBase):
             flow = openml.flows.sklearn_to_flow(clf)
             flow_exists = openml.flows.flow_exists(flow.name, flow.external_version)
             self.assertIsInstance(flow_exists, int)
+            self.assertGreater(flow_exists, 0)
             downloaded_flow = openml.flows.get_flow(flow_exists)
             setup_exists = openml.setups.setup_exists(downloaded_flow)
             self.assertIsInstance(setup_exists, int)
+            self.assertGreater(setup_exists, 0)
             run_ids = _run_exists(task.task_id, setup_exists)
+            self.assertGreater(len(run_ids), 0)
             run_id = random.choice(list(run_ids))
 
         # now the actual unit test ...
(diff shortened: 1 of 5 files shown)


tests/test_utils/test_utils.py
```diff
@@ -0,0 +1,18 @@
...
+
+    def test_list_all(self):
+        list_datasets = openml.datasets.functions._list_datasets
+        datasets = openml.utils.list_all(list_datasets)
+
+        self.assertGreaterEqual(len(datasets), 100)
+        for did in datasets:
+            self._check_dataset(datasets[did])
+
+        # TODO implement these tests
```
janvanrijn: Sure you didn't forget these?

ArlindKadra: you mean we should not verify the datasets?

janvanrijn: I mean there's still a TODO in the PR ;) 

(line 16)

ArlindKadra: Ah you are right, I did not see what you meant because of the lines selected. The TODO seems already implemented.

tests/test_runs/test_run_functions.py
```diff
@@ -527,10 +527,13 @@ def test_get_run_trace(self):
             flow = openml.flows.sklearn_to_flow(clf)
             flow_exists = openml.flows.flow_exists(flow.name, flow.external_version)
             self.assertIsInstance(flow_exists, int)
+            self.assertGreater(flow_exists, 0)
             downloaded_flow = openml.flows.get_flow(flow_exists)
             setup_exists = openml.setups.setup_exists(downloaded_flow)
             self.assertIsInstance(setup_exists, int)
+            self.assertGreater(setup_exists, 0)
```
ArlindKadra: Maybe we can just use assertGreater instead of also making use of assertIsInstance, since if the argument is False it will throw AttributeError: 'NoneType' object has no attribute 'fail'.

Issue #451, opened 2018-04-23
Confusing run_flow_on_task


joaquinvanschoren:
I've seen several people make the same error when using OpenML. They do
```
run_flow_on_task(flow, task)
```
instead of 
```
run_flow_on_task(task, flow)
```
I can see how this is confusing to beginners. Would it make sense to rename the method? Or at least add more info to the error message? Right now it returns with 
```
AttributeError: 'OpenMLTask' object has no attribute 'model'
```
This makes sense, but could we also add something like 'Please check whether you entered the task and flow objects in the right order'?

PGijsbers:
This is something I find myself doing wrong still, every now and then. What is the reasoning behind this order anyway?

I would suggest making sure the order in the method name is the same as the order of the parameters (no preference as to doing it by renaming the method or re-ordering the parameters). Alternatively you could also just allow both permutations of arguments (but this requires additional input validation and might not be as neat).

mfeurer:
There is no reasoning behind this order, it's sloppiness. I asked @ArlindKadra to do this once dataset uploading works. I'd actually prefer to allow both permutations. If @ArlindKadra didn't yet start on this, do you want to take this over @PGijsbers ?

PGijsbers:
In principle yes, but I'm also about to go on holiday. Considering it shouldn't be much work, I'll see if I can squeeze it in.

mfeurer:
Cool!

PGijsbers:
Do we want to phase out usage of the old ordering (e.g. give a deprecation warning for now)?

joaquinvanschoren:
Allowing both permutations and outputting a warning for the 'wrong' order seems a good way to go. Documentation should only document the more logical/consistent option IMHO.

Indeed, the same goes for `run_model_on_task`.

ArlindKadra:
@mfeurer I have not started on this, so there is no duplicate work happening :)

janvanrijn:
My personal preferences:
* I am highly in favor of strict type checking
* I don't know how exactly this will be implemented, but my feeling tells me that allowing a task as flow argument and vice versa, and inducing which one is which will lead to confusing and hard to maintain code, and potentially problems with experimental results 
* I don't care about argument order, so either of these is fine by me

My proposal would be to simply change the order of arguments (you seem to agree that it makes sense) and throw an error if someone uses the old order of arguments.

mfeurer:
Fixed in #452.

PGijsbers:
[...]

PR #452, opened 2018-04-23
Fix/451 (WIP?)

PGijsbers:
Worked on: #451 

* Functions `run_model_on_task` and `run_flow_on_task` now allow both ordering of the first two arguments ((task, model) or (model, task) and (task, flow) or (flow, task), respectively).

* Added two simple unit tests which test that the normal output with swapped arguments still works.

There are some outstanding questions, see my last comment in #451. I propose we continue the discussion there.

I made an early pull request to check with CI (I have some issues with my local environment atm and don't have the time to fix it).


PGijsbers:
note: AppveyorCI shouldn't work on this branch because it does not have the updated scripts.

PGijsbers:
Adding a TODO where? Open an issue or a `#TODO` in the code? Also, I can just add a deprecation warning right now instead. It doesn't really take an effort, but what should the text be? Something like
```
warnings.warn("The old argument order (task, model) is deprecated and will not be supported in the future"
              "Please use the order (model, task).", PendingDeprecationWarning)
```
whenever arguments are passed by the old order?

Again I'm very sorry this took so long. I learned not to take up a new issues right before a vacation because there's always a lot to catch up to when you get back anyway ^^;

mfeurer:
Yep, adding a `#TODO` as a comment was what I meant. And the deprecation warning would be great.

mfeurer:
Could you please resolve the conflicts and check the failing tests?

mfeurer:
Thanks! Test failures are due to missing tags on the test server. I will contact @janvanrijn about this.

Issue #457, opened 2018-04-25
Why does run_flow_or_task always upload the flow?


joaquinvanschoren:
The docs for 'run_flow_or_task' say:`In case a flow is not yet published, it is published after executing the run`.

I wonder why this is always done. I can imagine that people just want to run a set of flows on a task without uploading it later. Maybe they want to test very weird configurations that they will never upload runs for, and these will all live on the server as flows without runs.

Would it not make more sense to defer the uploading until a run is also uploaded? Hence, flow uploading would be done in run_publish if the flow is unknown to OpenML.

joaquinvanschoren:
Also, it makes it harder to run experiments without an internet connection? In some cases you want to run experiments without an internet connection, and just store the run objects for later uploading?

mfeurer:
Added via #516.

PR #459, opened 2018-04-30
added serialize run functionality

janvanrijn:
It allows run objects (including predictions and traces) to be serialized to disk, and reloaded. This functionality is almost used in all my projects, and I can imagine us using this for the benchmark study. 

Unit tests should pass, code should make sense, please check the unit test if run equality check is OK.


openml/runs/run.py
```diff
@@ -65,6 +66,49 @@ def __str__(self):
     def _repr_pretty_(self, pp, cycle):
         pp.text(str(self))
 
+    @classmethod
+    def from_filesystem(cls, folder):
```
mfeurer: Could you please add a docstring?

openml/runs/run.py
```diff
@@ -65,6 +66,49 @@ def __str__(self):
...
+
+        if os.path.isfile(trace_path):
+            with open(trace_path, 'r') as fp:
+                trace = arff.load(fp)
+                run.trace_attributes = trace['attributes']
+                run.trace_content = trace['data']
+
+        return run
+
+    def to_filesystem(self, output_directory):
```
mfeurer: Could you please add a docstring, here, too?

openml/runs/run.py
```diff
@@ -65,6 +66,49 @@ def __str__(self):
...
+        run_xml = self._create_description_xml()
+        predictions_arff = arff.dumps(self._generate_arff_dict())
+
+        with open(output_directory + '/description.xml', 'w') as f:
```
mfeurer: Could you please use `os.path.join` as above?

openml/runs/run.py
```diff
@@ -65,6 +66,49 @@ def __str__(self):
...
+            f.write(run_xml)
+        with open(output_directory + '/predictions.arff', 'w') as f:
```
mfeurer: Same here.

openml/runs/run.py
```diff
@@ -65,6 +66,49 @@ def __str__(self):
...
+            f.write(predictions_arff)
+
+        if self.trace_content is not None:
+            trace_arff = arff.dumps(self._generate_trace_arff_dict())
+            with open(output_directory + '/trace.arff', 'w') as f:
```
mfeurer: And here

tests/test_runs/test_run.py
```diff
@@ -66,3 +70,55 @@ def test_tagging(self):
...
+
+        task = openml.tasks.get_task(119)
+        run = openml.runs.run_model_on_task(task, model)
+
+        cache_path = os.path.join(self.workdir, 'runs', str(random.getrandbits(128)))
+        os.makedirs(cache_path)
+        run.to_filesystem(cache_path)
+
+        run_prime = openml.runs.OpenMLRun.from_filesystem(cache_path)
+        self._test_run_obj_equals(run, run_prime)
```
mfeurer: You should add a check here that the trace is available. The function `_test_run_obj_equals` does not guarantee this.

janvanrijn:
Agreed to all.

openml/runs/run.py
```diff
@@ -68,6 +69,21 @@ def _repr_pretty_(self, pp, cycle):
 
     @classmethod
     def from_filesystem(cls, folder):
+        """
+        The inverse of the to_filesystem method. Initiates a run based
```
mfeurer: I think that `initiate` is unfortunate wording here as it also means 'to start'. How about 'instantiates a run object'?

mfeurer:
I did not yet merge this to develop/master as you requested in my other PR because we can do this whenever we're ready to start up some experiments. I you need this badly we can merge this into master under a `0.7.1` version tag.

janvanrijn:
Sure. Where can I find the change log?

janvanrijn:
Ouch, I extended the unit tests with a publish statement, and apparently the model is always needed to be present in order to be able to upload. Don't know if we should really enforce this, as it seems just a sanity check, but I encountered some more discrepancies. Apparently the to/from xml functions upon which this relied did not work perfectly. 

I pushed a fix, also adding more checks to both the serialize/unserialize functions and unit tests. However, I have a feeling the Run to/from XML functionality could use some more extensive unit tests. Please have a critical look at this and feel free to extend.

mfeurer:
The changelog is a bit hidden and we need to re-launch it, but it's here: https://github.com/openml/openml-python/blob/master/doc/progress.rst.

Regarding the behavior of publish when no flow is present, I'm not sure if/how we can support this. Instead of having a flow in the run, we would keep the model in the run and change this once we want to upload the run to OpenML? @joaquinvanschoren also opened an issue about this: https://github.com/openml/openml-python/issues/457

janvanrijn:
This is different. The publish function obviously requires a flow id (which is why Joaquin opened the issue), but requires a model (which I didn't store). Not sure if we really should require this, but this is what I fixed in the last PR. 

Good to merge?

janvanrijn:
Almost forgot, but it's now in the change log 

Made an issue with some suggestions #465

mfeurer:
I think you forgot to push the commit.

janvanrijn:
I accidentally put it in the other pull request (listing)

PR #464, opened 2018-05-03
Fix listing

janvanrijn:
Fixed list_all function:
* changed handling of the catched error (more restrictive)
* changed handling of the number of results (required additional constant 'BATCH_SIZE_ORIG')
* indicated the constant variables by capitalized naming convention (LIMIT and BATCH_SIZE_ORIG)
* made it protected (naming convention to start with underscore)
* added unit tests for all individual functions

### DIFF
--- a/doc/progress.rst
+++ b/doc/progress.rst
@@ -9,6 +9,12 @@ Progress
 Changelog
 =========
 
+0.8.0
+~~~~~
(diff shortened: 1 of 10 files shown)


janvanrijn:
Agreed to all. 

Except that I don't know mock that well, and won't be able to invest too much time right now. I added this test to the server.

PR #467, opened 2018-05-05
pushed to fix serializing

janvanrijn:
Apparently, the previous serialization / deserialization breaks when using a random search with a non numeric hyperparameter in the grid. 

This is due to the serialization / deserialization in liac arff, which keeps on adding escape characters. As such, serializing once is OK, but deserializing and reserializing is a problem. 

Although I liked the previous approach (very clear and verbose) we can not rely on this, and I have created a new PR (with more extensive tests)

### DIFF
--- a/openml/runs/functions.py
+++ b/openml/runs/functions.py
@@ -391,11 +391,11 @@ def _run_task_get_arffcontent(model, task, add_local_measures):
     # this information is multiple times overwritten, but due to the ordering
     # of tne loops, eventually it contains the information based on the full
     # dataset size
-    user_defined_measures_per_fold = dict()
+    user_defined_measures_per_fold = collections.OrderedDict()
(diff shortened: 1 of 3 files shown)


mfeurer:
Sorry for the long waiting time. Is this still necessary if we fix #466 / https://github.com/renatopp/liac-arff/issues/69 ?

If it is no longer necessary, would you like to add some more unit tests to ensure that the data OpenML needs to handled is passed through liac-arff correctly?

janvanrijn:
With the aforementioned fix, this pr won't be necessary.

Haven't tested the fix, but if it works it is really awesome. I hated the way this pr handles serialization of runs

mfeurer:
As the [fix for liac-arff](https://github.com/renatopp/liac-arff/issues/69) appears to solve this issue, I close this PR for now.

Issue #431, opened 2018-04-04
Remove split pickling


mfeurer:
or fix it, but the pickle files are not used at the moment.

mfeurer:
This isn't a bug as the pickle files are actually used.

PR #470, opened 2018-05-15
Removing split pickling

ArlindKadra:
Issue #431

### DIFF
--- a/openml/tasks/split.py
+++ b/openml/tasks/split.py
@@ -127,10 +127,9 @@ class OpenMLSplit(object):
                             np.array(repetitions[repetition][fold][sample][0], dtype=np.int32),
                             np.array(repetitions[repetition][fold][sample][1], dtype=np.int32))
 
-            if cache:
-                with open(pkl_filename, "wb") as fh:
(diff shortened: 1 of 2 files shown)


openml/tasks/split.py
```diff
@@ -125,11 +125,12 @@ def _from_arff_file(cls, filename, cache=True):
                             np.array(repetitions[repetition][fold][sample][0], dtype=np.int32),
                             np.array(repetitions[repetition][fold][sample][1], dtype=np.int32))
 
+            '''
```
janvanrijn: PyCharm usually wants double quotes, right?

ArlindKadra:
@mfeurer since you mentioned removing or fixing split pickling, I only commented out the code as a simple solution. I can remove it completely if you are ok with it. @janvanrijn we are using double quotes for docstrings.

mfeurer:
I think removing it is fine. Reading those files isn't as costly as reading in the data.

ArlindKadra:
@mfeurer sure, but we will not have a way to disable loading from cache.

mfeurer:
I don't think we have this for anything else. As a split should never change I don't see a reason to keep this functionality.

PR #489, opened 2018-07-19
FIX replace if statement with assumptions by one without

### LINKED ISSUE #487: Liac-arff: Sparse ARFF / ARFF stored in same way
The following [line of code](https://github.com/openml/openml-python/blob/master/openml/datasets/dataset.py#L113) attempts to distinguish sparse arff from regular arff. 
(FFR, the openml-python code)
```
if isinstance(data['data'], tuple):
  X = data['data']
  X_shape = (max(X[1]) + 1, max(X[2]) + 1)
  X = scipy.sparse.coo_matrix((X[0], (X[1], X[2])), shape=X_shape, dtype=np.float32)
  X = X.tocsr()
elif isinstance(data['data'], list):
  X = np.array(data['data'], dtype=np.float32)
```

However, when executing the [following test](https://github.com/renatopp/liac-arff/blob/master/tests/test_loads.py) from the liac-arff reader, the sparse arff and regular arff are internally represented the same. 

Am I missing something or does this deprecate the line of code that checks for tuples, and stores them as csr matrices? 

@amueller relevant for sklearn openml loader

mfeurer:
#### Reference Issue
Fixes #487.

This PR replaces an if statement which assumes a specific return type from liac-arff with an if statement that is solely based on the specifics of a dataset.

Current unit tests.

### DIFF
--- a/openml/datasets/dataset.py
+++ b/openml/datasets/dataset.py
@@ -169,13 +169,13 @@ class OpenMLDataset(object):
                                    for name, type_ in data['attributes']]
                     attribute_names = [name for name, type_ in data['attributes']]
 
-                    if isinstance(data['data'], tuple):
+                    if format.lower() == 'sparse_arff':
                         X = data['data']
                         X_shape = (max(X[1]) + 1, max(X[2]) + 1)
                         X = scipy.sparse.coo_matrix(
                             (X[0], (X[1], X[2])), shape=X_shape, dtype=np.float32)
                         X = X.tocsr()
-                    elif isinstance(data['data'], list):
+                    elif format.lower() == 'arff':
                         X = np.array(data['data'], dtype=np.float32)
                     else:
                         raise Exception()


openml/datasets/dataset.py
```diff
@@ -169,13 +169,13 @@ def __init__(self, name, description, format, dataset_id=None,
                                    for name, type_ in data['attributes']]
                     attribute_names = [name for name, type_ in data['attributes']]
 
-                    if isinstance(data['data'], tuple):
+                    if format.lower() == 'sparse_arff':
```
janvanrijn: Is it formally defined that in case of 'dense' the format is stored as 'arff' and in case of spare the format is stored as 'sparse_arff'?

At least, I am not aware of this being tested. 

@joaquinvanschoren what do you think?

mfeurer: I thought that's what the field from OpenML is good for. Actually, the code is already depending on this (https://github.com/openml/openml-python/blob/781815ba036905e74c43f114f5d0e193435260d9/openml/datasets/dataset.py#L172).

PR #490, opened 2018-07-20
FIX store fold name in run

mfeurer:
This PR is required to make our benchmark study scripts run again. It should be tested by running the regular unit tests.

### DIFF
--- a/openml/runs/functions.py
+++ b/openml/runs/functions.py
@@ -123,8 +123,14 @@ def run_flow_on_task(flow, task, avoid_duplicate_runs=True, flow_tags=None,
     if not isinstance(flow.flow_id, int) or flow_id == False:
         _publish_flow_if_necessary(flow)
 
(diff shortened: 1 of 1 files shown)


openml/runs/functions.py
```diff
@@ -123,8 +123,14 @@ def run_flow_on_task(flow, task, avoid_duplicate_runs=True, flow_tags=None,
...
 
-    run = OpenMLRun(task_id=task.task_id, flow_id=flow.flow_id,
-                    dataset_id=dataset.dataset_id, model=flow.model, tags=tags)
+    run = OpenMLRun(
+        task_id=task.task_id,
+        flow_id=flow.flow_id,
+        dataset_id=dataset.dataset_id,
+        model=flow.model,
+        tags=tags,
+        flow_name=flow.name,
```
janvanrijn: This is the only difference, right?

mfeurer: Correct, the rest is just reformatting to make such changes look nicer in the future.

PR #502, opened 2018-08-09
Fix500

### LINKED ISSUE #500: Wrong check for publishing flow in run_flow_on_task
The following check will always be true if the function is called from `run_model_on_flow`:

```
if not isinstance(flow.flow_id, int) or flow_id == False:
    _publish_flow_if_necessary(flow)
```

yielding in an unnecessary function call to OpenML. I propose to swap it with:

```
if flow_id == False:
    _publish_flow_if_necessary(flow)
```

As flow id comes directly from the flow_exists function. I also propose to cross-check this against the flow id, as follows:
```
# in case the flow not exists, flow_id will be False (as returned by flow_exists)
    # also check whether there are no illegal flow.flow_id values (compared to result of
    # openml.flows.flow_exists)
    if flow_id is False:
        if flow.flow_id is not None:
            raise ValueError('flow.flow_id is not None, but the flow does not'
                             'exist on the server according to flow_exists')
        _publish_flow_if_necessary(flow)

    if not isinstance(flow.flow_id, int):
        if flow.flow_id != flow_id:
            # This should never happen, unless user made a flow-creation fault
            raise ValueError('Result flow_exists and flow.flow_id are not same. ')
```

janvanrijn:
Fixes #500

### DIFF
--- a/tests/test_runs/test_run_functions.py
+++ b/tests/test_runs/test_run_functions.py
@@ -754,6 +754,35 @@ class TestRun(TestBase):
         self.assertRaises(TypeError, openml.runs.run_model_on_task,
                           task=task, model=clf, avoid_duplicate_runs=False)
 
+    def test_run_with_illegal_flow_id(self):
+        # check the case where the user adds an illegal flow id to a non-existing flow
+        task = openml.tasks.get_task(1)
+        clf = DecisionTreeClassifier()
+        flow = sklearn_to_flow(clf)
+        flow, _ = self._add_sentinel_to_flow_name(flow, None)
+        flow.flow_id = -1
+
+        self.assertRaises(ValueError, openml.runs.run_flow_on_task,
+                          task=task, flow=flow, avoid_duplicate_runs=False)
+
+    def test_run_with_illegal_flow_id_1(self):
+        # check the case where the user adds an illegal flow id to an existing flow
+        # comes to a different value error than the previous test
+        task = openml.tasks.get_task(1)
+        clf = DecisionTreeClassifier()
+        flow_orig = sklearn_to_flow(clf)
+        try:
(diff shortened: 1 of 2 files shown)


janvanrijn:
The only error I see is the one due to the unicode dataset

tests/test_runs/test_run_functions.py
```diff
@@ -754,6 +754,35 @@ def test_run_with_classifiers_in_param_grid(self):
...
 
+    def test_run_with_illegal_flow_id(self):
+        # check the case where the user adds an illegal flow id to a non-existing flow
+        task = openml.tasks.get_task(1)
+        clf = DecisionTreeClassifier()
+        flow = sklearn_to_flow(clf)
+        flow, _ = self._add_sentinel_to_flow_name(flow, None)
+        flow.flow_id = -1
+
+        self.assertRaises(ValueError, openml.runs.run_flow_on_task,
```
mfeurer: Could you please make this a `self.assertRaisesRegex`? The function is quite complicated, so it would be good to make sure it fails exactly where we expect it to fail.

tests/test_runs/test_run_functions.py
```diff
@@ -754,6 +754,35 @@ def test_run_with_classifiers_in_param_grid(self):
...
+        try:
+            flow_orig.publish()  # ensures flow exist on server
+        except openml.exceptions.OpenMLServerException:
+            # flow already exists
+            pass
+        flow_new = sklearn_to_flow(clf)
+
+        flow_new.flow_id = -1
+
+        self.assertRaises(ValueError, openml.runs.run_flow_on_task,
```
mfeurer: Same here.

janvanrijn:
Good points, now they should be fine.

PR #508, opened 2018-08-28
Nominal values attribute fix

### LINKED ISSUE #507: Nominal features values
#### Description
The `nominal_values` field of a feature is left unpopulated, see [here](https://github.com/openml/openml-python/blob/7a612eaece2fd28315d99dda3e2393ada2f27478/openml/datasets/dataset.py#L84). The data seems to be available trough [REST](https://www.openml.org/api/v1/json/data/features/2?api_key=).

#### Steps/Code to Reproduce
Example:
```python
import openml
dataset = openml.datasets.get_dataset(2)
print(dataset.features[0].nominal_values)
```

#### Expected Results
`["GB","GK","GS","TN","ZA","ZF","ZH","ZM","ZS"]`

#### Actual Results
`None`

#### Versions
```
Darwin-17.7.0-x86_64-i386-64bit
Python 3.6.5 (default, Apr  5 2018, 10:39:42) 
[GCC 4.2.1 Compatible Apple LLVM 9.1.0 (clang-902.0.39.1)]
NumPy 1.14.2
SciPy 1.0.1
Scikit-Learn 0.19.2
OpenML 0.7.0
```
-

maxcopeland:
#### Reference Issue
Fixes #507 
Adds list of nominal values to be passed to `OpenMLDataFeature` constructor in setting the `OpenMLDataset.feature` attribute

Creates `nom_val` variable to store list of nominal variables if feature is nominal, stores `None` otherwise, then `nom_val` is passed to `OpenMLDataFeature` constructor

#### Any other comments?
`xmlfeature['oml:nominal_values']` contains list as string type (e.g. `"['A', 'B', 'C', 'D']"), thus the use of splitting, lambda

### DIFF
--- a/openml/datasets/dataset.py
+++ b/openml/datasets/dataset.py
@@ -78,10 +78,16 @@ class OpenMLDataset(object):
         if features is not None:
             self.features = {}
             for idx, xmlfeature in enumerate(features['oml:feature']):
+                # split string of nominal values into type list if feature is nominal
+                # otherwise passing none for nominal values
+                try:
+                    nom_vals = [str(x[1:-1]) for x in xmlfeature['oml:nominal_values'][1:-1].split(',')]
+                except KeyError:
+                    nom_vals = None
                 feature = OpenMLDataFeature(int(xmlfeature['oml:index']),
                                             xmlfeature['oml:name'],
                                             xmlfeature['oml:data_type'],
-                                            xmlfeature['oml:nominal_values'] ######## This must pass nominal values for feature
+                                            nom_vals,
                                             int(xmlfeature.get('oml:number_of_missing_values', 0)))
                 if idx != feature.index:
                     raise ValueError('Data features not provided in right order')
--- a/openml/datasets/dataset.py
+++ b/openml/datasets/dataset.py
@@ -81,7 +81,7 @@ class OpenMLDataset(object):
                 feature = OpenMLDataFeature(int(xmlfeature['oml:index']),
                                             xmlfeature['oml:name'],
                                             xmlfeature['oml:data_type'],
-                                            None,  # todo add nominal values (currently not in database)
+                                            xmlfeature['oml:nominal_values'] ######## This must pass nominal values for feature
                                             int(xmlfeature.get('oml:number_of_missing_values', 0)))
                 if idx != feature.index:
                     raise ValueError('Data features not provided in right order')


openml/datasets/dataset.py
```diff
@@ -78,10 +78,16 @@ def __init__(self, dataset_id=None, name=None, version=None, description=None,
         if features is not None:
             self.features = {}
             for idx, xmlfeature in enumerate(features['oml:feature']):
+                # split string of nominal values into type list if feature is nominal
+                # otherwise passing none for nominal values
+                try:
+                    nom_vals = [str(x[1:-1]) for x in xmlfeature['oml:nominal_values'][1:-1].split(',')]
```
mfeurer: Could you please add a brief explanation of this statement, preferably with an example string that is parsed?

mfeurer:
Thanks a lot for tackling this @maxcopeland. Please excuse the long wait time, but I was busy for the last few days finishing up a paper.

In addition to my comment, could you please:
* add a unit test
* from my understanding, a comma is legal in the nominal value (as long as it's properly quoted) (@janvanrijn please confirm). Therefore, I think you need to parse this a bit differently, for example using the csv module or `ast.literal_eval`.

@janvanrijn is there a reason that this is passed as a python-list-like string and not as a proper XML list? I assume the second one would be easier to handle by the client APIs as the parsing is handled by XML libraries.

janvanrijn:
@mfeurer no particular reason, if you prefer an xml list we can discuss it next week at the Hackathon?

mfeurer:
I definitively prefer the list to be in XML, because we can then rely on standard libraries to do the parsing. If the list is stored like it is stored right now, this requires custom parsing on the client side, which we should definitively avoid.

maxcopeland:
No problem. Should I rewrite the parsing to expect xml list in `xmlfeature['oml:nominal_values']`?

mfeurer:
Thanks for your offer. I'm afraid we first need to discuss how to expect such information from the server and then implement this. We will come back to you next week.

maxcopeland:
Okay, sounds good. Just let me know once there's a consensus and I'll make the necessary changes.

maxcopeland:
@mfeurer Great. Sure-- I'll get started.

mfeurer:
Superseded by #644.

Issue #511, opened 2018-08-31
Support for BayesSearch of SkOpt


janvanrijn:
Requires: 
- extension of sklean_to_flow and flow_to_sklearn
- unit tests and an additional dependency (for unit tests)

mfeurer:
I think this can now be done as an extension and would therefore not be part of the main library -> closing this for now.

PR #512, opened 2018-08-31
[WIP] Add #511 - support for scikit-optimize

janvanrijn:
Adds #511 ; it adds support for scikit-optimize BayesSearch

Open questions:
- How to ensure that the 'eval' method does not get abused
- Add formal dependency for skopt? (my vote: yes)

@mfeurer WDYT?

### DIFF
--- a/openml/flows/sklearn_converter.py
+++ b/openml/flows/sklearn_converter.py
@@ -47,7 +47,8 @@ def sklearn_to_flow(o, parent_model=None):
         rval = [sklearn_to_flow(element, parent_model) for element in o]
         if isinstance(o, tuple):
             rval = tuple(rval)
-    elif isinstance(o, (bool, int, float, six.string_types)) or o is None:
+    elif isinstance(o, (bool, int, float, six.string_types)) or o is None or \
+            np.issubdtype(o, np.integer) or np.issubdtype(o, np.float):
         # base parameter values
         rval = o
     elif isinstance(o, dict):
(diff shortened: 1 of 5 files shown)


janvanrijn:
> Is there any particular reason why you considered scikit-optimize? 

Because it works naturally well together with sklearn, and it was (almost) a ten liner to integrate with the current openml package. 

> So far I have not seen any results of scikit-optimize outperforming any other Bayesian optimization package.

I can imagine SMAC3 being faster, due to many nice optimizations, such as the pyrfr, but not better, right? I would assume that with correct parameterization, all results from SMAC3 should also be obtainable by scikit-optimize. 

Do you have another opinion?

mfeurer:
> Because it works naturally well together with sklearn, and it was (almost) a ten liner to integrate with the current openml package.

Yes, but those ten lines are ten lines too much in my opinion. We really should discuss this next week before investing any time in extending the scikit-learn converter to new packages.

> I can imagine SMAC3 being faster, due to many nice optimizations, such as the pyrfr, but not better, right? I would assume that with correct parameterization, all results from SMAC3 should also be obtainable by scikit-optimize.

My comment is not about SMAC3, it is mostly about other GP-based Bayesian optimization toolkits where GPyOpt, Spearmint and RoBO appear to perform best. All in all, it depends on the function(s) that you are trying to optimize, and I have not yet seen any results regarding scikit-optimize.

Issue #515, opened 2018-09-10
run_model_on_task has no docs, run_flow_on_tasks docs are wrong


amueller:
See [here](https://github.com/openml/openml-python/blob/7a612eaece2fd28315d99dda3e2393ada2f27478/openml/runs/functions.py#L37) and [here](https://github.com/openml/openml-python/blob/7a612eaece2fd28315d99dda3e2393ada2f27478/openml/runs/functions.py#L58).

The ``run_flow_on_task`` actually needs a flow, not a model, from what I can see.

ArlindKadra:
I referenced the second part here #498 and I have not done a commit yet as there is also some additional functionality to be added into the function.

mfeurer:
Added via #516.

Issue #498, opened 2018-08-03
Unclear function documentation and warning


ArlindKadra:
https://github.com/openml/openml-python/blob/5a1048d9087677b6a2d62d9e044a6c00d37795f0/openml/runs/functions.py#L52-L94

I was checking out this function and it confused me. 
https://github.com/openml/openml-python/blob/5a1048d9087677b6a2d62d9e044a6c00d37795f0/openml/runs/functions.py#L62
There is no model parameter, there is a flow and it is an OpenMLFlow.
https://github.com/openml/openml-python/blob/5a1048d9087677b6a2d62d9e044a6c00d37795f0/openml/runs/functions.py#L67
I am guessing it should be changed from second to first argument, as this is the second one ?
https://github.com/openml/openml-python/blob/5a1048d9087677b6a2d62d9e044a6c00d37795f0/openml/runs/functions.py#L93-L94
It seems wrong. It was OpenMLTask and OpenMLFlow, now it is reversed. The warning should display "It should be (flow, task) and not (task, flow) "?

If I understood things clearly, I can make a pull request with the minor changes.

ArlindKadra:
On a further note, I will also update the examples tutorial following the Deprecation.

mfeurer:
Yes, the parameter documentation is wrong. Line 67 can be removed in the future. Your note on the warning is correct, too.

mfeurer:
Added via #516.

PR #516, opened 2018-09-11
Refactoring run_flow_on_task and doc add for run_model

ArlindKadra:
#### Reference Issue
#515,  #498, #457 

The PR fixes the documentation for run_flow_on_task, it also tackles run_flow_or_task always uploading the flow.

Calling the function, getting a run object in the end without uploading the flow to OpenML.


PGijsbers:
Hi, I am interested in having this functionality and if wonder it's ok working on this. I am a bit of a loss of the changes made so far, so I have some questions to better understand your work:
- What is the `_to_dict` function for (in tasks/tasks.py). It seems to return some flow_container which is not referenced elsewhere, it seems to serialize flows but is located in `tasks.py`. Was it meant for flow serialization (for saving to disk)?
- Work on actually preventing flow checks/uploads is not handled yet, is that correct?

In general, I thought to update the process to the following:
- check if flow exists online, if so download. If not or no internet connection, use the local flow (possibly publish it if a flag is set and internet is available).
- do the run logic
- assuming still using a local flow, reference that when creating an OpenMLRun.

then upload the flow when:
- publishing the run, if a local flow was used, upload it first, update the OpenMLRun with updated flow information before uploading the run.
- or when `flow.publish` is called.

ArlindKadra:
Hey @PGijsbers 

>  What is the `_to_dict` function for (in tasks/tasks.py). It seems to return some flow_container which is not referenced elsewhere, it seems to serialize flows but is located in `tasks.py`. Was it meant for flow serialization (for saving to disk)?

It has been quite a bit of time. Considering the commit name, I guess I was implementing the generation of a task dictionary and I just copy pasted the code from the _to_dict function in the flow class. I started tweaking it and I never got back to the pr, that is why there are still flow parts there, so basically it is not useful. You should remove that function completely.

This is addressed  in #607 .

> Work on actually preventing flow checks/uploads is not handled yet, is that correct?

Yes, it is not handled yet. Preventing flow uploads is not easily done, since a flow id is needed to create the run, with the idea of storing the run locally.

As to how you want to proceed, it seems really good to me. We had a discussion on how to handle this in the last OpenML workshop and this was what we agreed upon. @mfeurer @janvanrijn can you confirm this was and still is the case ?

The functions to get the cached flows should be useful and the documentation change for the run_model_on_task is nice too.

ArlindKadra:
@mfeurer should the flow cache functions be moved in another PR ?

PGijsbers:
So basically undo the last commit 22b1e620459fe76933d4d132656ba053fc4fb096 before proceeding to work?

ArlindKadra:
Yes, the other commits should be useful.

mfeurer:
I think this is what we discussed in Paris. And flow caching should be fine within this PR as it would already test the new flow caching functionality.

PGijsbers:
Question about new expected behavior:

[1](https://github.com/openml/openml-python/blob/master/tests/test_runs/test_run.py#L158)) wants to store and load an `OpenMLRun` where no model is attached. With the new behavior, it is possible that the run is still tied to an `OpenMLFlow` which only exists locally, and thus the `parameter_settings` for the run can not be extracted from the model and serialized. This leads to the scenario where if the user would save the run with local flow without persisting the model, after loading it the `parameter_settings` will be lost. What should the behavior be? I think an error or warning might be in order.

mfeurer:
I don't think it is necessary for that function to sync with the server, and assume that check was in place to make sure that only runs are extracted for which the flow is published. As the workflow is supposed to change, this behavior can also be changed

PGijsbers:
Is it okay to exclude that change from this PR though? It's already getting quite big and I do not know if I have a good enough handle on what exactly subflows are supposed to be or what consistency assumptions there are. I can dive into it sometime, but it seems a bit daunting to tackle that right now, especially since I would prefer to round up this PR.

In place, I would either raise an error or warning if a run is being saved to disk without model if it is linked to a local flow. This can later be refactored into just storing the `parameter_settings`.

PGijsbers:
@mfeurer @janvanrijn 
I ran into an issue, and while I think it is separate from this PR, I wasn't completely sure.

What is the difference between a flow's `external_version` and `dependencies`? I ran `run_model_on_task` with the same model on two different computers which had a scikit-learn version differences (0.20.0 vs 0.20.2), and on publishing the flow I got the error:
```
ValueError: Flow category_encoders.target_encoder.TargetEncoder: values for attribute 'dependencies' differ: 'sklearn==0.20.2
numpy>=1.6.1
scipy>=0.9'
vs
'sklearn==0.20.0
numpy>=1.6.1
scipy>=0.9'.
```
This is an error regarding one of the subcomponents of my pipeline, [flow 9151](https://www.openml.org/f/9151).
The `external_version` is `'category_encoders==1.3.0,openml==0.8.0'` for both flows (and the name is the same for both, too), which means the server identifies them as the same. I am not sure if they _should_ have been regarded as the same flow; while the package the component came from is identical, as one of the dependecies changes different models may (theoretically) still be generated (I don't think that's the case here). 

Making the flow unique by its dependencies might really scatter run results. This makes me think it was a conscious choice to store the used dependencies but not create a new flow if they differ? If so, why are dependencies still forced to match the server side? If they are supposed to generate flows, why are they not incorporated in `flow_exists`? In case a difference dependencies should not create a new flow, they still should be stored in the `OpenMLRun`, which I do not think is currently the case?

openml/flows/functions.py
```diff
@@ -1,13 +1,75 @@
...
 
+from ..exceptions import OpenMLCacheException
 import openml._api_calls
 from . import OpenMLFlow
 import openml.utils
 
+FLOWS_CACHE_DIR_NAME = 'flows'
+
+
+def _get_cached_flows():
```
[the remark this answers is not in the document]
PGijsbers: In general, I might have forgotten, but if I'm not mistaken this is specific code was written by Arlind. I'll add type annotations.

examples/flows_and_runs_tutorial.py
```diff
@@ -89,6 +89,33 @@
...
+    pipe,
+    task,
+    avoid_duplicate_runs=False,
+    upload_flow=False)
+
+# The run may be stored offline, and the flow will be stored along with it:
+run.to_filesystem(output_directory='myrun')
+
+# They made later be loaded and uploaded
+run = openml.runs.OpenMLRun.from_filesystem(folder='myrun')
```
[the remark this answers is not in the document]
PGijsbers: This inconsistency was already present in the development branch, I did not change anything. I could however, add that change to the PR (I suggest naming the parameter `directory` both ways).

openml/flows/functions.py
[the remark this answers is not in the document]
mfeurer: > this is specific code was written by Arlind
Sorry, I forgot about the history of this PR.

examples/flows_and_runs_tutorial.py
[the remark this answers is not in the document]
mfeurer: :+1:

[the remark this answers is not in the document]
mfeurer: (to changing the name)

openml/exceptions.py
```diff
@@ -49,5 +49,14 @@ class OpenMLHashException(PyOpenMLError):
...
 class PrivateDatasetError(PyOpenMLError):
     """ Exception thrown when the user has no rights to access the dataset. """
-    def __init__(self, message):
-        super(PrivateDatasetError, self).__init__(message)
+    def __init__(self, message: str):
+        super().__init__(message)
+
+
+class RunsExistError(PyOpenMLError):
+    """ Indicates run(s) already exists on the server when avoiding them is preferred. """
```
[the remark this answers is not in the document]
PGijsbers: Agreed, done.

openml/runs/functions.py
```diff
@@ -46,46 +85,52 @@ def run_model_on_task(model, task, avoid_duplicate_runs=True, flow_tags=None,
...
 
 
-def run_flow_on_task(flow, task, avoid_duplicate_runs=True, flow_tags=None,
-                     seed=None, add_local_measures=True):
+def run_flow_on_task(flow: OpenMLFlow, task: OpenMLTask,
+                     avoid_duplicate_runs: bool = True,
+                     flow_tags: List[str] = None,
+                     seed: int = None,
+                     add_local_measures: bool = True,
+                     upload_flow: bool = False) -> OpenMLRun:
```
mfeurer: The formatting comment from above would also apply here.

openml/runs/functions.py
```diff
@@ -95,30 +140,48 @@ def run_flow_on_task(flow, task, avoid_duplicate_runs=True, flow_tags=None,
...
-        if ids:
-            raise PyOpenMLError("Run already exists in server. "
-                                "Run id(s): %s" % str(ids))
-        _copy_server_fields(flow_from_server, flow)
+    flow.model = _set_model_seed_where_none(flow.model, seed=seed)
+
+    # We only need to sync with the server right now if we want to upload the flow,
+    # or ensure no duplicate runs exist. Otherwise it can be synced at upload time.
+    flow_id = None
+    if upload_flow or avoid_duplicate_runs:
```
mfeurer: This looks super clean, I like it :)

openml/flows/functions.py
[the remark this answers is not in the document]
PGijsbers: I'm assuming you forgot an extra line-break in the comment. Either way, I added type hints to every function in `openml/flows/functions.py` now.

openml/flows/functions.py
```diff
@@ -26,13 +88,13 @@ def get_flow(flow_id, reinstantiate=False):
...
+            name='flows.functions.get_flow:%d' % flow_id,
+            lock_path=openml.utils._create_lockfiles_dir(),
+    ):
+        flow = _get_flow_description(flow_id)
 
-    flow_dict = xmltodict.parse(flow_xml)
-    flow = OpenMLFlow._from_dict(flow_dict)
-
-    if reinstantiate:
+    if reinstantiate and flow:
```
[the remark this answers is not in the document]
PGijsbers: Looking into this. It was introduced during merge ac343c58, but obviously does not exist on the development branch.

openml/runs/functions.py
```diff
@@ -892,23 +912,30 @@ def obtain_field(xml_obj, fieldname, from_server, cast=None):
     else:
         task_evaluation_measure = None
 
-    flow_id = int(run['oml:flow_id'])
+    if not from_server and run['oml:flow_id'] is None:
```
mfeurer: Could you please add the `from_server` argument to the docstring?

openml/flows/sklearn_converter.py
```diff
@@ -236,16 +236,18 @@ def openml_param_name_to_sklearn(openml_parameter, flow):
...
     Parameters
     ----------
     flow : OpenMLFlow
-        openml flow object (containing flow ids, i.e., it has to be downloaded
-        from the server)
+        OpenMLFlow object (containing flow ids, i.e., it has to be downloaded from the server)
+
+    model: object, optional (default=None)
+        The model from which to obtain the parameter values. Must match the flow signature.
+        If None, use the model specified in `OpenMLFlow.model`
```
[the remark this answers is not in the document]
PGijsbers: That's exactly what it is for. I think you mean the function summary? If so, you're correct it's outdated and I updated it based on this feedback.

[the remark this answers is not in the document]
PGijsbers: Changed it to:

    Extracts all parameter settings required for the flow from the model.
    If no explicit model is provided, the parameters will be extracted from `flow.model` instead.

Do you think this works?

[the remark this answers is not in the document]
mfeurer: Yes, thanks a lot.

openml/runs/functions.py
```diff
@@ -32,11 +33,49 @@
 RUNS_CACHE_DIR_NAME = 'runs'
 
 
-def run_model_on_task(model, task, avoid_duplicate_runs=True, flow_tags=None,
-                      seed=None, add_local_measures=True):
-    """See ``run_flow_on_task for a documentation``."""
-    # TODO: At some point in the future do not allow for arguments in old order
-    # (order changed 6-2018).
+def run_model_on_task(model: object, task: OpenMLTask,
```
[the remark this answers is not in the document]
PGijsbers: Done for this method and `run_flow_on_task`. I didn't know you could have a trailing comma in parameters, nifty.

openml/runs/functions.py
```diff
@@ -20,7 +21,7 @@
 from openml.flows.flow import _copy_server_fields
 from ..flows import sklearn_to_flow, get_flow, flow_exists, OpenMLFlow
 from ..setups import setup_exists, initialize_model
-from ..exceptions import OpenMLCacheException, OpenMLServerException
+from ..exceptions import OpenMLCacheException, OpenMLServerException, RunsExistError
```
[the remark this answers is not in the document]
PGijsbers: There is also an `PrivateDatasetError`. So far it looks like the `OpenML`-prefixed errors might be (somewhat) unexpected: consistency errors, cache errors, server errors. I felt `RunsExistError` would be closer to the `PrivateDatasetError`. That said, it's all a bit wishy-washy, so we are probably better off either prefixing everything or specifying guidelines.

openml/flows/functions.py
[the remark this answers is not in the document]
PGijsbers: I don't see how it resolves to `False` anymore (though I am sure it could in the past :D), I removed it (we'll see if travis proves us wrong).

openml/runs/functions.py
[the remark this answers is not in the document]
mfeurer: Yes, that would be great.

tests/test_runs/test_run.py
```diff
@@ -179,3 +183,31 @@ def test_to_from_filesystem_no_model(self):
...
+        task = openml.tasks.get_task(119)
+        run = openml.runs.run_model_on_task(
+            model=model,
+            task=task,
+            add_local_measures=False,
+            avoid_duplicate_runs=False,
+            upload_flow=False
+        )
+
+        cache_path = os.path.join(
```
mfeurer: Could you please check the base test class? It creates a temporary directory for each unit test. It would be great if you could use that instead of creating your own directory.

tests/test_runs/test_run.py
```diff
@@ -179,3 +183,31 @@ def test_to_from_filesystem_no_model(self):
...
+            ('classifier', DummyClassifier()),
+        ])
+        task = openml.tasks.get_task(119)
+        run = openml.runs.run_model_on_task(
+            model=model,
+            task=task,
+            add_local_measures=False,
+            avoid_duplicate_runs=False,
+            upload_flow=False
+        )
```
mfeurer: Would it be possible to add a check here that the flow is not uploaded?

openml/runs/functions.py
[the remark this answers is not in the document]
PGijsbers: So which do we do? Prefix both with `OpenML` for now?

tests/test_runs/test_run.py
```diff
@@ -179,3 +183,31 @@ def test_to_from_filesystem_no_model(self):
...
+
+        cache_path = os.path.join(
+            self.workdir,
+            'runs',
+            str(random.getrandbits(128)),
+        )
+        run.to_filesystem(cache_path)
+        # obtain run from filesystem
+        loaded_run = openml.runs.OpenMLRun.from_filesystem(cache_path)
+        loaded_run.publish()
```
mfeurer: And that it is instead uploaded here? Maybe you could add some checks on the run, too? Maybe using the mock package, then you do not even need to interact with the server.

mfeurer: (although it might be good to test this kind of behavior against the server)

tests/test_runs/test_run.py
```diff
@@ -179,3 +183,31 @@ def test_to_from_filesystem_no_model(self):
         # assert default behaviour is throwing an error
         with self.assertRaises(ValueError, msg='Could not find model.pkl'):
             openml.runs.OpenMLRun.from_filesystem(cache_path)
+
+    def test_publish_with_local_loaded_flow(self):
```
mfeurer: Do you think it would make sense to add a test which downloads a flow from the internet, caches it, does a run, and uploads the run afterwards? Or is the functionality used more or less the same?

openml/runs/functions.py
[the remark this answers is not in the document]
mfeurer: Yes, prefixing both would be good I think.

mfeurer:
I played a bit with the code and it appears that it cannot handle a simple random forest at the moment:
```
import sklearn.ensemble

import openml

openml.config.server = "https://test.openml.org/api/v1/xml"
openml.config.apikey = "610344db6388d9ba34f6db45a3cf71de"

model = sklearn.ensemble.RandomForestClassifier(n_estimators=33)

task = openml.tasks.get_task(12)

run = openml.runs.run_model_on_task(
    model=model,
    task=task,
    avoid_duplicate_runs=False,
    upload_flow=False,
)

print(run)
print(run.flow)
print(run.flow.flow_id)
print(run.flow_id)

run.publish()
```
as it results in:
```
[run id: None, task id: 12, flow id: None, flow name: sklearn.ensemble.forest.Ra...]
<openml.flows.flow.OpenMLFlow object at 0x7fe912411438>
None
None
  File "/home/feurerm/sync_dir/projects/openml/python/testing3.py", line 24, in <module>
    run.publish()
  File "/home/feurerm/sync_dir/projects/openml/python/openml/runs/run.py", line 402, in publish
    self.parameter_settings = openml.flows.obtain_parameter_values(self.flow, self.model)
  File "/home/feurerm/sync_dir/projects/openml/python/openml/flows/sklearn_converter.py", line 373, in obtain_parameter_values
    model = model if model else flow.model
  File "/home/feurerm/miniconda/3-4.5.4/envs/openml/lib/python3.6/site-packages/sklearn/ensemble/base.py", line 140, in __len__
    return len(self.estimators_)
AttributeError: 'RandomForestClassifier' object has no attribute 'estimators_'
```

tests/test_runs/test_run.py
PGijsbers: This follows [this existing test](https://github.com/openml/openml-python/blob/develop/tests/test_runs/test_run.py#L159). Why was the generated folder name used there then? Should it also be changed?

mfeurer: That is an excellent question. I'm afraid that this was oversight by me in reviewing the other test. It would be great if you could change them both.

tests/test_runs/test_run.py
PGijsbers: Good catch. Added (it works locally in isolation but I have to see if it works in the test suite too).

tests/test_runs/test_run.py
PGijsbers: Well this is where I am a little puzzled by the desire to both test against the server but also use mocking instead 😅  Anyway, I updated the test right now to ensure the run and flow exist on the server.

PGijsbers:
Weird. I can recreate the problem locally and am looking into this.

PGijsbers:
It seems that with absence of an explicit `__bool__` operator Python objects fall back on the `__len__` operator to handle 'truthiness', since `BaseEnsemble` had a `__len__` method but not a `__bool__` method, `__len__` was invoked. I now refactored the problematic line 
`model = model if model else flow.model`
to include explicit None checking:
`model = model if model is not None else flow.model`

tests/test_runs/test_run.py
PGijsbers: It's the same for running the experiment, but it's different at upload time (because the flow already exists). Added [this test](https://github.com/openml/openml-python/blob/fix498/tests/test_runs/test_run_functions.py#L1508).

tests/test_runs/test_run.py
PGijsbers: As far as I can tell, you are referring to `self.workdir`? The `to_filesystem` method [requires the targeted directory to be empty](https://github.com/openml/openml-python/blob/develop/openml/runs/run.py#L137). For this reason, it seems best to create an new unique subdirectory of `self.workdir` as we are doing here (it looks like @janvanrijn designed it like this originally). Granted we don't necessarily need two levels of subdirectories, one should suffice. If you are not referring to `self.workdir`, please point me in the right direction.

mfeurer: Yes, it is `self.workdir`. But actually, the test is executed in the workdir any way, so nevermind.

tests/test_runs/test_run.py
mfeurer: Yes, most likely it is good the way it is

PGijsbers:
But that is essentially the same as a user just creating the flow through `sklearn_to_flow` and then accessing the `None` fields, right? If so, I would agree that it's not really related to this PR so much as #634.

openml/flows/flow.py
```diff
@@ -326,16 +352,27 @@ def publish(self):
         # instantiate an OpenMLFlow.
         import openml.flows.functions
 
-        xml_description = self._to_xml()
+        flow_id = openml.flows.functions.flow_exists(self.name, self.external_version)
+        if not flow_id:
```
ArlindKadra: I think this would be simpler to understand if you control with self.flow_id and not flow_id. It organizes the checks a bit nicer in my opinion.

openml/flows/functions.py
```diff
@@ -1,13 +1,75 @@
...
+        openml.utils._remove_cache_dir_for_id(FLOWS_CACHE_DIR_NAME, fid_cache_dir)
+        raise OpenMLCacheException("Flow file for fid %d not "
+                                   "cached" % fid)
+
 
-def get_flow(flow_id, reinstantiate=False):
+def get_flow(flow_id: int, reinstantiate: bool = False) -> OpenMLFlow:
     """Download the OpenML flow for a given flow ID.
 
     Parameters
```
ArlindKadra: you have forgotten to finish the sentence for the `reinstantiate` parameter.

openml/flows/functions.py
```diff
@@ -167,8 +262,8 @@ def __list_flows(api_call):
     return flows
 
 
-def _check_flow_for_server_id(flow):
-    """Check if the given flow and it's components have a flow_id."""
+def _check_flow_for_server_id(flow: OpenMLFlow) -> None:
+    """ Raises a ValueError of the flow or any of its subflows has no flow id. """
```
ArlindKadra: for the flow, if any typos

ArlindKadra:
@mfeurer @PGijsbers except for a small suggestion, everything looks really good to me :)

openml/flows/flow.py
PGijsbers: I am not sure I follow. `flow_id` and `self.flow_id` are not the same. In this case `flow_id` is `False` if the flow does not yet exist on the server, or an integer representing the flow id as known by the server. `self.flow_id` is None if it has not yet been synced with the server, or an integer if it has. Or do you mean changing it to:
```
if self.flow_id and not flow_id:
    raise ...
elif not flow_id:
    ...
    rest of code
```
?

openml/flows/functions.py
PGijsbers: Unfortunately this is an artefact that exists in [develop](https://github.com/openml/openml-python/blob/develop/openml/flows/functions.py#L18). I am not sure what this extra clause would be so I can not complete it.

openml/flows/functions.py
PGijsbers: Changed it to `Raises a ValueError if the flow or any of its subflows has no flow id.`, good catch.

openml/flows/flow.py
ArlindKadra: I meant the second case that you gave a code example for and I was thinking of a different organization. But looking at it more, I do not see a clear way on how to not have extra checks.

PR #526, opened 2018-09-17
[MRG] MAINT: compatibility sklearn 0.20

glemaitre:
Make test pass with scikit-learn 0.20

### DIFF
--- a/tests/test_flows/test_flow.py
+++ b/tests/test_flows/test_flow.py
@@ -149,7 +149,7 @@ class TestFlow(TestBase):
         flow = openml.flows.sklearn_to_flow(clf)
         flow, _ = self._add_sentinel_to_flow_name(flow, None)
         flow.publish()
-        self.assertRaisesRegex(openml.exceptions.OpenMLServerException,
+        self.assertRaisesRegexp(openml.exceptions.OpenMLServerException,
                                 'flow already exists', flow.publish)
 
     def test_publish_flow_with_similar_components(self):
--- a/tests/test_flows/test_flow.py
+++ b/tests/test_flows/test_flow.py
@@ -349,12 +349,21 @@ class TestFlow(TestBase):
         openml.flows.functions.assert_flows_equal(new_flow, flow)
         self.assertIsNot(new_flow, flow)
 
-        estimators_names = (sentinel, 'RandomizedSearchCV', 'Pipeline',
-                            'OneHotEncoder', 'StandardScaler', 'FeatureUnion',
-                            'TruncatedSVD', 'SelectPercentile',
-                            'AdaBoostClassifier', 'DecisionTreeClassifier')
-        for keyword in estimators_names:
-            self.assertTrue(keyword in new_flow.name)
+        # OneHotEncoder was moved to _encoders module in 0.20
+        module_name_encoder = ('_encoders'
+                               if LooseVersion(sklearn.__version__) >= "0.20"
+                               else 'data')
+        fixture_name = '%ssklearn.model_selection._search.RandomizedSearchCV(' \
+                       'estimator=sklearn.pipeline.Pipeline(' \
+                       'ohe=sklearn.preprocessing.%s.OneHotEncoder,' \
+                       'scaler=sklearn.preprocessing.data.StandardScaler,' \
+                       'fu=sklearn.pipeline.FeatureUnion(' \
+                       'pca=sklearn.decomposition.truncated_svd.TruncatedSVD,' \
+                       'fs=sklearn.feature_selection.univariate_selection.SelectPercentile),' \
+                       'boosting=sklearn.ensemble.weight_boosting.AdaBoostClassifier(' \
+                       'base_estimator=sklearn.tree.tree.DecisionTreeClassifier)))' \
+                        % (sentinel, module_name_encoder)
+        self.assertEqual(new_flow.name, fixture_name)
         new_flow.model.fit(X, y)
 
     def test_extract_tags(self):
--- a/tests/test_flows/test_flow.py
+++ b/tests/test_flows/test_flow.py
@@ -4,6 +4,7 @@ import hashlib
 import re
 import sys
 import time
+from distutils.version import LooseVersion
 
 if sys.version_info[0] >= 3:
     from unittest import mock
@@ -22,6 +23,12 @@ import sklearn.pipeline
 import sklearn.preprocessing
 import sklearn.naive_bayes
 import sklearn.tree
+
+if LooseVersion(sklearn.__version__) < "0.20":
+    from sklearn.preprocessing import Imputer
+else:
+    from sklearn.impute import SimpleImputer as Imputer
+
 import xmltodict
 
 from openml.testing import TestBase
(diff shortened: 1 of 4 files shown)


tests/test_flows/test_flow.py
```diff
@@ -274,16 +283,18 @@ def test_existing_flow_exists(self):
...
 
     def test_sklearn_to_upload_to_flow(self):
         iris = sklearn.datasets.load_iris()
         X = iris.data
         y = iris.target
 
         # Test a more complicated flow
-        ohe = sklearn.preprocessing.OneHotEncoder(categorical_features=[1],
-                                                  handle_unknown='ignore')
+        ohe_params = {'handle_unknown': 'ignore'}
```
mfeurer: Is there a reason why the list of categorical feature is dropped here?

tests/test_flows/test_sklearn.py
```diff
@@ -304,15 +350,17 @@ def test_serialize_pipeline_clustering(self):
...
-        serialization =  sklearn_to_flow(fu)
-        self.assertEqual(serialization.name,
-                         'sklearn.pipeline.FeatureUnion('
-                         'ohe=sklearn.preprocessing.data.OneHotEncoder,'
-                         'scaler=sklearn.preprocessing.data.StandardScaler)')
+        fu = sklearn.pipeline.FeatureUnion(
+            transformer_list=[('ohe', ohe), ('scaler', scaler)])
+        serialization = sklearn_to_flow(fu)
+        estimators_names = ('FeatureUnion', 'OneHotEncoder', 'StandardScaler')
+        for keyword in estimators_names:
```
mfeurer: What's the reason to make this test more leniant?

tests/test_flows/test_sklearn.py
```diff
@@ -345,31 +395,29 @@ def test_serialize_feature_union(self):
 
         fu.set_params(scaler=None)
         serialization = sklearn_to_flow(fu)
-        self.assertEqual(serialization.name,
-                         'sklearn.pipeline.FeatureUnion('
-                         'ohe=sklearn.preprocessing.data.OneHotEncoder)')
+        estimators_names = ('FeatureUnion', 'OneHotEncoder')
+        for keyword in estimators_names:
+            self.assertTrue(keyword in estimators_names)
```
mfeurer: I think the same question as above applies here.

tests/test_flows/test_sklearn.py
```diff
@@ -345,31 +395,29 @@ def test_serialize_feature_union(self):
...
-            "scaler=sklearn.preprocessing.data.StandardScaler)")
-        self.assertEqual(
-            fu2_serialization.name,
-            "sklearn.pipeline.FeatureUnion("
-            "scaler=sklearn.preprocessing.data.OneHotEncoder,"
-            "ohe=sklearn.preprocessing.data.StandardScaler)")
+        estimators_names = ('FeatureUnion', 'OneHotEncoder', 'StandardScaler')
+        for keyword in estimators_names:
+            self.assertTrue(keyword in fu1_serialization.name)
+            self.assertTrue(keyword in fu2_serialization.name)
```
mfeurer: and here as well.

tests/test_flows/test_sklearn.py
```diff
@@ -378,22 +426,21 @@ def test_serialize_complex_flow(self):
...
-                       'scaler=sklearn.preprocessing.data.StandardScaler,' \
-                       'boosting=sklearn.ensemble.weight_boosting.AdaBoostClassifier(' \
-                       'base_estimator=sklearn.tree.tree.DecisionTreeClassifier)))'
-        self.assertEqual(serialized.name, fixture_name)
+        # check that the name of the estimator occur in the name of the flow
+        estimator_names = ('RandomizedSearchCV', 'Pipeline', 'OneHotEncoder',
+                           'StandardScaler', 'AdaBoostClassifier',
+                           'DecisionTreeClassifier')
+        for keyword in estimator_names:
+            self.assertTrue(keyword in serialized.name)
```
mfeurer: and here, too.

tests/test_flows/test_sklearn.py
```diff
@@ -652,110 +678,148 @@ def test_illegal_parameter_names(self):
...
-            sklearn.pipeline.Pipeline(steps=[('bag', sklearn.ensemble.BaggingClassifier(n_jobs=5))]),
-            sklearn.pipeline.Pipeline(steps=[('bag', sklearn.ensemble.BaggingClassifier(n_jobs=-1))]),
-            sklearn.model_selection.GridSearchCV(singlecore_bagging, legal_param_dist),
-            sklearn.model_selection.GridSearchCV(multicore_bagging, legal_param_dist)
+            sklearn.pipeline.Pipeline(
+                steps=[('bag', sklearn.ensemble.BaggingClassifier(n_jobs=1))]),
+            sklearn.pipeline.Pipeline(
+                steps=[('bag', sklearn.ensemble.BaggingClassifier(n_jobs=5))]),
+            sklearn.pipeline.Pipeline(
+                steps=[('bag', sklearn.ensemble.BaggingClassifier(n_jobs=-1))]),
```
mfeurer: Would it make sense to check for `n_jobs=None`, too?

tests/test_flows/test_flow.py
glemaitre: It is deprecated in scikit-learn 0.20. You should use a `ColumnTransformer` instead. The test being on numerical data (iris do not have category), I thought to just remove the parameters.

tests/test_flows/test_sklearn.py
glemaitre: We moved some of the transformer from module. Therefore, the serialization name will be incorrect across version of scikit-learn. In the future, it could be as well the case.

I have 2 rermarks:

* The test was intended to check the presence of the estimator in the serialization, doesn't it? Is it something else in the test which was important to be tested?
* Speaking with @janvanrijn IRL, currently all scikit-learn estimators are imported with only a single level and no more. All extra level are actually kinda private (some are explicit by putting some "_" at the front of the module name). I think that you should not rely on those extra levels.

To be more specific, you should rely that a user do `from sklearn.decompisition import PCA` and not `from sklearn.decomposition.pca import PCA` since `pca` should be private module.

tests/test_flows/test_sklearn.py
glemaitre: This is the first one by default

glemaitre: and adding `n_jobs=None` will not make the test pass in scikit-learn < 0.20

glemaitre:
@mfeurer I answer to the questions. Regarding the less restrictive test, I would be inclined to modify it but I would need a bit more details regarding what actually we should tests there.

tests/test_flows/test_sklearn.py
mfeurer: Okay, got it. Thanks for clarifying this.

tests/test_flows/test_flow.py
mfeurer: I think the categorical feature was mostly there to have it tested in addition. Anyway, it should not really be necessary as this is also tested at some other location.

tests/test_flows/test_sklearn.py
mfeurer: * The test also intends to check that the name is correctly serialized and does not change (also due to mistakes on our side). 
* Regarding the names: they extra module level might be private, but that's the information we get from the Python interpreter when we ask it for the full module path. Do you have any other suggestion on how to obtain the name(s)?

mfeurer: I think the idea of having such a mapping is a good idea from the point of view of the OpenML connector, but it needs to live inside of scikit-learn in order to get constant maintenance. For now I suggest to open a new issue in the OpenML-python issue tracker and the scikit-learn issue tracker to keep track of this and get this PR merged. Would you be okay to re-add the string asserts and add to add an if/else statement?

glemaitre:
Ok I'll do that. 

Sent from my phone - sorry to be brief and potential misspell.

Issue #327, opened 2017-10-09
move to pytest?


amueller:
Should we move to pyest given that nosetests is no longer maintained?

mfeurer:
That's a pity. Is pytest doing the same things?

amueller:
yes, more generally, but slightly differently.

Issue #447, opened 2018-04-20
Do not leave files after running unit tests


mfeurer:
Currently, the unit tests leave a few files. This clutters the development directory and makes it harder to keep an overview in git.

rth:
That would be fairly straightforward with pytest https://github.com/openml/openml-python/issues/327

Neeratyoy:
Hi,
On taking a preliminary look at `openml/testing.py`, the `setUp()` and `tearDown()` does a neat job of creating a temp directory and cleaning it up. It works.
However, all the excess files which show up untracked for git, are in the directory `test/files/`.
In that case, can I use `os.walk()` before and after the testing, to look for new files that were not there in this directory and get them cleaned? If not, can someone just nudge me to see where and how `test/files` is getting updated? 

I do need to take a deeper look but pytest still seems to have a related open [issue](https://github.com/pytest-dev/pytest/issues/1120), so this might need some book-keeping and forced deletions.

mfeurer:
@Neeratyoy I'm afraid that the current implementation does not work, i.e. it changes the working directory before running the tests, executing the 2nd `git status` outside of the git repository. Could you please create a new PR fixing this?

Neeratyoy:
Sure, I'll take a look at this.

Neeratyoy:
@mfeurer could you please guide me as to how I can test this locally?
When I execute `bash ci_scripts/test.sh`, or even check `echo $TEST_DIR` at the beginning of `run_tests()`, it gives a `mkdir: missing operand` error. The result of `echo $TEST_DIR` is blank.
How can I recreate locally how CI runs this?

>  executing the 2nd `git status` outside of the git repository.

Can I in that case store the current directory `` curr_dir=`pwd` `` and before the second `git status` execute `cd $curr_dir` ?

mfeurer:
> How can I recreate locally how CI runs this?

That's tricky. The easiest solution would be to do `export TEST_DIR=/tmp/test/` prior to running bash ci_scripts/test.sh.

> Can I in that case store the current directory curr_dir=`pwd` and before the second git status execute cd $curr_dir ?

That sounds like a good solution.

mfeurer:
Added in #721.

PR #527, opened 2018-09-17
Migrate to Pytest

rth:
Closes https://github.com/openml/openml-python/issues/327

nose is not longer maintained and most popular packages in the scientific python community have migrated to pytest. The existing tests will run without modifications with pytest, but it will make easier,
 - cleaning up temporary files (https://github.com/openml/openml-python/issues/447)
 - running some tests without actually connecting to the server without using mock responses, as it e.g. done in the openml loaded in scikit-learn.


rth:
CI is green. 

What do you think about this @mfeurer @janvanrijn ?

rth:
So, `pip install` is now the recommended way to install packages (even local packages) as compared to `python setup.py install` (see https://packaging.python.org/tutorials/installing-packages/#installing-from-a-local-src-tree).

```
pip install .  # dot important at the end
```
installs the local package.

To get the behaviour of `develop`, it's necessary to add the `-e` flag.

And to install extra dependencies, one would do,
```
pip install openml[test]
```
which installs extra "test" dependencies specified in the `setup.py`. So the sum of that in this case is,
```
pip install -e .[test]
```
with some paranthesis as I wasn't sure how bash would handle it.

mfeurer:
Thanks a lot for explaining this. I just talked to @janvanrijn and he's okay with merging this. Thanks a lot!

rth:
Thanks for the review @mfeurer !

Issue #538, opened 2018-09-20
get_task failing


ArlindKadra:
#### Description
Getting a certain clustering task is failing. Task shows fine on the server.
#### Steps/Code to Reproduce

```
import openml

result = openml.tasks.functions.get_task(126033)
print(result)

```
#### Expected Results
Actual task.
#### Actual Results
`FileNotFoundError: [Errno 2] No such file or directory:`

ArlindKadra:
input is not a list but a dictionary.
https://github.com/openml/openml-python/blob/34c06df237dc3f953b0be77acedfc1f28b82abd5/openml/tasks/functions.py#L351

ArlindKadra:
@janvanrijn  @mfeurer I am trying to fix the issue above and I have a question regarding the current implementation of tasks.
https://github.com/openml/openml-python/blob/811f9cee3bdbca535eb898282edab9ab6ee71594/openml/tasks/task.py#L10-L20

The base class upon which all of the tasks are built, has estimation procedure as a mandatory field, with type and parameters. The above issue is happening because the tasks has only 1 input. 

Are clustering tasks supposed to have a mandatory estimation procedure field and the above an outlier which should be removed or I should change the implementation because like this.

The easiest fix for the second option, would be to have the common args like that for all the other tasks and to pass none for the estimation procedure for the clustering task.

mfeurer:
I think removing the estimation procedure from the base class and having it in the `OpenMLSupervisedTask` would be a clean solution.

mfeurer:
Fixed by #541.

PR #541, opened 2018-09-20
Single input task partial fix

ArlindKadra:
#### Reference Issue
#538

Handling cases where tasks have only one input. Change to the cache dir.

Added unit test for get_task with different task types.


mfeurer:
@ArlindKadra could you please rebase this on the development branch and update for the fact that we have multiple task classes by now?

openml/config.py
```diff
@@ -19,7 +19,7 @@
     'apikey': None,
     'server': "https://www.openml.org/api/v1/xml",
     'verbosity': 0,
-    'cachedir': os.path.expanduser('~/.openml/cache'),
+    'cachedir': os.path.expanduser(os.path.join('~', '.openml', 'cache')),
```
mfeurer: In light of your change of `/` to `os.sep`, you should replace the `~` by `os.path.expanduser`.

ArlindKadra: @mfeurer according to the docs, "On Unix and Windows, return the argument with an initial component of ~ or ~user replaced by that user’s home directory.". So the ~ is necessary.

mfeurer: You're right, my mistake.

openml/tasks/functions.py
```diff
@@ -292,9 +292,12 @@ def get_task(task_id):
...
             dataset = get_dataset(task.dataset_id)
-            class_labels = dataset.retrieve_class_labels(task.target_name)
-            task.class_labels = class_labels
-            task.download_split()
+            # Clustering tasks do not have class labels
+            # and do not offer download_split
+            if not isinstance(task, OpenMLClusteringTask):
+                task.class_labels = \
+                    dataset.retrieve_class_labels(task.target_name)
+                task.download_split()
```
mfeurer: Would it make sense to move this into the task classes? If they have a split and class labels they retrieve them, otherwise, they don't. Also, how does this work for regression tasks (regarding class labels)?

tests/test_tasks/test_task_functions.py
```diff
@@ -156,6 +156,11 @@ def test_get_task_with_cache(self):
         task = openml.tasks.get_task(1)
         self.assertIsInstance(task, OpenMLTask)
 
+    def test_get_task_clustering(self):
```
mfeurer: Could you please add a unit test for regression, too.

ArlindKadra: yes I can

openml/tasks/functions.py
ArlindKadra: Nice point, it does not work for regression. What do you think about this

```
if isinstance(task, OpenMLSupervisedTask):
    task.download_split()
    if isinstance(task, OpenMLClassificationTask): 
        task.class_labels = \
            dataset.retrieve_class_labels(task.target_name)

```

mfeurer: Agreed. Your proposed solution is fine for this.

PR #543, opened 2018-09-21
[WIP] refactor flow interface

vigsterkr:
@mfeurer check the interface design. i thought i will have time for the run_task part but that'll take a bit longer but i thought it's still better that you look at the interface.
the tests in `tests/test_flows/test_sklearn.py` all pass.

### DIFF
--- a/openml/flows/__init__.py
+++ b/openml/flows/__init__.py
@@ -1,7 +1,7 @@
 from .flow import OpenMLFlow, _copy_server_fields
 
-from .sklearn_converter import sklearn_to_flow, flow_to_sklearn, _check_n_jobs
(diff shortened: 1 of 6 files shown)


mfeurer:
Hey, thanks for putting this up! I just had a brief look and think it would be great if you could add some comments about the workflow. Currently, `from_workflow` is static, while `to_flow` is not, and there are multiple class attributes, so it would be great if you could explain those design decisions.

mfeurer:
Superseded by #647.

vigsterkr:
@mfeurer heheh cool. would have been cool to know it earlier :) we'll have to amend our local patch to the extension as there are some already working things on top of it... will try to send in that PR

PR #544, opened 2018-09-21
Moretasks

joaquinvanschoren:

It generalizes tasks so that you can handle different tasks differently.
Right now every task was assumed to be a classification task

Standard unit test

### DIFF
--- a/openml/tasks/task.py
+++ b/openml/tasks/task.py
@@ -141,14 +141,14 @@ class ClassificationTask(SupervisedTask):
 class RegressionTask(SupervisedTask):
     def __init__(self, task_id, task_type_id, task_type, data_set_id, estimation_procedure_type,
                  estimation_parameters, evaluation_measure, target_name, data_splits_url):
-        super(RegressionTask, self)).__init__(task_id, task_type_id, task_type, data_set_id, estimation_procedure_type,
+        super(RegressionTask, self).__init__(task_id, task_type_id, task_type, data_set_id, estimation_procedure_type,
                  estimation_parameters, evaluation_measure, target_name, data_splits_url)
 
 
 class ClusteringTask(OpenMLTask):
     def __init__(self, task_id, task_type_id, task_type, data_set_id, estimation_procedure_type,
                  estimation_parameters, evaluation_measure, number_of_clusters=None):
-        super(ClusteringTask, self)).__init__(task_id, task_type_id, task_type, data_set_id, estimation_procedure_type,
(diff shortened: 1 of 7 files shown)


openml/tasks/task.py
```diff
@@ -7,38 +7,58 @@
...
-    def __init__(self, task_id, task_type_id, task_type, data_set_id,
-                 target_name, estimation_procedure_type, data_splits_url,
-                 estimation_parameters, evaluation_measure, cost_matrix,
-                 class_labels=None):
+    def __init__(self, task_id, task_type_id, task_type, data_set_id, estimation_procedure_type,
+                 estimation_parameters, evaluation_measure):
         self.task_id = int(task_id)
         self.task_type_id = int(task_type_id)
         self.task_type = task_type
         self.dataset_id = int(data_set_id)
```
janvanrijn: can we call this "source_data", in co-ordinance with the task descriptions (hidden in the XML)

openml/tasks/task.py
```diff
@@ -7,38 +7,58 @@
...
-        self.target_name = target_name
         self.estimation_procedure = dict()
         self.estimation_procedure["type"] = estimation_procedure_type
-        self.estimation_procedure["data_splits_url"] = data_splits_url
         self.estimation_procedure["parameters"] = estimation_parameters
         #
         self.estimation_parameters = estimation_parameters
```
janvanrijn: I would rather store the estimation procedure id (not available in api right now, let's add it)

openml/tasks/task.py
```diff
@@ -7,38 +7,58 @@
...
-        self.split = None
-
-        if cost_matrix is not None:
-            raise NotImplementedError("Costmatrix")
 
     def get_dataset(self):
         """Download dataset associated with task"""
         return datasets.get_dataset(self.dataset_id)
 
+    def push_tag(self, tag):
```
janvanrijn: didn't we have general tag functions?

janvanrijn: Otherwise we should create them. (and not duplicate tag functions code for each entity)

mfeurer: Great remark. I think we should do this in a new PR, though, as it is not directly related to this PR. I opened issue #549 for this.

openml/tasks/task.py
mfeurer: @janvanrijn I'm afraid that won't be backwards compatible... But we can probably transit at some point to store something different in the cache. Do you want to open a new issue on the OpenML issue tracker?

Issue #540, opened 2018-09-20
Improve dataset uploading


mfeurer:
Reviewing PR #539 which tests dataset uploading, I found four things which we could improve to make dataset uploading nicer for the user:

1. There should be no necessity to provide the format as this will always be `'arff'`.
2. The current example encodes the class label as a `REAL`, although it should actually be a categorical.
3. We could add an example for uploading a list of lists, which could then also contain strings and categories.
4. The function `create_dataset` should be imported into the `openml.datasets` namespace.

@ArlindKadra @glemaitre

mfeurer:
After discussion with @ArlindKadra @joaquinvanschoren there are some more open points:

1. we need to check if the data is sparse and change the data format accordingly.
2. we need an example regarding the upload of sparse data
3. we need a unit test regarding the upload of sparse data

glemaitre:
> we need to check if the data is sparse and change the data format accordingly.

Do you want to change the format depending if the user pass an array (dataframe) or a sparse array (dataframe)?

glemaitre:
I think this is OK. It is what I did in the loading stage https://github.com/openml/openml-python/pull/548

amueller:
is ``create_datasets`` listed in the api docs? I can't see any docs on uploading datasets at all :-/

ArlindKadra:
There is a tutorial at the docs -> user guide section. Check the docs at
/develop/ instead of dev. We have to remove the old dev folder as the
documentation is being deployed differently.

amueller:
ah, I was looking at /develop/

amueller:
even more off-topic: is there a reason the example gallery is not in the top level menu and that progress and changelog link to the same thing?

amueller:
the create_dataset is still missing from the API docs

amueller:
#553 #554 #556

mfeurer:
This was fixed by #547

PR #545, opened 2018-09-21
[MRG] EHN: allow to upload DataFrame and infer dtype and column name

glemaitre:
Partially adressed #540 

Allow to use a pandas DataFrame when creating and publishing a dataset on OpenML

Important point to be considerd:

- categorical columns are not automatically converted to string and will instead raise an error
- boolean columns are converted to categorical column encoding True and False as string
- integer and real are used depending of the dtype to allow preserving the dtype when loading with liac-arff.
- a dict can be passed to override the dtype inference for some specific columns.

TODO:

- [x] Add unit tests for the exceptions


glemaitre:
The PEP8 error is kinda unrelated sine that we can't avoid the issue of the import on the top of the file with mock.

examples/create_upload_tutorial.py
```diff
@@ -5,17 +5,24 @@
...
 import sklearn.datasets
 
 ############################################################################
-# For this example we will upload to the test server to not  pollute the live server with countless copies of the same dataset.
+# For this example we will upload to the test server to not  pollute the live
+# server with countless copies of the same dataset.
 openml.config.server = 'https://test.openml.org/api/v1/xml'
 
+############################################################################
+# Uploading a data set store in a NumPy array
```
mfeurer: Should be store**d**

examples/create_upload_tutorial.py
```diff
@@ -77,8 +89,98 @@
...
-    paper_url='https://www.spiedigitallibrary.org/conference-proceedings-of-spie/1905/0000/Nuclear-feature-extraction-for-breast-tumor-diagnosis/10.1117/12.148698.short?SSO=1'
+    original_data_url=('https://archive.ics.uci.edu/ml/datasets/Breast+Cancer'
+                       '+Wisconsin+(Diagnostic)'),
+    paper_url=('https://www.spiedigitallibrary.org/conference-proceedings-of'
+               '-spie/1905/0000/Nuclear-feature-extraction-for-breast-tumor-'
+               'diagnosis/10.1117/12.148698.short?SSO=1')
+)
+
+############################################################################
+try:
```
mfeurer: I actually think these `try/excepts` are no longer necessary and only add little extra information for the user. Could you please remove them?

examples/create_upload_tutorial.py
```diff
@@ -77,8 +89,98 @@
...
+df['play'] = df['play'].astype('category')
+print(df.info())
+
+############################################################################
+# We enforce the column 'outlook', 'winday', and 'play' to be a categorical
+# dtype while the column 'rnd_str' is kept as a string column. Then, we can
+# call :func:`create_dataset` by passing the dataframe and fixing the parameter
+# ``attributes`` to ``'auto'``.
+
+# force OpenML to infer the attributes from the dataframe
```
mfeurer: Is there a reason to define these arguments outside of the function you'll be calling?

tests/test_datasets/test_dataset_functions.py
```diff
@@ -379,3 +380,112 @@ def test_create_dataset_numpy(self):
             paper_url=paper_url
         )
         dataset.publish()
+
+    def test_create_dataset_pandas(self):
+        # pandas is only a optional dependency and we need to skip the test if
+        # it is not installed.
```
mfeurer: In the long run we will probably make pandas required (it is already required for the examples and documentation), so I'd totally okay if you add it as a dependency.

tests/test_datasets/test_dataset_functions.py
```diff
@@ -379,3 +380,112 @@ def test_create_dataset_numpy(self):
...
+            ignore_attribute=None,
+            citation=citation,
+            attributes='auto',
+            data=df,
+            format='arff',
+            version_label='test',
+            original_data_url=original_data_url,
+            paper_url=paper_url
+        )
+        dataset.publish()
```
mfeurer: Could you please download and check that everything was converted and uploaded properly?

examples/create_upload_tutorial.py
glemaitre: I found it more didactic to have a comment and an assignment instead of just assigning inside the function. WDYT?

tests/test_datasets/test_dataset_functions.py
glemaitre: I assumed that I can do that with something like:

```python
upload_id = dataset.publish()
dataset_uploaded = openml.datasets.get_dataset(upload_id)
```

However, I receive the following exception:

```pytb
openml.exceptions.OpenMLServerException: https://test.openml.org/api/v1/xml/data/features/5915 returned code 273: Dataset not processed yet
```

What is the usual way that you are doing it.

glemaitre: One way could be something like that:

```python
        upload_id = dataset.publish()
        while True:
            try:
                dataset_uploaded = openml.datasets.get_dataset(upload_id)
            except OpenMLServerException as msg_exception:
                if "code 273" in str(msg_exception):
                    pass
                else:
                    raise msg_exception
```

but I am not sure how long it can take for the dataset to be processed and we could get in some timeout nightmare.

ArlindKadra: @glemaitre We already had this problem in #547. We decided to only compare based on the arff and we also implemented 2 functions which get the arff and format based on the id (those are available). I have to refactor my pr a bit more, but it should be available tomorrow.

glemaitre: So do you mean to use this function:

https://github.com/openml/openml-python/pull/547/files#diff-2ffad3a1ae43bbcc0e841e2f1df8967bR692

So I would probably wait that your PR get first in and I will manage the conflict then.

ArlindKadra: yes, @mfeurer are you okay with this?

openml/datasets/functions.py
```diff
@@ -353,6 +353,39 @@ def get_dataset(dataset_id):
...
+        # categorical column can contain mixed type and we should therefore
+        # raise an error asking to convert all enties to string.
+        if not all([isinstance(cat, six.string_types) for cat in arff_dtype]):
+            raise ValueError("The column '{}' of the dataframe is of "
+                             "'category' dtype. Therefore, all values in this "
+                             "columns should be string. Please convert the "
+                             "entries which are not string."
+                             .format(column_name))
+    elif column_dtype.name == 'object':
+        arff_dtype = 'STRING'
```
amueller: Usually we want this to be category as well, right? kinda tricky ;)
And what about integers that encode categories?

glemaitre: > And what about integers that encode categories?

If they encode categories, then one should use the `category` dtype for those columns.

However, this is part needs to be modified. We should manage `INTEGER` and `REAL`/`NUMERIC` differently. We also have to think more in depth about the missing values. I'll use the expertise of @jorisvandenbossche

Basically, we are working on uploading the datasets from `dirty_cat` and we are confronted to those issues.

openml/datasets/functions.py
```diff
@@ -411,10 +447,25 @@ def create_dataset(name, description, creator, contributor, collection_date,
...
+
+    if attributes == 'auto':
+        if not hasattr(data, "columns"):
+            raise ValueError("Automatically inferring the attributes required "
+                             "a pandas DataFrame. A {!r} was given instead."
+                             .format(data))
+        # infer the type of data for each column of the DataFrame
+        attributes_ = [(col_name,
+                        _pandas_dtype_to_arff_dtype(data, col_name, col_dtype))
+                       for col_name, col_dtype in data.dtypes.iteritems()]
```
jorisvandenbossche: you don't necessarily need to pass the dtype here, you could also iterate only through the column names

openml/datasets/functions.py
jorisvandenbossche: Yes, I would also say that it is up to the user to use a categorical dtype. Actual string columns are not that rare (eg titanic dataset), and it is quite impossible to guess from a string column if it should be categorical or not. 

Of course, it could also be a keyword that the user can pass to indicate which columns to see as categorical without the need that the (s)he does the actual conversion themselves.

>  And what about integers that encode categories?

I think arff does not support integer categories?

examples/create_upload_tutorial.py
mfeurer: The current example has comments inside the function so everything is in one place. Not sure which one is better though as the current version has a pretty long function call.

tests/test_datasets/test_dataset_functions.py
mfeurer: This looks good at first glance.

openml/datasets/functions.py
mfeurer: Conceptually, arff should allow this (as long as the categories are integers, too). However, liac-arff does not support this. Can be reproduced with:
```
import arff

arff.dumps({
    'relation': 'test',
    'description': 'abc',
    'attributes': [['test', [1, 2, 3]]],
    'data': [[1], [2]],
})
```
I'm wondering if it's worth changing this. Formally, there's also no difference between REAL, INTEGER and NUMERIC, but I don't see a reason to not be more specific about the data type as the arff specification allows this.

glemaitre: I think that there is a difference between REAL and INTEGER since the dtype will be kept, isn't it.

```python
In [5]: X = arff.dumps({ 
   ...:   'relation': 'test', 
   ...:   'description': 'abc', 
   ...:   'attributes': [['test', 'REAL']], 
   ...:   'data': [[1], [2]] 
   ...: })                                                                                              

In [6]: arff.loads(X)                                                                                   
Out[6]: 
{'description': 'abc',
 'relation': 'test',
 'attributes': [('test', 'REAL')],
 'data': [[1.0], [2.0]]}

In [7]: X = arff.dumps({ 
   ...:   'relation': 'test', 
   ...:   'description': 'abc', 
   ...:   'attributes': [['test', 'INTEGER']], 
   ...:   'data': [[1], [2]] 
   ...: })                                                                                              

In [8]: arff.loads(X)                                                                                   
Out[8]: 
{'description': 'abc',
 'relation': 'test',
 'attributes': [('test', 'INTEGER')],
 'data': [[1], [2]]}
```

janvanrijn: Going by the official arff specification, all numeric attributes should go by NUMERIC 
https://www.cs.waikato.ac.nz/ml/weka/arff.html

In order to make OpenML datasets comply with as much as possible arff parsers, maybe it's best to enforce this too server side? In that case the Python uploader should of course also upload NUMERIC key. @mfeurer @joaquinvanschoren @berndbischl WDYT?

janvanrijn: Have to add that currently this is not enforced, server side

glemaitre:
I have to add much more tests but basically I change the way the dtype inference was done using pandas (only available for pandas >= 0.19, I think):
https://github.com/openml/openml-python/pull/545/files#diff-2ffad3a1ae43bbcc0e841e2f1df8967bR357

It should be robust to find the right dtype even in presence of NA values. It should also be extensible to DATE (datetime) dtype since pandas is also inferring those. I see that there is a PR in liac-arff which could be worth to push forward ;)

glemaitre:
The error on the CI is a bit cryptic to me. Anyone has an idea why is it failing only for some of the build and not others. I can use this function locally as well.

openml/datasets/functions.py
jorisvandenbossche: @janvanrijn you pointed to an outdated version of the arff standard (see the first sentence on the page you linked to), the newer one (https://waikato.github.io/weka-wiki/arff_stable/) actually mentions 'integer' and 'real' (although it says they are treated as numeric)

janvanrijn:
> The error on the CI is a bit cryptic to me. Anyone has an idea why is it failing only for some of the build and not others. I can use this function locally as well.

I also encountered this error. Opened issue #565 and PR #566. If you merge #566 into this branch it should be fixed.

examples/create_upload_tutorial.py
mfeurer: I discussed this with @ArlindKadra regarding #547 to have this based on the length. But that would distribute the creation of the arguments. I start to see why creating them beforehand is a good idea.

tests/test_datasets/test_dataset_functions.py
mfeurer: This could be deleted now, right?

openml/datasets/functions.py
```diff
@@ -353,6 +354,57 @@ def get_dataset(dataset_id):
...
+        # dropped before the inference instead.
+        column_dtype = pd.api.types.infer_dtype(df[column_name].dropna())
+
+        if column_dtype == 'categorical':
+            # for categorical feature, arff expects a list string. However, a
+            # categorical column can contain mixed type and we should therefore
+            # raise an error asking to convert all entries to string.
+            categories = df[column_name].cat.categories
+            categories_dtype = pd.api.types.infer_dtype(categories)
+            if categories_dtype != 'string':
```
jorisvandenbossche: You might need to allow 'unicode' here as well for python 2 compat ?

openml/datasets/functions.py
```diff
@@ -353,6 +354,57 @@ def get_dataset(dataset_id):
...
+        elif column_dtype == 'boolean':
+            # boolean are encoded as categorical.
+            attributes_arff.append((column_name, ['True', 'False']))
+        elif column_dtype in PD_DTYPES_TO_ARFF_DTYPE.keys():
+            attributes_arff.append((column_name,
+                                    PD_DTYPES_TO_ARFF_DTYPE[column_dtype]))
+        else:
+            raise ValueError("The dtype '{}' of the column '{}' is not "
+                             "currently supported by liac-arff. Supported "
+                             "dtypes are categorical, string, interger, "
```
jorisvandenbossche: interger -> integer

openml/datasets/functions.py
```diff
@@ -411,10 +468,29 @@ def create_dataset(name, description, creator, contributor, collection_date,
...
+
+    if attributes == 'auto' or isinstance(attributes, dict):
+        if not hasattr(data, "columns"):
+            raise ValueError("Automatically inferring the attributes required "
+                             "a pandas DataFrame. A {!r} was given instead."
+                             .format(data))
+        # infer the type of data for each column of the DataFrame
+        attributes_ = attributes_arff_from_df(data)
+        if isinstance(attributes, dict):
+            # override the attributes which was specified by the user
```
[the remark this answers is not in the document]
glemaitre: Yes. It is the mechanism that allow the user to overwrite some specific attribute. So it could be useful to force a specific data type or specify the categories (e.g. if there is some missing categories in the data column).

glemaitre:
@mfeurer @janvanrijn @ArlindKadra @amueller 

The PR is ready for a full round of review. With the previous work it should be almost ready to be merged (the CI should be green).

openml/datasets/functions.py
```diff
@@ -359,6 +361,57 @@ def get_dataset(dataset_id):
...
+        'floating': 'REAL',
+        'string': 'STRING'
+    }
+    attributes_arff = []
+    for column_name in df:
+        # skipna=True does not infer properly the dtype. The NA values are
+        # dropped before the inference instead.
+        column_dtype = pd.api.types.infer_dtype(df[column_name].dropna())
+
+        if column_dtype == 'categorical':
```
janvanrijn: I missed the discussion in the thread. is this a regular pandas data type? Is there any reason to not use the dtype str? 

(the comments below do not seem to help me too much, sorry)

openml/datasets/functions.py
```diff
@@ -359,6 +361,57 @@ def get_dataset(dataset_id):
...
+            categories_dtype = pd.api.types.infer_dtype(categories)
+            if categories_dtype not in ('string', 'unicode'):
+                raise ValueError("The column '{}' of the dataframe is of "
+                                 "'category' dtype. Therefore, all values in "
+                                 "this columns should be string. Please "
+                                 "convert the entries which are not string. "
+                                 "Got {} dtype in this column."
+                                 .format(column_name, categories_dtype))
+            attributes_arff.append((column_name, categories.tolist()))
+        elif column_dtype == 'boolean':
```
janvanrijn: why not use dtype=bool?

openml/datasets/functions.py
```diff
@@ -423,6 +481,24 @@ def create_dataset(name, description, creator, contributor,
...
+        if isinstance(attributes, dict):
+            # override the attributes which was specified by the user
+            for attr_idx in range(len(attributes_)):
+                attr_name = attributes_[attr_idx][0]
+                if attr_name in attributes.keys():
+                    attributes_[attr_idx] = (attr_name, attributes[attr_name])
+    else:
+        attributes_ = attributes
+
+    data = data.values if hasattr(data, "columns") else data
```
janvanrijn: This seems to imply that internally we store everything as arff. Wouldn't it be much cooler to store everything internally as pandas, that will make it much easier to do operations on the datasets. (And of course convert before uploading dataset / performing a run). However maybe that should be a next PR.

openml/datasets/functions.py
glemaitre: This is an available dtype in pandas to mention that a column is a categorical column.

> Is there any reason to not use the dtype str?

Actually I don't see why categories should be string :). Right now, the liac-arff does not allow otherwise, we could actually think about converting silently the categories to string if a column is of category dtype.

openml/datasets/functions.py
glemaitre: because we are using the `pd.api.types.infer_dtype` function which return string.

openml/datasets/functions.py
glemaitre: I would say that this is another PR.

I would imagine that it would require quite some changes but if the consensus is to use pandas then why not. Then, the data should be only tabular data (which is probably the main use case in openml)

openml/datasets/functions.py
janvanrijn: Cool, thanks for the clarification :)

tests/test_datasets/test_dataset_functions.py
```diff
@@ -671,3 +715,126 @@ def test_get_online_dataset_format(self):
...
+            paper_url=paper_url
+        )
+        upload_did = dataset.publish()
+        self.assertEqual(
+            _get_online_dataset_arff(upload_did),
+            dataset._dataset,
+            "Uploaded ARFF does not match original one"
+        )
+
+        # Check that we can overwrite the attributes
```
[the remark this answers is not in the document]
glemaitre: line 797, we checked that the attributes contain `'g'` which is not present in the original data. I would consider that as an assert. Do you want to test something else?

glemaitre:
Do you add entry to document changes in the library like what's new or changelog (I see a `progress.rst` maybe)?

tests/test_datasets/test_dataset_functions.py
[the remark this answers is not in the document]
mfeurer: Do you mean this?
```
        self.assertEqual(
            _get_online_dataset_arff(upload_did),
            dataset._dataset,
            "Uploaded ARFF does not match original one"
        )
```
Either I'm missing something, or there is no strict check that `'g'` is in the attributes of the downloaded file? Line 797 only checks that the download is equal to the upload, right?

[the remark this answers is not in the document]
glemaitre: OK I see we can add a strict check then.

glemaitre:
Anything else?

openml/datasets/functions.py
amueller: Sparse data might be an issue, right?

glemaitre: Sparse arrays do not have `values` attribute so we should be fine.

If you think about SparseDataFrame, I would say that we should not offer support. They are going to be [deprecated](https://github.com/pandas-dev/pandas/issues/19239) in favor of sparse dtype.

Is the liac-arff actually usable with sparse data?

glemaitre:
#response_container_BBPPID{font-family: initial; font-size:initial; color: initial;}  So it will not. Be an issue here                                                                                                                                                                        Sent from my phone - sorry to be brief and potential misspell.                                                                                                                                                                            From: notifications@github.comSent: 23 October 2018 09:26To: openml-python@noreply.github.comReply to: reply@reply.github.comCc: g.lemaitre58@gmail.com; mention@noreply.github.comSubject: Re: [openml/openml-python] [MRG] EHN: allow to upload DataFrame and infer dtype and column name (#545)  @mfeurer commented on this pull request.



In openml/datasets/functions.py:
> +        if not hasattr(data, "columns"):
+            raise ValueError("Automatically inferring the attributes required "
+                             "a pandas DataFrame. A {!r} was given instead."
+                             .format(data))
+        # infer the type of data for each column of the DataFrame
+        attributes_ = attributes_arff_from_df(data)
+        if isinstance(attributes, dict):
+            # override the attributes which was specified by the user
+            for attr_idx in range(len(attributes_)):
+                attr_name = attributes_[attr_idx][0]
+                if attr_name in attributes.keys():
+                    attributes_[attr_idx] = (attr_name, attributes[attr_name])
+    else:
+        attributes_ = attributes
+
+    data = data.values if hasattr(data, "columns") else data

Yes, it can handle scipy's COO data structure.

openml/datasets/functions.py
glemaitre: So this will not be an issue. @amueller could you elaborate on your thoughts.

mfeurer:
@janvanrijn @LishengSun

tests/test_datasets/test_dataset_functions.py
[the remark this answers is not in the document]
glemaitre: Actually we could support `SparseDataFrame` for now. Calling `to_coo` instead of `values` should be enough then.

glemaitre:
I can do the support for SparseDataFrame in an upcoming PR.

PR #547, opened 2018-09-21
Issue 540

ArlindKadra:
#### Reference Issue
#540 
Makes the arff attribute optional for the dataset, fixes the target type in the dataset publish doc to Nominal from Real. create_dataset is imported into the openml.datasets namespace and a unit test to upload a dataset as a list of lists is added.

Added unit test, other changes by the default ones.

### DIFF
--- a/examples/create_upload_tutorial.py
+++ b/examples/create_upload_tutorial.py
@@ -167,7 +167,7 @@ citation = (
 )
 
 weather_dataset = create_dataset(
-    name="Wind",
+    name="Weather",
     description=description,
     creator='I. H. Witten, E. Frank, M. A. Hall, and ITPro',
     contributor=None,
--- a/tests/test_datasets/test_dataset_functions.py
+++ b/tests/test_datasets/test_dataset_functions.py
@@ -480,7 +480,7 @@ class TestOpenMLDataset(TestBase):
         ]
 
         dataset = create_dataset(
-            name="Wind_dataset",
+            name="ModifiedWeather",
             description=(
                 'Testing dataset upload when the data is a list of lists'
             ),


examples/create_upload_tutorial.py
```diff
@@ -30,8 +30,8 @@
 data = np.concatenate((X, y.reshape((-1, 1))), axis=1)
 attribute_names = list(attribute_names)
 attributes = [
-    (attribute_name, 'REAL') for attribute_name in attribute_names
-] + [('class', 'REAL')]
+                 (attribute_name, 'REAL') for attribute_name in attribute_names
+             ] + [('class', 'INTEGER')]
```
mfeurer: Could you please rename this to 'target' as it is actually a regression dataset?

ArlindKadra: http://scikit-learn.org/stable/modules/generated/sklearn.datasets.load_breast_cancer.html 
It is classification

ArlindKadra:
@mfeurer when you have time you can take a look at the changes.

.travis.yml
```diff
@@ -23,6 +23,8 @@ env:
   - DISTRIB="conda" PYTHON_VERSION="3.6" SKLEARN_VERSION="0.18.2" RUN_FLAKE8="true" SKIP_TESTS="true"
   - DISTRIB="conda" PYTHON_VERSION="3.7" SKLEARN_VERSION="0.19.2"
 
+before_install:
+ - python -c "import fcntl; fcntl.fcntl(1, fcntl.F_SETFL, 0)"
```
glemaitre: What is it useful for?

openml/datasets/__init__.py
```diff
@@ -1,7 +1,14 @@
...
-__all__ = ['check_datasets_active', 'get_dataset', 'get_datasets',
-           'OpenMLDataset', 'OpenMLDataFeature', 'list_datasets']
+__all__ = [
+    'check_datasets_active',
+    'get_dataset',
+    'get_datasets',
+    'OpenMLDataset',
+    'OpenMLDataFeature',
+    'list_datasets',
+    'create_dataset',
```
glemaitre: We can organize those in alphabetic order since that we have a diff.

openml/datasets/dataset.py
```diff
@@ -86,23 +84,23 @@ class OpenMLDataset(object):
     dataset: string, optional
         Serialized arff dataset string.
     """
-    def __init__(self, name, description, format, dataset_id=None,
+    def __init__(self, name, description, data_format='arff', dataset_id=None,
```
glemaitre: Do we care about backward compatibility? Changing the name would require a deprecation warning.

openml/datasets/dataset.py
```diff
@@ -208,15 +206,31 @@ def remove_tag(self, tag):
...
-        ):
-            return True
         else:
-            return False
+            for field in self.__dict__:
+                if field not in server_fields:
+                    if field in other.__dict__:
+                        if self.__dict__[field] != other.__dict__[field]:
+                            return False
+                    else:
```
glemaitre: I personnally find that the following will be more readable:

```python
        server_fields = {'dataset_id', 'version', 'upload_data', 'url',
                         'dataset', 'data_file'}

        if type(other) != OpenMLDataset:
            return False

        # check that the keys are identical
        self_keys = set(self.__dict__.keys()).difference_update(server_fields)
        other_keys = set(other.__dict__.keys()).difference_update(server_fields)
        if self_keys != other_keys:
            return False

        # check that values of the common keys are identical
        return all(self.__dict__[key] == other.__dict__[key]
                   for key in self_keys)
```

openml/datasets/dataset.py
```diff
@@ -30,8 +30,6 @@ class OpenMLDataset(object):
         Name of the dataset.
     description : str
         Description of the dataset.
-    format : str
```
glemaitre: You need to keep the documentation for the new parameter.
If it is going to be deprecated, we can use `.. deprecated:: xxx` from sphinx

openml/datasets/functions.py
```diff
@@ -3,10 +3,10 @@
...
 import os
 import re
-import shutil
+import numpy as np
 import six
 import arff
-
+from scipy.sparse import coo_matrix
 from oslo_concurrency import lockutils
 import xmltodict
```
glemaitre: move this import above the `from xxx`

openml/datasets/functions.py
```diff
@@ -355,7 +355,7 @@ def get_dataset(dataset_id):
 
 def create_dataset(name, description, creator, contributor, collection_date,
                    language, licence, attributes, data, default_target_attribute,
-                   row_id_attribute, ignore_attribute, citation, format="arff",
+                   row_id_attribute, ignore_attribute, citation,
```
glemaitre: Same question as before: shall we deprecate the parameters instead of removing it?

openml/datasets/functions.py
```diff
@@ -418,22 +416,62 @@ def create_dataset(name, description, creator, contributor, collection_date,
         'data': data
     }
 
+    # Determine arff format from the dataset
```
glemaitre: move this block before building the `arff_object`

openml/datasets/functions.py
```diff
@@ -418,22 +416,62 @@ def create_dataset(name, description, creator, contributor, collection_date,
...
+    if isinstance(data, list):
+        if isinstance(data[0], list):
+            d_format = 'arff'
+        elif isinstance(data[0], dict):
+            d_format = 'sparse_arff'
+        else:
+            raise ValueError('Illegal dataset value, '
```
glemaitre: I would be a bit more specific:

```
raise ValueError(
    "When giving a list, the list should contain a list for dense "
    "data or a dictionary for sparse data. Got {!r} instead."
    .format(data[0])
)

File: openml/datasets/functions.py [hunk 11]
```diff
@@ -418,22 +416,62 @@ def create_dataset(name, description, creator, contributor, collection_date,
         'data': data
     }
 
+    # Determine arff format from the dataset
+    if isinstance(data, list):
+        if isinstance(data[0], list):
```
glemaitre: Actually you need to check all the element of data. If I pass:

```python
data = [[1, 2], {'whatever': [0, 1]}]
```
this will not raise an error.

openml/datasets/functions.py
```diff
@@ -418,22 +416,62 @@ def create_dataset(name, description, creator, contributor, collection_date,
...
+            d_format = 'sparse_arff'
+        else:
+            raise ValueError('Illegal dataset value, '
+                             'only list of lists/dicts is supported')
+    elif isinstance(data, np.ndarray):
+        d_format = 'arff'
+    elif isinstance(data, coo_matrix):
+        d_format = 'sparse_arff'
+    else:
+        raise ValueError('Illegal dataset value, '
```
glemaitre: I would have a more specific message here as well.

openml/datasets/functions.py
```diff
@@ -418,22 +416,62 @@ def create_dataset(name, description, creator, contributor, collection_date,
...
+                         'please check the function documentation')
+
     # serializes the arff dataset object and returns a string
     arff_dataset = arff.dumps(arff_object)
     try:
         # check if arff is valid
         decoder = arff.ArffDecoder()
-        decoder.decode(arff_dataset, encode_nominal=True)
+        if d_format == 'arff':
```
glemaitre: It might be shorter to write:

```python
return_type = arff.COO if d_format == 'sparse_arff' else arff.DENSE
decoder.decode(arff_dataset, encode_nominal=True, return_type=return_type)
```

openml/datasets/functions.py
```diff
@@ -621,28 +659,74 @@ def _create_dataset_from_description(description, features, qualities, arff_file
...
+        md5_checksum=description.get("oml:md5_checksum"),
         data_file=arff_file,
         features=features,
-        qualities=qualities)
+        qualities=qualities,
+    )
     return dataset
+
+
+def _get_online_dataset_arff(did):
```
glemaitre: Let's be specific with the name `dataset_id` would be nice :)

openml/datasets/functions.py
```diff
@@ -621,28 +659,74 @@ def _create_dataset_from_description(description, features, qualities, arff_file
...
+    from the OpenML website.
+
+    Parameters
+    ----------
+    did : int
+        A dataset id.
+
+    Returns
+    -------
+    str
```
glemaitre:
```
arff_dataset : str
    A string representation of an ARFF file.
```

openml/datasets/functions.py
```diff
@@ -621,28 +659,74 @@ def _create_dataset_from_description(description, features, qualities, arff_file
...
         data_file=arff_file,
         features=features,
-        qualities=qualities)
+        qualities=qualities,
+    )
     return dataset
+
+
+def _get_online_dataset_arff(did):
+    """Download the arff file for a given dataset id
```
glemaitre: arff -> ARFF

openml/datasets/functions.py
```diff
@@ -621,28 +659,74 @@ def _create_dataset_from_description(description, features, qualities, arff_file
...
+    """
+    dataset_xml = openml._api_calls._perform_api_call("data/%d" % did)
+    # build a dict from the xml.
+    # use the url from the dataset description and return the arff string
+    return openml._api_calls._read_url(
+        xmltodict.parse(dataset_xml)['oml:data_set_description']['oml:url']
+    )
+
+
+def _get_online_dataset_format(did):
```
glemaitre: the same comments than above

openml/datasets/functions.py
```diff
@@ -621,28 +659,74 @@ def _create_dataset_from_description(description, features, qualities, arff_file
...
+    from the OpenML website.
+
+    Parameters
+    ----------
+    did : int
+        A dataset id.
+
+    Returns
+    -------
+    str
```
glemaitre: same comments than above

openml/datasets/functions.py
```diff
@@ -621,28 +659,74 @@ def _create_dataset_from_description(description, features, qualities, arff_file
...
+        Dataset format.
+    """
+    dataset_xml = openml._api_calls._perform_api_call("data/%d" % did)
+    # build a dict from the xml and get the format from the dataset description
+    return xmltodict\
```
glemaitre: I am not a big fan of `\`. I would instead break the line in the paranthesis or reaffect the name of the keys into 2 variables. But this is more personnal thing I assume

tests/test_datasets/test_dataset_functions.py
```diff
@@ -31,7 +31,9 @@
                                        _get_dataset_arff,
                                        _get_dataset_features,
                                        _get_dataset_qualities,
-                                       DATASETS_CACHE_DIR_NAME)
+                                       DATASETS_CACHE_DIR_NAME,
+                                       _get_online_dataset_arff,
```
glemaitre: put them in alphabetic order

tests/test_datasets/test_dataset_functions.py
```diff
@@ -317,12 +319,19 @@ def test_deletion_of_cache_dir_faulty_download(self, patch):
...
+
         openml.datasets.get_dataset(3)
         file_path = os.path.join(openml.config.get_cache_directory(),
                                  "datasets", "3", "dataset.arff")
         dataset = OpenMLDataset(
-            "anneal", "test", "ARFF",
-            version=1, licence="public", default_target_attribute="class", data_file=file_path)
+            "anneal",
+            "test",
+            data_format="ARFF",
```
glemaitre: uhm, it should not be `"arff"` in lower case even this is converted after.

tests/test_datasets/test_dataset_functions.py
```diff
@@ -335,47 +344,248 @@ def test__retrieve_class_labels(self):
         self.assertEqual(labels, ['C', 'H', 'G'])
 
     def test_upload_dataset_with_url(self):
+
         dataset = OpenMLDataset(
-            "UploadTestWithURL", "test", "ARFF",
+            "UploadTestWithURL",
+            "test",
+            data_format="ARFF",
```
glemaitre: lower case?

glemaitre:
In addition, check:

https://codecov.io/gh/openml/openml-python/pull/547/diff

Basically a test checking the equality between OpenMLDataset(s) is not tested

openml/datasets/dataset.py
ArlindKadra: You are right, in a previous discussion we decided to remove format and I removed it from the documentation, it should be there along with the message.

openml/datasets/dataset.py
ArlindKadra: I am not sure about this one, it looks the other way around to me.

openml/datasets/functions.py
ArlindKadra: True, this way if there is a problem with the data we skip the object creation and gc.

openml/datasets/functions.py
ArlindKadra: Theoretically the liac-arff should throw an error regarding this. Just to be sure, I will check later "today". If it does, then I would probably try to avoid going through each row of the data.
The code above just wants to infer the format, I added the error raising statements since that is information that we already have from deciding on the format.

openml/datasets/functions.py
ArlindKadra: If I am not mistaken the convention is to use variable: value in the docs. Following that, I am not storing the result anywhere, that is why there is no variable name before str.

tests/test_datasets/test_dataset_functions.py
ArlindKadra: you mean it should be "arff".

ArlindKadra:
> In addition, check:
> 
> https://codecov.io/gh/openml/openml-python/pull/547/diff
> 
> Basically a test checking the equality between OpenMLDataset(s) is missing

@glemaitre I agree, that would be a good issue. Not sure if we should handle it in this PR. The reason that I changed the eq method in this PR is that originally I planned to use it and the way it was did not seem functional.

tests/test_datasets/test_dataset_functions.py
ArlindKadra: yep

openml/datasets/functions.py
ArlindKadra: Neither am I. Just trying to keep it as readable as possible.

openml/datasets/functions.py
glemaitre: OK so now it makes sense to let the construction of the `arff_object` above since it will serve of checking for the data variable.

openml/datasets/functions.py
glemaitre: Good point, I was not aware of the convention but you are right.

tests/test_datasets/test_dataset_functions.py
glemaitre: Yep

openml/datasets/functions.py
ArlindKadra: I checked and liac-arff does throw an error for the example above. I would say no, it is worth it to keep the checks which determine the format of the ARFF before the object creation, that way if the most simplest of cases are violated we throw an error and there is no arff object to be collected from gc.
Liac-arff can take care of the rest.

openml/datasets/functions.py
```diff
@@ -370,6 +375,8 @@ def create_dataset(name, description, creator, contributor, collection_date,
         Name of the dataset.
     description : str
         Description of the dataset.
+    format : str, optional
+        Format of the dataset. Only 'arff' for now.
```
glemaitre: "Only 'arff' for now" is not correct anymore.

openml/datasets/dataset.py
ArlindKadra: Now that I took a second look at this, format in my understanding is written as the attribute of the class, the attribute of the class is still there and it is not changed. However I changed the name of the parameter given to the init because it shadowed the builtin format. So a deprecation warning should do the trick inside the init.

tests/test_datasets/test_dataset.py
```diff
@@ -1,10 +1,11 @@
 import numpy as np
-from scipy import sparse
 import six
+import openml
```
mfeurer: Please rearrange the imports to be PEP8 compliant.

tests/test_datasets/test_dataset_functions.py
```diff
@@ -1,38 +1,36 @@
...
+import os
+import sys
+import numpy as np
+import scipy.sparse
+import random
+import six
+if sys.version_info[0] >= 3:
+    from unittest import mock
+else:
+    import mock
```
mfeurer: Please undo this reordering.

openml/datasets/dataset.py
```diff
@@ -208,15 +206,31 @@ def remove_tag(self, tag):
         openml._api_calls._perform_api_call("/data/untag", data=data)
 
     def __eq__(self, other):
+
+        server_fields = {
```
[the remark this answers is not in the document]
glemaitre: set would be better IMO :P

I still like my solution if this is doing the same thing :)
https://github.com/openml/openml-python/pull/547#discussion_r223218288

openml/datasets/dataset.py
```diff
@@ -1,20 +1,20 @@
-from collections import OrderedDict
+import arff
```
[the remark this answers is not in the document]
glemaitre: Actually this is PEP8 rule regarding the order:
https://www.python.org/dev/peps/pep-0008/#imports

What is usually impossible to follow is all import on top of file when you need to make a try .. except ...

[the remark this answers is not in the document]
mfeurer: Hm, I actually wanted to say that the order should follow:

1. Standard library imports.
2. Related third party imports.
3. Local application/library specific imports.

openml/datasets/functions.py
```diff
@@ -609,40 +667,86 @@ def _create_dataset_from_description(description, features, qualities, arff_file
...
+    Returns
+    -------
+    str
+        Dataset format.
+    """
+    dataset_xml = openml._api_calls._perform_api_call("data/%d" % dataset_id)
+    # build a dict from the xml and get the format from the dataset description
+    return xmltodict\
+        .parse(dataset_xml)['oml:data_set_description']['oml:format']\
+        .lower()
```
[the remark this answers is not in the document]
ArlindKadra: https://github.com/openml/openml-python/blob/e711267cc4acdfa82610e880fb7307ec9959a2bd/tests/test_datasets/test_dataset_functions.py#L468-L477

https://github.com/openml/openml-python/blob/e711267cc4acdfa82610e880fb7307ec9959a2bd/tests/test_datasets/test_dataset_functions.py#L511-L520

@mfeurer Now that I am thinking about it, the above checks are happening in test_create_dataset_sparse and test_create_dataset_list. Although they check that the datasets are uploaded correctly, they also check that the functions work correctly. In my opinion this already covers the functions, any other test where we do not upload, in the future might fail if information is removed from the test server.

examples/create_upload_tutorial.py
```diff
@@ -7,39 +7,70 @@
...
+# ========================
+#
+# Prepare dataset
+# ^^^^^^^^^^^^^^^
 # Load an example dataset from scikit-learn which we will upload to OpenML.org via the API.
+
 breast_cancer = sklearn.datasets.load_breast_cancer()
 name = 'BreastCancer(scikit-learn)'
-X = breast_cancer.data
+x = breast_cancer.data
```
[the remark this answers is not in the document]
ArlindKadra: but why do we want to use an uppercase variable name ?

.travis.yml
```diff
@@ -23,6 +23,8 @@ env:
   - DISTRIB="conda" PYTHON_VERSION="3.6" SKLEARN_VERSION="0.18.2" RUN_FLAKE8="true" SKIP_TESTS="true"
   - DISTRIB="conda" PYTHON_VERSION="3.7" SKLEARN_VERSION="0.19.2"
 
+before_install:
```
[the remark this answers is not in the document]
ArlindKadra: yes, I missed doing that.

examples/create_upload_tutorial.py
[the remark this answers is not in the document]
mfeurer: Convention. All ML books (I'm aware of) use upper case bold face variables for matrices. We can't use boldface, but we can at least write it upper case.

openml/datasets/functions.py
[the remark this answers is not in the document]
mfeurer: True, but this also suboptimal because it is not immediately clear where this is tested. How about a simple test like for all the other functions which retrieve dataset related stuff?

examples/create_upload_tutorial.py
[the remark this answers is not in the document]
glemaitre: I would say that this is more a scikit-learn convention (it is used in example and in all methods of estimators).
The other solution is to call `X` -> `data` and `y` -> `label` and you remove this issue and it is explicit.

examples/create_upload_tutorial.py
```diff
@@ -7,39 +7,70 @@
...
 y = breast_cancer.target
+target_names = breast_cancer.target_names
+y = np.array([target_names[i] for i in y])
```
glemaitre: I would force the array to be `object` dtype since this is containing strings.

examples/create_upload_tutorial.py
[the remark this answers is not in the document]
mfeurer: That's a good point, but we should do this in a separate PR.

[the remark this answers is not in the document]
ArlindKadra: @mfeurer then we should probably make a note also on the issue that the PR will solve so that it removes the ignore to capital variable names ?

openml/datasets/dataset.py
ArlindKadra: so, the reason that I think this is not good is that supposing that the object has 'size' fields. 
And that the server fields are of a size t.
`self_keys = set(self.__dict__.keys()).difference_update(server_fields) `
for this you have a complexity of 't'
`other_keys = set(other.__dict__.keys()).difference_update(server_fields)`
for this you have the complexity too
```
 if self_keys != other_keys:
            return False
```
Taking a guess and saying this has 'size-t' complexity checking if size is equal and going through each element.
```
return all(self.__dict__[key] == other.__dict__[key]
                   for key in self_keys)
```
You iterate over each key left after removing the server fields, complexity 'size-t'.
You then call all which still iterates over each item on the iterable, complexity 'size-t'.
So in the end your code has 'size+ size+ size -t complexity'

I might be wrong on my analysis, but I would say my code has a 'size' complexity and if there is a field which is not equal it stops right away, yours goes until the end.

examples/create_upload_tutorial.py
ArlindKadra: just noticed, I would agree with this, however instead we are going to go for another dataset.

openml/datasets/dataset.py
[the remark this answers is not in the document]
ArlindKadra: set is indeed better instead of the dictionary for the server_fields.

openml/datasets/functions.py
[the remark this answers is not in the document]
ArlindKadra: I agree that it is not clear where it is tested.. Are you thinking of checking an existing dataset on the server ?

openml/datasets/dataset.py
ArlindKadra:
```
if self_keys != other_keys:
            return False
```
I think this would be better expressed as
```
if len(self_keys) != len(other_keys):
            return False
```

examples/tasks_tutorial.py
```diff
@@ -58,7 +58,9 @@
 print(len(filtered_tasks))
 
 ############################################################################
-# Resampling strategies can be found on the `OpenML Website <http://www.openml.org/search?type=measure&q=estimation%20procedure>`_.
+# Resampling strategies can be found on
+# the `OpenML Website <http://www.openml.org
```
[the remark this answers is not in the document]
ArlindKadra: @mfeurer I checked something there, noticed that the line was too long and I fixed it. The problem now, would be that reversing it, would mean it is on the diff and then it would fail the build because it is >90. Or am I wrong ?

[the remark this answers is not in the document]
mfeurer: It should not appear in the diff because the file will stay the same compared to the development branch.

PR #548, opened 2018-09-21
[MRG] EHN: Add support for pandas DataFrame and SparseDataFrame when loading

### LINKED ISSUE #257: String feature support
Hi, this is more a question than a bug report. I tried downloading dataset 40748 that includes sting features and I get this error:

  File "/usr/local/lib/python2.7/dist-packages/openml-0.3.0-py2.7.egg/openml/datasets/dataset.py", line 197, in get_data
    raise PyOpenMLError('Dataset not compatible, PyOpenML cannot handle string features')
openml.exceptions.PyOpenMLError: Dataset not compatible, PyOpenML cannot handle string features

Given that `arff` module that is used internally already supports string features is it planned to add support for text datasets to openml-python?

### LINKED ISSUE #364: Return tables as dataframes
#363 adds pandas as a dependency and in the tutorials we convert dicts of dicts to dataframes anyway, therefore we could simply return some properties as dataframes. @amueller @janvanrijn what's your opinion on this?

### LINKED ISSUE #408: Getting non-encoded dataset
If I import, for instance, the adult dataset: https://www.openml.org/d/1590
`get_data` automatically encodes categorical features:
```
dataset = oml.datasets.get_dataset(1590) # Adult dataset
X, y, attribute_names = dataset.get_data(target=dataset.default_target_attribute, 
    return_attribute_names=True)
data = pd.DataFrame(X, columns=attribute_names)
display(data.head())
```
->'education' is encoded as numbers

Is there a way to get the original non-encoded features, so I can look at them with pandas?

Thanks!

glemaitre:
supersede #363 
closes #257
closes #364
closes #408.

Add support to load ARFF into pandas DataFrame and pandas SparseDataFrame.

TODO: 

- [x] Add unit tests
- [x] Add documentation
- [x] Add example
- [x] Add unit tests for exceptions
- [ ] reuse the `row_id_attribute`` to rename the dataframe index name if possible.

Check the dtype consistence:

- [x] boolean will be encoded as category 'True', and 'False' and need to be converted
- [x] check that integer dtype is preserved
- [x] check the behavior with missing values with the different dtype


glemaitre:
@mfeurer For the unit test, would it be fine to actually used the following dataset:

https://www.openml.org/d/40945

It contains all type of possible and this would be handy. But I don't know if this is robust enough (if there anybody going to remove it anytime?)

glemaitre:
@mfeurer @joaquinvanschoren @janvanrijn 
If you already have any comments even if this is still wip.

ci_scripts/flake8_diff.sh
```diff
@@ -125,7 +125,7 @@ check_files() {
     if [ -n "$files" ]; then
         # Conservative approach: diff without context (--unified=0) so that code
         # that was not changed does not create failures
-        git diff --unified=0 $COMMIT_RANGE -- $files | flake8 --diff --show-source $options
+        git diff --unified=0 $COMMIT_RANGE -- $files | flake8 --ignore E402 --diff --show-source $options
```
janvanrijn: Just asking, what exactly is this additional ignore?

mfeurer: https://lintlyci.github.io/Flake8Rules/rules/E402.html

glemaitre:
@mfeurer @janvanrijn I would need an hand for the testing here.
I want to be able to test the following the dtype inference in presence of missing values

I uploaded a specific [dataset](https://test.openml.org/d/4659) on the test server. However, this will go wrong when you are clearing the server. I was wondering if we could have somewhere couple of specific dataset to be able to test all scenario.

glemaitre:
ping @mfeurer @janvanrijn @joaquinvanschoren We will need to read those data since we can upload them :)

janvanrijn:
My apologies, I thought this was merged already, and missed your previous comment.

> I uploaded a specific dataset on the test server. However, this will go wrong when you are clearing the server. I was wondering if we could have somewhere couple of specific dataset to be able to test all scenario.

Actually, yes I have a collection of datasets that are always available on the test server (symlinking to datasets that are actually on the main server). Dataset ids 1 through 130 will always be available on the server, even after every reset. If you need any more / other datasets, please let me know.

mfeurer:
In addition to Jan's suggestions, as we can merge sparse dataset upload support in a bit, you could upload the dataset first and then check if loading it again works. Otherwise you could check them in into the directory `test/files` and test the load function without invoking server calls or by mocking them out.

openml/datasets/dataset.py
```diff
@@ -158,41 +159,106 @@ def __init__(self, name, description, format=None,
...
+                        (X[0], (X[1], X[2])), shape=X_shape, dtype=np.float32)
+                    X = X.tocsr()
+
+                elif self.format.lower() == 'arff':
+                    X = pd.DataFrame(data['data'], columns=attribute_names)
+                    X_null_count = X.isnull().sum()
+
+                    for column_name in X.columns:
+                        if attribute_dtype[column_name] == 'boolean':
+                            X[column_name] = self._create_column_bool(
```
mfeurer: I'm afraid that this could become extremely costly for large datasets with a large amount of columns. Do you think it would be feasible (in terms of memory) to store all columns in a list first and then concatenate them at the end of the function?

openml/datasets/dataset.py
```diff
@@ -158,41 +159,106 @@ def __init__(self, name, description, format=None,
...
+                                categories_names[column_name]
+                            )
+                        elif attribute_dtype[column_name] == 'categorical':
+                            X[column_name] = self._unpack_categories(
+                                X[column_name],
+                                categories_names[column_name]
+                            )
+                        elif (attribute_dtype[column_name] == 'integer' and
+                                pd.api.types.infer_dtype(
+                                    X[column_name]) != 'integer'):
```
mfeurer: Could you please explain this statement and the similar statement below?

openml/datasets/dataset.py
```diff
@@ -290,30 +353,98 @@ def decode_arff(fh):
...
+    def _create_column_bool(series, categories):
+        col = []
+        for x in series:
+            try:
+                val = categories[int(x)]
+                val = True if val.lower() == 'true' else False
+                col.append(val)
+            except (TypeError, ValueError):
+                col.append(np.nan)
+        return pd.Series(col, index=series.index, dtype=object)
```
mfeurer: Is there a reason to keep this a type object and not cast to categorical?

openml/datasets/dataset.py
```diff
@@ -158,41 +159,106 @@ def __init__(self, name, description, format=None,
...
+                            X[column_name] = X[column_name].astype(float)
+                else:
+                    raise Exception()
 
-                    with open(self.data_pickle_file, "wb") as fh:
-                        pickle.dump((X, categorical, attribute_names), fh, -1)
-                    logger.debug("Saved dataset %d: %s to file %s" %
-                                 (int(self.dataset_id or -1), self.name, self.data_pickle_file))
+                with open(self.data_pickle_file, "wb") as fh:
+                    pickle.dump((X, categorical, attribute_names), fh, -1)
```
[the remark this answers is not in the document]
mfeurer: Actually, one would have to replace the previous pickle files as they were numpy arrays.

tests/test_datasets/test_dataset.py
```diff
@@ -99,6 +107,58 @@ def test_get_data_with_ignore_attributes(self):
...
+        self.assertTrue(X.dtype == 'object')
+        self.assertTrue(y.dtype == int)
+
+    def test_get_data_pandas(self):
+        data = self.titanic.get_data(dataset_format='dataframe')
+        self.assertTrue(isinstance(data, pd.DataFrame))
+        self.assertEqual(data.shape[1], len(self.titanic.features))
+        self.assertEqual(data.shape[0], 1309)
+        col_dtype = {
+            'pclass': 'float64',
```
mfeurer: Shouldn't this be an integer?

tests/test_datasets/test_dataset.py
```diff
@@ -99,6 +107,58 @@ def test_get_data_with_ignore_attributes(self):
...
+            'name': 'object',
+            'sex': 'category',
+            'age': 'float64',
+            'sibsp': 'float64',
+            'parch': 'float64',
+            'ticket': 'object',
+            'fare': 'float64',
+            'cabin': 'object',
+            'embarked': 'category',
+            'boat': 'object',
```
mfeurer: Now that I'm seeing this, I don't think that casting integer columns with missing values to object is in the spirit of arff. The latest version mentioned that float and int should be both treated as numerical (see [here](https://github.com/Waikato/weka-3.8/blob/master/wekadocs/manual/arff.tex)), so I think we should do the same here.

mfeurer:
This actually implements #257 #364 and #408.

openml/datasets/dataset.py
```diff
@@ -158,41 +159,106 @@ def __init__(self, name, description, format=None,
...
+                    if (isinstance(type_, list) and
+                            self.format.lower() == 'sparse_arff'):
+                        try:
+                            np.array(type_, dtype=np.float32)
+                        except ValueError:
+                            raise ValueError(
+                                "Categorical data needs to be numeric when "
+                                "using sparse ARFF."
+                            )
+                    # string can only be supported with pandas DataFrame
```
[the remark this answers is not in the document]
glemaitre: You could always an object array while with sparse matrices, you cannot.

openml/datasets/dataset.py
glemaitre: I would rely on @jorisvandenbossche for the internal of pandas. But I don't think that it will be more costly. I am unsure that it will be more costly.

openml/datasets/dataset.py
glemaitre: There is no need to force the dtype here. It will be `bool` with there is not missing values and `object` if there is. `object` refers to `bool` and `float` (i.e., np.nan marker).

openml/datasets/dataset.py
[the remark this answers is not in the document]
mfeurer: Sorry, I rather meant that we need to decide whether we actually want to allow that. I'd prefer to not have any numpy array of type object.

openml/datasets/dataset.py
mfeurer: I don't know about the internals of pandas, but I made the experience that changing the columns inplace can be quite expensive:
```
import pandas as pd
import numpy as np
array = np.random.random(size=(1000, 10000))
df = pd.DataFrame(array)

def cast_array(df):
    df_copy = df.copy()
    for name in df_copy.columns:
        df_copy[name] = df_copy[name].astype(float)
    return df_copy

def cast_array_2(df):
    df_copy = df.copy()
    columns = []
    for name in df_copy.columns:
        columns.append(df_copy[name].astype(float))
    return pd.concat(columns, axis=1)

df_copy_1 = cast_array(df)
df_copy_2 = cast_array_2(df)
print(np.all(df_copy_1 == df_copy_2))
%timeit cast_array(df)
%timeit cast_array_2(df)
```
results in
```
## -- End pasted text --
True
1.6 s ± 75.8 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
1.06 s ± 75.5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

openml/datasets/dataset.py
mfeurer: But in the case the column is all bool, there should be only two allowed values in total. Therefore, the data frame should reflect that for two reasons:
1. to have all information about the dataset in the dataframe (object is ambiguous here)
2. to not allow adding any other types to that column

glemaitre: My previous comment was not clear enough. When I meant `There is no need to force the dtype here`, I mean that I should remove `dtype=object` because the type inference will work. In the case of pure `bool` the type will be `bool`.

openml/datasets/dataset.py
jorisvandenbossche: The second one is actually what `df.astype(..)` is doing under the hood if you provide a dict of {col: new_dtype} (this is indeed more efficient than actually astyping single columns on the full dataframe one by one (the first function)).

glemaitre: OK. Make sense.

tests/test_datasets/test_dataset.py
glemaitre: Does Weka cast integer to float when reading an integer column?

openml/datasets/dataset.py
mfeurer: Okay, got it. I just looked up the specification of arff again, and there is no definition of bool. Could you please state your opinion on casting boolean columns to categorical?

tests/test_datasets/test_dataset.py
mfeurer: I don't think so: https://github.com/Waikato/weka-3.8/blob/49865490cef763855ede07cd11331a7aeaecd110/weka/src/main/java/weka/core/converters/ArffLoader.java#L694

Also, reading this arff file:
```
@relation in-vs-float

@attribute attr1 real
@attribute attr2 integer
@attribute attr3 numeric

@data
1, 2, 3
1.0, 2.0, 3.0
1.1, 1.0, 1.1
```
results in the following being displayed in WEKA:
image

glemaitre: and what happens with 

```
@relation in-vs-float

@attribute attr1 real
@attribute attr2 integer
@attribute attr3 numeric

@data
1, 2, 3
1.0, 2, 3.0
1.1, 1, 1.1
```

openml/datasets/dataset.py
glemaitre: My only concern is that the round trip (upload-download) will not give the original data. If you are fine with that then we can just let the bool as categories and it will be encoded by any encoder after anyway.

tests/test_datasets/test_dataset.py
glemaitre: I am really wondering what is the point of putting some header then :)
So I would remove the type inference and only keep the unpacking of the categories then.

openml/datasets/dataset.py
mfeurer: Hm, why would a round-trip destroy categorical-encoded bools? If a category with `['True', 'False']` is passed it should be serialized/deserialized without any loss in information.

glemaitre: What I meant is that a user will pass a dataframe with `[True, False]` in a column with `dtype=bool` and will get back a `['True', 'False']` column with `dtype='categories'`.

glemaitre: So the dtype and the values are not the same even if they represnet the same information. Anyway, I would say that this is also a limitation of the ARFF format not having a proper dtype info.

tests/test_datasets/test_dataset.py
glemaitre: Apparently it has been uploaded as numeric which is real

openml/datasets/dataset.py
mfeurer: Got it, thanks for the explanation. I'd be fine with that behavior, given that this comes from the arff format not defining boolean data types.

tests/test_datasets/test_dataset.py
mfeurer: > I am really wondering what is the point of putting some header then :)

I stopped wondering about arff...

> So I would remove the type inference and only keep the unpacking of the categories then.

That should be sufficient.

examples/datasets_tutorial.py
```diff
@@ -55,7 +55,11 @@
 ############################################################################
 # Get the actual data.
 #
-# Returned as numpy array, with meta-info (e.g. target feature, feature names,...)
+# The dataset can be returned in 2 possible formats: as a NumPy array or as
```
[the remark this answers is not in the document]
mfeurer: Thanks for updating this, but could you please also update number in the text?

mfeurer:
Hey, the changes look good to me. I think the only thing which is still open is the issue that the cache can now stores a numpy array, but will store a data frame in the future. Maybe we should change the file name to contain a version number?

Is there anything that I can help you with for this pull request?

glemaitre:
I have to check why the tests are failing and I need to add more tests to cover all code.

glemaitre:
I'll post a series of question by trying to solve and writing some unit tests:

Before this PR, data were stored as float32. Categories were kept encoded as ordinal. Using the dataframe as a placeholder to store the data, we can store the categories directly. It implies that I can return an array with `dtype=object` containing the string. This is really inefficient but it contains the original categories instead of the encoded one.

So should I keep the previous behavior (meaning reencoding the categories read with ARFF) and return always an array of float32 or return a dtype array with strings inside?

mfeurer:
> So should I keep the previous behavior (meaning reencoding the categories read with ARFF) and return always an array of float32 or return a dtype array with strings inside?

From the top of my head I would expect the following behavior:
* user can choose between old return value (encoded numpy array) and new return value (data frame)
* a category is encoded as a category (for data frames)
* to differentiate between a category and a string (which would then be of dtype=object I assume).

What do you think of this?

glemaitre:
> user can choose between old return value (encoded numpy array) and new return value (data frame)

OK so you want backward compatibility.

> What do you think of this?

Because I changed this behavior to return string as well. So I will revert this behavior.

mfeurer:
> OK so you want backward compatibility.

Would be great. But didn't you have this already in this PR?

> Because I changed this behavior to return string as well. So I will revert this behavior.

Is there a reason to return the string over the category? Currently, arff defines mostly categories, so it would be natural to model them as categories in the data structure in my opinion.

glemaitre:
@mfeurer I think that we are not understanding each other.

Let's imagine the following dataset

```python
[[1., 'A', 'one string'],
 [2., 'B', 'two string']]
```

The second column will be category and the third just string.

1. numpy array

The old behavior would raise an error because of the third columns.
Considering only the 2 first columns, we would get:

```python
np.array([[1., 0.],
          [2., 1.]], dtype=np.float32)
```

where the categories get encoded.

The new behavior (which I just reverted) could return:

```python
np.array([[1., 'A', 'one string'],
          [2., 'B', 'two string']], dtype=object)
```

2. pandas dataframe

We will get:

```python
[[1., 'A', 'one string'],
 [2., 'B', 'two string']]
```

where the first column will be float, second column will be category ("A", "B") and the third will be object (for string).

 So

> Is there a reason to return the string over the category? Currently, arff defines mostly categories, so it would be natural to model them as categories in the data structure in my opinion.

I am using the category for the dataframe.

> Would be great. But didn't you have this already in this PR?

I was using the real categories stored in ARFF and not just an integer. We can keep the backward behavior forcing people to use dataframe whenever they need string support.

mfeurer:
Okay, I think I get it. You are proposing to use:
* Dataframe as the default
* dtype=object is on the data frame level and not the column level
* use categorical columns where appropriate
* only use numpy arrays as an option to keep backwards compatibility
Is that correct?

glemaitre:
> Dataframe as the default

Not really, you will break the backward compatibility. We can raise a FutureWarning that it will be the default in 2 versions if you want.

> dtype=object is on the data frame level and not the column level

Not sure to understand this comment. A dataframe as a dtype per column. A string column will be of object dtype. 

> use categorical columns where appropriate

Yes.

> only use numpy arrays as an option to keep backwards compatibility

Up to you.

mfeurer:
> We can raise a FutureWarning that it will be the default in 2 versions if you want.

That would be great.

> Not sure to understand this comment. A dataframe as a dtype per column. A string column will be of object dtype.

This is my fault, I think I talked nonsense here. Sorry for the confusion on my side.

> Yes.

Great :)

> Up to you.

I think this interdependent with the first quote.

glemaitre:
OK. I could add the FutureWarning. I want first the test to pass and check the coverage. I think that we are almost there.

mfeurer:
Cool :) I'm out for today, please ping me once this is ready. I'll do my best to make sure there's no delay from my side.

glemaitre:
Is there a reason that one of the worker is crashing, It seems that it cannot be related with this PR, isn't it?

janvanrijn:
> Is there a reason that one of the worker is crashing, It seems that it cannot be related with this PR, isn't it?

The error messages were not extremely helpful. 

I restarted 3 jobs, let's see what happens?

glemaitre:
@janvanrijn @mfeurer I need to test the following line:
https://github.com/openml/openml-python/pull/548/files#diff-ba5e66c898a59d9f8fca4a850f601196R216

I need a dataset with 'True' and 'False' as categories to be sure that the dumping-loading will transform Python boolean properly. Any idea which dataset is available on the test server?

glemaitre:
@joaquinvanschoren @janvanrijn @mfeurer 
The PR should be ready for review. The error in the CI is still cryptic to me.
The code is almost all covered; there is only missing the case previously mentioned due to boolean.

mfeurer:
I've never seen this before, too. It happens in the same unit tests, but different lines, right? As the tests are approaching runtimes of ~20 minutes, could it be that the pytest timeout of 600 seconds in ci_scripts/test.sh kicks in? Maybe you can add one of the argument suggested [over at stackoverflow](https://stackoverflow.com/a/27899853) to see if that's the case?

examples/datasets_tutorial.py
```diff
@@ -55,16 +55,28 @@
...
 #
-# Returned as numpy array, with meta-info
-# (e.g. target feature, feature names, ...)
-X, y, attribute_names = dataset.get_data(
+# The dataset can be returned in 2 possible formats: as a NumPy array, a SciPy
+# sparse matrix, or as a Pandas DataFrame (or SparseDataFrame). The format is
+# controlled with the parameter ``dataset_format`` which can be either 'array'
+# (default) or 'dataframe'. Let's first build our dataset from a NumPy array
+# and manually create a dataframe.
+X, y, attribute_names = dataset.
```
mfeurer: I believe this is the line where the examples fail due to a syntax error, right?

mfeurer:
@janvanrijn is it possible for you to search the testserver for a dataset containing the attributes `'True'` and `'False'`? Otherwise, I'm afraid we need to use mock to fake such a dataset.

@glemaitre could you please check the two or three open comments from one of my previous reviews, too?

joaquinvanschoren:
@glemaitre None of the original datasets on the test server has categories 'True' and 'False'
Dataset 51 has 'true' and 'false'
Dataset 49 has 'TRUE' and 'FALSE'
Dataset 6537 looks like a mock dataset created in the unit tests, and that one has 'True' and 'False'
On the production server, dataset 41150 (MiniBooNE) has 'True' and 'False'. It could be included on the test server.

glemaitre:
> Dataset 51 has 'true' and 'false'
> Dataset 49 has 'TRUE' and 'FALSE'

I think that we could easily support those case in addition `'True'.lower().capitalize()` should make the trick.

openml/datasets/dataset.py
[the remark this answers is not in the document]
glemaitre: Most probably true. I added a comment.

tests/test_datasets/test_dataset.py
glemaitre: As previously mentioned. It would be a numeric field.

openml/datasets/dataset.py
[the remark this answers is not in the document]
glemaitre: This is resolve

tests/test_datasets/test_dataset.py
glemaitre: We are fine here then.

glemaitre:
The failure is still the same. Tonight I will decorate the tests as suggested. Otherwise, I modified slightly the code and use the suggested dataset to check for the boolean column.

glemaitre:
It should be ready for review.

mfeurer:
I think I'm fine and will approve once you briefly explain what happens/happened to the open TODO and after you had a look at the unit tests. I will afterwards whether and how this collides with existing pickle files on the hard drives.

glemaitre:
@mfeurer You are probably right regarding the timeout. On the built that does not fail the test which is commonly failing is close the the timeout.

```
519.58s call     tests/test_runs/test_run_functions.py::TestRun::test_predict_proba_hardclassifier
```

mfeurer:
Do you have any idea what makes that test take so long? I guess it takes only a few seconds locally?

mfeurer:
Thinking about this more generally, it seems that the time required to execute the unit tests roughly doubles with this PR. Do you have any idea why this is the case?

glemaitre:
> Do you have any idea what makes that test take so long? I guess it takes only a few seconds locally?

It is very slow locally as well on my laptop which is kinda powerful. I can expect Travis to be slower.

```
321.45s call     tests/test_runs/test_run_functions.py::TestRun::test_predict_proba_hardclassifier
```

glemaitre:
> Thinking about this more generally, it seems that the time required to execute the unit tests roughly doubles with this PR. Do you have any idea why this is the case?

I can think about 2 things:

* The pickling could be slower with dataframe. We keep the dtypes and we have to pickle several numpy array then.
* When creating the dataframe, we have some kind of check/replace that go through the categorical data. So it would slower the data.

Since that we are using the dataframe by default and then convert is to numpy, all the current code will be affected by this slow down. However, I don't see any straightforward solution.

openml/datasets/dataset.py
```diff
@@ -303,28 +347,98 @@ def decode_arff(fh):
...
+        if array_format == "array" and not scipy.sparse.issparse(data):
+            # We encode the categories such that they are integer to be able
+            # to make a conversion to numeric for backward compatibility
+            def _encode_if_category(column):
+                if column.dtype.name == 'category':
+                    column = column.cat.codes
+                    column = column.apply(lambda x: np.nan if x == -1 else x)
+                return column
+            if data.ndim == 2:
+                for column_name in data.columns:
```
mfeurer: I just briefly profiled the code and it looks like this for loop is the culprit of the slowdown. Replacing it by:
```
columns = {
    column_name: _encode_if_category(data.loc[:, column_name]) 
    for column_name in data.columns
}
                data = pd.DataFrame(columns)
```
increases the test speed again (while the unit tests are still passing). Could you please check (and replace if this is the case)?

mfeurer:
Tests look good (except for a minor merge issue). Would you mind fixing the two remaining issues (import error and pep8) and then merge?

glemaitre:
@janvanrijn @joaquinvanschoren Do you want to have another look at the PR?

mfeurer:
Awesome, it's green :tada: :balloon: 

Let's give them a few hours, also @amueller (who opened the original PR for Pandas support) and @PGijsbers

openml/datasets/dataset.py
```diff
@@ -303,28 +347,100 @@ def decode_arff(fh):
             with io.open(filename, encoding='utf8') as fh:
                 return decode_arff(fh)
 
+    @staticmethod
+    def _convert_array_format(data, array_format, attribute_names):
```
PGijsbers: Maybe document supported conversions and how they ought to be specified. From reading code, it looks like there are only two specific conversions allowed:
 - non-sparse numeric data to numpy-array (`array_format="array"`)
 - sparse data to sparse dataframe (`array_format="dataframe"`)

no cross combinations are possible (sparse to non-sparse and vice versa).
Additionally it seems the method converts categorical columns by using their numeric representation? It might be good to document that as well.

PGijsbers:
Looks good to me! I left a remark with one particular internal function that I felt was underdocumented. Though if it is merged in its current state that would be fine with me too.

mfeurer:
Thanks for the feedback @PGijsbers, I just merged (to make sure that this is no longer delayed by merge issues) and created an issue (#639) to improve the docs.

PR #561, opened 2018-10-02
added documentation for running specific tests

joaquinvanschoren:
Adds documentation about running specific unit tests. I found this very useful and didn't know this beforehand.


mfeurer:
Hey, this is really useful, thanks a lot. Could you please extend this a bit by giving advice how to run all unit tests for a module, and also use a unit test which is already inside develop and which will most likely also stay there (for example something testing datasets/functions.py)?

mfeurer:
Hey, I think that my previous comment still applies and it would be great to have this additional information before merging this pull request.

PR #566, opened 2018-10-06
fixes issue #565 and removes future / deprecation warnings

### LINKED ISSUE #565: Test Run Get Trace unit test fails
```
Error
  File "/home/janvanrijn/anaconda3/envs/openml-python/lib/python3.6/unittest/case.py", line 59, in testPartExecutor
    yield
  File "/home/janvanrijn/anaconda3/envs/openml-python/lib/python3.6/unittest/case.py", line 605, in run
    testMethod()
  File "/home/janvanrijn/projects/openml-python/tests/test_runs/test_run_functions.py", line 610, in test_get_run_trace
    trace = openml.runs.functions._create_trace_from_arff(
AttributeError: module 'openml.runs.functions' has no attribute '_create_trace_from_arff'
```

Makes sense.

janvanrijn:
Fixes #565

### DIFF
--- a/openml/setups/functions.py
+++ b/openml/setups/functions.py
@@ -37,6 +37,9 @@ def setup_exists(flow, model=None):
 
     if model is None:
         model = flow.model
+        if flow.model is None:
+            raise ValueError('Could not locate model (neither given as'
+                             'argument nor available as flow.model)')
     else:
         exists = flow_exists(flow.name, flow.external_version)
         if exists != flow.flow_id:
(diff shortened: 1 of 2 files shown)


mfeurer:
It seems like the tests are failing with older scikit-learn versions.

janvanrijn:
I honestly have no clue why the code crashed before. According to my analysis, this should not have happened. I made some changes to the if/else structure of the `setup_exists` fn, as I had the impression that this was not according to the way it should be. These should have had no influence on the code crashing or not. @mfeurer what do you think?

PR #581, opened 2018-10-20
Fix #580

### LINKED ISSUE #580: Listing functions when not enough results are available
It occured to me that the listing function goes bananas when there are less results available that we wanted to obtain. 

I extended the unit test of `test_utils.py`:
```
class OpenMLTaskTest(TestBase):

    def mocked_perform_api_call(call):
        # TODO: JvR: Why is this not a staticmethod?
        url = openml.config.server + '/' + call
        return openml._api_calls._read_url(url)

    @mock.patch('openml._api_calls._perform_api_call', side_effect=mocked_perform_api_call)
    def test_list_all_few_results_available(self, _perform_api_call):
        # we want to make sure that the number of api calls is only 1.
        # Although we have multiple versions of the iris dataset, there is only
        # one with this name/version combination

        datasets = openml.datasets.list_datasets(size=1000,
                                                 data_name='iris',
                                                 data_version=1)
        self.assertEqual(len(datasets), 1)
        self.assertEqual(_perform_api_call.call_count, 1)
```
It actually makes more calls. I will push a fix

janvanrijn:
Fixes #580

### DIFF
--- a/openml/evaluations/functions.py
+++ b/openml/evaluations/functions.py
@@ -108,7 +108,7 @@ def __list_evaluations(api_call):
         run_id = int(eval_['oml:run_id'])
         array_data = None
         if 'oml:array_data' in eval_:
-            eval_['oml:array_data']
+            array_data = eval_['oml:array_data']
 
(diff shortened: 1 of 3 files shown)


janvanrijn:
Really appreciated, thanks!

PR #586, opened 2018-10-31
[MRG] EHN: inferred row_id_attribute from dataframe to create a dataset

glemaitre:
This allows to automatically inferred the name of the index column.

However, I break the backward compatibility since that the parameter is now a keyword argument. It was already done like that for `format`.

### DIFF
--- a/openml/datasets/functions.py
+++ b/openml/datasets/functions.py
@@ -515,12 +515,13 @@ def create_dataset(name, description, creator, contributor,
 
     if row_id_attribute is not None:
         is_row_id_an_attribute = any([attr[0] == row_id_attribute
-                                  for attr in attributes_])
+                                      for attr in attributes_])
         if not is_row_id_an_attribute:
-            raise ValueError("'row_id_attribute' should be one of the data "
-                            "attribute. Got '{}' while candidates are {}."
-                            .format(row_id_attribute,
-                                    [attr[0] for attr in attributes_]))
+            raise ValueError(
+                "'row_id_attribute' should be one of the data attribute. "
+                " Got '{}' while candidates are {}."
+                .format(row_id_attribute, [attr[0] for attr in attributes_])
+            )
 
     data = data.values if hasattr(data, "columns") else data
 
--- a/openml/datasets/functions.py
+++ b/openml/datasets/functions.py
@@ -488,7 +488,7 @@ def create_dataset(name, description, creator, contributor,
     class:`openml.OpenMLDataset`
         Dataset description."""
 
-    if hasattr(data, "index"):
+    if isinstance(data, (pd.DataFrame, pd.SparseDataFrame)):
         # infer the row id from the index of the dataset
         if row_id_attribute is None:
             row_id_attribute = data.index.name
@@ -513,13 +513,14 @@ def create_dataset(name, description, creator, contributor,
     else:
         attributes_ = attributes
 
-    is_row_id_an_attribute = any([attr[0] == row_id_attribute
+    if row_id_attribute is not None:
+        is_row_id_an_attribute = any([attr[0] == row_id_attribute
                                   for attr in attributes_])
-    if row_id_attribute is not None and not is_row_id_an_attribute:
-        raise ValueError("'row_id_attribute' should be one of the data "
-                         "attribute. Got '{}' while candidates are {}."
-                         .format(row_id_attribute,
-                                 [attr[0] for attr in attributes_]))
+        if not is_row_id_an_attribute:
+            raise ValueError("'row_id_attribute' should be one of the data "
+                            "attribute. Got '{}' while candidates are {}."
+                            .format(row_id_attribute,
+                                    [attr[0] for attr in attributes_]))
 
     data = data.values if hasattr(data, "columns") else data
 
--- a/openml/datasets/functions.py
+++ b/openml/datasets/functions.py
@@ -417,8 +417,9 @@ def attributes_arff_from_df(df):
 def create_dataset(name, description, creator, contributor,
                    collection_date, language,
                    licence, attributes, data,
-                   default_target_attribute, row_id_attribute,
-                   ignore_attribute, citation, format=None,
+                   default_target_attribute,
+                   ignore_attribute, citation,
+                   row_id_attribute=None, format=None,
                    original_data_url=None, paper_url=None,
                    update_comment=None, version_label=None):
     """Create a dataset.
@@ -433,11 +434,6 @@ def create_dataset(name, description, creator, contributor,
         Name of the dataset.
     description : str
         Description of the dataset.
-    format : str, optional
-        Format of the dataset which can be either 'arff' or 'sparse_arff'.
-        By default, the format is automatically inferred.
-        .. deprecated: 0.8
-            ``format`` is deprecated in 0.8 and will be removed in 0.10.
     creator : str
         The person who created the dataset.
     contributor : str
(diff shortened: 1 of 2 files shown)


glemaitre:
ping @mfeurer @janvanrijn @joaquinvanschoren Any feedback?

tests/test_datasets/test_dataset_functions.py
```diff
@@ -803,6 +803,65 @@ def test_create_dataset_pandas(self):
...
+            citation=citation,
+            attributes='auto',
+            data=df,
+            row_id_attribute='index_column',
+            format=None,
+            version_label='test',
+            original_data_url=original_data_url,
+            paper_url=paper_url
+        )
+        self.assertEqual(dataset.row_id_attribute, 'index_column')
```
janvanrijn: should we also test that the "publish" function correctly uploads the index column?

openml/datasets/functions.py
```diff
@@ -499,6 +504,9 @@ def create_dataset(name, description, creator, contributor,
     else:
         attributes_ = attributes
 
+    if row_id_attribute is None and hasattr(data, "index"):
```
janvanrijn: what happens with the dataframe index if `row_id_attribute` is set to a value other than the index? Will it be ignored at upload time? Can this be added to comments and somehow tested?

openml/datasets/functions.py
```diff
@@ -463,14 +459,23 @@ def create_dataset(name, description, creator, contributor,
...
         Reference(s) that should be cited when building on this data.
     version_label : str, optional
         Version label provided by user, can be a date, hash, or some other type of id.
+    row_id_attribute : str, optional
+        The attribute that represents the row-id column, if present in the
+        dataset. If ``data`` is a dataframe and ``row_id_attribute`` is not
+        specified, it will be inferred.
+        .. versionadded: 0.8
+           Inference of ``row_id_attribute`` from a dataframe.
+    format : str, optional
```
janvanrijn: Personally, I wouldn't mind just removing format rather than deprecating it. It seems that the function signature is changed anyway, making the function not backwards compatible (bad, but acceptable and nothing really that can be done about this)

janvanrijn:
Thanks for this PR! I left some questions. @mfeurer is currently out of office, hopefully he can also have a look at it, but I think this can be merged quite easily :)

openml/datasets/functions.py
glemaitre: OK, but I would do that in a separate PR.

openml/datasets/functions.py
glemaitre: Could you quickly show a code snippet (even pseudo-code) to illustrate what you mean? I am not sure to see the use-case?

openml/datasets/functions.py
```diff
@@ -463,14 +459,23 @@ def create_dataset(name, description, creator, contributor,
...
     ignore_attribute : str | list
         Attributes that should be excluded in modelling, such as identifiers and indexes.
     citation : str
         Reference(s) that should be cited when building on this data.
     version_label : str, optional
         Version label provided by user, can be a date, hash, or some other type of id.
+    row_id_attribute : str, optional
+        The attribute that represents the row-id column, if present in the
+        dataset. If ``data`` is a dataframe and ``row_id_attribute`` is not
+        specified, it will be inferred.
```
mfeurer: It would be great if you could be a bit more specific on how it will be inferred. Honestly I'm a bit confused right now. Doesn't a dataframe always have an index?
```
In [1]: import pandas as pd                                                                                                                                                                                  

In [2]: data = [['a'], ['b'], ['c'], ['d'], ['e']]                                                                                                                                                           

In [3]: column_names = ['rnd_str']                                                                                                                                                                           

In [4]: df = pd.DataFrame(data, columns=column_names)                                                                                                                                                        

In [5]: hasattr(data, "index")                                                                                                                                                                               
Out[5]: True
```
Unless the user specifies an index column it should not be uploaded. And I assume that the pandas index column is not uploaded right now.

tests/test_datasets/test_dataset_functions.py
mfeurer: That would indeed be great.

openml/datasets/functions.py
glemaitre: yes, it always has an `index` and the `index.name` is either a string or None. By default, it will be None.

Just to be sure, `row_id_attribute` corresponds to the name of the index column, isn't it?

mfeurer: > Just to be sure, row_id_attribute corresponds to the name of the index column, isn't it

Yes, but only if there is one. For most datasets there is none.

> yes, it always has an index and the index.name is either a string or None. By default, it will be None.

Shouldn't the code then be `hasattr(data, "index") and hasattr(data.index, "name")`?

glemaitre: > Shouldn't the code then be hasattr(data, "index") and hasattr(data.index, "name")?

No, `index` as always a `name` attribute. It is `None` by default which is the also the value required when `row_id_attribute` should not be specified.

To illustrate what I mean.

```python
In [2]: df = pd.DataFrame([1, 2])

In [3]: df
Out[3]: 
   0
0  1
1  2

In [4]: df.index.name

In [5]: print(df.index.name)
None
```

mfeurer: Okay, I got it. Do you think it would make sense to add a brief description about this behavior to the docstring?

openml/datasets/functions.py
mfeurer: I think @janvanrijn wanted to only point out a potential issue:
```
In [17]: dataset = openml.datasets.functions.create_dataset(
    ...:             name=name,
    ...:             description=description,
    ...:             creator=creator,
    ...:             contributor=None,
    ...:             collection_date=collection_date,
    ...:             language=language,
    ...:             licence=licence,
    ...:             default_target_attribute=default_target_attribute,
    ...:             ignore_attribute=None,
    ...:             citation=citation,
    ...:             attributes='auto',
    ...:             data=df,
    ...:             row_id_attribute='rnd_str',
    ...:             format=None,
    ...:             version_label='test',
    ...:             original_data_url=original_data_url,
    ...:             paper_url=paper_url,
    ...: )

In [18]: dataset.row_id_attribute
Out[18]: 'rnd_str'

In [19]: df.index.name
Out[19]: 'index'
```

mfeurer:
Looks good to me, but it would be great to have the additional unit test and maybe a warning if the row ID is given twice (in the form of an index and as an attribute to the create dataset function).

openml/datasets/functions.py
glemaitre: OK so the PR is wrong. The behaviour needs to be better defined:

* If data is an array and `row_id_attribute` is `None` then we let `row_id_attribute` at `None`.
* If data is an array and `row_id_attribute` is not `None`, we need to have a check that this is actually an attribute.
* If data is a dataframe and `row_id_attribute` is `None`, we need to infer it. If `df.index.name` is `None` then we are back to the first bullet. Otherwise, we take the name as `row_id_attribute` and we need to `df.reset_index()` to include the data into the array when converting to numpy `df.values`.

glemaitre:
> Looks good to me, but it would be great to have the additional unit test and maybe a warning if the row ID is given twice (in the form of an index and as an attribute to the create dataset function).

I am a bit lost then. How do you actually store the index column in ARFF if this is not listed in the attribute?

glemaitre:
@janvanrijn 

I am getting locally and on travis the following:

```
>           parser.Parse(xml_input, True)
E           xml.parsers.expat.ExpatError: junk after document element: line 71, column 6
```

Is it linked with the test server? This is a bit blocking to write test locally actually.

openml/datasets/functions.py
glemaitre: Python lists also have an index. So it would break for this case.

glemaitre:
The errors are unrelated. @mfeurer I modified the documentation to reflect the behaviour of the attributes.

PR #583, opened 2018-10-23
[MRG]  EHN: support SparseDataFrame when creating a dataset 

glemaitre:
Add support for `SparseDataFrame` when creating a matrix.

PR #592, opened 2018-11-14
[MRG] DEPR: remove the format parameter from create_dataset

glemaitre:
We mentioned in #586 that `format` could be deprecated even if it is breaking the backward compatibility:

https://github.com/openml/openml-python/pull/586#discussion_r231617719

It is what this PR is doing.


openml/datasets/functions.py
```diff
@@ -501,34 +496,26 @@ def create_dataset(name, description, creator, contributor,
...
-                raise ValueError(
-                    'When giving a list or a numpy.ndarray, '
-                    'they should contain a list/ numpy.ndarray '
-                    'for dense data or a dictionary for sparse '
-                    'data. Got {!r} instead.'
-                    .format(data[0])
-                )
-        elif isinstance(data, coo_matrix):
-            d_format = 'sparse_arff'
+    if isinstance(data, (list, np.ndarray)):
```
mfeurer: Hm, don't you need to ckeck for `pd.DataFrame` and `pd.SparseDataFrame`, too?

glemaitre: True

glemaitre:
I will fix this PR after merging #583

Issue #261, opened 2017-05-23
Clean up test server after running unit tests


mfeurer:
If a unit test is successful, delete everything it saved on the server. Otherwise, it's hard to have a clean server that is useful for testing purposes. One way to achieve this: add a 'sentinel' string, i.e. a unique random string, to everything that is uploaded. At unit test shutdown time, search for everything with that string and remove it from the server via an API call.

Neeratyoy:
On a successful publish, a new *entity_type/**entity_id*** is generated. On using `openml.utils._delete_entity(entity_type, entity_id)`, the *entity_type/**entity_id*** is no longer available on the test server. 
On re-publishing the same entity again, a new ID is generated, which is the previous ID+1. Is that an acceptable cleanup? Or is the expectation to not increase the file counter, i.e., on deletion, that ID's space will be re-used by subsequent uploads.

mfeurer:
> Is that an acceptable cleanup?

Yes.

mfeurer:
Added by #735.

PR #607, opened 2018-12-08
[WIP] Task upload

ArlindKadra:

- Implements task upload.

    1. Generates dictionary that represents object.
     2. Generates xml from the dictionary.

- Adds a function that allows tasks to be created  (facilitates task creation for users compared to the task constructor).
- Adds type annotations for the task object.

New unit tests for different kinds of tasks.

Addresses #657, #656


openml/tasks/task.py
```diff
@@ -42,11 +45,88 @@ def remove_tag(self, tag):
...
 
+    def _to_dict(self):
+
+        task_container = OrderedDict()
+        task_dict = OrderedDict([
+            ('@xmlns:oml', 'http://openml.org/openml')
+        ])
+        task_container['oml:task_inputs'] = task_dict
+
+        if self.task_id is not None:
```
mfeurer: Is there a reason to actually allow this?

openml/tasks/task.py
```diff
@@ -42,11 +45,88 @@ def remove_tag(self, tag):
...
+
+        task_dict['oml:task_type_id'] = self.task_type_id
+
+        source_data = OrderedDict([
+            ('@name', 'source_data'),
+            ('#text', str(self.dataset_id))
+
+        ])
+
+        # not a clustering task
```
mfeurer: Hm, could this task-specific code be moved into the actual per-task classes?

tests/test_tasks/test_stat_task.py
```diff
@@ -0,0 +1,38 @@
+from time import time
+
+import openml
+from openml.testing import TestBase
+
+
+class OpenMLStatTest(TestBase):
```
mfeurer: Could you please explain the naming here?

tests/test_tasks/test_task.py
```diff
@@ -1,77 +1,42 @@
...
+        if cls is OpenMLTaskTest:
+            raise unittest.SkipTest(
+                "Skip OpenMLTaskTest tests,"
+                " it's a base class"
+            )
+        super(OpenMLTaskTest, cls).setUpClass()
+
+    def test_download_task(self):
+
+        openml.tasks.get_task(self.task_id)
```
mfeurer: Could you please add a few asserts here to check that the downloaded content is as expected?

tests/test_tasks/test_supervised_task.py
```diff
@@ -0,0 +1,29 @@
...
+    @classmethod
+    def setUpClass(cls):
+        if cls is OpenMLSupervisedTaskTest:
+            raise unittest.SkipTest(
+                "Skip OpenMLSupervisedTaskTest tests,"
+                " it's a base class"
+            )
+        super(OpenMLSupervisedTaskTest, cls).setUpClass()
+
+    def test_get_X_and_Y(self):
```
mfeurer: Could you please rename this method? I find the name misleading.

tests/test_tasks/test_task.py
```diff
@@ -1,77 +1,42 @@
...
+
+    def test_upload_task(self):
```
mfeurer: Would it be possible to create a task which will never be on the server so that the upload can definitely be tested? For example by uploading a dataset prior to creating a task?

ArlindKadra: This is already done, the tasks in the unit tests were using different evalutation strategies which were not present in the test server and the upload was successful.

tests/test_tasks/test_supervised_task.py
ArlindKadra: I can, I did not change the naming since that was the previous name the method had.
What do you think about **test_get_examples_and_labels**?

tests/test_tasks/test_task.py
ArlindKadra: Yes, I can.

tests/test_tasks/test_stat_task.py
ArlindKadra: Yes, good point.

openml/tasks/task.py
ArlindKadra: I understand what you mean. It is doable.

tests/test_tasks/test_supervised_task.py
mfeurer: That would be a descriptive name.

tests/test_tasks/test_task.py
mfeurer: Yes, but they'll be there in the next iteration, right?

ArlindKadra: Yes, normally this would be easily doable if the remove_task functionality would be on the connector. However it is only available in the website. 
Are you suggesting to upload a dataset continuously so that the id always changes resulting in a new task?
Is there no plan for the moment to port the remove_task to the connector?

tests/test_tasks/test_supervised_task.py
ArlindKadra: @mfeurer Now that I am thinking about it, we are testing the class method and the method that we are testing is **get_X_and_y()**.
Should we still rename?

tests/test_tasks/test_task.py
ArlindKadra: Now that I am revisiting this comment, adding asserts here is not possible, since this method will be called only on the classes that extend the base class and that information is task specific.
What I can do is override this method on the extending classes and perform the check according to the tasks.

tests/test_tasks/test_supervised_task.py
mfeurer: Could you please provide more context? I can't follow you right now.

tests/test_tasks/test_task.py
mfeurer: I currently don't plan to add this, but I think @janvanrijn added the delete functionality in PR #620. However, I wouldn't mind if this was kept on the server for now.

tests/test_tasks/test_task.py
mfeurer: That would be great.

tests/test_tasks/test_task.py
mfeurer: We should actually address this in #261 and actually also remove datasets, flows and runs (at some point, not now).

tests/test_tasks/test_task.py
```diff
@@ -1,85 +1,66 @@
...
+        if cls is OpenMLTaskTest or OpenMLSupervisedTaskTest:
+            raise unittest.SkipTest(
+                "Skip OpenMLTaskTest tests,"
+                " it's a base class"
+            )
+        super(OpenMLTaskTest, cls).setUpClass()
+
+    def test_download_task(self) -> OpenMLTask:
 
-    @mock.patch('openml.tasks.functions.get_dataset', autospec=True)
```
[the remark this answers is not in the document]
ArlindKadra: yes, I think we discussed this and we agreed that we would remove it since we are actually checking the get_dataset function. Or do I remember wrong?

openml/tasks/task.py
```diff
@@ -161,27 +305,69 @@ def __init__(self, task_id, task_type_id, task_type, data_set_id,
...
-        self.number_of_clusters = number_of_clusters
+        self.target_name = target_name
+        self.estimation_procedure_id = estimation_procedure_id
+
+    def _to_dict(self) -> Dict[str, OrderedDict]:
+
+        task_container = super(OpenMLClusteringTask, self)._to_dict()
+        task_dict = task_container['oml:task_inputs']
+
+        task_dict['oml:input'].append(
```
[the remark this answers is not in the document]
ArlindKadra: Nice observation. It also made me realise that the code for some of the tasks can be incomplete since when the class is downloaded from the server it has more information concerning the estimation procedure.

openml/tasks/task.py
```diff
@@ -106,10 +204,12 @@ def __init__(self, task_id, task_type_id, task_type, data_set_id,
         self.estimation_procedure["parameters"] = estimation_parameters
         self.estimation_parameters = estimation_parameters
         self.estimation_procedure["data_splits_url"] = data_splits_url
+        self.estimation_procedure_id = estimation_procedure_id
         self.target_name = target_name
         self.split = None
 
-    def get_X_and_y(self):
+    def get_X_and_y(self) -> \
+            Tuple[np.ndarray, np.ndarray]:
```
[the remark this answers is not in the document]
ArlindKadra: No real reason. removing the linebreak should be more consistent with the way that I have been using it.

tests/test_tasks/test_task.py
```diff
@@ -1,85 +1,66 @@
...
-        self.assertEqual(Y.dtype, float)
+        task = openml.tasks.get_task(self.task_id)
+        # adding sentinel so we can have a new dataset
+        # hence a "new task" to upload
+        task_dataset = openml.datasets.get_dataset(task.dataset_id)
+        task.dataset_id = self._upload_dataset(task_dataset)
+        task.estimation_procedure_id = self.estimation_procedure
+        try:
+            task.publish()
+        except OpenMLServerException as e:
```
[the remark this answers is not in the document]
ArlindKadra: Yes, true. I should remove this.

tests/test_tasks/test_task.py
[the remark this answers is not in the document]
mfeurer: Unlikely, the reason I am asking is because removing this test also removes the only check that the class labels are loaded. Could you please add such a check again (obviously only for the classification and learning curve task)? Also, it would be good if that test actually used a task from the server instead of a task which is kept locally.

openml/tasks/task.py
```diff
@@ -109,32 +121,39 @@ def remove_tag(self, tag: str):
         data = {'task_id': self.task_id, 'tag': tag}
         openml._api_calls._perform_api_call("/task/untag", 'post', data=data)
 
-    def _to_dict(self) -> Dict[str, OrderedDict]:
+    def _to_dict(self) -> dict:
```
mfeurer: Why change this? Anyway, it should be `Dict` instead of `dict`. In case you changed it because of the ordered dict, you can use `'OrderedDict[str, OrderedDict]'`.

tests/test_tasks/test_task.py
```diff
@@ -65,10 +119,32 @@ def _upload_dataset(self, dataset: OpenMLDataset) -> int:
...
-            # published. Return old id.
-            return dataset.dataset_id
+        # Providing both dataset file and url
+        # raises an error when uploading.
+        dataset.url = None
+
+        return dataset.publish()
+
+    @staticmethod
+    def _wait_dataset_activation(
```
mfeurer: Are you still using this method? I couldn't find any usage of it.

tests/test_tasks/test_task.py
```diff
@@ -1,50 +1,104 @@
...
+            random_feature = random_dataset.features[random_feature_index]
+            if self.task_type_id == 2:
+                if random_feature.data_type == 'numeric':
+                    break
+            else:
+                if random_feature.data_type == 'nominal':
+                    break
+        return random_feature.name
+
+    def _reupload_dataset(self, dataset: OpenMLDataset) -> int:
```
mfeurer: Do you use this function anywhere?

openml/tasks/__init__.py
```diff
@@ -8,7 +8,12 @@
     TaskTypeEnum,
 )
 from .split import OpenMLSplit
-from .functions import (get_task, get_tasks, list_tasks)
+from .functions import (
+    create_task,
+    get_task,
+    get_tasks,
+    list_tasks
```
mfeurer: Please add a comma at the end of this line.

openml/tasks/task.py
```diff
@@ -1,26 +1,56 @@
 import io
 import os
+from collections import OrderedDict
+from typing import Union, Tuple, Dict, List, Optional
+from abc import ABC
```
mfeurer: Could you please sort the import alphabetically?

openml/tasks/functions.py
```diff
@@ -450,3 +450,60 @@ def _create_task_from_xml(xml):
...
+
+
+def create_task(
+        task_type_id: int,
+        dataset_id: int,
+        target_name: str,
+        estimation_procedure_id: int,
+        evaluation_measure: str = None,
+        **kwargs
+):
```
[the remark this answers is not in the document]
ArlindKadra: Initially, I had an `OpenMLTask` return type, since it is the base class. However, the class can return different task types based on the task type id, so `Union[task_type_1, ..., task_type_5]`  is a viable option, but too specific. I was not sure on what notation to use.

openml/tasks/functions.py
```diff
@@ -450,3 +450,60 @@ def _create_task_from_xml(xml):
...
+    """Create a task based on different given
+    attributes.
+
+    Builds different task objects based on the
+    task type.
```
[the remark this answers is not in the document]
ArlindKadra: > Could you please add a comment saying that the user should check the arguments of the class she is about to create to figure out how to create the task?

Sure thing

openml/tasks/task.py
ArlindKadra: I wanted to change it to `OrderedDict`, however I decided to check the notation used in flows and to be consistent I used the same one. Both functions return an `OrderedDict.`
https://github.com/openml/openml-python/blob/c7db12287aa25415c854d15549cfaa05835cd7d6/openml/flows/flow.py#L151

Personally I would want `OrderedDict[str, OrderedDict]`.

tests/test_tasks/test_task.py
ArlindKadra: Regarding this function and the one below, I am not using them. I created them with the initial implementation of task upload that we ended up not using. I left the functions since they might be of use in the future.
However, I can comment them out, or remove them.

openml/tasks/task.py
ArlindKadra: Ops, I missed that.

tests/test_tasks/test_task.py
mfeurer: Yes, please remove them then.

openml/tasks/task.py
mfeurer: Then go for it.

openml/tasks/functions.py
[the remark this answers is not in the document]
mfeurer: Why would that be too specific? I think the union would be the best option here.

openml/tasks/task.py
ArlindKadra: I already did it, I did not resolve the conversation so it would act as a reminder, in case you wanted to edit the flows `_to_dict`.

openml/tasks/functions.py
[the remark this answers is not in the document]
ArlindKadra: Then specific would be the wrong word used by me in this situation. What I meant, was that given that there are 4 possible classes, the return would be: 
`Union[class1, Union[class2, Union[class3, class4]]]`
In the future adding more classes would add more Unions. So I am not sure if this appears confusing compared to having the base class, which in turn does not showcase all the task types.
I think the Union is the best option too and since you agree, I am going with that one.

openml/tasks/task.py
mfeurer: Not now, this is about task upload.

openml/tasks/functions.py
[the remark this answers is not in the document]
mfeurer: I think that's the best decision at the moment. We can still update this if when there are more classes coming up.

Regarding the specific annotation, doesn't `Union[class1, class2, class3, class4]` work as well?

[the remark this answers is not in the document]
ArlindKadra: > I think that's the best decision at the moment. We can still update this if when there are more classes coming up.
> 
> Regarding the specific annotation, doesn't `Union[class1, class2, class3, class4]` work as well?

more than right, I read the documentation for the redundant arguments by mistake and I did not pay attention. I thought there was a limit. Good catch.

openml/extensions/sklearn/extension.py
```diff
@@ -1264,29 +1264,32 @@ def _prediction_to_probabilities(y: np.ndarray, classes: List[Any]) -> np.ndarra
...
-            if proba_y.shape[1] != len(task.class_labels):
-                message = "Estimator only predicted for {}/{} classes!".format(
-                    proba_y.shape[1], len(task.class_labels),
-                )
-                warnings.warn(message)
-                openml.config.logger.warn(message)
+                if task.class_labels is not None:
+                    proba_y = _prediction_to_probabilities(pred_y, list(task.class_labels))
+
+            if task.class_labels is not None:
```
mfeurer: I think we should make this fail if `task.class_labels is None`.

openml/runs/functions.py
```diff
@@ -452,11 +455,12 @@ def _calculate_local_measure(sklearn_fn, openml_name):
             for i, tst_idx in enumerate(test_indices):
 
                 arff_line = [rep_no, fold_no, sample_no, tst_idx]  # type: List[Any]
-                for j, class_label in enumerate(task.class_labels):
-                    arff_line.append(proba_y[i][j])
+                if task.class_labels is not None:
```
mfeurer: Same as above.

openml/runs/run.py
```diff
@@ -216,30 +216,39 @@ def _generate_arff_dict(self) -> 'OrderedDict[str, Any]':
...
+            class_labels = task.class_labels
+            instance_specifications = [
+                ('repeat', 'NUMERIC'),
+                ('fold', 'NUMERIC'),
+                ('sample', 'NUMERIC'),
+                ('row_id', 'NUMERIC')
+            ]
+
+            arff_dict['attributes'] = instance_specifications
+            if class_labels is not None:
```
mfeurer: same here.

openml/runs/run.py
```diff
@@ -216,30 +216,39 @@ def _generate_arff_dict(self) -> 'OrderedDict[str, Any]':
...
-                                       'NUMERIC')
-                                      for i in range(len(class_labels))]
-            prediction_and_true = [('prediction', class_labels),
-                                   ('correct', class_labels)]
-            arff_dict['attributes'] = (instance_specifications
-                                       + prediction_confidences
-                                       + prediction_and_true)
+
+            arff_dict['attributes'] = instance_specifications
+            if class_labels is not None:
```
mfeurer: and here as well.

openml/tasks/functions.py
```diff
@@ -441,3 +442,68 @@ def _create_task_from_xml(xml):
         raise NotImplementedError('Task type %s not supported.' %
                                   common_kwargs['task_type'])
     return cls(**common_kwargs)
+
+
+def create_task(
+        task_type_id: int,
+        dataset_id: int,
+        target_name: str,
```
mfeurer: I'm surprised this works, shouldn't it `target_name: Optional[str]` as the target is optional for clustering tasks?

openml/tasks/task.py
```diff
@@ -396,19 +398,24 @@ def get_X(
         )
         return X_and_y
 
-    def _to_dict(self) -> Dict[str, OrderedDict]:
+    def _to_dict(self) -> 'OrderedDict[str, OrderedDict]':
 
         task_container = super(OpenMLClusteringTask, self)._to_dict()
-        task_dict = task_container['oml:task_inputs']
 
+        # Right now not supported as a feture
```
mfeurer: feture -> feature

tests/test_tasks/test_clustering_task.py
```diff
@@ -1,46 +1,29 @@
...
 
         super(OpenMLClusteringTaskTest, self).setUp()
-        # no clustering tasks on test server
-        self.production_server = 'https://openml.org/api/v1/xml'
-        self.test_server = 'https://test.openml.org/api/v1/xml'
-        openml.config.server = self.production_server
         self.task_id = 146714
+        self.task_type_id = 5
         self.estimation_procedure = 17
+        self.production_server = "https://openml.org/api/v1/xml"
```
mfeurer: Isn't that available in the main testing class?

openml/tasks/functions.py
ArlindKadra: Nicely spotted, it should be optional. This raises a nice problem I guess, the unit test runs the `create_task` function with `target_name` available for all task types. 
I can modify it, to run clustering tasks without a target_name. With a `target_name` it covers the case with extra information, however in this case it backfired.
You think we should check for both? 
Or you think now that this is covered we can leave it like that.

tests/test_tasks/test_clustering_task.py
ArlindKadra: Yes, it is. Fixed.

openml/tasks/functions.py
mfeurer: > You think we should check for both?

Yes, it would be good to test both.

PR #609, opened 2018-12-19
Subclass all test classes from openml test helper

### LINKED ISSUE #392: CI runs on production server
It looks like something is wrong with the unit tests, because they upload runs to the production server: https://www.openml.org/r/8857778

Please don't :). Unless these experiments are actually useful?

mfeurer:
Fixes #392

### DIFF
--- a/openml/runs/functions.py
+++ b/openml/runs/functions.py
@@ -127,6 +127,11 @@ def run_flow_on_task(flow, task, avoid_duplicate_runs=True, flow_tags=None,
             raise ValueError('flow.flow_id is not None, but the flow does not'
                              'exist on the server according to flow_exists')
         _publish_flow_if_necessary(flow)
+        # if the flow was published successfully
+        # and has an id
+        if flow.flow_id is not None:
+            flow_id = flow.flow_id
+
 
     data_content, trace, fold_evaluations, sample_evaluations = res
     if not isinstance(flow.flow_id, int):
--- a/openml/testing.py
+++ b/openml/testing.py
@@ -77,16 +77,20 @@ class TestBase(unittest.TestCase):
                 raise
         openml.config.server = self.production_server
 
-    def _add_sentinel_to_flow_name(self, flow, sentinel=None):
+    def _get_sentinel(self, sentinel=None):
         if sentinel is None:
             # Create a unique prefix for the flow. Necessary because the flow is
             # identified by its name and external version online. Having a unique
             #  name allows us to publish the same flow in each test run
             md5 = hashlib.md5()
             md5.update(str(time.time()).encode('utf-8'))
+            md5.update(str(os.getpid()).encode('utf-8'))
             sentinel = md5.hexdigest()[:10]
             sentinel = 'TEST%s' % sentinel
+        return sentinel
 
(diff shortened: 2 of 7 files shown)


mfeurer:
Hey @janvanrijn there's an unexpected server error happening: https://api.travis-ci.org/v3/job/469988062/log.txt

janvanrijn:
Will immediately make that error more XML-able.

joaquinvanschoren:
Seems like you're uploading a dataset with an ID that's already taken:
<p>Error Number: 1062</p><p>Duplicate entry 'Pandas_testing_dataset-5133' for key 'nameID'</p>

janvanrijn:
Looks like this is a result of the following:
* Due to bad luck, two times a dataset with the same name got uploaded at the same time. Should not happen in practice.
* Usually, on the main server this would not give a mysql error as they are hidden away and replaced with a xml error message (at least, most, including this one, are). On the test server we run the database in debug mode, meaning that every error will result in a verbose error message like this one. Given that this was only an incident, I feel like we should keep it that way.

mfeurer:
It appears that there are some rather random test failures:
```
Unexpected failing examples:
/home/travis/build/openml/openml-python/examples/introduction_tutorial.py failed leaving traceback:
  File "/home/travis/build/openml/openml-python/examples/introduction_tutorial.py", line 70, in <module>
    run = openml.runs.run_flow_on_task(flow, task, avoid_duplicate_runs=False)
  File "/home/travis/build/openml/openml-python/openml/runs/functions.py", line 147, in run_flow_on_task
    "same: '%s' vs '%s'" % (str(flow.flow_id), str(flow_id))
ValueError: Result from API call flow_exists and flow.flow_id are not same: '58748' vs 'False'
```
and happens in
```
    # in case the flow not exists, flow_id will be False (as returned by
    # flow_exists). Also check whether there are no illegal flow.flow_id values
    # (compared to result of openml.flows.flow_exists)
    if flow_id is False:
        if flow.flow_id is not None:
            raise ValueError('flow.flow_id is not None, but the flow does not'
                             'exist on the server according to flow_exists')
        _publish_flow_if_necessary(flow)

    data_content, trace, fold_evaluations, sample_evaluations = res
    if not isinstance(flow.flow_id, int):
        # This is the usual behaviour, where the flow object was initiated off
        # line and requires some additional information (flow_id, input_id for
        # each hyperparameter) to be usable by this library
        server_flow = get_flow(flow_id)
        openml.flows.flow._copy_server_fields(server_flow, flow)
        openml.flows.assert_flows_equal(flow, server_flow,
                                        ignore_parameter_values=True)
    else:
        # This can only happen when the function is called directly, and not
        # through "run_model_on_task"
        if flow.flow_id != flow_id:
            # This should never happen, unless user made a flow-creation fault
            raise ValueError(
                "Result from API call flow_exists and flow.flow_id are not "
                "same: '%s' vs '%s'" % (str(flow.flow_id), str(flow_id))
            )
```
Could it be that we need to set `flow_id` after `_publish_flow_if_necessary` to avoid this error?

mfeurer:
Thanks @ArlindKadra this appears to solve the issue. The build failures are not related.

PR #618, opened 2019-02-14
Fix issues

mfeurer:
Fixes jamming the server with too many concurrent requests by adding a new flag.

### DIFF
--- a/doc/progress.rst
+++ b/doc/progress.rst
@@ -20,6 +20,8 @@ Changelog
   OpenML.
 * ADD #564: New helpers to access the structure of a flow (and find its
   subflows).
+* ADD #618: The software will from now on retry to connect to the server if a
+  connection failed. The number of retries can be configured.
 * FIX #538: Support loading clustering tasks.
 * FIX #464: Fixes a bug related to listing functions (returns correct listing
   size).
--- a/openml/_api_calls.py
+++ b/openml/_api_calls.py
@@ -1,9 +1,7 @@
-import io
-import os
+import time
 import requests
 import warnings
 
-import arff
 import xmltodict
 
 from . import config
@@ -71,7 +69,12 @@ def _read_url_files(url, data=None, file_elements=None):
         file_elements = {}
     # Using requests.post sets header 'Accept-encoding' automatically to
     # 'gzip,deflate'
-    response = requests.post(url, data=data, files=file_elements)
+    #response = requests.post(url, data=data, files=file_elements)
+    response = send_request(
+        request_method='post',
+        url=url,
+        kwargs={'data': data, 'files': file_elements},
+    )
     if response.status_code != 200:
         raise _parse_server_exception(response, url=url)
     if 'Content-Encoding' not in response.headers or \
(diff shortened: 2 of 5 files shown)


doc/progress.rst
```diff
@@ -20,6 +20,8 @@ Changelog
   OpenML.
 * ADD #564: New helpers to access the structure of a flow (and find its
   subflows).
+* ADD #618: The software will from now on retry to connect to the server if a
+  connection failed. The number of retries can be configured.
```
janvanrijn: can you also add #602 please? maybe close to columntransformer (sorry i forgot)

openml/_api_calls.py
```diff
@@ -102,12 +103,39 @@ def _read_url(url, data=None):
     return response.text
 
 
+def send_request(request_method, url, kwargs):
```
janvanrijn: i hate kwargs, it's so vague. Do we need this level of flexibility or can we just add optional arguments?

doc/progress.rst
mfeurer: As the column transformer is only added in this new release it makes no sense to also add a fix to this new features in the change log.

openml/_api_calls.py
mfeurer: Changed.

openml/_api_calls.py
```diff
@@ -1,22 +1,17 @@
...
-import arff
 import xmltodict
 
 from . import config
 from .exceptions import (OpenMLServerError, OpenMLServerException,
                          OpenMLServerNoResult)
 
 
-def _perform_api_call(call, data=None, file_elements=None,
-                      add_authentication=True):
```
joaquinvanschoren: don't you need the authentication flag? GET requests don't require authentication, POST/DELETE requests do.

openml/testing.py
```diff
@@ -65,6 +65,10 @@ def setUp(self):
                 with open(openml.config.config_file, 'w') as fh:
                     fh.write('apikey = %s' % openml.config.apikey)
 
+        # Increase the number of retries to avoid spurios server failures
+        self.connection_n_retries = openml.config.connection_n_retries
+        openml.config.connection_n_retries = 10
```
joaquinvanschoren: I'm a bit confused about the number of retries being stored in the _default dict, the openml.config file and self.connection_n_retries. So, by default it is 2, then when that doesn't work, it is increased to 10 and stored in the openml.config, and then it stops (it doesn't get higher than 10)?

openml/_api_calls.py
```diff
@@ -102,12 +104,44 @@ def _read_url(url, data=None):
...
+                    raise NotImplementedError()
+                break
+            except (
+                    requests.exceptions.ConnectionError,
+                    requests.exceptions.SSLError,
+            ) as e:
+                if i == n_retries:
+                    raise e
+                else:
+                    time.sleep(0.1 * i)
```
joaquinvanschoren: nice

openml/_api_calls.py
mfeurer: That is handled internally and does not need to be an argument to the openml-api call API.

openml/testing.py
mfeurer: Nope, it's increased to 10 for testing in every case. It is stored in the class to restore the original value afterwards.

mfeurer: I agree that it is unfortunate that the value of 2 is twice in the original file, I just made a note to change it later.

PR #620, opened 2019-02-16
[WIP] Add support for Studies

janvanrijn:


openml/study/functions.py
```diff
@@ -4,44 +4,49 @@
...
-        return result_dict[tag]
-    elif isinstance(result_dict[tag], dict):
-        return [result_dict[tag]]
-    else:
-        raise TypeError()
-
-
-def get_study(study_id, type=None):
-    '''
+def get_study(study_id, knowledge_type=None):
```
mfeurer: Could you please document the arguments? I have no idea what `knowledge_type` is.

openml/study/functions.py
```diff
@@ -53,7 +58,180 @@ def get_study(study_id, type=None):
...
+            flows=flows, 
+            setups=setups,
+            runs=runs
+    )
     return study
+
+
+def create_study(alias, benchmark_suite, name, description, run_ids):
+    """
+    Creates an OpenML study (collection of knowledge types, where the runs
```
mfeurer: Could you please add a description of what a knowledge type is?

openml/study/functions.py
```diff
@@ -53,7 +58,180 @@ def get_study(study_id, type=None):
...
+        brief description (meta-info)
+    task_ids : list
+        a list of task ids associated with this study
+    
+    Returns:
+    --------
+    OpenMLStudy
+        A local OpenML study object (call publish method to upload to server)
+    """
+    return OpenMLStudy(
```
mfeurer: Hm, I'm not sure about this one. Does the server already distinguish studies and suites? If yes, we should add a subclass `OpenMLSuite` to make the code easier to understand.

openml/study/study.py
```diff
@@ -1,19 +1,35 @@
...
+        Can be used to obtain all relevant information from a study at once.
+    
+        Parameters
+        ----------
+        study_id : int
             the study id
+        alias : str (optional)
+            a string ID, unique on server (url-friendly)
+        main_knowledge_type : str
+            the knowledge type (e.g., task, run) that is core in this study.
```
mfeurer: Is this only task or run? If yes, could you please explicitly state this? Also, I'm wondering whether this should be renamed to for example `main_entity_type` to highlight that this runs on different types of entities (which is in my opinion easier to relate to the different concepts like dataset, task etc. than knowledge).

tests/test_study/test_study_functions.py
```diff
@@ -1,26 +1,123 @@
...
+        study_id = study.publish()
+        self.assertGreater(study_id, 0)
+        
+        # verify main meta data
+        study_downloaded = openml.study.get_study(study_id)
+        self.assertEqual(study_downloaded.alias, fixture_alias)
+        self.assertEqual(study_downloaded.name, fixture_name)
+        self.assertEqual(study_downloaded.description, fixture_descr)
+        # verify resources
+        self.assertEqual(study_downloaded.flows, None)
```
mfeurer: Could you please replace these by the more descriptive `self.assertNone`?

tests/test_study/test_study_functions.py
```diff
@@ -1,26 +1,123 @@
...
+        
+        study = openml.study.create_study(
+            alias=fixt_alias,
+            benchmark_suite=None,
+            name=fixt_name,
+            description=fixt_descr,
+            run_ids=list(run_list.keys())
+        )
+        study_id = study.publish()
+        self.assertGreater(study_id, 0)
```
mfeurer: Maybe you should also check that the number of runs attached is larger than zero?

janvanrijn: I now added a line that checks the run list size, this implicitly ensures that the runs are greater than zero.

openml/study/functions.py
janvanrijn: The server does not distinguish between those. It is a parameter in the database (main knowledge type is either runs or tasks)

mfeurer: I assumed that. But wouldn't it from a usage perspective make sense to have two separate objects for this?

tests/test_study/test_study_functions.py
mfeurer: Thanks!

openml/study/functions.py
janvanrijn: Alternatively we could have 2 objects that both have a different contructor. However i figured it might be complicated because since the server always returns a "study" tag and we have to differentiate. Seems like more potential legacy and maintenance to me, but all together it doesn't make a great difference

mfeurer: Okay, let's then leave it like it is and see how things are going.

openml/study/study.py
janvanrijn: I think that the terms `entities` and `knowledge type` are not consistently documented within OpenML. Maybe the term `entity type` and `entities` would be better though

janvanrijn: Done. Please check the PR again, as I replaced the word knowledge type with entity type throughout the relevant files

mfeurer:
Looks good to me. Except that one line which basically says to not merge this PR.

mfeurer:
And it seems there are some minor flake8 errors in the push test. I really don't know what to do with the flake8 errors in the PR test as they are so many and would only happen after the merge.

janvanrijn:
> Looks good to me. Except that one line which basically says to not merge this PR.

we need to wait for the PHP PR to be merged. 

> And it seems there are some minor flake8 errors in the push test. I really don't know what to do with the flake8 errors in the PR test as they are so many and would only happen after the merge.

Bigger issue: some lines are OK, other still contain mistakes (legacy mistakes, and we don't really want to solve them all at once)

Clearly it would be beneficial to have the whole project flake compatible. We can maybe ask the programmer to do so on one of the first days? (Once all big PR's are merged) From there on flake can run on all lines, not just the diff, and we don't have these problems anymore. 

The problems introduced in this PR will be fixed by me.

janvanrijn:
```
openml/study/__init__.py:2:1: F401 '.functions.get_study' imported but unused
from .functions import get_study, create_study, create_benchmark_suite, \
^
openml/study/__init__.py:2:1: F401 '.functions.create_study' imported but unused
from .functions import get_study, create_study, create_benchmark_suite, \
^
openml/study/__init__.py:2:1: F401 '.functions.create_benchmark_suite' imported but unused
from .functions import get_study, create_study, create_benchmark_suite, \
^
openml/study/__init__.py:2:1: F401 '.functions.delete_study' imported but unused
from .functions import get_study, create_study, create_benchmark_suite, \
^
openml/study/__init__.py:2:1: F401 '.functions.attach_to_study' imported but unused
from .functions import get_study, create_study, create_benchmark_suite, \
^
openml/study/__init__.py:2:1: F401 '.functions.detach_from_study' imported but unused
from .functions import get_study, create_study, create_benchmark_suite, \
```

flake8 residuals. this is fine right? Can we inform it that this warning in `__init__` files is OK?

tests/test_runs/test_run_functions.py
```diff
@@ -456,7 +456,8 @@ def determine_grid_size(param_grid):
             # suboptimal (slow), and not guaranteed to work if evaluation
             # engine is behind.
             # TODO: mock this? We have the arff already on the server
-            self._wait_for_processed_run(run.run_id, 200)
+            print(run.run_id)
```
mfeurer: Could you please remove this call to the print function?

mfeurer:
Looks good to me except for the failing unit tests.

Issue #624, opened 2019-02-21
PEP8 Compliance


PGijsbers:
The `develop` branch is not PEP8 compliant. Specifically, after cloning the repo and running flake8, the following output is produced (ignoring E402 as per [script](https://github.com/openml/openml-python/blob/master/ci_scripts/flake8_diff.sh#L128)):

```
.\doc\conf.py:18:1: F401 'sphinx_gallery.sorting.ExplicitOrder' imported but unused
.\doc\conf.py:18:1: F401 'sphinx_gallery.sorting.FileNameSortKey' imported but unused
.\doc\conf.py:30:1: E265 block comment should start with '# '
.\doc\conf.py:38:1: E265 block comment should start with '# '
.\doc\conf.py:67:1: E265 block comment should start with '# '
.\doc\conf.py:90:1: E265 block comment should start with '# '
.\doc\conf.py:94:1: E265 block comment should start with '# '
.\doc\conf.py:96:1: E265 block comment should start with '# '
.\doc\conf.py:104:1: E265 block comment should start with '# '
.\doc\conf.py:107:1: E265 block comment should start with '# '
.\doc\conf.py:111:1: E265 block comment should start with '# '
.\doc\conf.py:115:1: E265 block comment should start with '# '
.\doc\conf.py:121:1: E265 block comment should start with '# '
.\doc\conf.py:124:1: E265 block comment should start with '# '
.\doc\conf.py:138:5: E265 block comment should start with '# '
.\doc\conf.py:201:1: E265 block comment should start with '# '
.\doc\conf.py:204:1: E265 block comment should start with '# '
.\doc\conf.py:208:1: E265 block comment should start with '# '
.\doc\conf.py:213:1: E265 block comment should start with '# '
.\doc\conf.py:223:1: E265 block comment should start with '# '
.\doc\conf.py:227:1: E265 block comment should start with '# '
.\doc\conf.py:231:1: E265 block comment should start with '# '
.\doc\conf.py:238:1: E265 block comment should start with '# '
.\doc\conf.py:241:1: E265 block comment should start with '# '
.\doc\conf.py:244:1: E265 block comment should start with '# '
.\doc\conf.py:247:1: E265 block comment should start with '# '
.\doc\conf.py:250:1: E265 block comment should start with '# '
.\doc\conf.py:253:1: E265 block comment should start with '# '
.\doc\conf.py:256:1: E265 block comment should start with '# '
.\doc\conf.py:261:1: E265 block comment should start with '# '
[...]

PR #625, opened 2019-02-22
[WIP] Fix624 pep8

PGijsbers:
edit: marked WIP as I need to ensure unit tests don't fail first.

The `develop` branch is not PEP8 compliant (see #624). This fixes that for the core module*, ignoring E402 as per the script.
Examples and unit tests are excluded from this PR\*\*, for the following reasons:
 - I have to set up documentation tools to see how and if documentation would be impacted by PEP8 changes.
 - I hope that we can have a discussion regarding the 80-character line limit.
 - Keeping the PR close to HEAD. For two reasons above, and a lot of test files also violating PEP8, including tests and documentation can incur serious delays. I would like to avoid another months-long PR if reasonable.

In/around code that needed to be changed for PEP8 reasons, I sparingly also refactored and fixed a bug. Most notably:
 - openml.tasks.split.OpenMLSplit.\_\_eq\_\_ featured a bug where they were only not equal if _all_ elements differed, rather than _any_ ([here](https://github.com/openml/openml-python/commit/0e5918f599ab088a9005c768c3da341ad43a98e6#diff-aac2ca0e70ca5eab23745f707a903134L52))
 - openml.flows.functions.[_get_fn_arguments_with_defaults](https://github.com/openml/openml-python/commit/d270d023e007de14e670c5f7bfb0ed6d1d952c7d#diff-b2251035552ccbb4c6cc9367ed6b035aL603) more pythonic and removed Py2 support in the process

When adapting the PEP8 check, has there been any discussion on the 80-character limit?
I personally think that a 100 or 120 character limit is typically better.
The 80 character limit limits reasonably verbose variable names, especially at higher indent levels.
Examples in the code that I think would be fine to be on one line:
 - [1](https://github.com/openml/openml-python/commit/d270d023e007de14e670c5f7bfb0ed6d1d952c7d#diff-b2251035552ccbb4c6cc9367ed6b035aR680) verbose names cause exceeding the limit by 3 characters.
 - [2](https://github.com/openml/openml-python/blob/fix624_pep8/openml/runs/run.py#L197) a simple operation, again too verbose because of the string. 
 - [3](https://github.com/openml/openml-python/blob/fix624_pep8/openml/utils.py#L100) documentation line. Possibly particularly problematic (how would documentation interact with #noqa flake8 exceptions?).
 - [4](https://github.com/openml/openml-python/blob/develop/openml/tasks/split.py#L26) [5](https://github.com/openml/openml-python/blob/develop/openml/tasks/split.py#L112) multi-level indexing with variables as indices quickly creates long lines that are still easy to understand.

If developers try to avoid excessively verbose names or chaining functions, I think these are fine exceptions. Modern screens can easily show two 100 or 120 character lines side-by-side. Limiting to 80 characters leads to scenarios that (in my opinion) hurt the legibility more like splitting lines with square brackets as seen in [4](https://github.com/openml/openml-python/blob/develop/openml/tasks/split.py#L26).

\* There is one exception [utils#100](https://github.com/openml/openml-python/blob/fix624_pep8/openml/utils.py#L100) violating the character limit (85>80). This again has to do with that I do not know how documentation will be affected by breaking up the change.

\*\* [this](https://github.com/openml/openml-python/pull/625/files#diff-087f76652d5a0b4b57b444fb1f89fd2d) is the exception that might lead to visible changes. After I realized my work _might_ impact the final rendering of documentation, I stopped. I could revert this change if desired.


mfeurer:
I'd be happy to merge that, and I also start too see why more than 80 characters per line might be a good idea. How about 100 and we see if this goes well?

@janvanrijn any objections on solving all PEP8 issues at once?

@PGijsbers I think it would be most compelling if you could also change the flake8 checking script so that it checks the whole library at once.

PGijsbers:
Invoking flake8 on all code instead of just diffs seems easy enough. Considering it's a quick operation anyway even for the whole repo, I don't know/think that it's bad practice (especially since you don't need to sort through diffs). 
Changing the max line length is also easy if we can agree on it (100 as a start is fine for me).

PGijsbers:
Not sure why the previous tests failed - the changed code only had [one minor refactor](https://github.com/openml/openml-python/pull/625/files#diff-efa6168096f57dec6cd3fa2c9aff403bL135) but might have introduced off-by-one errors? Even so, the actual numbers are also different in [pr](https://travis-ci.org/openml/openml-python/jobs/497071802#L1269) and [push](https://travis-ci.org/openml/openml-python/jobs/497071782#L1262).

janvanrijn:
This is really cool stuff @PGijsbers , thanks for taking care of this. 

> I'd be happy to merge that, and I also start too see why more than 80 characters per line might be a good idea. How about 100 and we see if this goes well?

I usually do 120, this seems the default for PyCharm. However keep in mind that for consistency with sklearn 80 is not a very bad idea. (Fine with both)

PGijsbers:
Flake seems to be failing on warnings, do we want this behavior?
This brings up another issue I noticed (but did not think was very important right now), in that there are two different incompatible warnings: W503 and W504 which forbid a linebreak before and after an operator, respectively (see [this](https://travis-ci.org/openml/openml-python/jobs/497099401#L873)).
Do we want to enforce one or the other?
I personally feel that we should not enforce neither, but only set a preference. In my opinion, the preference should be for a line break after the operator, unless it gives a bad visual indent (as is sometimes the case in if-statements for example).

PGijsbers:
Then I propose the following changes w.r.t. line length:
 - I will set the max length to 100.
 - For lines longer than 100 characters but shorter than 120 characters, refactoring is encouraged but disabling them with flake with `# noqa: E501` is okay too.
 - We evaluate at some point if those extra characters are generally warranted or not.

Can either of you two take a look at the [failed test](https://travis-ci.org/openml/openml-python/jobs/497143872#L1262)? I don't understand how it fails with [new code but not old code](https://github.com/openml/openml-python/pull/625/commits/af6d5c6980ba0c5d2fdeacbaea61b2dd97b783ce). As far as I can tell, in neither scenario should `BATCH_SIZE_ORIG` be modified in the first place, so the lines should not affect the unit test. Moreover it seems that every time the test gets executed, different amount of (increasing) datasets are found. My theory is that someone is simply uploading datasets, and my change got rejected by pure chance? Could this be possible?

mfeurer:
I think that this PR should not fail on warnings to keep it manageable.

I think that we should prefer to have operators at the beginning of a line (following math notation).

I'm fine with 100 and 120 character setup.

And @janvanrijn agrees on your assessment of the error in #626 .

PGijsbers:
I checked the failed tests and it seems to stem from issue #626 and flake8 checking the whole repo (and me only editing the ./openml directory). Looks good to merge to me.

joaquinvanschoren:
Thanks Pieter!

It’s great that the line length has been bumped. I often had to do weird
line breaks with chaining functions.

PR #630, opened 2019-02-24
Finish pep8 compliance

mfeurer:


mfeurer:
The script was huge because it performed flake8 on the difference between the latest commit and the common ancestor between the current branch and the development branch. These differences had to be fetched and assembled. Now the script simply checks everything.

ci_scripts/flake8_diff.sh
```diff
@@ -1,156 +1,3 @@
 #!/bin/bash
 
-# Inspired from https://github.com/scikit-learn/scikit-learn/blob/master/build_tools/travis/flake8_diff.sh
```
PGijsbers: I had left the original script in-tact originally since i wasn't sure if we wanted to remove everything (e.g. shortcutting if no new commits). But I would prefer this way as well, it's much easier to parse and flake8 is not the bottleneck in our CI procedure anyway.

examples/flows_and_runs_tutorial.py
```diff
@@ -58,7 +58,8 @@
 ############################################################################
 # Share the run on the OpenML server
 #
-# So far the run is only available locally. By calling the publish function, the run is sent to the OpenML server:
+# So far the run is only available locally. By calling the publish function,
```
PGijsbers: Did you double-check that introducing line breaks here does not mess with the layout of the generated documentation? I would expect not but good to make sure.

ci_scripts/flake8_diff.sh
mfeurer: The new script is even faster because it does not invoke any git commands.

examples/flows_and_runs_tutorial.py
mfeurer: Yes, they do not change the layout. In fact, some examples were already formatted correctly.

ci_scripts/flake8_diff.sh
PGijsbers: Makes sense 👍

Issue #304, opened 2017-09-01
Serilization of np integers and floats


janvanrijn:
Seems like the sklearn to flow serialization does not like it when a np.int64 is passed through:
```
  File "/home/vanrijn/projects/openml-python/openml/flows/sklearn_converter.py", line 81, in sklearn_to_flow
    raise TypeError(o, type(o))
TypeError: (2, <class 'numpy.int64'>)
```
Apparently, that case is not caught by
`elif isinstance(o, (bool, int, float, six.string_types)) or o is None:`

mfeurer:
Correct, but why can't you use normal integers?

PGijsbers:
I ran into this as well. In my case I wanted to specify all possible labels for the scikit-learn OrdinalEncoder, and the possible labels were retrieved from the dataset, which was stored as numpy floats. Extending for numpy types should be relatively simple, given we can retrieve them programmatically:
```
>>> np.sctypes
{'int': [<class 'numpy.int8'>, <class 'numpy.int16'>, <class 'numpy.int32'>, <class 'numpy.int64'>],
'uint': [<class 'numpy.uint8'>, <class 'numpy.uint16'>, <class 'numpy.uint32'>, <class 'numpy.uint64'>],
'float': [<class 'numpy.float16'>, <class 'numpy.float32'>, <class 'numpy.float64'>],
'complex': [<class 'numpy.complex64'>, <class 'numpy.complex128'>],
'others': [<class 'bool'>, <class 'object'>, <class 'bytes'>, <class 'str'>, <class 'numpy.void'>]}
```
e.g.
```
simple_numpy_types = [nptype for type_cat, nptypes in np.sctypes.items()
                             for nptype in nptypes
                             if type_cat != 'others']
simple_types = tuple([bool, int, float, str] + simple_numpy_types])
```
and then
`elif isinstance(o, simple_types) or o is None:`
I expected this to work, **but it does not**.
```
... traceback ...
  File "D:\repositories/openml-python\openml\flows\sklearn_converter.py", line 534, in _extract_information_from_model
    raise TypeError(msg)
TypeError: Second item of tuple does not match assumptions. Expected OpenMLFlow, got <class 'numpy.float32'>
```
I also tested just casting the numpy types to python native types but that did not work either. It looks like fixing this might need a deeper dive into the code. Nevertheless, I hope any of this helped whomever ends up working on this.

PGijsbers:
First off, a (artificial) minimal example that exposes the issue:
```
[...]

PR #635, opened 2019-03-03
Add support for serializing numpy data types.

PGijsbers:
Notes on the fix in #304.


PGijsbers:
@mfeurer I assumed you would get a ping automatically, but I am not sure if this is the case. I added the tests.

mfeurer:
I get a ping for a commit, but I usually do not check the commits because I don't know if that commit makes it ready for review again. Will check now.

Issue #636, opened 2019-03-03
`sklearn_to_flow` raises TypeError on simple flow


PGijsbers:
I found that the following simple example does not convert to an `OpenMLFlow` with `sklearn_to_flow` . I do not know *why* this happens (lists of lists?). The definition of a flow in code is also quite unclear and has a lots of gaps left with "TODO: add documentation".
```
import openml
from sklearn.preprocessing import OrdinalEncoder
openml.flows.sklearn_to_flow(OrdinalEncoder(categories=[[0,1],[0,1]]))
```
raises
```
... traceback ...
  File "D:\repositories/openml-python\openml\flows\sklearn_converter.py", line 534, in _extract_information_from_model
    raise TypeError(msg)
TypeError: Second item of tuple does not match assumptions. Expected OpenMLFlow, got <class 'numpy.float32'>
```

PGijsbers:
Fixed via #637.

PR #637, opened 2019-03-04
Fix636

PGijsbers:
Fixes an issue where a list of lists with simple types could not be serialized (because it was expected to be FeatureUnion, Pipeline or VotingClassifier), as described in #636.

Example flow generated: https://www.openml.org/f/9556
It deserializes correctly with
```
flow = openml.flows.get_flow(9556)
model = openml.flows.flow_to_sklearn(flow)
```

### DIFF
--- a/openml/flows/sklearn_converter.py
+++ b/openml/flows/sklearn_converter.py
@@ -502,13 +502,12 @@ def _extract_information_from_model(model):
         rval = sklearn_to_flow(v, model)
 
         def flatten_all(list_):
-            flattened = []
+            """ Flattens arbitrary depth lists of lists. """
             for el in list_:
                 if isinstance(el, (list, tuple)):
-                    flattened += flatten_all(el)
+                    yield from flatten_all(el)
                 else:
-                    flattened.append(el)
-            return flattened
+                    yield el
 
         if isinstance(rval, (list, tuple)):
             nested_list_of_simple_types = all([isinstance(el, (bool, str, int, float))
--- a/openml/flows/sklearn_converter.py
+++ b/openml/flows/sklearn_converter.py
@@ -501,11 +501,27 @@ def _extract_information_from_model(model):
     for k, v in sorted(model_parameters.items(), key=lambda t: t[0]):
         rval = sklearn_to_flow(v, model)
 
+        def flatten_all(list_):
+            flattened = []
+            for el in list_:
+                if isinstance(el, (list, tuple)):
+                    flattened += flatten_all(el)
+                else:
+                    flattened.append(el)
+            return flattened
+
+        if isinstance(rval, (list, tuple)):
+            nested_list_of_simple_types = all([isinstance(el, (bool, str, int, float))
+                                               for el in flatten_all(rval)])
+        else:
+            nested_list_of_simple_types = False
+
         if (isinstance(rval, (list, tuple))
             and len(rval) > 0
             and isinstance(rval[0], (list, tuple))
             and all([isinstance(rval[i], type(rval[0]))
-                     for i in range(len(rval))])):
+                     for i in range(len(rval))])
+                and not nested_list_of_simple_types):
 
             # Steps in a pipeline or feature union, or base classifiers in
             # voting classifier
@@ -588,7 +604,6 @@ def _extract_information_from_model(model):
             parameters[k] = json.dumps(component_reference)
 
         else:
-
             # a regular hyperparameter
             if not (hasattr(rval, '__len__') and len(rval) == 0):
                 rval = json.dumps(rval)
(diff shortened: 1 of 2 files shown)


PGijsbers:
The unit test uses `OrdinalEncoder` which is only available in `sklearn >= 0.20`, which I forgot about. I will switch in an older scikit-learn component.

PGijsbers:
I can't find a scikit-learn component in `0.19` or `0.18` which accepts a list of lists. Perhaps I overlooked one, can anyone think of any? If not, would it be fine to only test this against `scikit-learn>=0.20`?

mfeurer:
I'm not aware of one either. Unless @amueller or @janvanrijn now of one I think you can simply skip the tests for those sklearn versions.

janvanrijn:
Neural networks with layers structure param?

Op maandag 4 maart 2019 schreef Matthias Feurer <notifications@github.com>:

PGijsbers:
@janvanrijn thanks for the suggestion, unfortunately it looks like you just specify layer sizes through a tuple  of `n_layers` integers.

tests/test_flows/test_flow_functions.py
```diff
@@ -221,3 +225,13 @@ def test_are_flows_equal_ignore_if_older(self):
...
+
+    @unittest.skipIf(LooseVersion(sklearn.__version__) < "0.20",
+                     reason="OrdinalEncoder introduced in 0.20. "
+                            "No known models with list of lists parameters in older versions.")
+    def test_sklearn_to_flow_list_of_lists(self):
+        from sklearn.preprocessing import OrdinalEncoder
+        ordinal_encoder = OrdinalEncoder(categories=[[0, 1], [0, 1]])
+        flow = openml.flows.sklearn_to_flow(ordinal_encoder)
+        self._add_sentinel_to_flow_name(flow)
+        flow.publish()
```
mfeurer: Why do you need to publish this flow if the error happens in `sklearn_to_flow`?

openml/flows/sklearn_converter.py
```diff
@@ -501,11 +501,26 @@ def _extract_information_from_model(model):
...
+        def flatten_all(list_):
+            """ Flattens arbitrary depth lists of lists. """
+            for el in list_:
+                if isinstance(el, (list, tuple)):
+                    yield from flatten_all(el)
+                else:
+                    yield el
+
+        if isinstance(rval, (list, tuple)):
+            nested_list_of_simple_types = all([isinstance(el, (bool, str, int, float))
```
mfeurer: Could you please add a few comments here on what you are doing here? Even though I know the issue it tries to prevent I'm having trouble understanding this.

tests/test_flows/test_flow_functions.py
PGijsbers: I thought it might be useful to check the serialized flow is also valid for the server. You're right that it isn't strictly necessary. Should I remove it?

openml/flows/sklearn_converter.py
PGijsbers: I am not sure specifically if you mean line 513, or also the method above. Line 513 makes sure that each element within the list of lists is itself a 'simple type' that we can serialize (e.g. `bool`, `int`, ...). Rather than it needing special treatment according to the first case of the if/else statement directly below (which serializes e.g. steps in a pipeline). Though this did remind me that since #635 is merged, I should also allow for numpy types in 513. Update coming (will add comments too).

tests/test_flows/test_flow_functions.py
mfeurer: I think it is a good idea to ask the server for its opinion, but you should then also download the flow again and compare whether the deserialization works correctly.

openml/flows/sklearn_converter.py
```diff
@@ -506,14 +507,35 @@ def _extract_information_from_model(model):
...
             and len(rval) > 0
             and isinstance(rval[0], (list, tuple))
-            and all([isinstance(rval[i], type(rval[0]))
-                     for i in range(len(rval))])):
+            and all([isinstance(rval_i, type(rval[0])) for rval_i in rval])
+        )
 
-            # Steps in a pipeline or feature union, or base classifiers in
-            # voting classifier
+        nested_list_of_simple_types = (
```
mfeurer: Okay, so this iteratively checks if all elements in the list are simple types? If yes, could you maybe add a comment here that it does so?

PGijsbers: You are correct, I'll add the requested comment.

tests/test_flows/test_flow_functions.py
PGijsbers: Done.

Issue #149, opened 2016-08-30
OpenML doesn't return deactivated datasets any more by calling data/list


mfeurer:
This makes a test fail which checks whether dataset 17 is deactivated. Right now, it is not documented on OpenML how to retrieve deactivated datasets. Can only fix this issue once the documentation on OpenML is done.

joaquinvanschoren:
You mean this?
https://www.openml.org/api/v1/xml/data/list/status/deactivated

joaquinvanschoren:
Also documented:
https://www.openml.org/api_docs#!/data/get_data_list_status_status

I now notice how old this issue is. I just noticed because one of the unit tests is skipped because of it :)

mfeurer:
Yeah, we need to update the package at some point to reflect this change - I'm leaving this open for this reason.

PGijsbers:
Is it possible this issue is outdated? For deactivated dataset 478, I can currently fetch and discover it thus (all work):
```
d = openml.datasets.get_dataset(478)

ds = openml.datasets.list_datasets(status='all')
ds[478]  # No key error, description present

ds = openml.datasets.list_datasets(status='deactivated')
ds[478]  # No key error, description present
```
Deactivated datasets are by default *not* returned for `list_datasets`, but this is documented in the function's docstring.

PGijsbers:
Nevermind I noticed the unit test referenced in #382. Though I do think it is a separate issue (I think it is `check_datasets_active` that has the bug).

mfeurer:
Fixed via #642.

PR #642, opened 2019-03-17
Fix check_datasets_active and corresponding unit test

PGijsbers:
This solves an issue hinted at in #149. While retrieving deactivated datasets worked perfectly well, the `check_datasets_active` method was seemingly not updated for some latest changes, and its unit test was outdated too ([it was disabled](https://github.com/openml/openml-python/blob/master/tests/test_datasets/test_dataset_functions.py#L214)).

This PR:
 - Fixes a bug in `check_datasets_active` where deactivated datasets would never be found due to not specifying `status='all'` in the `list_datasets` call (presumably because the default behavior was changed at some point).
 - Correctly points the unit test to the production server, as there is no deactivated dataset on the test server.
 - Updated the dataset id that is supposed to refer to an active dataset from 1 to 2 in the unit test (since d/1 is now deactivated).
 - reactivates the fixed unit test
 - refactored the flow of `check_datasets_active` a little for legibility and to avoid "excessive" looping/evaluations (minor but still nice :) )

### DIFF
--- a/openml/datasets/functions.py
+++ b/openml/datasets/functions.py
@@ -278,7 +278,7 @@ def _load_features_from_file(features_file: str) -> Dict:
                                    force_list=('oml:feature', 'oml:nominal_value'))
         return xml_dict["oml:data_features"]
 
-      
+
 def check_datasets_active(dataset_ids: List[int]) -> Dict[int, bool]:
     """ Check if the dataset ids provided are active.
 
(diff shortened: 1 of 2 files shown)


PGijsbers:
Appveyor fails unrelated to this code change, see #643.

PGijsbers:
Issue with AppVeyor is fixed in #645.

Issue #643, opened 2019-03-17
Current dev branch fails on appveyor


PGijsbers:
See [here](https://ci.appveyor.com/project/PGijsbers/openml-python-gqsnv/builds/23134996#L120). Not sure why this happens, though there seems to be an update 2 days ago ([here](https://anaconda.org/anaconda/nbconvert/files?version=5.4.1&channel=main&type=conda)), and I imagine this is the cause.

mfeurer:
That's annoying. However, all tests in that test script are currently skipped. Therefore, we could remove the failing script. Would you like to do so?

PGijsbers:
> However, all tests in that test script are currently skipped. 

Which tests are you referring to? I was under the impression that before the AppVeyor tests worked as intended?

PGijsbers:
Turns out setuptools v23 is no longer required, and as long as we update conda, we can use the latest version. It is fixed in #644.

mfeurer:
Fixed via #644.

Issue #612, opened 2019-01-24
Can't download dataset metadata without downloading the dataset itself


joaquinvanschoren:
#### Description
There are situations where one only wants to see the meta-data of a given dataset. The current get_dataset always downloads and parses the dataset itself. This is slow(!), and can cause errors if the files are really large. 

It would be nice to have an option (or another call) to download the meta-data only.

To make matters worse, the arff parser requires many times the file size for the download and parsing. E.g. downloading a 2GB dataset requires more than 16GB disk space. This fails in colab notebooks: they crash because of disk space limits. I would download the csv version directly, but this requires the file url or file_id, neither of which are returned by list_dataset and thus have to be retrieved with get_dataset. So, there is currently no way to download larger OpenML dataset into colab notebooks without manually looking up the file_ids.

#### Steps/Code to Reproduce
Download any large dataset (e.g. SVHN): 
```
openml.datasets.get_dataset(41081)
```

#### Expected Results
An OpenMLDataset object without the actual dataset, or a dict with the meta-data.

#### Actual Results
get_dataset is very slow, or crashes if disk space is not sufficient.

#### Versions
Darwin-18.2.0-x86_64-i386-64bit
Python 3.6.4 |Anaconda custom (64-bit)| (default, Jan 16 2018, 12:04:33)
[GCC 4.2.1 Compatible Clang 4.0.1 (tags/RELEASE_401/final)]
NumPy 1.15.3
SciPy 1.1.0
Scikit-Learn 0.20.0
OpenML 0.7.0

janvanrijn:
Agreed. Similar for tasks, where the split file is commonly downloaded (potentially 10 times the size of the dataset)

mfeurer:
Agreed, being able to disable this would be a great addition. I think actually be a better solution than the one proposed in #346, what do you think @amueller ?

PGijsbers:
I agree that metadata should be accessible easily without downloading the whole dataset. I think adding a parameter to the `get_dataset` method makes sense, e.g.:
```
def get_dataset(dataset_id: int, download_data: bool=False) -> OpenMLDataset:
    """ Download the OpenML dataset representation, optionally also download actual data file. 

    This function is thread/multiprocessing safe.
    This function uses caching. A check will be performed to determine if the information has
    previously been downloaded, and if so be loaded from disk instead of retrieved from the server.

    Parameters
    ----------
    dataset_id : int
        Dataset ID of the dataset to download
    download_data : bool
[...]

PR #644, opened 2019-03-17
Fix612 lazy download dataset

PGijsbers:
Datasets can now be downloaded without downloading the `arff` file.
Function signature of `get_dataset` changed from
`def get_dataset(dataset_id: Union[int, str]) -> OpenMLDataset:`
to
`def get_dataset(dataset_id: Union[int, str], download_data: bool = True) -> OpenMLDataset:`
I chose to default to `True` so there are no breaking changes to existing code.
If `download_data=False`, only metadata will be downloaded (i.e. *all* data except the arff file).

Whenever a user invokes ~`retrieve_class_labels` or~ `get_data`, both of which require the arff-file, the arff-file is retrieved and processed as if it were downloaded from the server on initialization (e.g. pickled). This happens without warning or error/argument.
As long as only functionality is used which does not require the arff file, no additional data is downloaded.

I took the liberty to refactor `retrieve_class_labels` s.t. it uses the already downloaded feature metadata instead of reading the arff file. This makes it so that `retrieve_class_labels` can 50used without downloading the underlying data, and overall should really speed up the method in cases where a huge arff file was loaded just to check the header 👍 
As part of this I also addressed open issue #507 that was being worked on in #508 (I only realized afterwards). Though it looks like that code was from before @janvanrijn changed the xml so that the nominal values are a list. My solution is able to use this update and with only minor changes address the issue.

Fixes:
 #643 
 #612 
 #507 
 #446
 #346 in part

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -29,9 +29,8 @@ install:
   - rmdir C:\\cygwin /s /q
 
   # Update previous packages and install the build and runtime dependencies of the project.
-  # XXX: setuptools>23 is currently broken on Win+py3 with numpy
-  # (https://github.com/pypa/setuptools/issues/728)
-  - conda update --all --yes setuptools=23
+  - conda update conda --yes
+  - conda update --all --yes
 
   # Install the build and runtime dependencies of the project.
   - "cd C:\\projects\\openml-python"
--- a/examples/datasets_tutorial.py
+++ b/examples/datasets_tutorial.py
@@ -78,6 +78,15 @@ print(X.head())
 print(X.info())
 
 ############################################################################
+# Sometimes you only need access to a dataset's metadata.
+# In those cases, you can download the dataset without downloading the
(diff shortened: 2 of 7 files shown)


PGijsbers:
It looks like `continuous-integration/appveyor/pr` failed because the conda update stalled.
I can't figure out how to rerun it, without pushing a new commit to this PR. I tried to re-build the commit, but that only seems to trigger `continuous-integration/appveyor/branch` ([which worked but is now being rerun](https://ci.appveyor.com/project/PGijsbers/openml-python-gqsnv/builds/23154192)).

PGijsbers:
I can't seem to restart the `continuous-integration/appveyor/pr` build tests without submitting a new commit. If requested, I will add one, but I feel confident that it is not needed to have said build rerun and the code quality is fine (as all three other builds passed).

joaquinvanschoren:
Awesome! ::hooray::

Issue #410, opened 2018-03-01
Roadmap


mfeurer:
As we have a new student research assistant starting on the OpenML python connector, I briefly thought of a roadmap on how to proceed with the python interface:

* **0.6**: Bugfixes and first pypi release, ideally done before the workshop in the beginning of April
* **0.7**: Improved upload of datasets and tasks, ideally done after the workshop in the beginning of April
* **0.8**: Full support of pandas: handling of strings and correct handling of categorical attributes, returning DataFrames instead of dictionaries of dictionaries in listing calls.
* **0.9**: Supporting tasks other than classification (regression, clustering)
* **1.0**: Providing a clean interface to let other packages talk to the OpenML connector.

@janvanrijn @amueller @joaquinvanschoren @ArlindKadra what do you guys think?

janvanrijn:
What are the particular bugfixes you had in mind for the 0.6 version? Isn't 0.6 the current version btw?

mfeurer:
Good catch, so everything should be incremented (and instead of 1.0 and 1.1 we'd have 0.10 and 1.0). Fixes I had in mind (and @ArlindKadra will work on them):

* #392
* #378
* #375
* #369
* #358
* #354
* #324
* #185
* #149
* #43
* #232
* #397 
* #404

joaquinvanschoren:
It would be great to support regression as well: https://github.com/openml/openml-python/issues/312

Step 1 would be to upload a single regression run. The definition for the predictions is in the regression task type: https://www.openml.org/api_docs#!/tasktype/get_tasktype_id

Step 2 is to document how traces can be uploaded as well: https://github.com/openml/openml/issues/413

mfeurer:
@joaquinvanschoren I think regression and other tasks are a separate work package and worth getting done rather sooner than later, but in my opinion, right now it's more important to have the current functionality in a bug-free state again.

amueller:
dataset upload and pandas integration seem essential, generally this looks good.

mfeurer:
* **0.9**:
    * Handling of regression and clustering tasks. 
    * Removal of python2.
    * Upload/creation of tasks. 
    * Full support of pandas: handling of strings and correct handling of categorical attributes, returning DataFrames instead of dictionaries of dictionaries in listing calls.
    * Providing a clean interface to let other packages talk to the OpenML connector.
* **0.10**: Better handling of uploads when running flows and/or runs: #457, #219

PGijsbers:
[...]

PR #647, opened 2019-03-19
Extension interface

mfeurer:
This PR implements an extension interface to OpenML-Python which abstracts away the deep connection with scikit-learn and allows using OpenML-Python with other machine learning libraries as well.

The main interface can be found in `openml.extensions.extension_interface`. A possible use for scikit-learn is given in `openml.extensions.sklearn_extension`.

The whole PR is very raw and is only intended to provide feedback to me. Missing parts are:

* [x] unit tests
* [x] mention in the documentation
* [x] docstrings
* [x] full type hints
* [x] make the API implementation for sklearn its own submodule
* [ ] obtain full diff coverage

This supersedes #543 

@janvanrijn @joaquinvanschoren @amueller @ArlindKadra @PGijsbers @Bilgecelik


ci_scripts/flake8_diff.sh
```diff
@@ -1,3 +1,4 @@
 #!/bin/bash
 
 flake8 --ignore E402,W503 --show-source --max-line-length 100 $options
+mypy openml --ignore-missing-imports --follow-imports skip
```
PGijsbers: Didn't know about mypy :D sounds great!

openml/extensions/extension_interface.py
```diff
@@ -0,0 +1,91 @@
+from abc import ABCMeta, abstractmethod
+from typing import Any, List, Optional, Tuple
+
+from openml import OpenMLTask, OpenMLFlow
+from openml.runs.trace import OpenMLTraceIteration, OpenMLRunTrace
+
+
+class Extension:
+    __metaclass__ = ABCMeta
```
PGijsbers: [PEP3115](https://www.python.org/dev/peps/pep-3115/) switches the syntax for adding a meta-class with Py3 to `class Extension(metaclass=ABCMeta):`, see also [updated docs](https://docs.python.org/3/library/abc.html#abc.ABC).

PGijsbers: It even seems advised to subclass `abc.ABC` directly instead of specifying the metaclass from a legibility point [see notes](https://docs.python.org/3/whatsnew/3.4.html#abc). Eitherway works for me.

openml/extensions/extension_interface.py
```diff
@@ -0,0 +1,91 @@
...
+
+    ################################################################################################
+    # Abstract methods for hyperparameter optimization
+
+    @abstractmethod
+    def is_hpo_class(self, model: Any) -> bool:
+        pass
+
+    @abstractmethod
+    def assert_hpo_class(self, model: Any) -> None:
```
PGijsbers: Should this method be abstract? Leaving it non-abstract makes more sense to me. Perhaps create an else clause which raises NotImplementedError to make sure that any instantiated child has different behavior depending on whether it's a hpo class.

openml/extensions/extension_interface.py
```diff
@@ -0,0 +1,90 @@
+from abc import ABC, abstractmethod
+from typing import Any, List, Optional, Tuple
+
+from openml import OpenMLTask, OpenMLFlow
+from openml.runs.trace import OpenMLTraceIteration, OpenMLRunTrace
+
+
+class Extension(ABC):
```
PGijsbers: I had left some comments here before but they fell off when I edited the file.. Anyhow, this way to define an abstract base class is [preferred as of Python 3.4](https://docs.python.org/3/whatsnew/3.4.html#abc). See also [this](https://docs.python.org/3/library/abc.html#abc.ABC).

openml/flows/flow.py
```diff
@@ -165,8 +166,8 @@ def _to_dict(self) -> dict:
             Flow represented as OrderedDict.
 
         """
-        flow_container = OrderedDict()
-        flow_dict = OrderedDict([('@xmlns:oml', 'http://openml.org/openml')])
+        flow_container = OrderedDict()  # type: 'OrderedDict[str, OrderedDict]'
+        flow_dict = OrderedDict([('@xmlns:oml', 'http://openml.org/openml')])  # type: 'OrderedDict[str, Union[List, str]]'  # noqa E501
```
PGijsbers: I am a fan of type hints for functions. I am not sure if I am a fan of providing (in-line) type hints for variables. I think it should generally be preferred to remove most ambiguity through documentation and variable names.

openml/flows/functions.py
```diff
@@ -23,7 +24,7 @@ def _get_cached_flows() -> OrderedDict:
     flows : OrderedDict
         Dictionary with flows. Each flow is an instance of OpenMLFlow.
     """
-    flows = OrderedDict()
+    flows = OrderedDict()  # type: 'OrderedDict[int, OpenMLFlow]'
```
PGijsbers: For instance, here it is quite clear what the dictionary contains. It is noted directly above in the function summary. There is only one place where elements are added to the dictionary, where the type of the key is cast directly above it, and a function is called that has (or will have) type hints specified. I understand that this is also probably to guarantee `mypy` can do type checking, but I fear it might come at the cost of legibility.

openml/flows/flow.py
mfeurer: I only used inline type annotations where mypy couldn't guess the type itself.

openml/flows/functions.py
mfeurer: I agree that this is rather clear. But do you mean readability instead of legibility?

openml/flows/functions.py
```diff
@@ -94,11 +103,8 @@ def get_flow(flow_id: int, reinstantiate: bool = False) -> OpenMLFlow:
     ):
         flow = _get_flow_description(flow_id)
 
-    if reinstantiate:
-        if not (flow.external_version.startswith('sklearn==')
-                or ',sklearn==' in flow.external_version):
-            raise ValueError('Only sklearn flows can be reinstantiated')
-        flow.model = openml.flows.flow_to_sklearn(flow)
+    if reinstantiate and extension is not None:
```
PGijsbers: Does it make sense to drop the reinstantiate parameter? Providing the extension with `reinstantiate=False` does not use the extension at all, and providing no extension with `reinstantiate=True` does not reinstantiate the model (and without warning, no less).

openml/runs/run.py
```diff
@@ -155,13 +156,14 @@ def to_filesystem(self, directory: str, store_model: bool = True) -> None:
         run_xml = self._create_description_xml()
         predictions_arff = arff.dumps(self._generate_arff_dict())
 
-        with open(os.path.join(directory, 'description.xml'), 'w') as f:
-            f.write(run_xml)
-        with open(os.path.join(directory, 'predictions.arff'), 'w') as f:
-            f.write(predictions_arff)
+        # It seems like typing does not allow to define the same variable multiple times
+        with open(os.path.join(directory, 'description.xml'), 'w') as fh_1:  # type: TextIO
```
PGijsbers: Here too, I feel like the context (and one-time use) of the variable does not warrant explicit typing.

openml/extensions/sklearn_extension.py
```diff
@@ -0,0 +1,467 @@
...
+        return arff_line
+
+    def is_hpo_class(self, model):
+        return isinstance(model, sklearn.model_selection._search.BaseSearchCV)
+
+    def assert_hpo_class(self, model):
+        if not isinstance(model, sklearn.model_selection._search.BaseSearchCV):
+            raise ValueError(
+                'Flow model is not an instance of sklearn.model_selection._search.BaseSearchCV'
+            )
```
PGijsbers: Still not sure if `assert_hpo_class` should have been non-abstract. I now see that custom error message, which is nice, but we also see that you e.g. now also stopped re-using code from `is_hpo_class` (whereas the base class does this well). It might be that I don't entirely understand what's going on, though.

openml/extensions/sklearn_extension.py
```diff
@@ -0,0 +1,467 @@
...
+        base_estimator.set_params(**trace_iteration.get_parameters())
+        return base_estimator
+
+    def obtain_arff_trace(self, extension, model, trace_content):
+        if not isinstance(model, sklearn.model_selection._search.BaseSearchCV):
+            raise ValueError('model should be instance of'
+                             ' sklearn.model_selection._search.BaseSearchCV')
+        if not hasattr(model, 'cv_results_'):
+            raise ValueError('model should contain `cv_results_`')
+
```
PGijsbers: Code re-use!
```
self.assert_hpo_class()
self.asserthpo_class_has_trace()
```

openml/runs/functions.py
```diff
@@ -1,10 +1,10 @@
-import collections
+from collections import OrderedDict
 import io
 import json
 import os
 import sys
-import time
-from typing import List, Union, Tuple
+from typing import List, Optional, Tuple, Union  # noqa F401
+import typing
```
PGijsbers: Where do we make a distinction in importing the types or specifying them as `typing`? E.g. why do we use `typing.Any` but don't use `typing.List`?

openml/extensions/extension_interface.py
```diff
@@ -0,0 +1,90 @@
...
+    @abstractmethod
+    def is_estimator(self, model: Any) -> bool:
+        pass
+
+    @abstractmethod
+    def seed_model(self, model: Any, seed: Optional[int]) -> Any:
+        pass
+
+    @abstractmethod
+    def run_model_on_fold(
```
janvanrijn: I'm undecided whether this function should be in this interface or in the `openml-python` connector (making use of call_back functions `model_predict` and `model_fit`).

openml/extensions/extension_interface.py
janvanrijn: What does this method do? (I don't understand the assert part)

openml/extensions/extension_interface.py
```diff
@@ -0,0 +1,90 @@
...
+        add_local_measures: bool,
+        extension: 'Extension',
+    ) -> Tuple:
+        pass
+
+    ################################################################################################
+    # Abstract methods for hyperparameter optimization
+
+    @abstractmethod
+    def is_hpo_class(self, model: Any) -> bool:
```
janvanrijn: What does this method do? Check whether it can generate a trace?

openml/extensions/extension_interface.py
```diff
@@ -0,0 +1,90 @@
...
+    @abstractmethod
+    def instantiate_model_from_hpo_class(
+        self,
+        model: Any,
+        trace_iteration: OpenMLTraceIteration,
+    ) -> Any:
+        pass
+
+    @abstractmethod
+    def obtain_arff_trace(
```
janvanrijn: I wonder whether this method is really necessary. In fact, this method might as well be the same for all trace iterations, and make calls to `extract_trace_data` (`extract_trace_data` is of course necessary).

janvanrijn:
I really like where this is going. Additional bonus: introduction of type hints. I marked down my first comments. Feel free to ping me for a next review.

openml/extensions/extension_interface.py
PGijsbers: The assert part reveals the intention that rather than just indicating whether or not it is a hyperparameter optimization class, it should actually raise an error if it isn't. This can be helpful for a piece of code that uses this object *needs/expects* it to be hpo_class, it can first assert that it is and then continue. If it turns out it isn't a hpo_class, the error will be raised which should lead to a quick and clear fail (matches with a fail-early, fail-loudly philosophy which tends to help identify bugs).

openml/flows/flow.py
mfeurer: We discussed offline to keep them for now.

openml/flows/functions.py
mfeurer: We discussed offline to keep them for now.

openml/flows/functions.py
mfeurer: Good catch, I'll drop it.

openml/runs/run.py
mfeurer: We discussed offline to keep them for now.

openml/extensions/sklearn_extension.py
mfeurer: I will update the child class to use the `is_hpo_class` method, but will continue to provide a custom error messgae. Nevertheless, the method should not be static.

openml/runs/functions.py
mfeurer: I prefer `from typing import List` etc. to have shorter type annotations. This was a mistake of mine, I'll remove the `import typing`.

openml/extensions/extension_interface.py
mfeurer: I was undecided too until I figured out that there's more interaction going on here:

* the model needs to copied/cloned which is specific to each library
* there's some code specific to pipelines included to figure out which classes are actually used by the scikit-learn estimator
* we would need `model_predict()` and `model_predict_proba()`
* as soon as we start adding timing for parallelized grid search and random search, this will require library specific code as well.

openml/extensions/extension_interface.py
mfeurer: Correct.

openml/extensions/extension_interface.py
mfeurer: I'm not sure if I get your comment on how this could be the same for all trace iterations. This method is actually used by `openml.runs.functions` and I guess it is pretty helpful to assemble the final trace. The method `extract_trace_data` on the other hand is not necessary in the abstract extension. Therefore, I removed it.

openml/flows/functions.py
PGijsbers: > I agree that this is rather clear. But do you mean readability instead of legibility?

I wasn't aware there was a distinction, I had always thought they were synonymous. But yes, I meant readability 👍

mfeurer:
Hi together,

I achieved two milestones: the tests are passing again and the sklearn interface is now in its own submodule.

In order to achieve these I had to update the extension interface itself (adding methods to get the version information, a setup string, parameter values and whether the method will train in parallel) and pass the extension interface around at a few places which can for example be seen in the test file tests/test_study/test_study_examples.py.

Therefore, I propose to add a 'default' extension mechanism to OpenML-Python. Users can specify a default tool to work with (the package default will be scikit-learn), but also overwrite this when calling a function which interacts with a machine learning package. Furthermore, I would make the extension a run attribute. Currently, runs are created in three places: when running a model, when downloading a run and when reading a run from disk. In the first case, the user already passes the extension, in the second case, the user already has to pass the correct extension to desialize a flow, and in the third case, the user would have to remember which extension is necessary to read the run. Not storing the extension in the run would require to pass the extension class to more methods as can be seen in the aforementioned unit test.

In order to not need to know whether a method will be run in parallel, I also propose to move the check whether time measurement is possible into the extension specific code.

Please let me know what you think so I can continue with open milestones.

PGijsbers:
> Furthermore, I would make the extension a run attribute. 

Or do we a expect a one-to-many relation for flow to extensions? Otherwise I am not a fan (though we'll see if a look at the code changes that). More precisely, in the case of a local run, either the flow is stored locally as well (and should contain all extension information), or references a server flow which should have a specified extension.

openml/flows/functions.py
```diff
@@ -69,18 +72,20 @@ def _get_cached_flow(fid: int) -> OpenMLFlow:
                                    "cached" % fid)
 
 
-def get_flow(flow_id: int, reinstantiate: bool = False) -> OpenMLFlow:
+def get_flow(
+    flow_id: int,
+    extension: 'Extension' = None,
```
PGijsbers: the type hint shouldn't need to be a string literal because of the conditional import above, right?

openml/runs/run.py
```diff
@@ -128,7 +132,7 @@ def from_filesystem(cls, directory, expect_model=True):
 
         return run
 
-    def to_filesystem(self, directory: str, store_model: bool = True) -> None:
+    def to_filesystem(self, directory: str, extension, store_model: bool = True) -> None:
```
PGijsbers: typehint

openml/runs/run.py
```diff
@@ -242,7 +247,7 @@ def _generate_arff_dict(self):
 
         return arff_dict
 
-    def get_metric_fn(self, sklearn_fn, kwargs=None):
+    def get_metric_fn(self, sklearn_fn, extension, kwargs=None):
```
PGijsbers: typehint

openml/runs/run.py
```diff
@@ -369,7 +374,7 @@ def _attribute_list_to_dict(attribute_list):
                 scores.append(sklearn_fn(y_true, y_pred, **kwargs))
         return np.array(scores)
 
-    def publish(self):
+    def publish(self, extension=None):
```
PGijsbers: typehint

openml/setups/functions.py
```diff
@@ -1,18 +1,22 @@
...
 from openml.flows import flow_exists
 import openml.exceptions
 import openml.utils
 
+if TYPE_CHECKING:
+    from openml.extensions.extension_interface import Extension
+
 
-def setup_exists(flow):
+def setup_exists(flow, extension):
```
PGijsbers: typehints

openml/testing.py
```diff
@@ -24,15 +26,15 @@ class TestBase(unittest.TestCase):
     Hopefully soon allows using a test server, not the production server.
     """
 
-    def setUp(self):
+    def setUp(self, n_levels=1):
```
PGijsbers: typehints

PGijsbers:
I looked at the code, and agree there is an excessive passing around to `extensions`. However to better know how to deal with it, I'll wait on an answer to my last remark.

----

Good progress! I noticed the following though:
 - the function signature of `extension.get_version_information()` says to return a `str` but the sklearn implementation currently returns a list (of strings).
 - it feels weird that in `SklearnExtension` the flow (de-)serialization is split from methods to perform a run. At the same time we should kind of have a look on how to avoid these monolithic files proper. Will have to think about it a bit more, but I don't think it should hold this work back necessarily (if we change things it is probably just moving around some code).

Other than that, not sure what more to add right now :)

mfeurer:
* I expect that there will only be one deserialization method for each flow. Indeed, instead of adding the extension to the run it should be attached to the flow. How about the following plugin mechanism: an extension module can register as an extension module. When creating a flow without an extension module, the external version string is passed to all registered extension modules and they can then decide whether they're the right model for the job.
* I was also thinking about removing all actual code from the sklearn extension module and putting in into separate files. Maybe now is the best time to do.

PGijsbers:
> How about the following plugin mechanism: ...

That make sense. Perhaps additionally, you could add directory(/ies) in the configuration files that tells where to look for extension modules so that we can also register them dynamically?

mfeurer:
> Perhaps additionally, you could add directory(/ies) in the configuration files that tells where to look for extension modules so that we can also register them dynamically?

I I would provide a function to which the extensions can register themselves, probably in the config module.

PGijsbers:
So you would always have to re-register extensions (i.e. they don't get saved)? Otherwise I don't see the distinction.

mfeurer:
Ah, I think I got it. I'm not sure if we want to automate it that much. Instead, I would actually prefer to keep it easy and would be fine with importing the extension when necessary.

openml/flows/functions.py
mfeurer: If it is not a string literal it will not work because then `Extension` will not be defined.

PGijsbers:
The idea would be that you could install/register the extension once and not every time you run Python (also enables e.g. packages to auto-register themselves on install). However, this is also something that does require any major changes if we want to add it later, so I see no reason to hold up any work over this either way :)

mfeurer:
> I was also thinking about removing all actual code from the sklearn extension module and putting in into separate files. Maybe now is the best time to do.

Done!

mfeurer:
Okay, @janvanrijn @PGijsbers @Bilgecelik @joaquinvanschoren I think the interface is now in good and usable shape. I added a plugin mechanism which makes the new extension mechanism mostly transparent for OpenML users. The most interesting changes/files at this moment are:

* tests/test_study/test_study_examples.py: to see that the overall interface for running models on tasks does not change
* openml/extensions/__init__.py: to see how simple the 'plugin' mechanism is
* openml/extensions/extension_interface.py: the actual extension interface
* openml/extensions/sklearn/extension.py: implementation of the extension interface for sklearn. The last line of the file registers the extension as plugin
* openml/flows/flow.py: see how the flow class retrieves the extension at the end of the constructor

openml/extensions/extension_interface.py
janvanrijn: I think I agree. Still I would consider making this an easier function. For example, looking at its signature, it currently has a wide range of arguments. I would almost guess that we don't need `task`, `rep_no`, `fold_no` and `sample_no` since we can provide `X` and `y`, and we should probably be a bit more clear on the `tuple` that is returned (I would propose: a numpy array with predictions corresponding to the `y` variable and a dict containing the local measures).

openml/extensions/sklearn/extension.py
```diff
@@ -0,0 +1,134 @@
...
+    def model_to_flow(self, model: Any) -> 'OpenMLFlow':
+        return sklearn_to_flow(model)
+
+    def flow_to_parameters(self, flow: Any) -> List:
+        return obtain_parameter_values(flow)
+
+    def get_version_information(self) -> List[str]:
+        return get_version_information()
+
+    def create_setup_string(self, model: Any) -> str:
```
janvanrijn: I don't understand what this function does. `setup_string` as in the legacy OpenML setup string? This can potentially be removed.

openml/extensions/sklearn/extension.py
```diff
@@ -0,0 +1,134 @@
+from typing import Any, Dict, List, Optional, Tuple, TYPE_CHECKING
+
+import sklearn.base
+
+from openml.extensions import Extension, register_extension
+from openml.extensions.sklearn.flow_functions import (
```
janvanrijn: why put these functions in different packages? why not implement them in this file?

janvanrijn:
Good stuff!

This is a big PR. We should probably aim to merge it asap as there will be many merge conflicts otherwise. Are there any things that you would like to do on it, or are you ready for a full review?

I was only able to do a quick stroll through it to orient myself a bit, but I can block 2 hours for detailed reviewing on it. Although it generally all looks good, I think it's good to have a second pair of eyes on it.

openml/extensions/extension_interface.py
mfeurer: Great idea regarding `X` and `y`, I'll take it into account. I'll be more specific regarding the tuple once I start working on the documentation.

openml/extensions/sklearn/extension.py
mfeurer: `setup_string` as it can be uploaded for a run. Is this legacy? If it is, I think removing it from the Python connector should be part of a different pull request.

openml/extensions/sklearn/extension.py
mfeurer: To keep the file rather small. Also, they're actually taking care of different things, namely serializing/deserializing flows and running flows.

mfeurer:
> We should probably aim to merge it asap

Yes!

> Are there any things that you would like to do on it, or are you ready for a full review

I'll add documentation next and then it'll be ready for a full review.

openml/extensions/extension_interface.py
mfeurer: I just tried to apply this simplification and stumbled upon a few issues which suggest that this should be implemented separately in a future pull request:

1. The task object is actually required so that the run function can decide on the call args for the fit function.
2. Simplifying the return value probably requires adding functionality to the `Task` class on how to transform predictions in valid content for uploading. I would like to avoid adding this in this PR as this will require larger refactoring regarding the way the prediction files are created.

Would that be okay for you for now?

openml/extensions/sklearn/extension.py
```diff
@@ -0,0 +1,318 @@
...
+    def run_model_on_fold(
+        self,
+        model: Any,
+        task: 'OpenMLTask',
+        rep_no: int,
+        fold_no: int,
+        sample_no: int,
+        add_local_measures: bool,
+    ) -> Tuple[List[List], List[List], 'OrderedDict[str, float]', Any]:
+        """Run a model on a repeat,fold,subsample triplet of the task and return prediction
```
PGijsbers: I understand this function signature is legacy. That said, seeing this again got me thinking - should we consider replacing parameters `task, rep_no, fold_no, sample_no` with `X_train, X_test, y_train=None`? Which would allow anyone writing an extension minimal knowledge of the intricacies of fetching the right data. It would centralize [this code](https://github.com/openml/openml-python/blob/extensions/openml/extensions/sklearn/run_functions.py#L182), so people don't need to keep re-implementing it. I might very easily be missing something, but at first glance this seems cleaner?

PGijsbers: Oh, I noticed the task is also needed for creating the arff content. I guess there's not much advantage then (still might be nicer if it is `task, X_train, X_test, y_train=None`... but now I'm not so sure anymore :)

PGijsbers:
Looking good :+1: I agree that we should make merging this a priority.

openml/extensions/sklearn/extension.py
mfeurer: Please check my answer to [Jan's comment above](https://github.com/openml/openml-python/pull/647#discussion_r269159706). I suggested to do this in a new PR because it will require major refactoring again.

mfeurer:
Okay, I checked the coverage, and the coverage diff is mostly in the dataset submodule and thereby nothing which should be addressed by this pull request. @PGijsbers @joaquinvanschoren @janvanrijn @Bilgecelik this pull request is ready for review.

openml/extensions/extension_interface.py
PGijsbers: Thanks for pointing me back to this thread @mfeurer, I had missed the follow-up discussion. I am in favor of continuing this PR as is, and then opening an issue so that we remind ourselves to refactor this afterwards 👍

openml/extensions/__init__.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+def get_extension_by_flow(
+    flow: 'OpenMLFlow',
+    raise_if_no_extension: bool = False,
+) -> Optional[Extension]:
+    """Get an extension which can handle the given flow.
+
+    Iterates all registered extensions and checks whether they can handle the presented flow.
+    In case an extension can, it is immediately returned.
```
PGijsbers: I would word it slightly differently, from
`In case an extension can, it is immediately returned.`
to
`The first extension which can handle to flow is returned (even if multiple extensions can handle the flow).`
The part in parenthesis can be included or not based on what you feel is best.

Edit: I would probably leave out the word "all" from the first sentence too, since it does not necessarily iterate all extensions (as indicated by the second sentence).

openml/extensions/__init__.py
```diff
@@ -0,0 +1,92 @@
...
+
+    Returns
+    -------
+    Extension or None
+    """
+    for extension_class in extensions:
+        if extension_class.can_handle_flow(flow):
+            return extension_class()
+    if raise_if_no_extension:
+        raise ValueError('No extension registered which can handle flow: %s' % flow)
```
PGijsbers: Let's start using `.format` string formatting as per [roadmap](https://github.com/openml/openml-python/issues/410#issuecomment-466527824)?

openml/extensions/__init__.py
```diff
@@ -0,0 +1,92 @@
...
+
+    Returns
+    -------
+    Extension or None
+    """
+    for extension_class in extensions:
+        if extension_class.can_handle_model(model):
+            return extension_class()
+    if raise_if_no_extension:
+        raise ValueError('No extension registered which can handle model: %s' % model)
```
PGijsbers: `.format`

openml/extensions/sklearn/extension.py
PGijsbers: Do you think it might make sense to use double inheritance instead? E.g. basically make all functions in `run_functions.py` part of a `SklearnRunExtension`-class and all functions in `flow_functions` part of a `SklearnFlowExtension`-class and then simply use inheritance instead (both of these would inherit from `Extension`). Functions which are implemented in neither can be implemented here (currently only `can_handle_model`). So the rewritten extension would look like:

```
class SklearnExtension(SklearnRunExtension, SklearnFlowExtension):

    @staticmethod
    def can_handle_model(model: Any) -> bool:
        """Check whether a model is an instance of ``sklearn.base.BaseEstimator``.

        Parameters
        ----------
        model : Any

        Returns
        -------
        bool
        """
        return isinstance(model, sklearn.base.BaseEstimator)
```
?

I don't have much experience with multiple inheritance but I think it would work (I tried a toy example and that worked). It would help by:
 - Removing a lot of double documentation, since now you document the function twice, e.g. [here](https://github.com/openml/openml-python/blob/extensions/openml/extensions/sklearn/extension.py#L228) and [here](https://github.com/openml/openml-python/blob/extensions/openml/extensions/sklearn/flow_functions.py#L262). This can lead to inconsistencies (indeed, even the linked example has inconsistencies, though only a minor ones regarding spacing.
 - making it much clearer which functions are actually implemented in this file.
 - maintenance when e.g. function signatures change. Currently we should change the `Extension` interface, the method implementation in e.g. `flow_functions.py` and the redirect in `extension.py`. Using multiple inheritance saves us one step (since the method is not redeclared in `extension.py`)

openml/runs/run.py
```diff
@@ -150,26 +158,27 @@ def to_filesystem(self, directory: str, store_model: bool = True) -> None:
 
         os.makedirs(directory, exist_ok=True)
         if not os.listdir(directory) == []:
-            raise ValueError('Output directory should be empty')
+            raise ValueError('Output directory %s should be empty' % os.path.abspath(directory))
```
PGijsbers: `.format`

openml/extensions/extension_interface.py
```diff
@@ -0,0 +1,291 @@
...
+
+        Parameters
+        ----------
+        model : Any
+
+        Returns
+        -------
+        bool
+        """
+        pass
```
janvanrijn: rather than a pass statement, should we throw a `NotImplementedError`? That is way more explicit and gives better feedback to users when they accidentally use this abstract extension.

openml/extensions/sklearn/extension.py
janvanrijn: Agreed.

openml/extensions/sklearn/extension.py
janvanrijn: To add to this. 

I see how it might be sub-optimal to handle this in the current PR, as this increases the size of the diff significantly, but this is something that we should have resolved before making a new release.

openml/extensions/sklearn/flow_functions.py
```diff
@@ -474,15 +491,16 @@ def _get_external_version_string(model, sub_components):
     for visitee in sub_components.values():
         for external_version in visitee.external_version.split(','):
             external_versions.add(external_version)
-    external_versions = list(sorted(external_versions))
-    external_version = ','.join(external_versions)
-    return external_version
+    return ','.join(list(sorted(external_versions)))
```
janvanrijn: For future PR: this string is not json parsable.(let's open an issue?)

openml/extensions/sklearn/flow_functions.py
```diff
@@ -875,17 +895,59 @@ def _serialize_cross_validator(o):
...
+                          module_name[1])
+    for parameter in parameters:
+        parameters[parameter] = flow_to_sklearn(
+            parameters[parameter], recursion_depth=recursion_depth + 1
+        )
+    return model_class(**parameters)
+
+
+def _format_external_version(model_package_name: str, model_package_version_number: str) -> str:
+    return '%s==%s' % (model_package_name, model_package_version_number)
```
janvanrijn: future pr: not json parsable. (issue)

openml/extensions/sklearn/flow_functions.py
```diff
@@ -875,17 +895,59 @@ def _serialize_cross_validator(o):
...
+    scipy_version = 'SciPy_{}.'.format(scipy.__version__)
+
+    return [python_version, sklearn_version, numpy_version, scipy_version]
+
+
+def is_hpo_class(model: Any) -> bool:
+    return isinstance(model, sklearn.model_selection._search.BaseSearchCV)
+
+
+def check_n_jobs(model: Any) -> bool:
```
janvanrijn: seems like you lost the docs:
```

    Returns True if the parameter settings of model are chosen s.t. the model
    will run on a single core (if so, openml-python can measure runtimes)
```

openml/extensions/sklearn/run_functions.py
```diff
@@ -0,0 +1,453 @@
+from collections import OrderedDict
```
janvanrijn: since this file is completely new anyway, I would go ahead and copy the functions in the `SklearnExtension` class already. This saves us time later for the refactoring  (in particular for reviewing, as the diff will be smaller).

openml/extensions/sklearn/run_functions.py
```diff
@@ -0,0 +1,453 @@
...
+import json
+import time
```
janvanrijn: Also, as according to git all functions here are `new` I did not have a thorough look at them as I assumed they will still behave the same. 

Did any of the functions changed / do you want me to look at any of them in depth? Otherwise I think we should rely on the unit tests.

openml/runs/run.py
```diff
@@ -435,7 +449,7 @@ def _create_description_xml(self):
         # tags = run_environment + [well_formatted_time] + ['run_task'] + \
         #     [self.model.__module__ + "." + self.model.__class__.__name__]
         description = _to_dict(taskid=self.task_id, flow_id=self.flow_id,
-                               setup_string=_create_setup_string(self.model),
+                               setup_string=self.setup_string,
```
janvanrijn: before you used the function to create a setup string on the fly, however now you rely on the fact that there is already a setup string and that it's up to date. Are you sure this is intended? Since the function is rather fast, I would make a function call nonetheless (to keep the function easily readible)

openml/testing.py
```diff
@@ -119,5 +120,58 @@ def _check_dataset(self, dataset):
         self.assertIn(dataset['status'], ['in_preparation', 'active',
                                           'deactivated'])
 
+    def _check_fold_evaluations(
```
janvanrijn: rename to `_check_fold_timing_evaluations`?

openml/testing.py
```diff
@@ -24,15 +27,15 @@ class TestBase(unittest.TestCase):
     Hopefully soon allows using a test server, not the production server.
     """
 
-    def setUp(self):
+    def setUp(self, n_levels: int = 1):
```
janvanrijn: can you please add a tiny comment string explaining `n_levels`? (seems important for understanding unit tests)

tests/test_extensions/test_sklearn_extension/test_sklearn_flow_functions.py
```diff
@@ -57,17 +54,17 @@ def fit(self, X, y):
         pass
 
 
-class TestSklearn(TestBase):
+class TestFunctions(TestBase):
```
janvanrijn: name -> `TestSklearnExtensionFlowFunctions`?

tests/test_extensions/test_sklearn_extension/test_sklearn_run_functions.py
```diff
@@ -0,0 +1,228 @@
...
+    seed_model,
+    run_model_on_fold,
+    obtain_arff_trace,
+)
+
+from openml.testing import TestBase
+from openml.runs.trace import OpenMLRunTrace
+
+
+class TestRun(TestBase):
```
janvanrijn: name -> `TestSklearnExtensionRunFunctions`?

openml/extensions/extension_interface.py
```diff
@@ -0,0 +1,291 @@
+from abc import ABC, abstractmethod
+from collections import OrderedDict  # noqa: F401
+from typing import Any, Dict, List, Optional, Tuple, TYPE_CHECKING
+
+
+if TYPE_CHECKING:
```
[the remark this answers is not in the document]
PGijsbers: I'll let Matthias correct me if I'm wrong, but short answer: this helps prevent cyclic imports.

Long answer: `TYPE_CHECKING` is a flag that is used to determine if a module is being loaded as part of a (static) type checking run. In order to do good type checking (e.g. proper use/availability of methods), normally you annotate the types directly (`can_handle_flow(flow : OpenMLFlow) -> bool`). However, to do this, you would have to (in this case) import `OpenMLFlow`, which itself would import (indirectly) again this file, leading to cyclic imports. Now you can still indicate the intended type by just specifying a string literal (`'OpenMLFlow'`), and while this helps for readability, the type checker itself does not know exactly what type it is, an import is still needed. So, using the `TYPE_CHECKING` flag the import can be done in type checking time but not at runtime (I am not sure about how type checkers avoid the cyclic imports exactly, but that's the worry for them (or anyone setting the `typing.TYPE_CHECKING` flag)). So, to answer the second question, `TYPE_CHECKING` should always be `False` unless a package (such as `mypy`) specifically sets it otherwise, in which case they are responsible for how things are handled (the file wouldn't normally load with `TYPE_CHECKING=True` due to the cyclic imports).

openml/extensions/extension_interface.py
mfeurer: > I assume we have the same idea about this, right?

Yes, I agree that we can put the merging of the traces into the general part of the library. I will add this to the proposed changes list for the updates to the extension interface.

openml/extensions/sklearn/extension.py
mfeurer: I can see your points, especially regarding double documentation, and clarity about where the functions are.

I'll move all code into the class. This shouldn't make this PR bigger as it does not make a difference to the total PR size where the code is copied to.

openml/extensions/__init__.py
mfeurer: I totally agree and will add this.

openml/extensions/__init__.py
mfeurer: Sorry, I'll change this one and the other ones.

openml/extensions/__init__.py
```diff
@@ -0,0 +1,92 @@
...
+from .extension_interface import Extension
+
+if TYPE_CHECKING:
+    from openml.flows import OpenMLFlow
+
+
+extensions = []
+
+
+def register_extension(extension: Type[Extension]) -> None:
```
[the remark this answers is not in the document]
mfeurer: This appeared to be an appropriate place to put these functions. I'll move them to a different file.

openml/extensions/extension_interface.py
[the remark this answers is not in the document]
mfeurer: Thanks @PGijsbers for adding the explanation, I believe it is correct.

I'll add a one-line comment to the if statement to make it easier to look this up.

openml/extensions/extension_interface.py
```diff
@@ -0,0 +1,291 @@
...
+
+    ################################################################################################
+    # General setup
+
+    @staticmethod
+    @abstractmethod
+    def can_handle_flow(flow: 'OpenMLFlow') -> bool:
+        """Check whether a given flow can be handled by this extension.
+
+        This is typically done by parsing the ``external_version`` field.
```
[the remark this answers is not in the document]
mfeurer: > I wonder if there are any possible clashes possible, as anyone can put anything in external version.

An adversary can definitively make the program crash

> Can we ensure that for example when downloading a weka flow (which sometimes has a md5 hash in external version) this does not crash?

Is the hash in hex form? If yes, it won't contain the `s` of `sklearn`.

> I agree that this is as much an OpenML problem

Most likely, I'm not sure what else to do except for checking what the server provides.

openml/extensions/__init__.py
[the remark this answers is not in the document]
mfeurer: This appeared to be an appropriate place to put these functions. I'll move them to a different file.

openml/extensions/extension_interface.py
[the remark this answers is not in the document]
mfeurer: Please see my answer above.

openml/extensions/extension_interface.py
```diff
@@ -0,0 +1,291 @@
...
+        task: 'OpenMLTask',
+        rep_no: int,
+        fold_no: int,
+        sample_no: int,
+        add_local_measures: bool,
+    ) -> Tuple[List[List], List[List], 'OrderedDict[str, float]', Any]:
+        """Run a model on a repeat,fold,subsample triplet of the task and return prediction information.
+
+        Returns the data that is necessary to construct the OpenML Run object. Is used by
+        run_task_get_arff_content. Do not use this function unless you know what you are doing.
```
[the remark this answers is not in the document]
mfeurer: Definitively! I'll change that.

openml/extensions/extension_interface.py
mfeurer: It's actually possible to remove the `pass`, too. Making sure that the abstract method is not used is done by the decorator which will raise an exception if you try to create a subclass which does not implement the method.

openml/extensions/sklearn/flow_functions.py
mfeurer: I chose this formatting to be in line with python's version formatting, but I agree that it becomes hard to maintain and impractical.

openml/extensions/sklearn/run_functions.py
mfeurer: No, I think I didn't touch the functions themselves, only the signatures (adding type hints).

openml/runs/run.py
mfeurer: Retrieving the setup string depends on the model. I moved this to the creation of the run object (after actually running the model) where we can be sure that the model is available. Therefore, I think it's correct.

openml/testing.py
mfeurer: Yes!

openml/testing.py
mfeurer: Yes, sure.

tests/test_extensions/test_sklearn_extension/test_sklearn_run_functions.py
mfeurer: Yes.

openml/extensions/sklearn/extension.py
PGijsbers: I don't currently foresee the problems that Jan mentioned could be an issue, but that is probably  due a lack of working with that particular code. Either way I trust the judgement, and the adjustments Jan proposed seem fine with me otherwise too.

openml/extensions/sklearn/flow_functions.py
mfeurer: Done, it's #658.

mfeurer:
Okay, I have addressed all issues raised so for, please go ahead for a (hopefully) final review. Regarding failing CI:

* appveyor/branch: this has been fixed on the development branch which is why appveyor/pr does not fail
* travis-ci/pr: only fails for the PEP8 check, which does not fail for travis/push. I suggest fixing this with the next PR to the development branch
* travis-ci/push: fails because of https://github.com/openml/OpenML/issues/948, which has been fixed in the development branch already (and which is why they don't fail for travis-ci/pr).

openml/extensions/sklearn/extension.py
PGijsbers: The idea was to still have the distinction between run/flow operations through two classes (and for me, preferably in two files). Now everything is back in one monolithic file/class.
I think Jan's version of the proposal was that there are two classes:
`class SklearnFlowExtension(Extension):` and `class SklearnExtension(SklearnFlowExtension)`. Here, the former implements the functions previously in `flow_functions` and the latter implements the functions previously in `run_functions` and `sklearn\extension.py`. Having the structure clear both in classes and files makes it easier to navigate the code base, has less chance for merge conflicts, and more explicitly defines the assumed dependencies.

Is there are reason you went with putting it all in one class/file, or is it a misunderstanding (possibly on my part)?

PGijsbers:
I would accept but I don't feel [this issue](https://github.com/openml/openml-python/pull/647#discussion_r271679675) is not resolved yet. Would like some clarification.

openml/extensions/sklearn/extension.py
mfeurer: I think the original proposal by @janvanrijn was to implement all functions related to the sklearn extension in one file. Your proposal with double inheritance would solve the issue with duplicated docstrings, but in my opinion would not help to make it clearer where the functionality is implemented compared to having two files with functions. Since having everything in a single file appears to be conceptually simpler and also easier maintain for the near future I went with this solution.

PGijsbers: Hmm. I don't really agree but at least this still solves the double documentation issues. If you prefer this and @janvanrijn has no objections then no further changes are needed in my opinion.

openml/extensions/extension_interface.py
janvanrijn: Where can I find this list?

openml/extensions/extension_interface.py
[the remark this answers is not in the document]
janvanrijn: what I am kind of getting at is that we have 2 sorts of dependencies:
* connector dependencies (`openml-python`, potentially later `openml-extension-multiflow` or something similar)
* external dependencies (sklearn, maybe numpy, scipy)
In many cases we don't need the exact same version to be able to:
* load a model
* ensure exact same predictions
The highest dependency level here is ofc scikit-learn (as it usually figures out numpy, scipy, etc) but maybe the openml-python connector can also change behaviour over version (e.g., when fixing a bug).

Maybe we should proceed and merge, and add this to the list of things to figure out before we release.

openml/extensions/extension_interface.py
mfeurer: In an unsaved tab in gedit on my laptop ;)

Here it is:
* Remove `obtain_arff_trace` and return the trace headers in each fold and later on check that they're all the same.
* Simplify `run_model_on_fold` to accept X and y and do not split the data itself
* Simplify the return value of `run_model_on_fold` to have functionality for transforming the predictions in the actual task classes.
* Figure out how to best handle dependencies (as raised [here](https://github.com/openml/openml-python/pull/647#discussion_r272686627))

mfeurer:
Here's the list of TODOs:

* Remove obtain_arff_trace and return the trace headers in each fold and later on check that they're all the same.
* Simplify run_model_on_fold to accept X and y and do not split the data itself
* Simplify the return value of run_model_on_fold to have functionality for transforming the predictions in the actual task classes.
* Figure out how to best handle dependencies (as raised here)

mfeurer:
Here's the list of TODOs:

* Remove obtain_arff_trace and return the trace headers in each fold and later on check that they're all the same.
* Simplify run_model_on_fold to accept X and y and do not split the data itself
* Simplify the return value of run_model_on_fold to have functionality for transforming the predictions in the actual task classes.
* Figure out how to best handle dependencies (as raised here)

Issue #646, opened 2019-03-18
Datasets cached pre-#548 can no longer be used for `run_model_on_task`


PGijsbers:
It looks like pre-#548 pickled data was in `np.array` format, where now the default is `pandas.dataframe`. When I now try to use `run_model_on_task` for which I still have a cached dataset with `np.array` as data instead of `pd.DataFrame`, [this line](https://github.com/openml/openml-python/blob/develop/openml/datasets/dataset.py#L520) is called with `(data, dataset_format=="array", [some list of attribute names)`. This raises an error because as far as I can tell `_convert_array_format` assumes that the input data is `pd.DataFrame` if specified dataformat is `"array"`, which makes [this line](https://github.com/openml/openml-python/blob/develop/openml/datasets/dataset.py#L370) raise an error because `np.array` does not have an attribute columns.

The fix seems easy enough, just check if `data` is already of the preferred type, e.g. start the function with
```
def _convert_array_format(data, array_format, attribute_names):
    if array_format == "array" and not scipy.sparse.issparse(data):
        if isinstance(data, np.ndarray):
            return data
        ...
```

Does this make sense? Shall I set up a PR?
@glemaitre @mfeurer

mfeurer:
Thanks for raising this issue. I would suggest that we break backwards compatibility here and simply delete and re-download the data in case the pickle file cannot be read. What do you think?

PGijsbers:
Not sure, on the one hand I am currently performing some runs with this hack and it seems pretty stable (so far on 90+ different tasks already done). It's not that the pickle file can't be read. It can be and seems to be done correctly. It is just after loading the data, `run_model_on_task` wants to have it in `np.ndarray` format and in the conversion assumes it is not yet in `np.ndarray` format which causes the problem.
So I don't think there is anything inherently wrong/breaking with not invalidating peoples cache, which makes the 'hack' ok.

 On the other hand, we probably don't want to worry about this scenario anymore later. And writing code that locally changes the stored data (I think it should be possible without redownloading? I didn't look too closely) probably takes more time than just forcing a cache refresh.

mfeurer:
The reason why I proposed this rather aggressive strategy is that new features (such as string support in datasets) will only be available if the data is stored as a pandas array. 

I also wouldn't like to reformat the data locally and therefore suggest to either download it again or re-read the arff file to create the pickle of a pandas dataframe.

Would you like to work on this @PGijsbers

PGijsbers:
[...]

PR #651, opened 2019-03-22
Fixes a bug that prevents openml from finding the config file

joaquinvanschoren:
This is a pretty serious issue that breaks the current release.

There was a comma missing in config.py which prevented openml from finding the config file. Uploads failed with a 102 'Authentication required' error.

Uploading anything to openml, with the API key only available in the config file, should not return an error anymore.

### DIFF
--- a/openml/config.py
+++ b/openml/config.py
@@ -24,7 +24,7 @@ _defaults = {
     'connection_n_retries': 2,
 }
 
-config_file = os.path.expanduser(os.path.join('~', '.openml' 'config'))
+config_file = os.path.expanduser(os.path.join('~', '.openml', 'config'))
 
 # Default values are actually added here in the _setup() function which is
 # called at the end of this module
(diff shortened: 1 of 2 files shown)


joaquinvanschoren:
And kudos to Jan for actually finding this bug.

janvanrijn:
Thanks for this fix. I am a bit surprised that this was able to get through the unit tests. Apparently, there are no unit tests that check whether the config file is obtained from the right place. @mfeurer any idea how we can test this to avoid regression?

mfeurer:
Thanks for this fix. In order to test it one could, if on travis, add some content to the known location of the config file, and then see if the read configuration functionality is able to read it.

joaquinvanschoren:
Thanks. Do you need we need a hotfix for this in the master branch?

mfeurer:
I plan to merge soon anyway. The only thing that's missing at the moment is a fix for #646

PR #654, opened 2019-03-30
Fix backwards compatibility #646.

PGijsbers:
See #646. Reprocess ARFF file if outdated datatype was used in pickle.

No unit tests are currently added. To add unit tests we would also need to add more data files (of outdated data) to the repository. That just doesn't seem worth it. Changes were tested locally by updating my local cache :) I will add unit tests if requested. The code changed in this patch is so minimal that I feel confident enough if the current unit tests don't break.

### DIFF
--- a/openml/datasets/dataset.py
+++ b/openml/datasets/dataset.py
@@ -174,97 +174,106 @@ class OpenMLDataset(object):
     def _data_arff_to_pickle(self, data_file):
         data_pickle_file = data_file.replace('.arff', '.pkl.py3')
         if os.path.exists(data_pickle_file):
-            logger.debug("Data pickle file already exists.")
-            return data_pickle_file
-        else:
-            try:
-                data = self._get_arff(self.format)
-            except OSError as e:
-                logger.critical("Please check that the data file %s is "
-                                "there and can be read.", data_file)
-                raise e
+            with open(data_pickle_file, "rb") as fh:
+                data, categorical, attribute_names = pickle.load(fh)
 
-            ARFF_DTYPES_TO_PD_DTYPE = {
-                'INTEGER': 'integer',
-                'REAL': 'floating',
-                'NUMERIC': 'floating',
-                'STRING': 'string'
-            }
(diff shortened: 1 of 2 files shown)


PGijsbers:
I get a consistent error, but this also happens on the current develop head (tested locally):
```
   def test_get_runs_list_by_tag(self):
        # TODO: comes from live, no such lists on test
        runs = openml.runs.list_runs(tag='curves')
>       self.assertGreaterEqual(len(runs), 1)
E       AssertionError: 0 not greater than or equal to 1
```
These seem to be server issues. is my best bet. If I try to filter runs with the tag through the website then over 6k results are returned quickly, but through the API I got the following response:
image
image
@janvanrijn any clue?

janvanrijn:
Seems like joining to the run tag column is extremely slow (I guess due to the large size). 

This could be solved by adding an index on the tag name, at the expense of additional disk cost. We currently have 23M run tag records, mostly indicating stuff that can be obtained through other ways as well. I think we should review the utility of (run) tags, and when a library automatically adds a tag to a run.

PGijsbers:
1. But the website loads them extremely quickly, e.g. [runs with 'curves'](https://www.openml.org/search?type=run&from=100&q=tags.tag%3Acurves)?
2. It seemed to have worked before? `git blame` tells me this code was last edited 2017-03-02 14:34:54.

janvanrijn:
website runs on ES, which is an index build upon the database. 

I am surprised that the query returns no results. It should either crash or return 6000+ results.

PGijsbers:
It does seem odd. Either way, I think we're best served by skipping this unit test for now as it is not an issue with the Python code. So I'll go ahead and do that (also @mfeurer so he knows what's up too).

janvanrijn:
@PGijsbers feel free to open an issue about this in the main tracker, with tags 'CoreSystem' and 'highest priority' assigned to me. I can't solve it this week (ECML deadline) but it will be high on my priority list

Issue #330, opened 2017-10-09
make tests run quicker


amueller:
right now there's some really slow tests. We should have nose list the top 10 slowest tests using the nose timing module.

PGijsbers:
(Leaving a note so I don't forget) - PR #644 introduces lazy loading for datasets, and tests should likely be updated to use lazy loading whenever possible.

mfeurer:
Printing the 20 slowest tests was added in #548. What would be the next step?

PGijsbers:
The best thing to do is probably systematically considering where we want to have integrations tests (i.e. tests which actually perform calls on the test server), or verify the query is correct instead of actually querying the server. We also have to look at which calls slow down tests and whether or not we should solve this in `openml-python` or on the server. Hopefully we can move away from using the server as a default, and only use it just enough to ensure our communication layer is still sound. We could [mark](https://docs.pytest.org/en/stable/mark.html) those specific tests, so it easy to avoid running them during development (and then just run them before a PR). 

The other part just involves looking at individual tests ([current 20 slowest tests](https://travis-ci.org/github/openml/openml-python/jobs/736919606#L1521)), see if we can speed them up either by removing redundant server calls, or by refactoring some of the code to make it more easily testable.

For instance, looking at the slowest test we have [`test_get_flow_id`](https://github.com/openml/openml-python/blob/0def226c736395451688472173e1fb6050a145cf/tests/test_flows/test_flow_functions.py#L336) clocking in 167sec (for reference included at the bottom of this post) to test [`get_flow_id`](https://github.com/openml/openml-python/blob/0def226c736395451688472173e1fb6050a145cf/openml/flows/functions.py#L267).

I see a couple things here that I would consider changing or discussing.
We see that the unit test calls `openml.flows.get_flow_id(model=clf, exact_version=True)` as well as `openml.flows.get_flow_id(name=flow.name, exact_version=True)`, same for `exact_version=False`. The resulting call to `flow_exists` or `list_flows` is then the same, and those calls are well over 95% of the total time taken for the test.

I would consider:
 - testing whether or not the model gets converted to the correct `name/external_version` separately (extract the functionality as method first).
 - testing if we find a correct flow from the dataframe separately (extract the functionality as method first).
 - mocking the response values ([see pytest docs](https://docs.pytest.org/en/stable/monkeypatch.html?highlight=patch), have to take a look at threat safety)
[...]

PR #655, opened 2019-03-30
Prefer lazy loading in unit tests

PGijsbers:
Prefer lazy loading for all unit tests that don't explicitly need the arff file.
Incremental progress towards #330.

I don't think it matters much for most unit tests (many call `get_data` after downloading the data regardless), but it helps for some and otherwise hopefully encourages future additions to think about lazily loading data :)

If there are any other suggestions that should be included in the PR I'd like to hear them.


PGijsbers:
Same fails as #654, see that PR.

Issue #346, opened 2017-10-11
be lazy in downloading datasets and splits


amueller:
I think the get_dataset and get_task functions should not download any big items. That allows you to inspect the attributes of the data without having to download possibly large amounts of data.

We can just download the data in the ``get_data`` method - similarly, in ``get_task`` we shouldn't download the splits, but only download them later when they are needed.

mfeurer:
Functionality for lazy downloading of datasets added in #644, `get_task` still needs a download flag.

mfeurer:
Functionality for lazy downloading of tasks added in #659.

PR #659, opened 2019-04-04
Lazy download of data splits 

Neeratyoy:
Addresses the pending part from #346 . Wherein, the splits and the data are not downloaded when `get_task` is called. Only when a run is executed, the data and the respective splits are downloaded.
A unit test has been added for the lazy fetch of a task.

Calling `get_task` or `get_tasks` with the additional parameter of `download_data=False` downloads only the task.xml file. Which can be verified by monitoring the cache directory. When a task is run, the corresponding datasplits.arff file is downloaded and available in the cache directory.

#### Any other comments?
- Earlier: `task.class_labels` assignment was happening in a nested manner after `task.download_split`
- Now: `task.class_labels`is being assigned even if splits are not being downloaded, by fetching it from dataset's description

### DIFF
--- a/openml/tasks/functions.py
+++ b/openml/tasks/functions.py
@@ -307,7 +307,11 @@ def get_task(task_id, download_data=True):
     -------
     task
     """
-    task_id = int(task_id)
+    try:
+        task_id = int(task_id)
+    except (ValueError, TypeError):
+        raise ValueError("Dataset ID is neither an Integer nor can be "
+                         "cast to an Integer.")
 
     with lockutils.external_lock(
             name='task.functions.get_task:%d' % task_id,
@@ -323,8 +327,6 @@ def get_task(task_id, download_data=True):
             # List of class labels availaible in dataset description
             # Including class labels as part of task meta data handles
             #   the case where data download was initially disabled
-            # The other alternative would be to abstract away this check
-            #   as part of download_split()
             if isinstance(task, OpenMLClassificationTask):
                 task.class_labels = \
                     dataset.retrieve_class_labels(task.target_name)
--- a/openml/tasks/functions.py
+++ b/openml/tasks/functions.py
@@ -275,29 +275,37 @@ def __list_tasks(api_call):
     return tasks
 
 
-def get_tasks(task_ids):
-    """Download tasks.
+def get_tasks(task_ids, download_data=True):
+    """Download task representation and data splits optionally.
     This function iterates :meth:`openml.tasks.get_task`.
     Parameters
     ----------
     task_ids : iterable
-        Integers representing task ids.
+        Integers/Strings representing task ids.
+    download_data : bool
+        Option to trigger download of data along with the meta data.
     Returns
     -------
     list
     """
     tasks = []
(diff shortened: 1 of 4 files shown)


tests/test_tasks/test_task_functions.py
```diff
@@ -129,6 +129,24 @@ def test_get_task(self):
...
+        task = openml.tasks.get_task(2, download_data=False)
+        self.assertIsInstance(task, OpenMLTask)
+        self.assertTrue(os.path.exists(os.path.join(
+            self.workdir, 'org', 'openml', 'test', "tasks", "2", "task.xml",
+        )))
+
+        self.assertFalse(os.path.exists(os.path.join(
+            self.workdir, 'org', 'openml', 'test', "tasks", "2", "datasplits.arff"
+        )))
+        # Not checking datasets/ as other test cases might interfere
```
mfeurer: The cache directories are unique for each unit test so this should not be an issue: https://github.com/openml/openml-python/blob/master/openml/testing.py#L59

tests/test_tasks/test_task_functions.py
```diff
@@ -129,6 +129,24 @@ def test_get_task(self):
             self.workdir, 'org', 'openml', 'test', "datasets", "1", "dataset.arff"
         )))
 
+    def test_get_task_lazy(self):
+        task = openml.tasks.get_task(2, download_data=False)
+        self.assertIsInstance(task, OpenMLTask)
```
mfeurer: Could you please add a check that the class labels are retrieved despite the data not downloaded?

openml/tasks/functions.py
```diff
@@ -275,31 +275,43 @@ def __list_tasks(api_call):
...
     for task_id in task_ids:
-        tasks.append(get_task(task_id))
+        tasks.append(get_task(task_id, download_data))
     return tasks
 
 
-def get_task(task_id):
-    """Download the OpenML task for a given task ID.
+def get_task(task_id, download_data=True):
+    """Download the OpenML task representation for a given task ID, optionally
```
mfeurer: Could you please try to keep the first line of the docstring to be a single line and add information about optional arguments afterwards?

openml/tasks/functions.py
```diff
@@ -275,31 +275,45 @@ def __list_tasks(api_call):
...
 
 
-def get_task(task_id):
-    """Download the OpenML task for a given task ID.
+def get_task(task_id, download_data=True):
+    """Download task representation for a given task ID
     Parameters
     ----------
-    task_id : int
+    The task representation is downloaded while the download of data
```
mfeurer: Sorry, that's not how I meant it. Could you please check for example the doc strings of the [scikit-learn random forest](https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/ensemble/forest.py#L750)? There's a single line, then some additional information (where I think the sentence about the optional download of the split and data should go) and then the `Parameters`.

Neeratyoy: sorry about that, should have looked up an example, will make the changes

PR #663, opened 2019-04-07
Issue #621 - better error messages on listing queries.  Adding check …

### LINKED ISSUE #621: Better error messages on listing queries
#### Description
When you want to list items using the listing calls, you have to provide lists (e.g. a list of tasks). If you don't know this you get a cryptic error message 'TypeError: 'int' object is not iterable'

#### Steps/Code to Reproduce

```python
import openml as oml
runs = oml.runs.list_runs(task=2)
```

#### Expected Results
Either detect that it is an integer instead of a list and handle internally, or provide a clearer error message that a list must be provided.

#### Actual Results
```python

/usr/local/lib/python3.6/dist-packages/openml/runs/functions.py in _list_runs(id, task, setup, flow, uploader, display_errors, **kwargs)
   1002         api_call += "/run/%s" % ','.join([str(int(i)) for i in id])
   1003     if task is not None:
-> 1004         api_call += "/task/%s" % ','.join([str(int(i)) for i in task])
   1005     if setup is not None:
   1006         api_call += "/setup/%s" % ','.join([str(int(i)) for i in setup])

TypeError: 'int' object is not iterable
```

#### Versions
openml 0.8.0 (fresh pip install)

timandrews335:
…for list type and return error message.

#### Reference Issue

Fixes #621 

In the list_runs function, if a parameter is passed in that should by type list but is not, an exception is raised

import openml as oml
runs = oml.runs.list_runs(task=2)

the above should result in a more user-friendly error message

### DIFF
--- a/openml/runs/functions.py
+++ b/openml/runs/functions.py
@@ -1094,6 +1094,19 @@ def list_runs(offset=None, size=None, id=None, task=None, setup=None,
         List of found runs.
     """
 
+    if id is not None and (not type(id) is list):
+        raise Exception('id must be of type list.')
+        return None
+    if task is not None and (not type(task) is list):
+        raise Exception('task must be of type list.')
+        return None
+    if setup is not None and (not type(setup) is list):
+        raise Exception('task must be of type list.')
+        return None
+    if uploader is not None and (not type(uploader) is list):
+        raise Exception('task must be of type list.')
+        return None
+
     return openml.utils._list_all(
         _list_runs, offset=offset, size=size, id=id, task=task, setup=setup,
         flow=flow, uploader=uploader, tag=tag, display_errors=display_errors,
--- a/openml/runs/functions.py
+++ b/openml/runs/functions.py
@@ -1094,18 +1094,14 @@ def list_runs(offset=None, size=None, id=None, task=None, setup=None,
         List of found runs.
     """
 
-    if id is not None and (not type(id) is list):
-        raise Exception('id must be of type list.')
-        return None
-    if task is not None and (not type(task) is list):
-        raise Exception('task must be of type list.')
-        return None
-    if setup is not None and (not type(setup) is list):
-        raise Exception('task must be of type list.')
-        return None
-    if uploader is not None and (not type(uploader) is list):
-        raise Exception('task must be of type list.')
-        return None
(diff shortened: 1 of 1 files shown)


openml/runs/functions.py
```diff
@@ -1094,6 +1094,19 @@ def list_runs(offset=None, size=None, id=None, task=None, setup=None,
         List of found runs.
     """
 
+    if id is not None and (not type(id) is list):
+        raise Exception('id must be of type list.')
+        return None
```
PGijsbers: Returning None after raising an exception is not necessary. 
As soon as the exception is raised, control is returned to the calling function.

openml/runs/functions.py
```diff
@@ -1094,6 +1094,19 @@ def list_runs(offset=None, size=None, id=None, task=None, setup=None,
         List of found runs.
     """
 
+    if id is not None and (not type(id) is list):
+        raise Exception('id must be of type list.')
```
PGijsbers: It is good practice to raise a  [`TypeError`](https://docs.python.org/3.6/library/exceptions.html#TypeError) when the passed argument is not of the correct type. This makes it easier for the caller to check against certain types of errors, and the nature of the error is more immediately clear.

PGijsbers:
Thank you for making a contribution to this package! 🎉  It is much appreciated :) 
I left some remarks on how the code could be improved further before a merge.

@mfeurer I would argue that the existing parameter names for this method are rather poor and should be changed too. When expecting lists, I would expect plurals to be used in parameternames, e.g. `tasks` instead of `task`, `ids` instead of `id`.

openml/runs/functions.py
```diff
@@ -1094,6 +1094,19 @@ def list_runs(offset=None, size=None, id=None, task=None, setup=None,
         List of found runs.
     """
 
+    if id is not None and (not type(id) is list):
```
PGijsbers: When checking for types in Python, the best practice is to use the `isinstance(object, type)` method, e.g. `isinstance(id, list)`, this is because [isinstance also allows for the object to be a subclass](https://premaseem.wordpress.com/2017/10/11/python-difference-between-type-and-isinstance/).

openml/runs/functions.py
timandrews335: Made that change

PR #672, opened 2019-04-15
Add wallclock measurements

janvanrijn:

### DIFF
--- a/openml/extensions/sklearn/extension.py
+++ b/openml/extensions/sklearn/extension.py
@@ -923,6 +923,8 @@ class SklearnExtension(Extension):
                 result.extend(SklearnExtension._get_parameter_values_recursive(sub_grid,
                                                                                parameter_name))
             return result
+        else:
+            raise ValueError('Param_grid should either be a dict or list of dicts')
 
     def _prevent_optimize_n_jobs(self, model):
         """
--- a/openml/extensions/sklearn/extension.py
+++ b/openml/extensions/sklearn/extension.py
@@ -918,9 +918,11 @@ class SklearnExtension(Extension):
                     result.append(value)
             return result
         elif isinstance(param_grid, list):
-            result = []
-            result.extend(SklearnExtension._get_parameter_values_recursive(
-                sub_grid, parameter_name) for sub_grid in param_grid)
+            result = list()
+            for sub_grid in param_grid:
+                result.extend(SklearnExtension._get_parameter_values_recursive(sub_grid,
+                                                                               parameter_name))
+            return result
 
     def _prevent_optimize_n_jobs(self, model):
         """
@@ -947,9 +949,9 @@ class SklearnExtension(Extension):
                 print('Warning! Using subclass BaseSearchCV other than '
                       '{GridSearchCV, RandomizedSearchCV}. '
                       'Should implement param check. ')
-
-            if len(SklearnExtension._get_parameter_values_recursive(param_distributions,
-                                                                    'n_jobs')) > 0:
+            n_jobs_vals = SklearnExtension._get_parameter_values_recursive(param_distributions,
+                                                                           'n_jobs')
+            if len(n_jobs_vals) > 0:
                 raise PyOpenMLError('openml-python should not be used to '
                                     'optimize the n_jobs parameter.')
 
--- a/openml/extensions/sklearn/extension.py
+++ b/openml/extensions/sklearn/extension.py
@@ -1013,10 +1013,9 @@ class SklearnExtension(Extension):
         # of n_jobs (the negate will make this fn return false). For that
         # reason, we need to add clause 2 that returns True if n_jobs does not
         # exist in the flow
-        return not SklearnExtension._check_parameter_value_recursive(
-            model.get_params(), 'n_jobs', [-1]) or \
-               SklearnExtension._check_parameter_value_recursive(
-                   model.get_params(), 'n_jobs', None)
+        clause1 = not SklearnExtension._check_parameter_value_recursive(model.get_params(), 'n_jobs', [-1])
+        clause2 = SklearnExtension._check_parameter_value_recursive(model.get_params(), 'n_jobs', None)
+        return clause1 or clause2
 
     ################################################################################################
     # Methods for performing runs with extension modules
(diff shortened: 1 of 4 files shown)


openml/extensions/sklearn/extension.py
```diff
@@ -934,12 +955,55 @@ def check(param_grid, restricted_parameter_name, legal_values):
                       '{GridSearchCV, RandomizedSearchCV}. '
                       'Should implement param check. ')
 
-            if not check(param_distributions, 'n_jobs', None):
+            if not SklearnExtension._check_parameter_value_recursive(param_distributions, 'n_jobs', None):
                 raise PyOpenMLError('openml-python should not be used to '
                                     'optimize the n_jobs parameter.')
 
+    def _can_measure_cputime(self, model: Any) -> bool:
+        """
```
PGijsbers: Maybe we should document that `n_jobs` may not be optimized, and also that time measurements can only be performed with `n_jobs != -1`? The latter might incentivize users to explicitly define their `n_jobs` for multicore runs.

openml/extensions/sklearn/extension.py
```diff
@@ -1133,17 +1200,29 @@ def _prediction_to_probabilities(
...
                 model_copy.fit(train_x)
 
-            if can_measure_runtime:
-                modelfit_duration = (time.process_time() - modelfit_starttime) * 1000
-                user_defined_measures['usercpu_time_millis_training'] = modelfit_duration
+            if can_measure_cputime:
+                modelfit_duration_cputime = (time.process_time() - modelfit_start_cputime) * 1000
+                user_defined_measures['usercpu_time_millis_training'] = modelfit_duration_cputime
+            if can_measure_wallclocktime:
+                modelfit_duration_walltime = (time.time() - modelfit_start_walltime) * 1000
```
PGijsbers: [`time.time`](https://docs.python.org/3/library/time.html#time.time) is in seconds, I think the `* 1000` is a mistake.

openml/extensions/sklearn/extension.py
janvanrijn: > Maybe we should document that n_jobs may not be optimized

Agreed. 

> and also that time measurements can only be performed with n_jobs != -1? 

Agreed that it should be documented. Will do so. For the review, please note that
* cpu time can be measured with `n_jobs = 1` only
* wall time can be measured with `n_jobs` >= 0

openml/extensions/sklearn/extension.py
janvanrijn: we record millis :)

openml/extensions/sklearn/extension.py
```diff
@@ -1133,17 +1200,29 @@ def _prediction_to_probabilities(
...
+            modelfit_duration_cputime = None
+            modelpredict_start_cputime = None
+
+            modelfit_start_walltime = None
+            modelfit_duration_walltime = None
+            modelpredict_start_walltime = None
+            if can_measure_cputime:
+                modelfit_start_cputime = time.process_time()
+            if can_measure_wallclocktime:
+                modelfit_start_walltime = time.time()
```
PGijsbers: Why not define these as
```
modelfit_start_cputime = time.process_time() if can_measure_cputime else None
```
I also don't think you have to define `modelfit_duration_X` here, as they are only used within their guarded if-statements.

openml/extensions/sklearn/extension.py
```diff
@@ -888,35 +888,56 @@ def _format_external_version(
...
+            return True
+        elif isinstance(param_grid, list):
+            return all(
+                SklearnExtension._check_parameter_value_recursive(sub_grid, parameter_name, legal_values)
+                for sub_grid in param_grid
+            )
+
+    def _prevent_optimize_n_jobs(self, model):
+        """
+        Ensures that HPO classess will not optimize the n_jobs hyperparameter
```
PGijsbers: typo: classes

openml/extensions/sklearn/extension.py
```diff
@@ -1112,8 +1176,11 @@ def _prediction_to_probabilities(
         # but not desirable if we want to upload to OpenML).
 
         model_copy = sklearn.base.clone(model, safe=True)
+        # security check
```
PGijsbers: 'Security Check' is too vague. Something explicit like 'prohibit users from optimizing `n_jobs`' seems better to me.

PGijsbers: But if you think that's too explicit I still think I would prefer 'Sanity check' over 'Security Check'.

openml/extensions/sklearn/extension.py
janvanrijn: Because PyCharm does not like that. obviously it does not error but having no warnings ensures a good overview

openml/extensions/sklearn/extension.py
```diff
@@ -934,12 +958,61 @@ def check(param_grid, restricted_parameter_name, legal_values):
...
+        if not (
+                isinstance(model, sklearn.base.BaseEstimator) or self.is_hpo_class(model)
+        ):
+            raise ValueError('model should be BaseEstimator or BaseSearchCV')
+
+        n_jobs_not_specified = \
+            SklearnExtension._check_parameter_value_recursive(model.get_params(), 'n_jobs', None)
+        n_jobs_is_minus_one = \
+            SklearnExtension._check_parameter_value_recursive(model.get_params(), 'n_jobs', [-1])
+        return n_jobs_not_specified or not n_jobs_is_minus_one
```
PGijsbers: There was a flake8 fail here (>100 characters), so I took the liberty of refactoring the code s.t. it complies and the variable names say enough (I think) so we don't need a 5-line block comment. I hope that's fine.

PGijsbers: I was also wondering, if I understand correctly, if different components have different `n_jobs` values (e.g. `2, 3, 1, ...`) then we currently would measure wallclock time, right?

janvanrijn: Good catch. The code is actually flawed. I will right

janvanrijn: For clarity, the condition fails when there are various occurrences of n_jobs, both -1 and positive.

PR #673, opened 2019-04-15
Improve extension interface

mfeurer:
Continuation of #647:

* [x] Remove obtain_arff_trace and return the trace headers in each fold and later on check that they're all the same.
* [x] Simplify run_model_on_fold to accept X and y and do not split the data itself
* [x] Simplify the return value of run_model_on_fold
* ~~[ ] Have functionality for transforming predictions into arff lines in the different task classes.~~ (As discussed with JvR offline, we'll keep task-specific code here until the maintenance becomes an issue as for now the code would be more complex if such transformations would be in the task classes).
* ~~[ ] Figure out how to best handle dependencies (as raised here)~~ (this is beyond the scope of this PR)
* [x] add more unit tests


openml/extensions/sklearn/extension.py
```diff
@@ -1029,11 +1031,15 @@ def _run_model_on_fold(
         self,
         model: Any,
         task: 'OpenMLTask',
+        X_train: Union[np.ndarray, scipy.sparse.spmatrix, pd.DataFrame],
+        y_train: np.ndarray,
         rep_no: int,
```
janvanrijn: can be removed, together with `fold_no`, `sample_no`

openml/extensions/sklearn/extension.py
```diff
@@ -1029,11 +1031,15 @@ def _run_model_on_fold(
...
         fold_no: int,
         sample_no: int,
         add_local_measures: bool,
```
janvanrijn: can be removed as it can be handled a level up

openml/extensions/extension_interface.py
```diff
@@ -1,12 +1,15 @@
...
+
+import numpy as np
+import scipy.sparse
 
 # Avoid import cycles: https://mypy.readthedocs.io/en/latest/common_issues.html#import-cycles
 if TYPE_CHECKING:
     from openml.flows import OpenMLFlow
     from openml.tasks.task import OpenMLTask
-    from openml.runs.trace import OpenMLRunTrace, OpenMLTraceIteration
+    from openml.runs.trace import OpenMLRunTrace, OpenMLTraceIteration  # noqa F401
```
PGijsbers: Why is OpenMLRunTrace even imported in the first place? It does not seem to be used in the file?

openml/extensions/sklearn/extension.py
```diff
@@ -1119,17 +1123,12 @@ def _run_model_on_fold(
...
             The repeat of the experiment (0-based; in case of 1 time CV, always 0)
         fold_no : int
             The fold nr of the experiment (0-based; in case of holdout, always 0)
-        sample_no : int
-            In case of learning curves, the index of the subsample (0-based; in case of no
-            learning curve, always 0)
-        add_local_measures : bool
-            Determines whether to calculate a set of measures (i.e., predictive accuracy)
-            locally,
-            to later verify server behaviour.
```
PGijsbers: Looks like `y_train`, `X_test` and `classes` are missing from docstring?

openml/extensions/sklearn/extension.py
```diff
@@ -1268,154 +1249,50 @@ def _prediction_to_probabilities(
...
+                proba_y = model_copy.predict_proba(X_test)
             except AttributeError:
-                proba_y = _prediction_to_probabilities(pred_y, list(model_classes))
+                proba_y = _prediction_to_probabilities(pred_y, list(classes))
+
+            if proba_y.shape[1] != len(classes):
+                # Remap the probabilities in case there was a class missing at training time
+                # By default, the classification targets are mapped to be zero-based indices to the
+                # actual classes. Therefore, the model_classes contain the correct indices to the
+                # correct probability array (the actualy array might be incorrect if there are some
```
PGijsbers: typo: actualy

mfeurer: Just corrected this.

openml/extensions/sklearn/extension.py
mfeurer: Thanks, I forgot to copy them from the parent class.

openml/extensions/extension_interface.py
mfeurer: Fixed, it's now used.

openml/extensions/sklearn/extension.py
```diff
@@ -1268,154 +1249,50 @@ def _prediction_to_probabilities(
...
+                # classes not present during train time).
+                proba_y_new = np.zeros((proba_y.shape[0], len(classes)))
+                for idx, model_class in enumerate(model_classes):
```
PGijsbers: I don't understand what you're trying to do here.
If targets are mapped to be zero-based indices, wouldn't `enumerate(model_classes)` just evaluate to `(1,1), (2,2), ..., (Kt, Kt)` where `Kt` is the number of classes in the training data?
If so why not do either
```
proba_y_new = np.zeros((proba_y.shape[0], len(classes)))
proba_y_new[:proba_y.shape[0], :proba_y.shape[1]] = proba_y
```
or use `np.hstack` to pad zero-columns?

openml/extensions/sklearn/extension.py
```diff
@@ -1268,154 +1249,50 @@ def _prediction_to_probabilities(
...
+                    proba_y_new[:, model_class] = proba_y[:, idx]
+                proba_y = proba_y_new
 
             if proba_y.shape[1] != len(task.class_labels):
-                warnings.warn(
-                    "Repeat %d Fold %d: estimator only predicted for %d/%d classes!"
-                    % (rep_no, fold_no, proba_y.shape[1], len(task.class_labels))
+                message = "Estimator only predicted for {}/{} classes!".format(
```
PGijsbers: in what scenario should be expect `len(classes)` to differ from `len(task.class_labels)`? (Also, what even is the difference between `classes` and `task.class_labels`?)

openml/extensions/sklearn/extension.py
```diff
@@ -1268,154 +1249,50 @@ def _prediction_to_probabilities(
...
-        # consistency check, only useful for supervised tasks
-        def _calculate_local_measure(sklearn_fn, openml_name):
-            user_defined_measures[openml_name] = sklearn_fn(test_y, pred_y)
-
-        # Task type specific outputs
-        arff_datacontent = []
-
         if isinstance(task, (OpenMLClassificationTask, OpenMLLearningCurveTask)):
 
+            if classes is None:
```
PGijsbers: I would personally prefer argument checking to be done at the start of the method call such that all constraints are clear, and `TypeError` lead to a quick fail (not e.g. after running a whole grid search!). Even though I know it introduces an extra line of code :)

openml/runs/functions.py
```diff
@@ -391,26 +395,103 @@ def _run_task_get_arffcontent(
     # TODO use different iterator to only provide a single iterator (less
     # methods, less maintenance, less confusion)
     num_reps, num_folds, num_samples = task.get_split_dimensions()
+    classes = None
 
+    n_fit = 0
     for rep_no in range(num_reps):
```
PGijsbers: maybe replace this with `itertools.product(num_reps, num_folds, num_samples)` so we get rid of two levels of indentation and can maybe spare some ugly line-breaks below.

openml/runs/functions.py
```diff
@@ -391,26 +395,103 @@ def _run_task_get_arffcontent(
...
         for fold_no in range(num_folds):
             for sample_no in range(num_samples):
+                n_fit += 1
```
PGijsbers: I don't see a reason not to use `enumerate` (it looks like you want to start at 1, which is done with `enumerate(Sequence, start=1)`

openml/runs/functions.py
```diff
@@ -391,26 +395,103 @@ def _run_task_get_arffcontent(
...
+
+                elif isinstance(task, OpenMLClusteringTask):
+                    for i in range(0, len(test_indices)):
+                        arff_line = [test_indices[i], pred_y[i]]  # row_id, cluster ID
+                        arff_datacontent.append(arff_line)
+
+                else:
+                    raise TypeError(type(task))
+
                 arff_datacontent.extend(arff_datacontent_fold)
```
PGijsbers: What is the use `arff_datacontent_fold`? It is defined in line 446 but I don't see it being used anywhere else. Is this an error? Or is it legacy? If it is legacy, I would prefer `arff_datacontent.extend([])  # legacy` or something equivalent.

openml/runs/functions.py
```diff
@@ -432,10 +513,13 @@ def _run_task_get_arffcontent(
...
                         sample_no] = user_defined_measures_fold[measure]
 
-    # Note that we need to use a fitted model (i.e., model_fold, and not model)
-    # here, to ensure it contains the hyperparameter data (in cv_results_)
-    if extension.is_hpo_class(model):
-        trace = extension.obtain_arff_trace(model_fold, arff_tracecontent)  # type: Optional[OpenMLRunTrace]  # noqa E501
+    if len(traces) > 0:
+        if len(traces) != n_fit:
+            raise ValueError(
+                'Did not find enough traces (expected %d, found %d)' % (n_fit, len(traces))
```
PGijsbers: `.format`

openml/runs/trace.py
```diff
@@ -346,9 +348,23 @@ def trace_from_xml(cls, xml):
...
+        for trace in traces:
+            for iteration in trace:
+                merged_trace[(iteration.repeat, iteration.fold, iteration.iteration)] = iteration
+
+        return cls(None, merged_trace)
+
     def __str__(self):
         return '[Run id: %d, %d trace iterations]' % (
-            self.run_id,
+            -1 if self.run_id is None else self.run_id,
```
PGijsbers: `.format`

openml/runs/trace.py
```diff
@@ -448,3 +464,14 @@ def __str__(self):
             self.evaluation,
             self.selected,
         )
+
+    def __eq__(self, other):
+        if not isinstance(other, OpenMLTraceIteration):
+            return False
+        attributes = [
+            'repeat', 'fold', 'iteration', 'setup_string', 'evaluation', 'selected', 'paramaters',
```
PGijsbers: I noticed the typo `paramaters` here, and also noticed this is present all throughout the init method. Could you fix that up there too?
Also, it looks like this code probably isn't tested then? Since `getattr` should raise an exception as this object does not have an attribute `paramater` as (somehow) the attribute to the object itself _is_ spelled correctly :)

openml/tasks/task.py
```diff
@@ -177,6 +177,20 @@ def __init__(self, task_id, task_type_id, task_type, data_set_id,
         )
         self.number_of_clusters = number_of_clusters
 
+    def get_X(self, dataset_format='array'):
+        """Get data associated with the current task.
+
```
PGijsbers: type hint, doc string with legal values for `dataset_format` parameter.

openml/tasks/task.py
```diff
@@ -108,7 +108,7 @@ def __init__(self, task_id, task_type_id, task_type, data_set_id,
         self.target_name = target_name
         self.split = None
 
-    def get_X_and_y(self):
+    def get_X_and_y(self, dataset_format='array'):
         """Get data associated with the current task.
```
PGijsbers: type hint, doc string with legal values for `dataset_format` parameter.

openml/runs/trace.py
```diff
@@ -346,9 +348,23 @@ def trace_from_xml(cls, xml):
 
         return cls(run_id, trace)
 
+    @classmethod
+    def merge_traces(cls, traces: List['OpenMLRunTrace']):
```
PGijsbers: return type not specified in typehint

tests/test_extensions/test_sklearn_extension/test_sklearn_extension.py
```diff
@@ -1254,101 +1273,259 @@ def test_seed_model_raises(self):
...
+        fold_evaluations = collections.defaultdict(lambda: collections.defaultdict(dict))
+        for measure in user_defined_measures:
+            fold_evaluations[measure][0][0] = user_defined_measures[measure]
+
         # trace. SGD does not produce any
-        self.assertIsInstance(arff_tracecontent, list)
-        self.assertEqual(len(arff_tracecontent), 0)
+        self.assertIsNone(trace)
+
+        self._check_fold_timing_evaluations(fold_evaluations, num_repeats, num_folds,
```
PGijsbers: since neither `num_repeats` nor `num_folds` seem to be used throughout the test other than here, I would prefer the call to just be
```
self._check_fold_timing_evaluations(fold_evaluations, 
    num_repeats=1, num_folds=1,
    task_type=task.task_type_id, check_scores=False)
```
which makes that clear.

tests/test_extensions/test_sklearn_extension/test_sklearn_extension.py
```diff
@@ -1254,101 +1273,259 @@ def test_seed_model_raises(self):
...
+        # check user defined measures
+        fold_evaluations = collections.defaultdict(lambda: collections.defaultdict(dict))
         for measure in user_defined_measures:
             fold_evaluations[measure][0][0] = user_defined_measures[measure]
 
+        # check that it produced and returned a trace object of the correct length
+        self.assertIsInstance(trace, OpenMLRunTrace)
+        self.assertEqual(len(trace.trace_iterations), 2)
+
         self._check_fold_timing_evaluations(fold_evaluations, num_repeats, num_folds,
```
PGijsbers: same note on `num_repeats` and `num_folds`

tests/test_extensions/test_sklearn_extension/test_sklearn_extension.py
```diff
@@ -1254,101 +1273,259 @@ def test_seed_model_raises(self):
...
+                model=clf1,
+                task=task,
+                X_train=X_train,
+                y_train=y_train,
+                X_test=X_test,
+                fold_no=0,
+                rep_no=0,
+                classes=task.class_labels,
+            )
+            pred_2, proba_2, _, _ = self.extension._run_model_on_fold(
```
PGijsbers: `proba_1` and `proba_2` are unused? so no need to name them either.
You can ignore any amount of return arguments with `pred_1, *_`.

tests/test_extensions/test_sklearn_extension/test_sklearn_extension.py
```diff
@@ -1254,101 +1273,259 @@ def test_seed_model_raises(self):
...
+
+        # check user defined measures
+        fold_evaluations = collections.defaultdict(lambda: collections.defaultdict(dict))
+        for measure in user_defined_measures:
+            fold_evaluations[measure][0][0] = user_defined_measures[measure]
+
+        # trace. SGD does not produce any
+        self.assertIsNone(trace)
+
+        self._check_fold_timing_evaluations(fold_evaluations, num_repeats, num_folds,
```
PGijsbers: `num_repeats`

tests/test_extensions/test_sklearn_extension/test_sklearn_extension.py
```diff
@@ -1254,101 +1273,259 @@ def test_seed_model_raises(self):
...
```
PGijsbers: `num_repeats`

openml/extensions/sklearn/extension.py
mfeurer: There's no difference, thanks for spotting this issue, it'll make the API a bit easier.

openml/extensions/sklearn/extension.py
mfeurer: That check will go away when using `task.class_labels`. I will add a check that `X_test` and `y_train` are given at the top of the function.

openml/extensions/sklearn/extension.py
mfeurer: The zero-columns might need to be in the middle of the array. I added an example to the comment.

openml/runs/functions.py
mfeurer: Done

openml/runs/functions.py
mfeurer: Done, thanks for the tip with `start=1`.

openml/runs/functions.py
mfeurer: It's not used any more so I removed it.

openml/runs/functions.py
mfeurer: Done

openml/runs/trace.py
mfeurer: Done

openml/runs/trace.py
mfeurer: I fixed the attribute and removed the equals function as it was not used.

openml/tasks/task.py
mfeurer: Done.

openml/tasks/task.py
mfeurer: Done.

openml/runs/trace.py
mfeurer: Done.

tests/test_extensions/test_sklearn_extension/test_sklearn_extension.py
mfeurer: Done.

tests/test_extensions/test_sklearn_extension/test_sklearn_extension.py
mfeurer: Done.

tests/test_extensions/test_sklearn_extension/test_sklearn_extension.py
mfeurer: I added some additional asserts to actually use them.

tests/test_extensions/test_sklearn_extension/test_sklearn_extension.py
mfeurer: Done.

tests/test_extensions/test_sklearn_extension/test_sklearn_extension.py
mfeurer: Done.

openml/extensions/sklearn/extension.py
mfeurer: `rep_no` and `fold_no` are necessary to create trace objects.

openml/extensions/sklearn/extension.py
mfeurer: Done

openml/runs/trace.py
```diff
@@ -346,9 +348,35 @@ def trace_from_xml(cls, xml):
...
+                            'Cannot merge traces because the parameters are not equal: {} vs {}'.
+                            format(
+                                list(merged_trace[previous_iteration].parameters.keys()),
+                                list(iteration.parameters.keys()),
+                            )
+                        )
+                merged_trace[key] = iteration
+                previous_iteration = key
+
+        return cls(None, merged_trace)
```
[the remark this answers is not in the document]
mfeurer: Thanks for the suggestion. I'm not 100% sure whether this will allow the same flexibility, for example, the first statement would only check the iterations in one trace, and it would be hard to extract a useful error message out of it. Therefore, i suggest leaving the code as it is.

mfeurer:
Okay @PGijsbers I found the issue. After some offline discussion with @janvanrijn I figured out that we can see the reason why there is no trace in the error log of the run. It turns out that there were three bugs:

1. there was an indexing error which caused the predictions to have wrong indices and therefore make the evaluation fail. I fixed this.
2. in case there was no evaluation, after 200 seconds the test continued and raised the cryptic error message that there was no trace associated. I added a useful error message.
3. In case there were no evaluations, due to recently added run caching, the test could not retrieve any new evaluations by the server because it would continue to check the cached run. I added a flag to ignore the cached run and download it again from the server.

All in all, this should be good to merge now.

openml/runs/trace.py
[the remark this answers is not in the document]
PGijsbers: Right, I missed that.
```python
all_iterations = [iter for trace in traces for iter in trace]
for (iter1, iter2) in zip(all_iterations, all_iterations[1:]):
   if sorted(iter1.parameters) != sorted(iter2.parameters):
       raise ValueError(...)

```
should then work fine? But I am fine leaving this as is (and perhaps refactoring it later if we can find something we all like).

PR #676, opened 2019-04-16
Fix dataset parsing for categories

mfeurer:
The recent PR #548 changed the behavior of retrieving the categories of a categorical variable by inducing it from the data. In case the data does not contain a category, we loose information about its potential existence. This PR uses the category information provided by the arff file to determine all legal categories of a categorical variable.


janvanrijn:
I can attest that all tests are green btw

mfeurer:
Okay, I did so.

PR #677, opened 2019-04-16
[MRG] Fix402

### LINKED ISSUE #402: openml.datasets.get_data() has unpredictable number of return values
MWE:
```
dataset = openml.datasets.get_dataset(40978)

res = dataset.get_data(target=dataset.default_target_attribute)
print(len(res))
dataset = openml.datasets.get_dataset(2)

res = dataset.get_data(target=dataset.default_target_attribute)
print(len(res))
```
This depends on whether a default target attribute has been specified.

PGijsbers:
This PR contributes:
 - Fixing #402: `dataset.get_data` now always returns a quaternary tuple of `data, targets, categorical, attribute names`.
 - Removing the `FutureWarning` for moving to `pd.DataFrame` and adjust default behavior.
 - Update tests to reflect both new default behaviors.

Second contribution and related test updates really should have been a separate PR.
I am sorry for putting it into one, it kinda happened.

Closes #402.
Will need to create a new issue, because I removed [this TODO](https://github.com/openml/openml-python/blob/develop/tests/test_datasets/test_dataset.py#L134) (all TODOs should rather be centralized on the issue tracker, right?).


tests/test_datasets/test_dataset.py
```diff
@@ -27,114 +27,26 @@ def setUp(self):
...
-        )
         self.assertEqual(len(categorical), 39)
         self.assertTrue(all([isinstance(cat, bool) for cat in categorical]))
-        rval, attribute_names = self.dataset.get_data(
-            dataset_format='array', return_attribute_names=True
-        )
         self.assertEqual(len(attribute_names), 39)
         self.assertTrue(all([isinstance(att, str)
                              for att in attribute_names]))
+        self.assertEqual(_, None)
```
mfeurer: Could you please use `self.assertIsNone(_)`?

tests/test_datasets/test_dataset.py
```diff
@@ -171,14 +83,80 @@ def test_get_data_pandas(self):
...
+
+    def test_get_data_with_rowid(self):
+        self.dataset.row_id_attribute = "condition"
+        rval, _, categorical, _ = self.dataset.get_data(include_row_id=True)
+        self.assertEqual(rval.shape, (898, 39))
+        self.assertEqual(len(categorical), 39)
+
+        rval, _, categorical, _ = self.dataset.get_data()
+        self.assertEqual(rval.shape, (898, 38))
+        self.assertEqual(len(categorical), 38)
```
mfeurer: Hm, what do you think about adding some type checks here?

tests/test_datasets/test_dataset.py
```diff
@@ -234,61 +212,48 @@ def setUp(self):
...
         self.assertEqual(X.dtype, np.float32)
         self.assertIsInstance(y, np.ndarray)
         self.assertIn(y.dtype, [np.int32, np.int64])
         self.assertEqual(X.shape, (600, 20000))
-        X, y, attribute_names = self.sparse_dataset.get_data(
-            dataset_format='array',
-            target="class",
-            return_attribute_names=True,
-        )
         self.assertTrue(sparse.issparse(X))
```
mfeurer: I assume this check is no longer necessary?

tests/test_datasets/test_dataset.py
```diff
@@ -234,61 +212,48 @@ def setUp(self):
...
-        self.assertTrue(sparse.issparse(rval))
+
         self.assertEqual(len(attribute_names), 20001)
-        self.assertTrue(all([isinstance(att, str)
-                             for att in attribute_names]))
+        self.assertTrue(all([isinstance(att, str) for att in attribute_names]))
 
     def test_get_sparse_dataframe(self):
-        rval = self.sparse_dataset.get_data(dataset_format='dataframe')
+        rval, *_ = self.sparse_dataset.get_data(dataset_format='dataframe')
```
mfeurer: Isn't dataframe the new default? Could you adapt this for all unit test?

openml/datasets/dataset.py
```diff
@@ -419,29 +419,31 @@ def _download_data(self) -> None:
...
         include_ignore_attributes : boolean (default=False)
             Whether to include columns that are marked as "ignore"
             on the server in the dataset.
-        return_categorical_indicator : boolean (default=False)
-            Whether to return a boolean mask indicating which features are
-            categorical.
-        return_attribute_names : boolean (default=False)
-            Whether to return attribute names.
-        dataset_format : string, optional
+        dataset_format : string, optional (default='dataframe')
```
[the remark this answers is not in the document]
PGijsbers: Why not? It has a default value, so specifying it is optional? Would there be other requirements?

tests/test_datasets/test_dataset.py
PGijsbers: I am a little annoyed by the fact that this file seems to be 80% type checking in almost the same format every time.
But you are right that type checking is probably sensible here, in particular the column types as if anything went wrong with omitting the column, column types might be mismatched too.

I guess this is something to refactor at some point though.

tests/test_datasets/test_dataset.py
PGijsbers: There is no check here. This is just a call to retrieve attribute names, which is now always done by default anyway.

tests/test_datasets/test_dataset.py
PGijsbers: It is. I am actually not entirely sure what is the best thing to do in unit tests; specify optional parameters or not.
I am actually leaning to specifying them if they are relevant to the test/expected output.

This way, you clearly create a unit test for each possible behavior explicitly. Whatever is then the default behavior should still work.

What do you think?

openml/datasets/dataset.py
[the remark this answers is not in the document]
mfeurer: Doesn't optional imply that one can pass `None` as defined [here](https://docs.python.org/3/library/typing.html#typing.Optional)? However, this method requires a dataset format.

tests/test_datasets/test_dataset.py
mfeurer: Yes, that makes sense. Actually, it doesn't really work if you don't specifiy anything (as noted above), so this is totally fine.

openml/datasets/dataset.py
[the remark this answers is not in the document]
PGijsbers: You are right. I forgot for a second that this is about _typing_ here. It's an optional argument, but not `Optional` :)

PR #680, opened 2019-04-18
Easy access test server

PGijsbers:
Two functions which make it easier to switch between a test environment and a production environment.
This is so we can create examples which operate on the test server and upload things, and have them easily be reproducible for people trying to follow the examples.

In the old style, you would have to set the server and apikey yourself. This is a little bit more verbose, possibly requires having to look up what the apikey is, and having the apikey in more plain sight.

With the changes we can instead use:
```
>>> import openml
>>> len(openml.tasks.list_tasks())
18742
>>> openml.config.start_use_example_configuration()
>>> len(openml.tasks.list_tasks())
1527
>>> openml.config.stop_use_example_configuration()
>>> len(openml.tasks.list_tasks())
18742
```

The apikey for the test server is from the openml tutorials, and is different from the one we use for unit testing. This is in the event we later need to purge specific subset of data from the test server (e.g. we find someone abused the example apikey and it is affecting unit tests).

Yay or nay?

I'll add a unit test if the concept is approved.

### DIFF
--- a/openml/config.py
+++ b/openml/config.py
@@ -38,8 +38,8 @@ avoid_duplicate_runs = True if _defaults['avoid_duplicate_runs'] == 'True' else 
 connection_n_retries = _defaults['connection_n_retries']
 
 
-class ExampleConfiguration:
-    """ Allows easy switching to and from a test configuration. """
+class ConfigurationForExamples:
+    """ Allows easy switching to and from a test configuration, used for examples. """
     _last_used_server = None
     _last_used_key = None
     _start_last_called = False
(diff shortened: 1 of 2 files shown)


openml/config.py
```diff
@@ -37,6 +37,33 @@
...
+# variables to keep track of last used settings for when `use_example_configuration` is activated
+_last_used_key = None
+_last_used_server = None
+
+
+def start_use_example_configuration():
+    global server
+    global apikey
+    global _last_used_server
+    global _last_used_key
```
PGijsbers: For some reason, I could not access the variables local to the module, so I have to work with global variables here as well :( 
Either that or introduce some class which saves the information in class attributes, (e.g. `openml.config.someclass.start_use_example_configuration()`, but that seemed uglier to me).

PGijsbers:
I cancelled the `pr` tests because its not final yet, to free resources for other PRs.

PGijsbers:
Changed it, but to `ConfigurationForExamples` as I thought this would be even clearer? If you disagree, I will just make it `ConfigurationExample`.

openml/config.py
```diff
@@ -140,8 +188,12 @@ def set_cache_directory(cachedir):
     cache_directory = cachedir
 
 
+start_use_example_configuration = ConfigurationForExamples.start_use_example_configuration
```
mfeurer: Sorry for not being specific enough, could you please also change the name of these functions to make their use clearer?

mfeurer:
Taking over this PR.

PR #683, opened 2019-04-19
Fix59

PGijsbers:
Getting datasets by name!
Before I add unit tests, does this like the desired behavior?
I am unsure about the error types to use.
Looking for feedback.

### DIFF
--- a/openml/datasets/functions.py
+++ b/openml/datasets/functions.py
@@ -298,6 +298,32 @@ def check_datasets_active(dataset_ids: List[int]) -> Dict[int, bool]:
     return active
 
 
+def _name_to_id(dataset_name: str, error_if_multiple: bool = False) -> int:
+    """ Attempt to find the dataset id of the dataset with the given name.
+
+    If multiple datasets with the name exist, and `error_if_multiple` is `False`,
+    then return the least recent still active dataset.
+
+    Raises an error if no dataset is found.
+
+    Parameters
+    ----------
+    dataset_name : str
+        The name of the dataset for which to find its id.
+    error_if_multiple : bool (default=False)
+        If `False`, if multiple datasets match, return the least recent active dataset.
+        If `True`, if multiple datasets match, raise an error.
+
+    Returns
+    -------
+    int
+       The id of the dataset.
+    """
+    
+
+
+
+
 def get_datasets(
         dataset_ids: List[Union[str, int]],
         download_data: bool = True,
--- a/openml/datasets/functions.py
+++ b/openml/datasets/functions.py
@@ -324,13 +324,14 @@ def _name_to_id(dataset_name: str, version: int = None, error_if_multiple: bool 
     """
     candidates = list_datasets(data_name=dataset_name, status='active', data_version=version)
     if error_if_multiple and len(candidates) > 1:
-        raise ValueError("Multiple active datasets exist")
+        raise ValueError("Multiple active datasets exist with name {}".format(dataset_name))
     if len(candidates) == 0:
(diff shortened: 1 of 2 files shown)


openml/datasets/functions.py
```diff
@@ -328,13 +366,22 @@ def get_datasets(
 
 
 @openml.utils.thread_safe_if_oslo_installed
-def get_dataset(dataset_id: Union[int, str], download_data: bool = True) -> OpenMLDataset:
+def get_dataset(dataset_id: Union[int, str],
```
mfeurer: Could you please format like:
```
def func(
   arg1: type,
   arg2: type,
) -> return type:
  pass
```?

File: openml/datasets/functions.py [hunk 2]
```diff
@@ -247,19 +247,19 @@ def __list_datasets(api_call):
...
     datasets = dict()
     for dataset_ in datasets_dict['oml:data']['oml:dataset']:
-        did = int(dataset_['oml:did'])
-        dataset = {'did': did,
-                   'name': dataset_['oml:name'],
-                   'format': dataset_['oml:format'],
-                   'status': dataset_['oml:status']}
+        ignore_attributes = ['oml:file_id', 'oml:quality']
+        # [4:] skips the 'oml:' prefix.
+        dataset = {k[4:]: v for (k, v) in dataset_.items() if k not in ignore_attributes}
```
mfeurer: You could also do `k.replace('oml:', '')` to get rid of the comment.

openml/datasets/functions.py
```diff
@@ -298,6 +298,43 @@ def check_datasets_active(dataset_ids: List[int]) -> Dict[int, bool]:
...
+    Raises an error if a version is specified but it could not be found.
+
+    Parameters
+    ----------
+    dataset_name : str
+        The name of the dataset for which to find its id.
+    version : int
+        Version to retrieve. If not specified, the oldest active version is returned.
+    error_if_multiple : bool (default=False)
+        If `False`, if multiple datasets match, return the least recent active dataset.
```
mfeurer: For rst you have to use '``' for inline code formatting.

openml/datasets/functions.py
```diff
@@ -328,13 +366,22 @@ def get_datasets(
 
 
 @openml.utils.thread_safe_if_oslo_installed
-def get_dataset(dataset_id: Union[int, str], download_data: bool = True) -> OpenMLDataset:
+def get_dataset(dataset_id: Union[int, str],
+                download_data: bool = True,
+                version: int = None,
```
mfeurer: needs to be `Optional[int]`

mfeurer:
The error types are fine, but you could also replace the runtime error by a value error I guess.

PGijsbers:
The reason I opted for a RuntimeError is less because it is more appropriated, but more because then you can test against the two types of errors without parsing their messages. What do you think?

functions#414 also gives a mypy error, because for some reason it can't infer that `dataset_id` is a `str` and still treats it as `Union[str, int]`... should I `cast` here? or use `str(dataset_id)`, even though we really know it is a string already?

PGijsbers:
@mfeurer pretty much wrapped up now. Currently deactivated datasets can not be retrieved by name (not even if the version is specified). Is this an issue? I am fine with having it like this.

PGijsbers:
Also I opted not to repeat all the `get_dataset`  and `get_datasets`  unit tests. Just tested obvious input. The choice to do this was that everything should behave exactly the same, as long as the correct id is retrieved. For that reason tested the behavior of `_name_to_id` directly instead.

PR #692, opened 2019-05-20
Apidocs

Neeratyoy:
* Adds missing classes and functions to the API docs
* Adds missing docstrings for classes and functions featuring in the API
* Handles the case of missing docstrings for functions wrapped by decorators
* Edits the Pull Request template instruction, asking for new functions/classes to be included alphabetically in api.rst

* Viewing the api page

#### Any other comments?
* For some functions/classes the docstrings are placeholders and can be much improved.

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -13,14 +13,21 @@ Top-level Classes
    :toctree: generated/
    :template: class.rst
 
+   OpenMLClassificationTask
+   OpenMLClusteringTask
+   OpenMLDataFeature
    OpenMLDataset
-   OpenMLRun
-   OpenMLTask
-   OpenMLSplit
-   OpenMLFlow
    OpenMLEvaluation
-   OpenMLSetup
+   OpenMLFlow
+   OpenMLLearningCurveTask
    OpenMLParameter
+   OpenMLRegressionTask
+   OpenMLRun
+   OpenMLSetup
+   OpenMLSplit
+   OpenMLStudy
+   OpenMLSupervisedTask
+   OpenMLTask
 
 .. _api_extensions:
 
@@ -42,9 +49,10 @@ Extensions
    :toctree: generated/
    :template: function.rst
 
-    register_extension
-    get_extension_by_model
     get_extension_by_flow
+    get_extension_by_model
+    register_extension
+
 
 Modules
 -------
@@ -83,10 +91,10 @@ Modules
    :toctree: generated/
    :template: function.rst
 
+    assert_flows_equal
     flow_exists
     get_flow
     list_flows
-    assert_flows_equal
 
 :mod:`openml.runs`: Run Functions
 ----------------------------------
@@ -127,20 +135,20 @@ Modules
    :toctree: generated/
    :template: function.rst
 
-    get_study
-    get_suite
-    create_study
-    create_benchmark_suite
-    update_study_status
-    update_suite_status
-    attach_to_study
-    attach_to_suite
-    detach_from_study
-    detach_from_suite
-    delete_study
-    delete_suite
-    list_studies
-    list_suites
+   attach_to_study
+   attach_to_suite
+   create_benchmark_suite
+   create_study
+   delete_study
+   delete_suite
+   detach_from_study
+   detach_from_suite
+   get_study
+   get_suite
+   list_studies
+   list_suites
+   update_study_status
+   update_suite_status'
 
 :mod:`openml.tasks`: Task Functions
 -----------------------------------
@@ -153,16 +161,3 @@ Modules
     get_task
     get_tasks
     list_tasks
-
-
-:mod:`openml.extensions`: Extension Functions
------------------------------------
-.. currentmodule:: openml.extensions
-
-.. autosummary::
-   :toctree: generated/
-   :template: function.rst
-
-    register_extension
-    get_extension_by_model
-    get_extension_by_flow
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -19,6 +19,8 @@ Top-level Classes
    OpenMLSplit
    OpenMLFlow
    OpenMLEvaluation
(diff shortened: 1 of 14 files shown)


joaquinvanschoren:
Hi Neeratyoy, Matthias, seems like the unit tests broke because of the release of scipy 1.3.0:
https://github.com/scipy/scipy/releases

which adds this:
```
For the Attribute-Relation File Format (ARFF) scipy.io.arff.loadarff
now supports relational attributes.
```

and breaks this:
https://travis-ci.org/openml/openml-python/jobs/534987620#L1420

Is this easy to fix, or shall we update the dependencies to:  scipy>0.13.3,<1.3.0?
https://github.com/openml/openml-python/blob/4152f91254c31d45dcc761870a209b42ac4df234/setup.py#L45

mfeurer:
@joaquinvanschoren thanks for having a look at this. I'd actually prefer to use liac-arff for this instead. First, it's much faster now since Joel Nothman from the sklearn team improved it, and second, we'd use the same arff reader in all places. However, I suggest that we use a different PR for that. @Neeratyoy could you please put that on your todo list?

Neeratyoy:
@mfeurer alright I shall look into this first.

doc/api.rst
```diff
@@ -100,6 +113,7 @@ Modules
     list_runs
     run_model_on_task
     run_flow_on_task
+    run_exists
```
mfeurer: This isn't in the `__all__` variable of `runs/__init__.py`. Could you please add it theres?

openml/__init__.py
```diff
@@ -37,10 +37,11 @@
 from .runs import OpenMLRun
 from . import flows
 from .flows import OpenMLFlow
-from . import setups
 from . import study
-from .study import OpenMLStudy
+from .study import BaseStudy, OpenMLStudy, OpenMLBenchmarkSuite
```
mfeurer: Please do not make `BaseStudy` available to the user(s).

openml/study/__init__.py
```diff
@@ -1,4 +1,4 @@
-from .study import OpenMLStudy, OpenMLBenchmarkSuite
+from .study import OpenMLStudy, OpenMLBenchmarkSuite, BaseStudy
```
mfeurer: Please do not make `BaseStudy` available to the user here.

openml/tasks/split.py
```diff
@@ -10,6 +10,14 @@
 
 
 class OpenMLSplit(object):
+    """OpenML Split object.
+
+       Parameters
+       ----------
+       name : int or str
+       description : str
+       split : int
```
mfeurer: This seems to be incorrect. Shouldn't this be a dictionary?

openml/tasks/task.py
```diff
@@ -17,6 +17,14 @@
 
 
 class OpenMLTask(ABC):
+    """OpenML Task object.
+
+       Parameters
+       ----------
+       task_type_id : int or str
```
mfeurer: Seing this I'm surprised that we allow strings here and cast them to int inside the constructor. Could you please update the code so that:
1. the casting happens outside
2. the docstring reflects this
3. the type annotation reflects this, too

openml/tasks/task.py
```diff
@@ -200,6 +208,14 @@ def publish(self) -> int:
 
 
 class OpenMLSupervisedTask(OpenMLTask, ABC):
+    """OpenML Supervised Classification object.
+
+       Inherited from :class:`openml.OpenMLTask`
+
+       Parameters
+       ----------
+       target_name : str
```
mfeurer: Could you please add some further explanation, like `Name of the target feature used for supervised learning`?

openml/tasks/task.py
```diff
@@ -351,6 +380,14 @@ def __init__(
 
 
 class OpenMLClusteringTask(OpenMLTask):
+    """OpenML Clustering object.
+
+       Inherited from :class:`openml.OpenMLTask`
+
+       Parameters
+       ----------
+       target_name : str
```
mfeurer: This should be optional, right? Also, could you add a bit of description here.

doc/api.rst
Neeratyoy: If I understand this right, I believe it's there?
https://github.com/openml/openml-python/blob/apidocs/openml/runs/__init__.py#L25

openml/tasks/split.py
Neeratyoy: Yeah, it should be `dict`.

openml/tasks/task.py
Neeratyoy: By casting outside, are you referring to an explicit cast made by the user or in any function invoking the constructor? Just checked some other constructors, and they accept ids only as `int`. 
Shall I make the id parameters take only `int` here too?

openml/__init__.py
Neeratyoy: Removed.

openml/study/__init__.py
Neeratyoy: Removed.
Pushed the changes.

openml/tasks/task.py
Neeratyoy: Added.

doc/api.rst
mfeurer: Not sure how I missed that, sorry.

openml/study/__init__.py
mfeurer: Sorry, but this one is still there.

openml/tasks/task.py
mfeurer: I mean 'any function invoking the constructor'. Yes, it would be good to only accepts integers here, too.

openml/study/__init__.py
Neeratyoy: Yes, sorry my mistake. Just noticed. Pushed the changes.

PR #693, opened 2019-05-21
Removing dependency on scipy.io.arff

Neeratyoy:
`_from_arff_file()` in _tasks/split.py_ uses _liac-arff_ instead of _scipy.io.arff_.

Running pytest or checking for unit test status.

#### Comments.
Some code cleanup will be performed once the unit tests pass.

### DIFF
--- a/openml/tasks/split.py
+++ b/openml/tasks/split.py
@@ -77,24 +77,12 @@ class OpenMLSplit(object):
                 raise FileNotFoundError(
                     'Split arff %s does not exist!' % filename
                 )
-            # splits, meta = scipy.io.arff.loadarff(filename)
-            # name = meta.name
             file_data = arff.load(open(filename))
             splits = file_data['data']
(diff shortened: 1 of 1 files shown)


openml/tasks/split.py
```diff
@@ -108,7 +110,11 @@ def _from_arff_file(cls, filename: str) -> 'OpenMLSplit':
                     repetitions[repetition][fold][sample] = ([], [])
                 split = repetitions[repetition][fold][sample]
 
-                type_ = line[type_idx].decode('utf-8')
+                if not isinstance(line[type_idx], str):
```
mfeurer: Is this case actually triggered with liac-arff?

Neeratyoy: Hi,
The decode was required for how the file was being read earlier. With liac-arff it reads as str. So yes you are correct. This if-statement is not really triggered.
Removing it.

PR #698, opened 2019-05-28
Fix bugs by using live server due to reduced volatility

mfeurer:
This PR fixes two bugs which were 'introduced' recently by data changes on the test server. This does not include the fix included in #693 and will therefore not fix all unit tests.


mfeurer:
(tests pass on appveyor, presumably due to a different scipy version used for Windows)

PGijsbers:
What's the reason for not merging in the latest version of dev (with mentioned fix)?

mfeurer:
Actually none, except that I'm not a fan of merging. The PR checks are green now, which should give us enough confidence that this PR does what it's supposed to do.

PR #700, opened 2019-05-29
Fix a bug

mfeurer:
Fixes a bug which makes sure that uploading a task does not fail due to the task already existing on the test server.

### DIFF
--- a/tests/test_tasks/test_task.py
+++ b/tests/test_tasks/test_task.py
@@ -62,6 +62,9 @@ class OpenMLTaskTest(TestBase):
                 # success
                 break
             except OpenMLServerException as e:
+                # Error code for 'task already exists'
+                # Should be 533 according to the docs
+                # (# https://www.openml.org/api_docs#!/task/post_task)
                 if e.code == 614:
                     continue
                 else:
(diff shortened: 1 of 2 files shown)


tests/test_tasks/test_task.py
```diff
@@ -42,16 +43,30 @@ def test_download_task(self):
...
+                    dataset_id=dataset_id,
+                    target_name=self._get_random_feature(dataset_id),
+                    estimation_procedure_id=self.estimation_procedure
+                )
+
+                task_id = task.publish()
+                # success
+                break
+            except OpenMLServerException as e:
+                if e.code == 614:
```
PGijsbers: Please document this error code.

tests/test_tasks/test_task.py
```diff
@@ -42,16 +43,30 @@ def test_download_task(self):
...
+                    continue
+                else:
+                    raise e
```
PGijsbers: Maybe use an `else` statement (as part of [`for` `else`](http://book.pythontips.com/en/latest/for_-_else.html#else-clause)) to fail the test explicitly with according descriptive message (rather than having to go from `task_id not defined`).

tests/test_tasks/test_task.py
mfeurer: Done

tests/test_tasks/test_task.py
mfeurer: Great idea! I just added this.

Issue #653, opened 2019-03-26
Quick summaries of openml objects


joaquinvanschoren:
#### Description
I often see users struggle with interpreting what an openml object (e.g. a dataset or a pipeline) really is and what it contains. For beginning users, it is also unclear what they can do to get more insight. Nobody likes to run `print(vars(flow)`, and going to the website to get more information is a bit troublesome (and not obvious).

What about we provide a summary() function for dataset, tasks, flows, and runs that provides a quick but rich overview of the object?

#### Expected Results
For instance, running something like
```python
import openml
data = openml.flows.get_dataset(1)
data.summary()
```
could output a snippet of the description, a small table with statistics, maybe even a small table with the first couple of rows and columns.

Likewise, running
```python
flow = openml.flows.get_flow(1)
flow.summary()
```
Could print out a list (or graph) of the pipeline components.

Likewise, `task.summary()` could return the task details, and `run.summary()` could return some run details and evaluation results.

In a notebook environment, you could even include some nice pandas tables and plots, but that will always work.

janvanrijn:
I agree that we need a string representation of objects, the absence of these seemed to be confusing for our aspiring new users on the tutorial last Saturday. 

Isn't this what the `__str__` and `__repr__` functions do? It is for example implemented for `OpenMLRun` objects, but as far as I know not for other objects.

joaquinvanschoren:
If the output is pure text, then we can do it with standard __str__. I just wanted to think more broadly about what would be really useful and intuitive.

print(run) currently returns:
[run id: 2, task id: 72, flow id: 75, flow name: weka.AdaBoostM1_DecisionSt...]

What I would love to see is something like:
Task: ID: 72, breast-cancer (version 1)
Flow: ID: 75, weka.AdaBoostM1_DecisionStump() ...
Uploader: Jan van Rijn
AUC: 0.8
Accuracy: 0.9

mfeurer:
How would that be different than the output of `__repr__` or `__str__` (under the assumption that someone implements that)?

Potential duplicate of #75.

PGijsbers:
It definitely overlaps with #75. Though I think a `__repr__` implementation should be more succinct and this should be either `__str__` or its own method. To me, the differences could be e.g.:

__run__
`__repr__`: `{Run {id}|Local run} of {Flow {id}|local flow} on Task {id} (by uploader {username}).`
Example: `Run 2 of Flow 75 on Task 72 by Jan van Rijn.`
[...]

PR #701, opened 2019-05-30
Adding object summary

Neeratyoy:
#### Reference Issue
Addresses #653.

Adds `__str__` for top-level classes that returns a custom string for displaying brief, necessary object information.

By a `print()` of the object variable.

Examples:
- `d = openml.datasets.get_dataset(523); print(d)`
- `f = openml.flows.get_flow(3740); print(f)`
- `r = openml.runs.get_run(10000); print(r)`
- `t = openml.tasks.get_task(31); print(t)`
- `s = openml.study.get_study(16); print(s)`
- `b = openml.study.get_suite(218); print(b)`
- `s = openml.setups.get_setup(10); print(s)`
  - `print(s.parameters[190])`
- `e = openml.evaluations.list_evaluations(function='predictive_accuracy', task=[31], size=10)`
  - `print(e[51])`

#### Any other comments?
- The reason for including URLs (which are not object attributes) is to nudge the user to visit the website for additional information since certain _descriptions_ can get arbitrarily long.
- I have tested with only a handful of IDs. I would be keen to test these on datasets, tasks, flows, etc. that are anomalous and can break this.
- Please provide suggestions, corrections on how the above can be done better.
- Also, if the examples should be edited to show that object summaries can be _pretty printed_.

### DIFF
--- a/openml/datasets/dataset.py
+++ b/openml/datasets/dataset.py
@@ -189,7 +189,7 @@ class OpenMLDataset(object):
         num_instances = ''
         if object_dict['qualities']['NumberOfInstances'] is not None:
             num_instances = '%-14s: %d\n' % ("# of instances",
-                                            object_dict['qualities']['NumberOfInstances'])
+                                             object_dict['qualities']['NumberOfInstances'])
         num_features = '%-14s: %d\n' % ("# of features", len(object_dict['features']))
         output_str = name + version + format + date + licence + d_url + w_url + local_file + \
             pickle_file + num_instances + num_features
--- a/openml/datasets/dataset.py
+++ b/openml/datasets/dataset.py
@@ -178,23 +178,23 @@ class OpenMLDataset(object):
         header = '{}\n{}\n'.format(header, '=' * len(header))
 
         base_url = "{}".format(openml.config.server[:-len('api/v1/xml')])
-        fields = pd.Series({"Name": self.name,
-                            "Version": self.version,
-                            "Format": self.format,
-                            "Upload Date": self.upload_date.replace('T', ' '),
-                            "Licence": self.licence,
-                            "Download URL": self.url,
-                            "OpenML URL": "{}d/{}".format(base_url, self.dataset_id),
-                            "Data file": self.data_file,
-                            "Pickle file": self.data_pickle_file,
-                            "# of features": len(self.features)})
-
+        fields = {"Name": self.name,
+                  "Version": self.version,
+                  "Format": self.format,
+                  "Upload Date": self.upload_date.replace('T', ' '),
+                  "Licence": self.licence,
+                  "Download URL": self.url,
+                  "OpenML URL": "{}d/{}".format(base_url, self.dataset_id),
+                  "Data file": self.data_file,
+                  "Pickle file": self.data_pickle_file,
+                  "# of features": len(self.features)}
         if self.qualities['NumberOfInstances'] is not None:
-            fields.append(pd.Series({"# of instances": int(self.qualities['NumberOfInstances'])}))
+            fields["# of instances"] = int(self.qualities['NumberOfInstances'])
 
+        # determines the order in which the information will be printed
         order = ["Name", "Version", "Format", "Upload Date", "Licence", "Download URL",
-                 "OpenML URL", "Data File", "Pickle File", "# of features"]
-        fields = list(fields.reindex(order).dropna().iteritems())
+                 "OpenML URL", "Data File", "Pickle File", "# of features", "# of instances"]
+        fields = [(key, fields[key]) for key in order if key in fields]
 
         longest_field_name_length = max(len(name) for name, value in fields)
         field_line_format = "{{:.<{}}}: {{}}".format(longest_field_name_length)
--- a/openml/datasets/dataset.py
+++ b/openml/datasets/dataset.py
@@ -174,29 +174,32 @@ class OpenMLDataset(object):
             self.data_pickle_file = None
 
     def __str__(self):
-        object_dict = self.__dict__
-        output_str = ''
         header = "OpenML Dataset"
         header = '{}\n{}\n'.format(header, '=' * len(header))
-        name = '{:.<14}: {}\n'.format("Name", object_dict['name'])
-        version = '{:.<14}: {}\n'.format("Version", object_dict['version'])
-        format = '{:.<14}: {}\n'.format("Format", object_dict['format'])
-        date = '{:.<14}: {}\n'.format("Upload Date", object_dict['upload_date'].replace('T', ' '))
-        licence = '{:.<14}: {}\n'.format("Licence", object_dict['licence'])
-        d_url = '{:.<14}: {}\n'.format("Download URL", object_dict['url'])
-        base_url = 'https://www.openml.org/d/'
-        w_url = '{:.<14}: {}\n'.format("OpenML URL", base_url + str(self.dataset_id))
-        local_file = '{:.<14}: {}\n'.format("Data file", object_dict['data_file'])
-        pickle_file = '{:.<14}: {}\n'.format("Pickle file", object_dict['data_pickle_file'])
-        num_features = '{:.<14}: {}\n'.format("# of features", len(object_dict['features']))
-        num_instances = ''
-        if object_dict['qualities']['NumberOfInstances'] is not None:
-            num_instances = '{:.<14}: {}\n'.format("# of instances",
-                                                   object_dict['qualities']['NumberOfInstances'])
 
-        output_str = '\n' + header + name + version + format + date + licence + d_url + w_url + \
-            local_file + pickle_file + num_features + num_instances + '\n'
-        return output_str
+        base_url = "{}".format(openml.config.server[:-len('api/v1/xml')])
+        fields = pd.Series({"Name": self.name,
+                            "Version": self.version,
+                            "Format": self.format,
+                            "Upload Date": self.upload_date.replace('T', ' '),
+                            "Licence": self.licence,
+                            "Download URL": self.url,
+                            "OpenML URL": "{}d/{}".format(base_url, self.dataset_id),
+                            "Data file": self.data_file,
+                            "Pickle file": self.data_pickle_file,
+                            "# of features": len(self.features)})
+
+        if self.qualities['NumberOfInstances'] is not None:
+            fields.append(pd.Series({"# of instances": int(self.qualities['NumberOfInstances'])}))
+
+        order = ["Name", "Version", "Format", "Upload Date", "Licence", "Download URL",
+                 "OpenML URL", "Data File", "Pickle File", "# of features"]
+        fields = list(fields.reindex(order).dropna().iteritems())
+
+        longest_field_name_length = max(len(name) for name, value in fields)
+        field_line_format = "{{:.<{}}}: {{}}".format(longest_field_name_length)
+        body = '\n'.join(field_line_format.format(name, value) for name, value in fields)
+        return header + body
 
     def _data_arff_to_pickle(self, data_file):
         data_pickle_file = data_file.replace('.arff', '.pkl.py3')
--- a/openml/datasets/dataset.py
+++ b/openml/datasets/dataset.py
@@ -181,13 +181,15 @@ class OpenMLDataset(object):
         fields = {"Name": self.name,
                   "Version": self.version,
                   "Format": self.format,
-                  "Upload Date": self.upload_date.replace('T', ' '),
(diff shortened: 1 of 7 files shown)


openml/datasets/dataset.py
```diff
@@ -173,6 +173,28 @@ def __init__(self, name, description, format=None,
...
+    def __str__(self):
+        object_dict = self.__dict__
+        output_str = ''
+        name = '\n%-14s: %s\n' % ("Name", object_dict['name'])
+        version = '%-14s: %s\n' % ("Version", object_dict['version'])
+        format = '%-14s: %s\n' % ("Format", object_dict['format'])
+        date = '%-14s: %s\n' % ("Upload Date", object_dict['upload_date'].replace('T', ' '))
+        licence = '%-14s: %s\n' % ("Licence", object_dict['licence'])
+        d_url = '%-14s: %s\n' % ("Download URL", object_dict['url'])
+        base_url = 'https://www.openml.org/d/'
```
mfeurer: Could you please use the server URL given by `openml.config.server`? In case the user works against the test server, this URL would be wrong.

openml/datasets/dataset.py
```diff
@@ -173,6 +173,28 @@ def __init__(self, name, description, format=None,
         else:
             self.data_pickle_file = None
 
+    def __str__(self):
+        object_dict = self.__dict__
+        output_str = ''
+        name = '\n%-14s: %s\n' % ("Name", object_dict['name'])
```
mfeurer: I don't think it is necessary to add the identifier via string interpolation, you can simply have it in the main string.

openml/datasets/dataset.py
```diff
@@ -173,6 +173,28 @@ def __init__(self, name, description, format=None,
...
+        date = '%-14s: %s\n' % ("Upload Date", object_dict['upload_date'].replace('T', ' '))
+        licence = '%-14s: %s\n' % ("Licence", object_dict['licence'])
+        d_url = '%-14s: %s\n' % ("Download URL", object_dict['url'])
+        base_url = 'https://www.openml.org/d/'
+        w_url = '%-14s: %s\n' % ("OpenML URL", base_url + str(self.dataset_id))
+        local_file = '%-14s: %s\n' % ("Data file", object_dict['data_file'])
+        pickle_file = '%-14s: %s\n' % ("Pickle file", object_dict['data_pickle_file'])
+        num_instances = ''
+        if object_dict['qualities']['NumberOfInstances'] is not None:
+            num_instances = '%-14s: %d\n' % ("# of instances",
```
mfeurer: Could you please move the "# of instances" to the next line?

openml/datasets/dataset.py
```diff
@@ -173,6 +173,28 @@ def __init__(self, name, description, format=None,
...
+                                             object_dict['qualities']['NumberOfInstances'])
```
mfeurer: Could you please add the number of features, too? And also the number of classes if that's available?

openml/tasks/task.py
```diff
@@ -42,6 +42,34 @@ def __init__(
...
+        target = ''
+        class_labels = ''
+        cost_matrix = ''
+        if object_dict['target_name'] is not None:
+            target = '%-20s: %s\n' % ("Target Feature", object_dict['target_name'])
+            if 'class_labels' in object_dict:
+                class_labels = '%-20s: %s\n' % ("# of Classes", len(object_dict['class_labels']))
+            if 'cost_matrix' in object_dict:
+                cost_matrix = '%-20s: %s\n' % ("Cost Matrix", "Available")
+        output_str = task_type + task_id + task_url + evaluation_measure + estimation_procedure + \
```
mfeurer: Could you please switch the estimation procedure and the evaluation measure?

openml/flows/flow.py
```diff
@@ -132,6 +132,28 @@ def __init__(self, name, description, model, components, parameters,
...
+
+        binary = ''
+        if object_dict['binary_url'] is not None:
+            binary = '%-16s: %s\n\n' % ('Binary URL', object_dict['binary_url'])
+
+        upload = '%-16s: %s\n' % ('Upload Date', object_dict['upload_date'].replace('T', ' '))
+        language = '%-16s: %s\n' % ('Language', object_dict['language'])
+        dependencies = '%-16s: %s\n' % ('Dependencies', object_dict['dependencies'])
+        # 3740 for example
+        output_str = id + version + url + name + description + binary + upload + \
```
mfeurer: The name should probably be above the URL? Also, what do you think of `Flow ID: %d (version %d)` to make the display of the version more compact?

openml/flows/flow.py
```diff
@@ -132,6 +132,28 @@ def __init__(self, name, description, model, components, parameters,
...
+                               'https://www.openml.org/f/' + str(object_dict['flow_id']))
+        name = '%-16s: %s\n' % ('Flow Name', object_dict['name'])
+        description = '%-16s: %s\n\n' % ('Flow Description', object_dict['description'])
+
+        binary = ''
+        if object_dict['binary_url'] is not None:
+            binary = '%-16s: %s\n\n' % ('Binary URL', object_dict['binary_url'])
+
+        upload = '%-16s: %s\n' % ('Upload Date', object_dict['upload_date'].replace('T', ' '))
+        language = '%-16s: %s\n' % ('Language', object_dict['language'])
```
mfeurer: Please remove the language, it is kind of unimportant (it is mostly English anyway).

openml/study/study.py
```diff
@@ -89,6 +89,36 @@ def __init__(
...
+        if object_dict['runs'] is not None:
+            runs = '%-16s: %s\n' % ("# of Runs", len(object_dict['runs']))
+
+        url = 'https://www.openml.org/u/' + str(object_dict['creator'])
+        creator = '\n%-16s: %s\n' % ("Creator", url)
+        upload_time = '%-16s: %s\n' % ("Upload Time",
+                                       object_dict['creation_date'].replace('T', ' '))
+        output_str = id + name + status + main_entity_type + study_url + data + \
+            tasks + flows + runs + creator + upload_time
+        return(output_str)
```
mfeurer: Please no brackets around the return values.

openml/study/study.py
```diff
@@ -89,6 +89,36 @@ def __init__(
...
+            flows = '%-16s: %s\n' % ("# of Flows", len(object_dict['flows']))
+        runs = ''
+        if object_dict['runs'] is not None:
+            runs = '%-16s: %s\n' % ("# of Runs", len(object_dict['runs']))
+
+        url = 'https://www.openml.org/u/' + str(object_dict['creator'])
+        creator = '\n%-16s: %s\n' % ("Creator", url)
+        upload_time = '%-16s: %s\n' % ("Upload Time",
+                                       object_dict['creation_date'].replace('T', ' '))
+        output_str = id + name + status + main_entity_type + study_url + data + \
```
mfeurer: What do you think about adding a header to the output string, for example 
```
OpenML Dataset
==============
ID: 5 (version 1)
...
```
obviously, this needs to be adapted for the different entities. Also, you'd need to create separate functions for the study and the suite class.

Neeratyoy: I have added __str__ functions for the other two classes which inherit this class, OpenMLStudy and OpenMLBenchmarkSuite. The __str__ appends the relevant header to the base class' __str__ string.

openml/datasets/dataset.py
Neeratyoy: I am unsure of what to do here.
Shall I change the ` base_url = 'https://www.openml.org/d/'` to ` base_url = 'https://test.openml.org/d/'` ? 
However,  [https://www.openml.org/d/31](https://www.openml.org/d/31)  and [https://test.openml.org/d/31](https://test.openml.org/d/31) don't point to the same resource.

Neeratyoy: Understood. Made the changes.

Neeratyoy:
> I don't think the string representations should have leading and trailing newlines. Was this ever discussed? Other large pretty prints like `pandas.DataFrame` or `numpy.ndarray` don't use them either.

No, it wasn't discussed. These implementations were my independent design choices. Have corrected them as per your suggestions.

> 
> What is the reason for referencing `self` through `object_dict` instead?

Oh, this was a silly mistake wherein I was using the `object_dict` to study which attributes can be printed for the objects. Continued using that for the rest of the function instead of referencing through `self`. Sorry about that.

> There are some magic numbers going on here, such as the length of the left column (i.e. number of trailing periods). I suspect it has to do with however long the longest line is. There is also a lot of repetition. Let me suggest some refactoring, for e.g. dataset:
> 
> ```python
>     def __str__(self):
>         header = "OpenML Dataset"
>         header = '{}\n{}\n'.format(header, '=' * len(header))
> 
>         openml_url = "{}d/{}".format(openml.config.server[:-len('api/v1/xml')], self.dataset_id)
>         fields = [("Name", self.name),
>                   ("Version", self.version),
>                   ("Format", self.format),
>                   ("Upload Date", self.upload_date.replace('T', ' ')),
>                   ("Licence", self.licence),
>                   ("Download URL", self.url),
>                   ("OpenML URL", openml_url),
>                   ("Data file", self.data_file),
>                   ("Pickle file", self.data_pickle_file),
>                   ("# of features", len(self.features))]
> 
>         if self.qualities['NumberOfInstances'] is not None:
>             fields.append(("# of instances", self.qualities['NumberOfInstances']))
> 
>         longest_field_name_length = max(len(name) for name, value in fields)
>         field_line_format = "{{:.<{}}}: {{}}".format(longest_field_name_length)
>         body = '\n'.join(field_line_format.format(name, value) for name, value in fields)
>         return header + body
> ```
> 
> Note that this refactor incorporates the advice to look at `openml.config.server` to determine the `base_url` , discards leading/trailing newlines and directly uses `self`.

This looks really good. I wanted to do something about the character padding. Wasn't aware of the escape formatting with `{{`. This is what I was looking for. Thanks a lot!

> Use `list.insert` instead of `list.append` for optional fields that might be inserted in the middle of a body.
> Other than that, it gives identical output (as long as the magic number was indeed determined by the longest field name).

I understood the reasoning behind `insert` and not `append`. However, with `insert` the index number needed to be specified again, which would again be _ghost numbers_ that might become incomprehensible later. 
I am therefore now using Pandas Series to create the fields to be printed. Along with a list of strings which independently determines the order in which the field appears and is able to handle a missing value for a (for optional fields). 
Just felt that the slightly redundant list of strings is a worthy trade-off for explicit insert indexes. Also, this seems more customizable, in terms of addition/deletion of fields, or reordering. Kindly advise.

PGijsbers:
I understand where you're coming from. I perhaps should have been more explicit. I would have something like this for flow, where possibly a binary url would come before upload date:
```python
if self.binary_url is not None:
    before_idx = [name for name, value in fields].index("Upload Date")
    fields.insert(before_idx , ("Binary URL": self.binary_url))
```
that said, it does introduce some work for the developer if they want to find out the full order. Additionally they also have to be careful about ordering if multiple such optional elements should be printed one after another. I think that's where your solution is better.

But can I ask why use `pd.Series` and not a regular `dict`?
It adds a lot of additional verbosity (`pd.Series({...})` everywhere), and the only use for the series object itself seems to be (unless I missed it) in 
`fields = list(fields.reindex(order).dropna().iteritems())`
Which, if `fields` was a `dict` instead, would be replaced by:
```
fields = [(name, fields[name]) for name in order if fields.get(name) is not None]
```
if you don't want to print _any_ None values, or
```
fields = [(name, fields[name]) for name in order if name in fields]
```
if you want to print `None` values if they have been explicitly added to the `fields` dict.

I haven't tested it myself, but I think this small change should keep equivalent behavior while using standard python over `pd.Series`.

Neeratyoy:
Well, this is completely down to my history with R :/ Ordering by 'rownames' is quite common there and I just rabbit-holed into this particular solution. You're right, I can do the same with a dict representation and use a list comprehension to create the equivalent `field` variable. Rest remaining the same.
And I think with the current design I don't need to check for a `None` since I am adding the optional attributes if available and just checking if the relevant key is present or not should suffice.
I'll make this change.

PGijsbers:
No worries :) I'm still learning to write cleaner Python myself. But whenever I spot easy improvements I really would like them to be done, it just helps so much down the line (easier for maintenance, extending, newcomers, etc.). Sorry for being a hard-ass 😅

Neeratyoy:
Haha! Absolutely understandable and much needed, to be honest. I am learning well ;)

PGijsbers:
Everything looks good. As these changes are not invasive I would like to hold off a merge until we fix the dev branch. I'll look at that sooner rather than later.

PR #710, opened 2019-06-12
Fixing ignoring of features on upload

### LINKED ISSUE #589: ignore features ignored when uploaded
(code doesn't run because dataset is not processed on downloading yet, but the idea is clear)

```
import openml

openml.config.server = 'https://test.openml.org/api/v1/'


data = [
    ['sunny', 85, 85, 'FALSE', 'no'],
    ['sunny', 80, 90, 'TRUE', 'no'],
    ['overcast', 83, 86, 'FALSE', 'yes'],
    ['rainy', 70, 96, 'FALSE', 'yes'],
    ['rainy', 68, 80, 'FALSE', 'yes'],
    ['rainy', 65, 70, 'TRUE', 'no'],
    ['overcast', 64, 65, 'TRUE', 'yes'],
    ['sunny', 72, 95, 'FALSE', 'no'],
    ['sunny', 69, 70, 'FALSE', 'yes'],
    ['rainy', 75, 80, 'FALSE', 'yes'],
    ['sunny', 75, 70, 'TRUE', 'yes'],
    ['overcast', 72, 90, 'TRUE', 'yes'],
    ['overcast', 81, 75, 'FALSE', 'yes'],
    ['rainy', 71, 91, 'TRUE', 'no'],
]

attribute_names = [
    ('outlook', ['sunny', 'overcast', 'rainy']),
    ('temperature', 'REAL'),
    ('humidity', 'REAL'),
    ('windy', ['TRUE', 'FALSE']),
    ('play', ['yes', 'no']),
]

description = (
    'The weather problem is a tiny dataset that we will use repeatedly'
    ' to illustrate machine learning methods. Entirely fictitious, it '
    'supposedly concerns the conditions that are suitable for playing '
    'some unspecified game. In general, instances in a dataset are '
    'characterized by the values of features, or attributes, that measure '
    'different aspects of the instance. In this case there are four '
    'attributes: outlook, temperature, humidity, and windy. '
    'The outcome is whether to play or not.'
)

citation = (
    'I. H. Witten, E. Frank, M. A. Hall, and ITPro,'
    'Data mining practical machine learning tools and techniques, '
    'third edition. Burlington, Mass.: Morgan Kaufmann Publishers, 2011'
)
ignore = ['outlook', 'temperature']
weather_dataset = openml.datasets.create_dataset(
    name="Weather",
    description=description,
    creator='I. H. Witten, E. Frank, M. A. Hall, and ITPro',
    contributor=None,
    collection_date='01-01-2011',
    language='English',
    licence=None,
    default_target_attribute='play',
    row_id_attribute=None,
    ignore_attribute=ignore,
    citation=citation,
    attributes=attribute_names,
    data=data,
    version_label='example',
)

upload_did = weather_dataset.publish()
print('URL for dataset: %s/data/%d' % (openml.config.server, upload_did))

dataset = openml.datasets.get_dataset(upload_did)
assert ignore == dataset.ignore_attributes

```

Neeratyoy:
#### Reference Issue
Fixes #589. 

The `_to_xml()` function iterates over all attributes to construct the XML string. There was a typo w.r.t. the `ignore_attributes` attribute. The _s_ was missing at the end.

By reproducing the same example in #589.

### DIFF
--- a/doc/progress.rst
+++ b/doc/progress.rst
@@ -21,10 +21,12 @@ Changelog
 * ADD #659: Lazy loading of task splits.
 * ADD #516: `run_flow_on_task` flow uploading is now optional.
 * ADD #680: Adds `openml.config.start_using_configuration_for_example` (and resp. stop) to easily connect to the test server.
+* ADD #75, #653: Adds a pretty print for objects of the top-level classes.
 * FIX #642: `check_datasets_active` now correctly also returns active status of deactivated datasets.
 * FIX #304, #636: Allow serialization of numpy datatypes and list of lists of more types (e.g. bools, ints) for flows.
 * FIX #651: Fixed a bug that would prevent openml-python from finding the user's config file.
 * FIX #693: OpenML-Python uses liac-arff instead of scipy.io for loading task splits now.
+* DOC #639: More descriptive documention for function to convert array format.
 * DOC #678: Better color scheme for code examples in documentation.
 * DOC #681: Small improvements and removing list of missing functions.
 * DOC #684: Add notice to examples that connect to the test server.
--- a/doc/progress.rst
+++ b/doc/progress.rst
@@ -6,6 +6,11 @@
 Changelog
 =========
 
+0.10.0
+~~~~~~
+* FIX #589: Fixing a bug that did not successfully upload the columns to ignore when creating and publishing a dataset.
+* DOC #639: More descriptive documention for function to convert array format.
+
 0.9.0
 ~~~~~
 * ADD #560: OpenML-Python can now handle regression tasks as well.
@@ -26,8 +31,6 @@ Changelog
 * FIX #304, #636: Allow serialization of numpy datatypes and list of lists of more types (e.g. bools, ints) for flows.
 * FIX #651: Fixed a bug that would prevent openml-python from finding the user's config file.
 * FIX #693: OpenML-Python uses liac-arff instead of scipy.io for loading task splits now.
-* FIX #589: Fixing a bug that did not successfully upload the columns to ignore when creating and publishing a dataset.
-* DOC #639: More descriptive documention for function to convert array format.
 * DOC #678: Better color scheme for code examples in documentation.
 * DOC #681: Small improvements and removing list of missing functions.
 * DOC #684: Add notice to examples that connect to the test server.
--- a/doc/progress.rst
+++ b/doc/progress.rst
@@ -26,6 +26,7 @@ Changelog
 * FIX #304, #636: Allow serialization of numpy datatypes and list of lists of more types (e.g. bools, ints) for flows.
 * FIX #651: Fixed a bug that would prevent openml-python from finding the user's config file.
 * FIX #693: OpenML-Python uses liac-arff instead of scipy.io for loading task splits now.
+* FIX #589: Fixing a bug that did not successfully upload the columns to ignore when creating and publishing a dataset.
 * DOC #639: More descriptive documention for function to convert array format.
 * DOC #678: Better color scheme for code examples in documentation.
 * DOC #681: Small improvements and removing list of missing functions.
--- a/openml/datasets/dataset.py
+++ b/openml/datasets/dataset.py
@@ -680,7 +680,7 @@ class OpenMLDataset(object):
         props = ['id', 'name', 'version', 'description', 'format', 'creator',
                  'contributor', 'collection_date', 'upload_date', 'language',
                  'licence', 'url', 'default_target_attribute',
-                 'row_id_attribute', 'ignore_attribute', 'version_label',
+                 'row_id_attribute', 'ignore_attributes', 'version_label',
                  'citation', 'tag', 'visibility', 'original_data_url',
                  'paper_url', 'update_comment', 'md5_checksum']
 
--- a/openml/datasets/dataset.py
+++ b/openml/datasets/dataset.py
@@ -690,6 +690,8 @@ class OpenMLDataset(object):
 
         for prop in props:
             content = getattr(self, prop, None)
+            if prop == 'ignore_attributes':
+                prop = "ignore_attribute"
             if content is not None:
                 data_dict["oml:" + prop] = content
 
--- a/openml/datasets/dataset.py
+++ b/openml/datasets/dataset.py
@@ -132,9 +132,9 @@ class OpenMLDataset(object):
         self.default_target_attribute = default_target_attribute
         self.row_id_attribute = row_id_attribute
         if isinstance(ignore_attribute, str):
-            self.ignore_attributes = [ignore_attribute]
+            self.ignore_attribute = [ignore_attribute]
         elif isinstance(ignore_attribute, list) or ignore_attribute is None:
-            self.ignore_attributes = ignore_attribute
+            self.ignore_attribute = ignore_attribute
         else:
             raise ValueError('Wrong data type for ignore_attribute. '
                              'Should be list.')
@@ -423,7 +423,7 @@ class OpenMLDataset(object):
             self,
             target: Optional[Union[List[str], str]] = None,
             include_row_id: bool = False,
-            include_ignore_attributes: bool = False,
+            include_ignore_attribute: bool = False,
             dataset_format: str = "dataframe",
     ) -> Tuple[
             Union[np.ndarray, pd.DataFrame, scipy.sparse.csr_matrix],
@@ -440,7 +440,7 @@ class OpenMLDataset(object):
             Splitting multiple columns is currently not supported.
         include_row_id : boolean (default=False)
             Whether to include row ids in the returned dataset.
-        include_ignore_attributes : boolean (default=False)
+        include_ignore_attribute : boolean (default=False)
             Whether to include columns that are marked as "ignore"
             on the server in the dataset.
         dataset_format : string (default='dataframe')
@@ -479,11 +479,11 @@ class OpenMLDataset(object):
             elif isinstance(self.row_id_attribute, Iterable):
                 to_exclude.extend(self.row_id_attribute)
 
-        if not include_ignore_attributes and self.ignore_attributes is not None:
-            if isinstance(self.ignore_attributes, str):
-                to_exclude.append(self.ignore_attributes)
-            elif isinstance(self.ignore_attributes, Iterable):
-                to_exclude.extend(self.ignore_attributes)
+        if not include_ignore_attribute and self.ignore_attribute is not None:
+            if isinstance(self.ignore_attribute, str):
+                to_exclude.append(self.ignore_attribute)
+            elif isinstance(self.ignore_attribute, Iterable):
+                to_exclude.extend(self.ignore_attribute)
 
         if len(to_exclude) > 0:
             logger.info("Going to remove the following attributes:"
@@ -566,7 +566,7 @@ class OpenMLDataset(object):
         return None
 
     def get_features_by_type(self, data_type, exclude=None,
-                             exclude_ignore_attributes=True,
+                             exclude_ignore_attribute=True,
                              exclude_row_id_attribute=True):
         """
         Return indices of features of a given type, e.g. all nominal features.
@@ -579,7 +579,7 @@ class OpenMLDataset(object):
         exclude : list(int)
             Indices to exclude (and adapt the return values as if these indices
                         are not present)
-        exclude_ignore_attributes : bool
+        exclude_ignore_attribute : bool
             Whether to exclude the defined ignore attributes (and adapt the
             return values as if these indices are not present)
         exclude_row_id_attribute : bool
@@ -593,9 +593,9 @@ class OpenMLDataset(object):
         """
         if data_type not in OpenMLDataFeature.LEGAL_DATA_TYPES:
             raise TypeError("Illegal feature type requested")
-        if self.ignore_attributes is not None:
-            if not isinstance(self.ignore_attributes, list):
-                raise TypeError("ignore_attributes should be a list")
+        if self.ignore_attribute is not None:
+            if not isinstance(self.ignore_attribute, list):
+                raise TypeError("ignore_attribute should be a list")
         if self.row_id_attribute is not None:
             if not isinstance(self.row_id_attribute, str):
                 raise TypeError("row id attribute should be a str")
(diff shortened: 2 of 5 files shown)


Neeratyoy:
An extremely minor change was required. I checked the XML string being returned and it now features the ignored attributes, whereas it wasn't earlier.

However, when I try to `publish()` the example given in #589 it returns a 103 saying authentication failed. Though my `openml.config.apikey` contains my _apikey_. I even reset my _apikey_ on the website and tried but it still returned the authentication error.

janvanrijn:
Please note that we have separate keys for test and live server.

Neeratyoy:
Oh alright, thanks a lot! 
I registered at https://test.openml.org/home but am currently unable to sign in and it says 'Temporarily Locked Out. Try again later.'. Guess I'll try after some time and drop a note here.
Thanks!

Neeratyoy:
I would need some advice here. I am observing the following:
-  If `_to_xml()` defined for datasets check for _ignore\_attribute**s**_ attribute, then the generated XML string contains the ignored attributes, since the [constructor](https://github.com/openml/openml-python/blob/fix_589/openml/datasets/dataset.py#L134) creates the attribute _ignore\_attribute**s**_
    - However, on `publish()` this returns a code 131: Problem validating uploaded description file
- When the XML string contains _< oml:ignore_attribute >_ it is a successful POST request
    - However, it returns 131 with _< oml:ignore_attribute**s** >_

Currently, I've put in a  fix where I check specifically for this attribute and change the oml:tag while constructing the XML string. This seems to return datasets with the list of ignored datasets. The example on #589 works too.

PGijsbers:
It seems like there were not any unit tests covering this issue. The only unit test I see is that a dataset like this can be made, but not if uploading/downloading is successful. Is that correct?
If so, do you mind adding a unit test?

Also, I forgot this with the other PRs. Could you add your changes to the [changelog](https://github.com/openml/openml-python/blob/develop/doc/progress.rst)? You can add changes of the previous PRs into the changelog with this PR.

Unfortunately I can't answer the more detailed issues right now. It's not something I am very familiar with so I would need a bit of time to look into it. It seems like it might be a server/schema thing in which case @janvanrijn might be able to help.

Finally, if you quickly want to test on the test server, you can use `openml.config.start_using_configuration_for_example`. It automatically changes your configuration to connect to the test server with a valid apikey. You can stop this configuration with the respective stop method.

mfeurer:
@Neeratyoy please check the xsd definition here https://github.com/openml/OpenML/blob/develop/openml_OS/views/pages/api_new/v1/xsd/openml.data.upload.xsd it seems that `ignore_attribute` is correct for the upload. So there's a discrepancy between the name in the xsd and the class attribute. So we should probably change the class attribute name itself. What do you think @Neeratyoy @PGijsbers ?

PGijsbers:
I think the class attribute `ignore_attributes` is fine as this refers to the sequence and not the individual `ignore_attribute` nodes (those would be the elements of the list). On the flipside, we don't do that for e.g. `row_id_attribute` either (it's not called `row_id_attributes`) though for that field you maybe expect more commonly to find only one column specified.

As long as we are consistent, I am fine with either renaming. However, since we expect the fields to be lists, I personally would prefer using the plural. I would go so far as to also refactor e.g. `row_id_attribute` to be plural and by default coerce the input into a list [like is done for `ignore_attributes`](https://github.com/openml/openml-python/blob/fix_589/openml/datasets/dataset.py#L134).

Neeratyoy:
I don't know if there was a specific reason for it but it does look odd since the named argument for the [constructor](https://github.com/openml/openml-python/blob/fix_589/openml/datasets/dataset.py#L134) is in _singular_ while the attribute created is plural. Do I refactor just _ignore\_attribute_ to be the singular form across the code base?

PGijsbers:
:+1: Seems like a plan

PGijsbers:
There is one setup (Py3.6, Sklearn 0.18) where there seems to be a failure:
```
self = <tests.test_datasets.test_dataset_functions.TestOpenMLDataset testMethod=test_publish_fetch_ignore_attribute>
    def test_publish_fetch_ignore_attribute(self):
        data = [
            ['a', 'sunny', 85.0, 85.0, 'FALSE', 'no'],
            ['b', 'sunny', 80.0, 90.0, 'TRUE', 'no'],
            ['c', 'overcast', 83.0, 86.0, 'FALSE', 'yes'],
            ['d', 'rainy', 70.0, 96.0, 'FALSE', 'yes'],
            ['e', 'rainy', 68.0, 80.0, 'FALSE', 'yes']
        ]
        column_names = ['rnd_str', 'outlook', 'temperature', 'humidity',
                        'windy', 'play']
        df = pd.DataFrame(data, columns=column_names)
        # enforce the type of each column
        df['outlook'] = df['outlook'].astype('category')
        df['windy'] = df['windy'].astype('bool')
        df['play'] = df['play'].astype('category')
        # meta-information
        name = '%s-pandas_testing_dataset' % self._get_sentinel()
        description = 'Synthetic dataset created from a Pandas DataFrame'
        creator = 'OpenML tester'
        collection_date = '01-01-2018'
        language = 'English'
        licence = 'MIT'
        default_target_attribute = 'play'
        citation = 'None'
        original_data_url = 'http://openml.github.io/openml-python'
        paper_url = 'http://openml.github.io/openml-python'
    
        # pass a list to ignore_attribute
        ignore_attribute = ['outlook', 'windy']
        dataset = openml.datasets.functions.create_dataset(
            name=name,
            description=description,
            creator=creator,
            contributor=None,
            collection_date=collection_date,
            language=language,
            licence=licence,
            default_target_attribute=default_target_attribute,
            row_id_attribute=None,
            ignore_attribute=ignore_attribute,
            citation=citation,
            attributes='auto',
            data=df,
            version_label='test',
            original_data_url=original_data_url,
            paper_url=paper_url
        )
    
        # publish dataset
        upload_did = dataset.publish()
        # test if publish was successful
        self.assertIsInstance(dataset.dataset_id, int)
    
        trials = 0
        timeout_limit = 1000
        dataset = None
        # fetching from server
        # loop till timeout or fetch not successful
        while True:
            if trials > timeout_limit:
                break
            try:
                dataset = openml.datasets.get_dataset(upload_did)
                break
            except Exception as e:
                trials += 1
                if str(e).split(':')[-1].strip() == "Dataset not processed yet":
                    # if returned code 273: Dataset not processed yet
                    continue
                else:
>                   raise RuntimeError(str(e))
E                   RuntimeError: https://test.openml.org/api/v1/xml/data/qualities/17315 returned code 362: No qualities found
/home/travis/build/openml/openml-python/tests/test_datasets/test_dataset_functions.py:1131: RuntimeError
```

Neeratyoy:
@PGijsbers can you please advise me on the design of this test here?

Unlike other test cases where after a `publish()` only the data arff or features are retrieved, this particular tets is calling the `get_dataset()` which means it tries to retrieve the qualities too, which takes time to compute on the server end. I am therefore looping and trying to `get` the dataset `timeout_limit` times and am raising a RUNTIME error if it then fails.
I initially had the `timeout_limit` as *100*. However, sometimes it passed while other times it ran into the RUNTIME error -- 100 fetches were not enough. I increased it to *1000* for now. 

Is this a good design or there are other alternatives to induce a _wait_? Hope this doesn't create a bottleneck in the test case execution time.

doc/progress.rst
```diff
@@ -21,10 +21,13 @@ Changelog
 * ADD #659: Lazy loading of task splits.
 * ADD #516: `run_flow_on_task` flow uploading is now optional.
 * ADD #680: Adds `openml.config.start_using_configuration_for_example` (and resp. stop) to easily connect to the test server.
+* ADD #75, #653: Adds a pretty print for objects of the top-level classes.
 * FIX #642: `check_datasets_active` now correctly also returns active status of deactivated datasets.
 * FIX #304, #636: Allow serialization of numpy datatypes and list of lists of more types (e.g. bools, ints) for flows.
 * FIX #651: Fixed a bug that would prevent openml-python from finding the user's config file.
 * FIX #693: OpenML-Python uses liac-arff instead of scipy.io for loading task splits now.
+* FIX #589: Fixing a bug that did not successfully upload the columns to ignore when creating and publishing a dataset.
+* DOC #639: More descriptive documention for function to convert array format.
```
PGijsbers: Please add new additions under a new header (just call the header 'newest release' for now). The 0.9.0 release is already out, so this change is not contained in it.

PGijsbers:
I'll take a look tomorrow to see if there's a way around that.

PGijsbers:
This unit test just took me 1m12s to complete, which is terrible (no fault of yours, of course). For reference, that's about 2x what the next slowest test is, and about 10x what our 20th slowest test is.

I think this probably requires a solution we should apply more generally across unit tests, and since there is an open option on slow unit tests, I posted a response regarding solutions there: #330.

Neeratyoy:
@PGijsbers 
Running this unit test took me a little above 59s. Running all the unit tests in `test_dataset_functions.py`, and skipping this unit test took above 100s. 
From what I could understand, the `yield` behaves in a round-robin fashion to induce concurrency. The time slices for each test is determined by how we position the `yield` in our unit test. 

Now, for this particular unit test, I believe the bottleneck is the server side computation of the qualities for the dataset? Can anything be done about that? 
If not, then round robin (`yield`) for this particular unit test might be beneficial. Instead of waiting for the qualities to be fetched, for a minute, that minute can be used to execute other unit tests. The apparent waiting time can be used by the server for calculating the qualities. We could place this unit test at the beginning and mark each failure to `get` from the server with a `yield`. 

I need to check how the flow of execution will be with certain unit tests marked with `yield` and others without. Also to generalize for all test cases, I would need to better understand [this](https://warehouse-camo.cmh1.psfhosted.org/c3d619775d5ed8f0110754af034247d022d4dbbc/68747470733a2f2f7261772e67697468756275736572636f6e74656e742e636f6d2f6465766f76612f7079746573742d7969656c642f623063376161303538646635663530636239613035323732666365303166633632613738626265652f686f772d69742d776f726b732d7079746573742d7969656c642e7376673f73616e6974697a653d74727565). 

Could we segregate slower unit tests (maybe using the ordering by names design you mentioned), and have them start at the beginning and then move on using `yield`? However, I feel that depends on the reasons for bottlenecks in speed. If a test case block is slow due to certain operations in between `yields` then we kind of run into the same problem.

For this particular slow unit test, this looks promising, since the server side process is not being stalled. Shall I refactor and give this a try or look for a general design to scale across all slow unit tests?

PGijsbers:
I think there are two different bottlenecks for server operations; the server updating its internal status or the server computing a response. To clarify, in this case the server is updating its internal status with new information on the dataset, this takes time. The other case is e.g. finding all runs with a certain tag sorted by some order, where a single roundtrip takes long because the server has to process data before returning a result.

The former should be able to be sped up by breaking the unit test in two (either by `yield` or explicitly with method ordering), the latter can not as from Python's point of view the server roundtrip is atomic. For these disk/network intensive operations we could technically make use of the [`asyncio`](https://docs.python.org/3/library/asyncio.html) built-in module. However as far as I can tell asyncio does not work with unit tests in the sense that awaiting in a test would allow us to execute a different one (to preserve unit test independence).

I propose the following temporary fix: Since unit tests by default are ordered alphabetically, we explicitly split the test in two (uploading, checking) and rename the upload part `test___upload_fetch_ignore_attribute` and the check part `test_upload_fetch_ignore_attribute`. The triple underscore will order it the first test alphabetically, and naming it `upload_f...` will sort the second test last. We can store the necessary information in the unit test *class* (`TestOpenMLDataset._ignore_attribute_test_info = (dataset.dataset_id, ignore_attribute)`). Trying this locally this decreased the total duration of the test to a little over one second.

Since there is no straightforward fix, I would like to have Matthias involved in the decision on how to structurally address this problem.

PGijsbers:
Also please include comments in the unit tests to clearly indicate they are not independent but depend on each other.

Neeratyoy:
@PGijsbers I did as you suggested. Split the test into two.
Please note that the unit test independence didn't allow variables to be shared across two unit tests. And for this test, the dataset ID on upload is needed to retrieve it, in a different test.
This [thread](https://stackoverflow.com/questions/21447740/persist-variable-changes-between-tests-in-unittest) has some interesting discussions and I went with the accepted answer and it works. The whole file test takes about a 100s down from 250+s previously. However, the solution (to share variables) is a little hacky.
Pytest has these [fixtures](https://docs.pytest.org/en/latest/fixture.html#sharing-a-fixture-across-tests-in-a-module-or-class-session) which could possibly allow us to share modules across unit tests, however, the unit tests need to call the fixture module. On trying this it failed since I believe the class calling the unit tests don't pass any parameters. (I didn't edit anything else other than the unit test in question in this PR)

PGijsbers:
The monkeypatching `hack' is what I suggested initially :) so I'm fine with that. Don't get me wrong, I am not happy with it, but for now I think it's ok. Rather than looking into a neater way of making dependent unit tests, we should look into asynchronous unit tests.

Neeratyoy:
haha alright :stuck_out_tongue: 
guess we can tackle that as a new issue -- refactoring unit testing entirely.

PGijsbers:
Restarting 3 runs after a fail (should be server issues):
```
/home/travis/build/openml/openml-python/tests/test_runs/test_run_functions.py:576: 
/home/travis/build/openml/openml-python/tests/test_runs/test_run_functions.py:455: in _run_and_upload_classification
    sentinel=sentinel,
/home/travis/build/openml/openml-python/tests/test_runs/test_run_functions.py:419: in _run_and_upload
    raise e
/home/travis/build/openml/openml-python/tests/test_runs/test_run_functions.py:415: in _run_and_upload
    fold=0,
/home/travis/build/openml/openml-python/openml/runs/functions.py:330: in initialize_model_from_trace
    run_trace = get_run_trace(run_id)
/home/travis/build/openml/openml-python/openml/runs/functions.py:272: in get_run_trace
    'get')
/home/travis/build/openml/openml-python/openml/_api_calls.py:51: in _perform_api_call
    return _read_url(url, request_method, data)
url = 'https://test.openml.org/api/v1/xml/run/trace/34900'
request_method = 'get', data = {'api_key': '610344db6388d9ba34f6db45a3cf71de'}
    def _read_url(url, request_method, data=None):
        data = {} if data is None else data
        if config.apikey is not None:
            data['api_key'] = config.apikey
    
        response = send_request(request_method=request_method, url=url, data=data)
        if response.status_code != 200:
>           raise _parse_server_exception(response, url)
E           openml.exceptions.OpenMLServerException: https://test.openml.org/api/v1/xml/run/trace/34900 returned code 572: No successful trace associated with this run.
/home/travis/build/openml/openml-python/openml/_api_calls.py:98: OpenMLServerException
```

PGijsbers:
Going to merge as 2 out of 3 tests now also succeeded and [its a known issue](https://github.com/openml/OpenML/issues/960).

Issue #695, opened 2019-05-21
Get list of data qualities in python API


joaquinvanschoren:

Neeratyoy:
Hello,
Is the goal to get a `list_data_qualities(dataset_id=x)` function that returns an array/dict/dataframe with the qualities for that dataset?

janvanrijn:
Seems like you need this URL:
https://www.openml.org/api/v1/data/qualities/list

docu seems missing. cc @joaquinvanschoren

PGijsbers:
Added in #711.

PR #711, opened 2019-06-12
Function to list all data qualities

Neeratyoy:
#### Reference Issue
Addresses #695.

Adds a new function to list all available data qualities.

`openml.datasets.list_qualities()`
`openml.datasets.list_qualities(True)`

#### Any other comments?
The function returns a list in the default form. Have included an optional verbosity parameter which additionally prints the list with its corresponding index. Since the list is over 100 elements long, thought that retrieving the index visually might help with list access.

### DIFF
--- a/doc/progress.rst
+++ b/doc/progress.rst
@@ -21,6 +21,7 @@ Changelog
 * ADD #659: Lazy loading of task splits.
 * ADD #516: `run_flow_on_task` flow uploading is now optional.
 * ADD #680: Adds `openml.config.start_using_configuration_for_example` (and resp. stop) to easily connect to the test server.
+* ADD #695: A function to retieve all the data quality measures available.
 * FIX #642: `check_datasets_active` now correctly also returns active status of deactivated datasets.
 * FIX #304, #636: Allow serialization of numpy datatypes and list of lists of more types (e.g. bools, ints) for flows.
 * FIX #651: Fixed a bug that would prevent openml-python from finding the user's config file.
--- a/doc/progress.rst
+++ b/doc/progress.rst
@@ -6,6 +6,10 @@
 Changelog
 =========
 
+0.10.0
+~~~~~~
+* ADD #695: A function to retrieve all the data quality measures available.
+
 0.9.0
 ~~~~~
 * ADD #560: OpenML-Python can now handle regression tasks as well.
@@ -21,7 +25,6 @@ Changelog
 * ADD #659: Lazy loading of task splits.
 * ADD #516: `run_flow_on_task` flow uploading is now optional.
 * ADD #680: Adds `openml.config.start_using_configuration_for_example` (and resp. stop) to easily connect to the test server.
-* ADD #695: A function to retieve all the data quality measures available.
 * FIX #642: `check_datasets_active` now correctly also returns active status of deactivated datasets.
 * FIX #304, #636: Allow serialization of numpy datatypes and list of lists of more types (e.g. bools, ints) for flows.
 * FIX #651: Fixed a bug that would prevent openml-python from finding the user's config file.
--- a/openml/datasets/functions.py
+++ b/openml/datasets/functions.py
@@ -171,6 +171,10 @@ def list_qualities() -> list:
     The function performs an API call to retrieve the entire list of
     data qualities that are computed on the datasets uploaded.
 
+    Returns
+    -------
+    list
+    
     """
     api_call = "data/qualities/list"
     xml_string = openml._api_calls._perform_api_call(api_call, 'get')
--- a/openml/datasets/functions.py
+++ b/openml/datasets/functions.py
@@ -174,7 +174,6 @@ def list_qualities() -> list:
     Returns
     -------
     list
-    
     """
     api_call = "data/qualities/list"
     xml_string = openml._api_calls._perform_api_call(api_call, 'get')
--- a/openml/datasets/functions.py
+++ b/openml/datasets/functions.py
@@ -165,33 +165,24 @@ def _get_cache_directory(dataset: OpenMLDataset) -> str:
     return _create_cache_directory_for_id(DATASETS_CACHE_DIR_NAME, dataset.dataset_id)
 
 
-def list_qualities(verbose=False):
+def list_qualities() -> list:
     """ Return list of data qualities available.
 
     The function performs an API call to retrieve the entire list of
-    data qualities that are available are computed on the datasets uploaded.
-
-    Parameters
-    ----------
-    verbose : bool (default=False)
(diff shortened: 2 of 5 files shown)


openml/datasets/functions.py
```diff
@@ -165,6 +165,36 @@ def _get_cache_directory(dataset: OpenMLDataset) -> str:
...
+        raise ValueError('Error in return XML, does not contain '
+                         '"oml:data_qualities_list"')
+    assert type(qualities['oml:data_qualities_list']['oml:quality']) == list
+    qualities = qualities['oml:data_qualities_list']['oml:quality']
+    if verbose:
+        header = "List of available data qualities:"
+        print(header)
+        print("=" * len(header))
+        for i, quality in enumerate(qualities):
+            print("{:>3}....{}".format(i + 1, quality))
```
PGijsbers: If the purpose is to help with list indexing, I would print the index and not index + 1 (i.e. 0-based and not 1-based).

PGijsbers: If there's a good reason to start the list at 1, let me know. It's not a necessity for me.

openml/datasets/functions.py
```diff
@@ -165,6 +165,36 @@ def _get_cache_directory(dataset: OpenMLDataset) -> str:
     return _create_cache_directory_for_id(DATASETS_CACHE_DIR_NAME, dataset.dataset_id)
 
 
+def list_qualities(verbose=False):
+    """ Return list of data qualities available.
+
+    The function performs an API call to retrieve the entire list of
+    data qualities that are available are computed on the datasets uploaded.
```
mfeurer: This sentence appears to be incorrect, I believe removing the 'are available' solves this.

openml/datasets/functions.py
```diff
@@ -165,6 +165,36 @@ def _get_cache_directory(dataset: OpenMLDataset) -> str:
...
+
+    Parameters
+    ----------
+    verbose : bool (default=False)
+        If True, prints out the list with an index
+
```
mfeurer: The `Returns` heading is missing here.

openml/datasets/functions.py
```diff
@@ -165,6 +165,36 @@ def _get_cache_directory(dataset: OpenMLDataset) -> str:
...
+    """
+    api_call = "data/qualities/list"
+    xml_string = openml._api_calls._perform_api_call(api_call, 'get')
+    qualities = xmltodict.parse(xml_string)
+    # Minimalistic check if the XML is useful
+    if 'oml:data_qualities_list' not in qualities:
+        raise ValueError('Error in return XML, does not contain '
+                         '"oml:data_qualities_list"')
+    assert type(qualities['oml:data_qualities_list']['oml:quality']) == list
```
mfeurer: Please use `isinstance()` here. Also, there's an argument to xmltodict called enforce_list which converts a part of the XML to a list so you don't need this check.

openml/datasets/functions.py
mfeurer: Hm, could you post an example of the output? I'm right now not convinced about adding a verbosity argument to one of function and think the user can do this much better by printing the returned pandas dataframe. However, for that I think the return type argument is missing.

PGijsbers: I was wondering this myself. The return is a `list`, so a user just using `print(enumerate(openml.datasets.list_qualities()))` is nearly identical. 
This function's output (the str representation of the list):
```['AutoCorrelation', 'CfsSubsetEval_DecisionStumpAUC', ...]```
With verbosity (the printed messages):
```
List of available data qualities:
=================================
  1....AutoCorrelation
  2....CfsSubsetEval_DecisionStumpAUC
  ...
```
using `print(list(enumerate(openml.datasets.list_qualities())))`:
```
[(0, 'AutoCorrelation'), (1, 'CfsSubsetEval_DecisionStumpAUC'),  ...]
```




ps. @Neeratyoy  also a tip in general, `enumerate` has a `start` parameter so you could have used `enumerate(qualities, start=1)` instead and not have to manually add 1 each time.

Comment (Neeratyoy) [reply openml/datasets/functions.py]: To be honest, in hindsight, I too find the verbosity feature quite redundant and unnecessary. Initially, I wanted to have an option of an `output_format` like other listing functions. Since this function is just a simple list of strings, I kind of force-fitted this verbosity. 

@PGijsbers @mfeurer shall I omit the verbosity feature and return a list as is or enhance the printing as @PGijsbers mentioned and add the relevant docstring for the return?

Comment (PGijsbers) [reply openml/datasets/functions.py]: I'm in favor of dropping it. Keeping it might start a trend where we have to add more or less trivial verbosity to other functions if users find them in one place but not another.

Comment (Neeratyoy) [reply openml/datasets/functions.py]: Alright, makes sense.

File: openml/datasets/functions.py (hunk 4 above)
Comment (Neeratyoy) [reply openml/datasets/functions.py]: Alright. I was following the structure for other listing functions which uses `assert`. Should it be changed for the other listing functions too for this PR?

File: openml/datasets/functions.py [hunk 5]
```diff
@@ -165,6 +165,30 @@ def _get_cache_directory(dataset: OpenMLDataset) -> str:
     return _create_cache_directory_for_id(DATASETS_CACHE_DIR_NAME, dataset.dataset_id)
 
 
+def list_qualities() -> list:
```
PGijsbers: Sorry for not catching this the first time, but please indicate it is a `List[str]` in the type hint. Currently it is hard to tell what is being returned exactly.

PGijsbers:
Please also update the changelog and add a simple short unit test so we can automatically detect if something breaks this functionality.

tests/test_datasets/test_dataset_functions.py
```diff
@@ -1165,7 +1165,11 @@ def test_create_dataset_attributes_auto_without_df(self):
         creator = 'OpenML tester'
         collection_date = '01-01-2018'
         language = 'English'
-        licence = 'MIT'
+        licence = 'MIT'   def test_list_evaluation_measures(self):
```
PGijsbers: What's going on here? this doesn't seem right.

doc/progress.rst
```diff
@@ -21,6 +21,7 @@ Changelog
 * ADD #659: Lazy loading of task splits.
 * ADD #516: `run_flow_on_task` flow uploading is now optional.
 * ADD #680: Adds `openml.config.start_using_configuration_for_example` (and resp. stop) to easily connect to the test server.
+* ADD #695: A function to retieve all the data quality measures available.
```
PGijsbers: Please add new additions under a new header (just call the header 'newest release' for now). The 0.9.0 release is already out, so this change is not contained in it.

PR #715, opened 2019-06-17
add sort to list_evaluations

sahithyaravi1493:
Added sort option to list evaluations

from openml import evaluations
##### Sort descending:
evaluations.list_evaluations(function='predictive_accuracy', task=[task_id], sort='desc', offset=0, size=10)
##### Sort ascending:
evaluations.list_evaluations(function='predictive_accuracy', task=[task_id], sort='asc', offset=0, size=10)

#### Any other comments?
sort_by option discussed before is irrelevant for now, as the api only supports a sort_by value.

### DIFF
--- a/tests/test_evaluations/test_evaluation_functions.py
+++ b/tests/test_evaluations/test_evaluation_functions.py
@@ -116,8 +116,7 @@ class TestEvaluationFunctions(TestBase):
         for run_id in evaluations.keys():
             self.assertIsNotNone(evaluations[run_id].value)
             self.assertIsNone(evaluations[run_id].values)
-
-
+            
     def test_evaluation_list_sort(self):
         openml.config.server = self.test_server
         size = 10
--- a/tests/test_evaluations/test_evaluation_functions.py
+++ b/tests/test_evaluations/test_evaluation_functions.py
@@ -118,7 +118,7 @@ class TestEvaluationFunctions(TestBase):
             self.assertIsNone(evaluations[run_id].values)
 
     def test_evaluation_list_sort(self):
-        openml.config.server = self.production_server
+        openml.config.server = self.test_server
         size = 10
         task_id = 1769
         # Get all evaluations of the task
@@ -126,7 +126,7 @@ class TestEvaluationFunctions(TestBase):
             "predictive_accuracy", offset=0, task=[task_id])
         # Get top 10 evaluations of the same task
         sorted_eval = openml.evaluations.list_evaluations(
-            "predictive_accuracy", size=size, offset=0, task=[task_id], sort="desc")
+            "predictive_accuracy", size=size, offset=0, task=[task_id], sort_order="desc")
 
         sorted_output = []
         unsorted_output = []
--- a/tests/test_evaluations/test_evaluation_functions.py
+++ b/tests/test_evaluations/test_evaluation_functions.py
@@ -116,7 +116,7 @@ class TestEvaluationFunctions(TestBase):
         for run_id in evaluations.keys():
             self.assertIsNotNone(evaluations[run_id].value)
             self.assertIsNone(evaluations[run_id].values)
-            
+
     def test_evaluation_list_sort(self):
         openml.config.server = self.test_server
         size = 10
--- a/tests/test_evaluations/test_evaluation_functions.py
+++ b/tests/test_evaluations/test_evaluation_functions.py
@@ -127,7 +127,7 @@ class TestEvaluationFunctions(TestBase):
         sorted_eval = openml.evaluations.list_evaluations(
             "predictive_accuracy", size=size, offset=0, task=[task_id], sort_order="desc")
         self.assertEqual(len(sorted_eval), size)
-        self.assertGreater(len(unsorted_eval),0)
+        self.assertGreater(len(unsorted_eval), 0)
         sorted_output = []
         unsorted_output = []
         for eval in sorted_eval.values():
--- a/tests/test_evaluations/test_evaluation_functions.py
+++ b/tests/test_evaluations/test_evaluation_functions.py
@@ -128,12 +128,8 @@ class TestEvaluationFunctions(TestBase):
             "predictive_accuracy", size=size, offset=0, task=[task_id], sort_order="desc")
         self.assertEqual(len(sorted_eval), size)
         self.assertGreater(len(unsorted_eval), 0)
-        sorted_output = []
-        unsorted_output = []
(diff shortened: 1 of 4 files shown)


janvanrijn:
Please wait a little bit longer with merging. I would like to maybe change the specification to:
`sort_by` and `sort_order`, as @joaquinvanschoren asked for multiple sort options.

janvanrijn:
I changed the definition of `sort` to `sort_order` so we can later easily add `sort_by` :)

joaquinvanschoren:
@sahithyaravi1493 can you update the code to use the renamed `sort_order` filter?
@janvanrijn if that is done, can we merge this soon and add support for `sort_by` in a separate PR?

PGijsbers:
Don't forget to add the change to the progress/changelog :) `doc/progress.rst`

tests/test_evaluations/test_evaluation_functions.py
```diff
@@ -117,6 +117,31 @@ def test_evaluation_list_per_fold(self):
             self.assertIsNotNone(evaluations[run_id].value)
             self.assertIsNone(evaluations[run_id].values)
 
+    def test_evaluation_list_sort(self):
+        openml.config.server = self.test_server
```
PGijsbers: The test server is the default server when running unit tests, so it's unnecessary to specify.

tests/test_evaluations/test_evaluation_functions.py
```diff
@@ -117,6 +117,31 @@ def test_evaluation_list_per_fold(self):
...
+        # Get top 10 evaluations of the same task
+        sorted_eval = openml.evaluations.list_evaluations(
+            "predictive_accuracy", size=size, offset=0, task=[task_id], sort_order="desc")
+
+        sorted_output = []
+        unsorted_output = []
+        for run_id in sorted_eval.keys():
+            sorted_output.append(sorted_eval[run_id].value)
+        for run_id in unsorted_eval.keys():
+            unsorted_output.append(unsorted_eval[run_id].value)
```
PGijsbers: Why not use `Dict.values()` instead?

tests/test_evaluations/test_evaluation_functions.py
PGijsbers: Actually, I think you meant to specify the production server. The test server does not have the referenced task. Because of this, no run results are retrieved, and the unit test only succeeds because empty dictionaries are retrieved.

PGijsbers: Oh. I see `sort_order` only works on the test server. Then please update with a task that has runs on the test server.

tests/test_evaluations/test_evaluation_functions.py
```diff
@@ -117,6 +117,31 @@ def test_evaluation_list_per_fold(self):
...
+        openml.config.server = self.test_server
+        size = 10
+        task_id = 1769
+        # Get all evaluations of the task
+        unsorted_eval = openml.evaluations.list_evaluations(
+            "predictive_accuracy", offset=0, task=[task_id])
+        # Get top 10 evaluations of the same task
+        sorted_eval = openml.evaluations.list_evaluations(
+            "predictive_accuracy", size=size, offset=0, task=[task_id], sort_order="desc")
+
```
PGijsbers: Please add a sanity check that ensures no empty dictionaries are returned. As is, it seems like the current test only passes because no results are returned.

tests/test_evaluations/test_evaluation_functions.py
```diff
@@ -117,6 +117,31 @@ def test_evaluation_list_per_fold(self):
...
+        sorted_output = []
+        unsorted_output = []
+        for run_id in sorted_eval.keys():
+            sorted_output.append(sorted_eval[run_id].value)
+        for run_id in unsorted_eval.keys():
+            unsorted_output.append(unsorted_eval[run_id].value)
+
+        # Check if output from sort is sorted in the right order
+        self.assertTrue(sorted(sorted_output, reverse=True) == sorted_output)
```
PGijsbers: I don't think this works. I think the `sorted_output` in this case would contain `OpenMLEvaluation` objects which do not have a built-in comparison methods. You'll have to specify how you want to sort them, (e.g. specify a function through the `key` parameter).

tests/test_evaluations/test_evaluation_functions.py
sahithyaravi1493: Yes,  After the change from sort to sort_order, I did not check if the task exists on test server. I will add a sanity check and update this

tests/test_evaluations/test_evaluation_functions.py
sahithyaravi1493: sorted_output contains a list of values of the score. So its a list that looks like [1,0.98,0.9....].
(I checked these when I did this on the production server.)

PGijsbers: My bad. I read it wrong :) I didn't notice you mapped to the `value` property of evaluations.

tests/test_evaluations/test_evaluation_functions.py
```diff
@@ -117,6 +117,31 @@ def test_evaluation_list_per_fold(self):
...
+        sorted_eval = openml.evaluations.list_evaluations(
+            "predictive_accuracy", size=size, offset=0, task=[task_id], sort_order="desc")
+        self.assertEqual(len(sorted_eval), size)
+        self.assertGreater(len(unsorted_eval), 0)
+        sorted_output = []
+        unsorted_output = []
+        for eval in sorted_eval.values():
+            sorted_output.append(eval.value)
+        for eval in unsorted_eval.values():
+            unsorted_output.append(eval.value)
```
PGijsbers: Sorry for missing this the first time. Would you mind using a list comprehension instead here? i.e.:
```python
        sorted_output = [evaluation.value for evaluation in sorted_eval.values()]
        unsorted_output = [evaluation.value for evaluation in unsorted_eval.values()]
```
It's equivalent in function (but faster), and even being a bit more verbose in variable names (to avoid collision with the `eval` builtin), it fits perfectly fine in two lines :)
After that I think this looks good to go :+1:

tests/test_evaluations/test_evaluation_functions.py
PGijsbers: Yep, my bad. I missed you mapped to the `OpenMLEvaluation.value` and not the `OpenMLEvaluation`.

PGijsbers:
🎉

Issue #608, opened 2018-12-12
dataset_id referenced before assignment


ArlindKadra:
#### Description

While trying to get the run with id 186276, the error dataset_id referenced before assignment was raised. 

The error is happening here:

https://github.com/openml/openml-python/blob/7c0a77d232a64da76be18bbce76e4bb8aeb4155b/openml/runs/functions.py#L883

Taking a quick look into the function throwing it, this seems to be the problem:
https://github.com/openml/openml-python/blob/7c0a77d232a64da76be18bbce76e4bb8aeb4155b/openml/runs/functions.py#L804-L807

The run at https://www.openml.org/api/v1/run/186276 does not seem to have a dataset field at all. 

Should this run be removed ? @mfeurer  @janvanrijn 
#### Steps/Code to Reproduce
```
import openml

run = openml.runs.get_run(186276)
```

mfeurer:
I assume that there's an else statement missing?

ArlindKadra:
That would be the easy hack yes, but do we really want to have runs without
datasets ?It seems to me the code is correctly trying to enforce having a
dataset id if the run is from the server.

janvanrijn:
I think it's a problem with the run (will check why it's invalidated later)

Although not enforced by design, all current task types work on a dataset.

Op donderdag 13 december 2018 heeft Arlind Kadra <notifications@github.com>
het volgende geschreven:

PGijsbers:
To clarify; is there a valid reason that a run on the server should not have a linked dataset?
If not, adding an else clause [here](https://github.com/openml/openml-python/blob/7c0a77d232a64da76be18bbce76e4bb8aeb4155b/openml/runs/functions.py#L804-L807) here which raises an error is proper and not hacky imo.

janvanrijn:
> To clarify; is there a valid reason that a run on the server should not have a linked dataset?

You mean task. No valid reason, bad design.

ArlindKadra:
> To clarify; is there a valid reason that a run on the server should not have a linked dataset?
> If not, adding an else clause [here](https://github.com/openml/openml-python/blob/7c0a77d232a64da76be18bbce76e4bb8aeb4155b/openml/runs/functions.py#L804-L807) here which raises an error is proper and not hacky imo.

I meant having an else statement and returning the run would be hacky, in case the `dataset_id` is not needed further in the function code. Adding an else and raising a more detailed error is proper.

ArlindKadra:
[...]

PR #717, opened 2019-06-19
Fixing dataset_id referenced before assignment 

Neeratyoy:
#### Reference Issue

Addreses #608.

Adds an else to handle the condition and raises an error.

`import openml`
`run = openml.runs.get_run(186276)`

#### Any other comments?
Enters an else block if input_data key is not present. Error is raised if inside else and `from_server=False`.

### DIFF
--- a/doc/progress.rst
+++ b/doc/progress.rst
@@ -6,6 +6,10 @@
 Changelog
 =========
 
+0.10.0
+~~~~~~
+* FIX #608: Fixing dataset_id referenced before assignment error in get_run function.
+
 0.9.0
 ~~~~~
 * ADD #560: OpenML-Python can now handle regression tasks as well.
--- a/openml/runs/functions.py
+++ b/openml/runs/functions.py
@@ -579,7 +579,7 @@ def get_run(run_id: int, ignore_cache: bool = False) -> OpenMLRun:
 
     try:
         if not ignore_cache:
-            _get_cached_run(run_id)
+            return _get_cached_run(run_id)
         else:
             raise OpenMLCacheException(message='dummy')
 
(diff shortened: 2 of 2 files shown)


Neeratyoy:
> ... should be reported to us (so that we can sanitize the server).

Should the message be to raise an issue [here](https://github.com/openml/openml-python/issues) or report at openml@googlegroups.com?

PGijsbers:
Sorry for not pointing you in the right direction immediately. I think the best is to open an issue at https://github.com/openml/OpenML/issues
@janvanrijn agreed?

Neeratyoy:
> Actually, perhaps just having an issue that we can refer to directly might be best, so people don't open a new issue for every run (and also, so that it's easier for them to check which ids have been reported).

That makes a lot of sense. Do I open an issue and update the URL in the warning message?

PGijsbers:
I got in touch with Jan over this. There seems to have been some miscommunication.
It is not only okay for a run not to reference a dataset, but we do not want to change the instances where this is not present either.

This means we should retrieve the task to determine the dataset id.
I would still only do this if the `run['oml:input_data']['oml:dataset']['oml:did']` data is not provided (i.e. where currently the exception is raised), as this requires an extra server roundtrip.

Sorry for mixed instructions.

Neeratyoy:
To clarify, instead of raising an error and terminating, we shall now fetch the dataset via Run -> Task -> Dataset and populate `run['oml:input_data']` ?

PGijsbers:
> To clarify, instead of raising an error and terminating, we shall now fetch the dataset via Run -> Task -> Dataset 

Yes. The task id is already [parsed earlier in the function](https://github.com/openml/openml-python/blob/develop/openml/runs/functions.py#L633).

> and populate run['oml:input_data']

No. It just needs to be stored in `dataset_id` as it would if it had received a complete xml description.

PR #719, opened 2019-06-21
Added documentation for creating tasks

joaquinvanschoren:
Documents the task upload functionality.


PGijsbers:
Is this PR ready for review?

examples/tasks_tutorial.py
```diff
@@ -142,3 +142,69 @@
 ids = [2, 1891, 31, 9983]
 tasks = openml.tasks.get_tasks(ids)
 pprint(tasks[0])
+
+############################################################################
+# Creating tasks
+# ^^^^^^^^^^^^^^^^^
```
mfeurer: The header formatting should only be as long as the header.

examples/tasks_tutorial.py
```diff
@@ -142,3 +142,69 @@
...
+#
+# You can also create new tasks. Take the following into account:
+#
+# * You can only create tasks on _active_ datasets
+# * For now, only the following tasks are supported: classification, regression,<br>
+# clustering, and learning curve analysis.
+# * For now, tasks can only be created on a single dataset.
+# * The exact same task should not already exist.
```
mfeurer: should -> must ?

examples/tasks_tutorial.py
```diff
@@ -142,3 +142,69 @@
...
 pprint(tasks[0])
+
+############################################################################
+# Creating tasks
+# ^^^^^^^^^^^^^^^^^
+#
+# You can also create new tasks. Take the following into account:
+#
+# * You can only create tasks on _active_ datasets
+# * For now, only the following tasks are supported: classification, regression,<br>
```
mfeurer: Is the `<br>` necessary to produce good layout? I'd prefer to not have html tags in there as the bootstrap layout of the website or the jupyter layout should take care of breaking the lines, right?

examples/tasks_tutorial.py
```diff
@@ -142,3 +142,69 @@
...
+# clustering, and learning curve analysis.
+# * For now, tasks can only be created on a single dataset.
+# * The exact same task should not already exist.
+#
+# Creating a task requires the following input:
+#
+# * task_type_id: The task type ID, required (see below). Required.
+# * dataset_id: The dataset ID. Required.
+# * target_name: For supervised tasks, the name of the attribute you aim to predict.<br>
```
mfeurer: Same here.

The target name can also be used for clustering tasks, right?

examples/tasks_tutorial.py
```diff
@@ -142,3 +142,69 @@
...
+# Optional.
+# * estimation_procedure_id : The ID of the estimation procedure used to create train-test<br>
```
mfeurer: One more html tag.

examples/tasks_tutorial.py
```diff
@@ -142,3 +142,69 @@
...
+    "TaskCreationTestDataset",
+    "test",
+    data_format="arff",
+    url="https://www.openml.org/data/download/61/dataset_61_iris.arff",
+)
+dataset.publish()
+openml.datasets.status_update(dataset.dataset_id, "active")
+
+tasktypes = openml.tasks.TaskTypeEnum
+my_task = openml.tasks.create_task(task_type_id=tasktypes.SUPERVISED_CLASSIFICATION,
```
mfeurer: Could you please format this function call the same way you formatted the creation of the dataset above?

examples/tasks_tutorial.py
```diff
@@ -142,3 +142,69 @@
...
+
+############################################################################
+# Example
+# #######
+#
+# Let's create a classification task on a new dataset. We first upload the dataset
+# (just a copy of the Iris dataset in this example) and then create a new classification
+# task on it. We'll use 10-fold cross-validation (ID=1), without a predefined measure.
+
+dataset = openml.OpenMLDataset(
```
mfeurer: Could you please use the function `openml.datasets.create_dataset`? We kind of advertise to not use the constructors.

examples/tasks_tutorial.py
```diff
@@ -142,3 +142,69 @@
...
+my_task = openml.tasks.create_task(task_type_id=tasktypes.SUPERVISED_CLASSIFICATION,
+                                   dataset_id=dataset.dataset_id,
+                                   target_name="class",
+                                   estimation_procedure_id=1)
+my_task.publish()
+
+############################################################################
+# Complete list of task types:
+
+vars(tasktypes)
```
mfeurer: Could you please refer to for a list of task types as long as they're not nicely visible in the Python API ? https://www.openml.org/search?type=task_type

examples/tasks_tutorial.py
```diff
@@ -142,3 +142,69 @@
...
+
+
+############################################################################
```
mfeurer: Could you maybe create a list of the three input types necessary to create tasks? I guess this would make it more readable.

examples/tasks_tutorial.py
joaquinvanschoren: `openml.datasets.create_dataset` currently doesn't allow upload via URL? Also, it requires 10+ attributes, which would be distracting for this example. I decided to not upload a new dataset, but rather that it will return an error if the tasks exists and how to deal with that.

joaquinvanschoren:
Ehm, tests now fail because`pytest: error: unrecognized arguments: --duration=20`
What's up with that?

Also, when you said "it does not execute on the test server", did you mean that the test fails or it doesn't even get executed? Do I need to register it somewhere?

PGijsbers:
> Ehm, tests now fail becausepytest: error: unrecognized arguments: --duration=20
What's up with that?

This is fixed on develop. It has to do with an upgrade to Pytest 5.0, see #728. I'm not sure if it's better to merge with develop or just cherry-pick f66c27c3bdf8c7d1bb0f4cec271ce87b222cb889 (@mfeurer).

mfeurer:
> I'm not sure if it's better to merge with develop or just cherry-pick

I'd prefer rebase, but merging the development branch into this branch should also be fine.

examples/tasks_tutorial.py
```diff
@@ -177,34 +177,35 @@
...
-# Complete list of task types:
-
-vars(tasktypes)
-
-
-############################################################################
+# Let's create a classification task on a dataset. In this example we will do this on the 
+# Iris dataset (ID=61). We'll use 10-fold cross-validation (ID=1), and predicive accuracy
+# as the predefined measure (as explained above, this can also be left open). 
+# Note that, since it already has this task, this will return an exception, but if
```
mfeurer: > Note that, since it already has this task, this will return an exception

Could you please write this a bit more explicit?

examples/tasks_tutorial.py
```diff
@@ -177,34 +177,35 @@
...
+    my_task.publish()
+except openml.exceptions.OpenMLServerException as e:
+    # Error code for 'task already exists'
+    if e.code == 614:
+        # Lookup task
+        tasks = openml.tasks.list_tasks(data_id=61, output_format='dataframe')
+        tasks = tasks.query("task_type == 'Supervised Classification' " \
+                            "and estimation_procedure == '10-fold Crossvalidation' " \
+                            "and evaluation_measures == 'predictive_accuracy'")
+        task_id = tasks["tid"].values[0]
```
mfeurer: Pandas has a new [to_numpy](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.to_numpy.html) functionality, could you please use that?

examples/tasks_tutorial.py
mfeurer: Looks good.

mfeurer:
Looks good overall, I'd be happy to merge once the tests pass

examples/tasks_tutorial.py
```diff
@@ -134,11 +133,83 @@
...
+# Let's create a classification task on a dataset. In this example we will do this on the
+# Iris dataset (ID=128 (on test server)). We'll use 10-fold cross-validation (ID=1),
+# and _predictive accuracy_ as the predefined measure (this can also be left open).
+# If a task with these parameters exist, we will get an appropriate exception.
+# If such a task doesn't exist, a task will be created and the corresponding task_id
+# will be returned.
+
+
+# using only test server for example uploads
+openml.config.start_using_configuration_for_example()
```
mfeurer: You also need to call the complementary un-use function in the end.

Neeratyoy: Done.

Issue #718, opened 2019-06-20
Issues instantiating/running models from flows


hp2500:
Hi there, 

I am trying to use a flow to instantiate a sklearn model that can then be run on a task. From the documentation it seems like this should be possible in several different ways, but haven't had any success so far trying to get a model from a flow with flow_to_model or from a setup with initialize_model. 
![image](https://user-images.githubusercontent.com/27285925/59817431-cf0ab280-92ed-11e9-81f2-ce8efe853ff2.png)

Running a flow directly on a task doesn't seem to work either.
![image](https://user-images.githubusercontent.com/27285925/59816887-c618e180-92eb-11e9-8be7-b3456a44b38c.png)
Do you have any suggestions?

@amueller

PGijsbers:
Sorry for the delay.
To reproduce the issue, please let us know which flow and/or setup you are working with.

That said, running a flow on a task is easiest done in the following way (e.g. `task 3` and `flow 7684`):

```python
import openml

task = openml.tasks.get_task(3)
flow = openml.flows.get_flow(7684, reinstantiate=True)  # Necessary otherwise no model is made.

run = openml.runs.run_flow_on_task(flow, task)
```
Note the one line in the middle, where the model is reinstantiated after downloading the flow. I think it's fair to say that we should check for this automatically in the `run_flow_on_task` method.

hp2500:
Hi, thanks a lot for your reply! I have tried your code, but it seems to require an older sklearn version. 

![image](https://user-images.githubusercontent.com/27285925/60113096-0e4a6080-973f-11e9-803f-cf5a4c10e4b1.png)

I am using sklearn version 0.21.2, because I am interested in some functions that were added only recently. Do you think there is a workaround? 

P.S. I originally used flow #6970, but I don't think it matters because I am getting the same errors for different flow ids. 

@amueller

hp2500:
But maybe, the required sklearn version depends on which version was used in the original flow... When I try a different flow, the error mentions a different sklearn version. 

![image](https://user-images.githubusercontent.com/27285925/60113614-2bcbfa00-9740-11e9-8975-b9afc814170d.png)

@amueller

PGijsbers:
[...]

PR #722, opened 2019-06-28
Reinstantiate model if needed. Better errors if can't.

PGijsbers:
Addresses some concerns raised in #718. Mainly:
 - If `run_flow_on_task` is called, and the flow has no model, it is now automatically reinstantiated if requirements are satisfied.
 - An error with descriptive message is raised if requirements are not satisfied.

Not addressed:
 - Reinstantiating models if versions mismatch but would otherwise seem possible.

### DIFF
--- a/tests/test_flows/test_flow_functions.py
+++ b/tests/test_flows/test_flow_functions.py
@@ -268,7 +268,11 @@ class TestFlowFunctions(TestBase):
 
     def test_get_flow_reinstantiate_model_no_extension(self):
         # Flow 10 is a WEKA flow
-        self.assertRaises(RuntimeError, openml.flows.get_flow, flow_id=10, reinstantiate=True)
+        self.assertRaisesRegex(RuntimeError,
+                               "No extension could be found for flow 10: weka.SMO",
+                               openml.flows.get_flow,
+                               flow_id=10,
+                               reinstantiate=True)
 
     @unittest.skipIf(LooseVersion(sklearn.__version__) == "0.20.0",
                      reason="No non-0.20 scikit-learn flow known.")
--- a/tests/test_flows/test_flow_functions.py
+++ b/tests/test_flows/test_flow_functions.py
@@ -256,3 +256,22 @@ class TestFlowFunctions(TestBase):
         server_flow = openml.flows.get_flow(flow.flow_id, reinstantiate=True)
         self.assertEqual(server_flow.parameters['categories'], '[[0, 1], [0, 1]]')
         self.assertEqual(server_flow.model.categories, flow.model.categories)
+
+    def test_get_flow_reinstantiate_model(self):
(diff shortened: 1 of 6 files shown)


tests/test_flows/test_flow_functions.py
```diff
@@ -256,3 +256,23 @@ def test_sklearn_to_flow_list_of_lists(self):
         server_flow = openml.flows.get_flow(flow.flow_id, reinstantiate=True)
         self.assertEqual(server_flow.parameters['categories'], '[[0, 1], [0, 1]]')
         self.assertEqual(server_flow.model.categories, flow.model.categories)
+
+    def test_get_flow_reinstantiate_model(self):
```
mfeurer: This appears to be a simple version of `tests.test_flows.test_flows:TestFlow.test_sklearn_to_upload_to_flow`. Could you please check that?

mfeurer: But it might be good to have such a test next to the one below.

tests/test_flows/test_flow_functions.py
```diff
@@ -256,3 +256,23 @@ def test_sklearn_to_flow_list_of_lists(self):
...
+        extension = openml.extensions.get_extension_by_model(model)
+        flow = extension.model_to_flow(model)
+        flow.publish(raise_error_if_exists=False)
+
+        downloaded_flow = openml.flows.get_flow(flow.flow_id, reinstantiate=True)
+        self.assertIsInstance(downloaded_flow.model, sklearn.ensemble.RandomForestClassifier)
+
+    def test_get_flow_reinstantiate_model_no_extension(self):
+        # Flow 10 is a WEKA flow
+        self.assertRaises(RuntimeError, openml.flows.get_flow, flow_id=10, reinstantiate=True)
```
mfeurer: Could you please use the more specific `assertRaisesRegex`? Then you can check for the new error message you added.

tests/test_flows/test_flow_functions.py
PGijsbers: I don't think `tests.test_flows.test_flows:TestFlow.test_sklearn_to_upload_to_flow` ever actually tests if the model is reinstantiated properly. I think the model is ignored by both `assert_flows_equal` and `_to_xml`.
Of course, it is easy to add the check there. But I think it's good to test for this separately from such a complicated flow setup. If this unit test fails, it's very likely due to an error in reinstantiation. If the other test fails, it can be due to any number of reasons, such as serializing subflows, generating xml, etc.

tests/test_flows/test_flow_functions.py
PGijsbers: done

PGijsbers:
Seem to be random errors with `E           openml.exceptions.OpenMLServerException: https://test.openml.org/api/v1/xml/task/ returned code 614: Task already exists.` and trace...

Issue #412, opened 2018-03-04
Short name for flows (pipelines)


joaquinvanschoren:
The flow names are currently very descriptive - which is good - but is often also troublesome to read.

E.g. if I try to plot an overview of sklearn pipelines on a task I get this:
![image](https://user-images.githubusercontent.com/1724503/36951877-ec2e9ec0-2008-11e8-973b-90a96e06029c.png)

Maybe we can implement a simple 'short name' for pipelines? E.g. this:

sklearn.pipeline.Pipeline(imputation=openmlstudy14.preprocessing.ConditionalImputer,hotencoding=sklearn.preprocessing.data.OneHotEncoder,variencethreshold=sklearn.feature_selection.variance_threshold.VarianceThreshold,classifier=sklearn.ensemble.forest.RandomForestClassifier)

could become this (only show the names of the components):
ConditionalImputer-OneHotEncoder-VarianceThreshold-RandomForestClassifier

mfeurer:
You mean using the field [`custom_name`](https://github.com/openml/OpenML/blob/7a1e4cfb96d58c5d20b4438e1c1102024dfd442b/openml_OS/views/pages/api_new/v1/xsd/openml.implementation.upload.xsd#L14)?

PGijsbers:
Looks like we can create a custom name and provide it to the generated flow for `scikit-learn` flows [here](https://github.com/openml/openml-python/blob/develop/openml/extensions/sklearn/extension.py#L418).

joaquinvanschoren:
I was thinking more about a way to shorten the existing names. If there is a custom name, great, but if we have these long automatic names we also need a way to shorten if for proper display.

PGijsbers:
Here's a gist with an implementation: [link](https://gist.github.com/PGijsbers/4940fc6fd7dacfe777f985ab89b22789).
Running it gives the output:
```
Flow name 6651: sklearn.linear_model.logistic.LogisticRegression
Short name: sklearn.LogisticRegression
Flow name 6969: sklearn.pipeline.Pipeline(imputation=openmlstudy14.preprocessing.ConditionalImputer,hotencoding=sklearn.preprocessing.data.OneHotEncoder,variencethreshold=sklearn.feature_selection.variance_threshold.VarianceThreshold,classifier=sklearn.ensemble.forest.RandomForestClassifier)
Short name: sklearn.Pipeline(ConditionalImputer,OneHotEncoder,VarianceThreshold,RandomForestClassifier)
[...]

PR #723, opened 2019-07-02
Function to trim flownames for scikit-learn flows.

PGijsbers:
There is a desire for shorter flow names for scikit-learn flows.
I added a function which uses the flow name string, and trims it down to its major components.
See #412 for more.


PGijsbers:
@mfeurer what's the reason for the dummy learners and/or why they are linked to the scikit-learn extension? 

Currently the generated flow names are not scikit-learn standard names (i.e. they don't start with `sklearn...`). Consequently these long names are not valid input to the `trim_pipeline_name` method. I raise an error for invalid strings currently, but I could also opt to have an optional parameter that specifies to return the full name instead.

openml/extensions/sklearn/extension.py
```diff
@@ -87,6 +87,79 @@ def can_handle_model(cls, model: Any) -> bool:
...
+            svc=sklearn.svm.classes.SVC)
+        ->
+        sklearn.Pipeline(ColumnTransformer,VarianceThreshold,SVC)
+
+        Parameters
+        ----------
+        long_name : str
+        extra_trim_length: int (default=100)
+            If the trimmed name would exceed `extra_trim_length` characters, additional trimming
+            of the long name is performed. This reduces the produced short name length.
```
mfeurer: of the long name -> of the short name

PGijsbers:
This breaks `OpenMLFlow.publish` on a flow which already exists on the server due to an introduced inconsistency where to local flow will have a `custom_name` and the remote flow does not. However, they are enforced to have the same `custom_name`. See below for a failing unit test. What do we do here?
 - Update server flows
 - Don't check for `custom_name`
 - Allow `custom_name` to be unequal only if the server field is `None`

---
The unit test ` test_to_from_filesystem_search` fails here:

```python
    def publish(self, raise_error_if_exists: bool = False) -> 'OpenMLFlow':
        """ Publish this flow to OpenML server.
    
        Raises a PyOpenMLError if the flow exists on the server, but
        `self.flow_id` does not match the server known flow id.
    
        Parameters
        ----------
        raise_error_if_exists : bool, optional (default=False)
            If True, raise PyOpenMLError if the flow exists on the server.
            If False, update the local flow to match the server flow.
    
        Returns
        -------
        self : OpenMLFlow
    
        """
        ... <removed code> ...
        try:
            openml.flows.functions.assert_flows_equal(
                self, flow, flow.upload_date, ignore_parameter_values=True
            )
        except ValueError as e:
            message = e.args[0]
            raise ValueError("Flow was not stored correctly on the server. "
                             "New flow ID is %d. Please check manually and "
                             "remove the flow if necessary! Error is:\n'%s'" %
>                            (flow_id, message))
E           ValueError: Flow was not stored correctly on the server. New flow ID is 22101. Please check manually and remove the flow if necessary! Error is:
E           'Flow sklearn.model_selection._search.GridSearchCV(estimator=sklearn.pipeline.Pipeline(imputer=sklearn.preprocessing.imputation.Imputer,classifier=sklearn.tree.tree.DecisionTreeClassifier)): values for attribute 'custom_name' differ: 'sklearn.GridSearchCV(Pipeline(Imputer,DecisionTreeClassifier))'
E           vs
E           'None'.'
```

This is expected as the existing flow does not have the `custom_name`. Come to think of it, I also think perhaps we should have a different error message there. I propose to change it to:

```
"The flow on the server is inconsistent with the local flow.  The server flow ID is %d. Please check manually and remove the flow if necessary! Error is:\n'%s'" % (flow_id, message))
```

mfeurer:
I think the 3rd option, together with an argument to the assert function to enable this behavior would be best.

And the different error message looks fine.

PGijsbers:
Getting
> 
```================================== FAILURES ===================================
_________ OpenMLTaskTest.test_list_datasets_with_high_size_parameter __________
[gw3] win32 -- Python 3.5.6 c:\miniconda35-x64\python.exe
self = <tests.test_utils.test_utils.OpenMLTaskTest testMethod=test_list_datasets_with_high_size_parameter>
    def test_list_datasets_with_high_size_parameter(self):
        datasets_a = openml.datasets.list_datasets()
        datasets_b = openml.datasets.list_datasets(size=np.inf)
        # note that in the meantime the number of datasets could have increased
        # due to tests that run in parallel.
        # instead of equality of size of list, checking if a valid subset
        a = set(datasets_a.keys())
        b = set(datasets_b.keys())
>       self.assertTrue(b.issubset(a))
E       AssertionError: False is not true
tests\test_utils\test_utils.py:54: AssertionError
```
Maybe restart tests tomorrow

PGijsbers:
Flow 20 seems to have been removed from the test server. @janvanrijn do you know what's up?

openml/extensions/sklearn/extension.py
```diff
@@ -128,19 +136,44 @@ def remove_all_in_parentheses(string: str) -> str:
...
-        if not name.startswith('sklearn'):
-            raise ValueError("Expected 'sklearn' in as start of flow name. "
-                             "Make sure a sklearn-flow is provided.")
-        if 'sklearn.model_selection' in name and not name.startswith('sklearn.model_selection'):
-            raise ValueError("Model Selection is not outer scope. "
-                             "This is unexpected, create a new issue with the flow id.")
-        short_name = 'sklearn.{}'
+        # Generally, we want to trim all hyperparameters, the exception to that is for model
+        # selection, as the `estimator` hyperparameter is very indicative of what is in the flow.
+        # So we first trim pipeline names of the `estimator` parameter. For reference:
```
mfeurer: I'm having trouble to match

> So we first trim pipeline names of the `estimator` parameter.

with the code. The code does not really check for pipelines below, but for model selection. I think the issue is that you do not mean a scikit-learn pipeline object by pipeline, is that correct? Could you update the comment to reflect this more clearly?

tests/test_extensions/test_sklearn_extension/test_sklearn_extension.py
```diff
@@ -1639,4 +1639,22 @@ def test_trim_flow_name(self):
         long_stripped, _ = re.subn(r'\s', '', long)
         self.assertEqual(short, SklearnExtension.trim_flow_name(long_stripped))
 
-        self.assertRaises(ValueError, SklearnExtension.trim_flow_name, "weka.IsolationForest")
+        long = """sklearn.pipeline.Pipeline(
```
mfeurer: Could you please add unit tests for feature union and simple non-pipeline, non-feature selection models here, too?

tests/test_flows/test_flow_functions.py
```diff
@@ -283,9 +283,11 @@ def test_get_flow_reinstantiate_model_no_extension(self):
...
+                     reason="Can't retrieve desired 0.20 run, "
+                            "see https://github.com/openml/OpenML/issues/992")
     def test_get_flow_reinstantiate_model_wrong_version(self):
-        # 20 is scikit-learn ==0.20.0
-        # I can't find a != 0.20 permanent flow on the test server.
-        self.assertRaises(ValueError, openml.flows.get_flow, flow_id=20, reinstantiate=True)
+        openml.config.server = self.production_server
+        _, sklearn_major, _ = LooseVersion(sklearn.__version__).version
+        flow = 8175 if sklearn_major != 19 else 8784
+        self.assertRaises(ValueError, openml.flows.get_flow, flow_id=flow, reinstantiate=True)
```
mfeurer: I know it wasn't there, but could you please change this to an `assertRaisesRegex`?

openml/extensions/sklearn/extension.py
PGijsbers: That's right. I meant any number of components specified under `estimators=` of the model selection. I understand it's confusing and I'll change it.

tests/test_extensions/test_sklearn_extension/test_sklearn_extension.py
PGijsbers: Good catch. FeatureUnion did not work as expected. Added the tests and fixed the issue.

tests/test_flows/test_flow_functions.py
PGijsbers: Done, also found that flow 8175  is `0.19.1` and not `0.19.2` so in terms of CI we can actually use it for all scikit-learn versions.

PGijsbers:
appveyor/branch: `[gw1] FAILED tests/test_utils/test_utils.py::OpenMLTaskTest::test_list_datasets_with_high_size_parameter`

Issue #727, opened 2019-07-05
refuses to `pip install openml`


azrdev:
#### Description
Installing openml [like the guide says](https://openml.github.io/openml-python/master/#how-to-get-openml-for-python)  using `pip install openml` fails deliberatly, and says `pip install -e .` (i.e. on a local checkout) should be used:
>     ERROR: Please install this package with pip: `pip install -e .` Installation requires pip>=10.0.

Apparently this was added to setup.py to prevent setuptools building numpy: https://github.com/openml/openml-python/blob/develop/setup.py#L9..L14
But since pip is able to pull binary numpy from wheels, I suspect this test should not fire if setup.py is called through pip.

#### Versions

pip 19.1.1
openml on pypi is currently 0.9.0

mfeurer:
Thanks for reporting this. We'll have a look into this.

PGijsbers:
I can not reproduce the issue locally. Does the full output of the `pip install openml` command provide more information (and if so, could you provide the full output)?

azrdev:
@PGijsbers full stdout/stderr below. As I wrote, it prints that warning from `setup.py`.

~~~sh
$ pip install openml
Collecting openml
  Using cached https://files.pythonhosted.org/packages/d6/36/4b733c0b76b5e2716d2b281b62bc6e4c8986032e5d00ab14595afd71383f/openml-0.9.0.tar.gz
Collecting liac-arff>=2.4.0 (from openml)
  Using cached https://files.pythonhosted.org/packages/e9/35/fbc9217cfa91d98888b43e1a19c03a50d716108c58494c558c65e308f372/liac-arff-2.4.0.tar.gz
Collecting xmltodict (from openml)
  Using cached https://files.pythonhosted.org/packages/28/fd/30d5c1d3ac29ce229f6bdc40bbc20b28f716e8b363140c26eff19122d8a5/xmltodict-0.12.0-py2.py3-none-any.whl
Collecting requests (from openml)
  Using cached https://files.pythonhosted.org/packages/51/bd/23c926cd341ea6b7dd0b2a00aba99ae0f828be89d72b2190f27c11d4b7fb/requests-2.22.0-py2.py3-none-any.whl
Collecting scikit-learn>=0.18 (from openml)
[...]

PR #731, opened 2019-07-09
Fixing failure of 'python setup.py install'

Neeratyoy:
#### Reference Issue

Addresses #727. 

Fixes a system exit on install.

`git clone https://github.com/openml/openml-python.git`
`python setup.py install`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -10,7 +10,8 @@ with open("openml/__version__.py") as fh:
 # very time consuming anyway.
 if len(sys.argv) > 1 and sys.argv[1] == 'install':
     import pip
-    if pip.__version__ < '10.0.':
+    pip_version = int(pip.__version__.split('.')[0])
+    if pip_version < 10:
         print('Please install this package with pip: `pip install -e .` '
               'Installation requires pip>=10.0.')
         sys.exit(1)
--- a/setup.py
+++ b/setup.py
@@ -9,9 +9,11 @@ with open("openml/__version__.py") as fh:
 # Using Python setup.py install will try to build numpy which is prone to failure and
 # very time consuming anyway.
 if len(sys.argv) > 1 and sys.argv[1] == 'install':
-    print('Please install this package with pip: `pip install -e .` '
-          'Installation requires pip>=10.0.')
-    sys.exit(1)
+    import pip
+    if pip.__version__ < '10.0.':
+        print('Please install this package with pip: `pip install -e .` '
+              'Installation requires pip>=10.0.')
+        sys.exit(1)
 
 if sys.version_info < (3, 5):
     raise ValueError(


mfeurer:
I don't think this PR keeps the original behavior which is only allowing installation via pip. Could you please check that installation via `python setup.py install` does not work, while it works with pip?

Neeratyoy:
I think I understood the issue wrong, in that case.

On my local, `pip install openml` works fine. While a `python setup.py install` after a `git clone` fails.
From #727 and the link he shared, it looked like issue was with the steps described [here](https://openml.github.io/openml-python/master/contributing.html#installation).

Neeratyoy:
@PGijsbers could you please opine on what is the expected behaviour in the installation from source for openml.

Currently following the steps in the [doc](https://openml.github.io/openml-python/master/contributing.html#installation-from-github) do not work owing to [this](https://github.com/openml/openml-python/blob/b0485b0877b84aed5f8f5d1cad7cf3f2284201a1/setup.py#L17).
 Is there a specific reason why we are not allowing `python setup.py install` to work.

Should we remove the documentation instruction then, or do something better than a direct `system.exit()`.
@mfeurer

mfeurer:
We could edit the docs to be `pip install . -e` and remove the if statement?

PGijsbers:
Installing from source should be done using `pip install -e .` not with `python setup.py install`. Looks like I forgot to update the docs 🤦‍♂ 

We moved away from `python setup.py install` because it would start building `numpy`, this is not necessary when installing through `pip`. 

However, it looks like `pip` (in some cases) will install `openml` through calling `python setup.py install`. In this case the user *is* installing `openml` through `pip` but our setup script prevents the user from doing so. So rather than checking if the setup script is called through `python setup.py install`, we should see if `python setup.py install` is called by `pip`. I am not sure how to do that.

pganssle:
> However, it looks like `pip` (in some cases) will install `openml` through calling `python setup.py install`. In this case the user _is_ installing `openml` through `pip` but our setup script prevents the user from doing so. So rather than checking if the setup script is called through `python setup.py install`, we should see if `python setup.py install` is called by `pip`. I am not sure how to do that.

Yes, this is the case - even `setup.py bdist_wheel` might invoke `setup.py install` at some point, because of some implementation details of `bdist_wheel`.

This is one of the reasons we have been unable to just remove the `setup.py install` command, despite the fact that we are actively discouraging it.

I will comment in the issue thread, but I think the best way to fix this would be to stop enforcing the "don't run `setup.py install`" in your `setup.py` script, and possibly raise exceptions if some specific conditions are not met.

mfeurer:
This is superseded by #750.

PR #732, opened 2019-07-10
Changing directories for git status in ci_scripts/test.sh

Neeratyoy:

`ci_scripts/test.sh` tests if all files generated have been deleted. However, the `git status` to check for files happens in another directory. This PR attempts to fix this.

By running unit tests.

### DIFF
--- a/ci_scripts/test.sh
+++ b/ci_scripts/test.sh
@@ -3,6 +3,8 @@ set -e
 # check status and branch before running the unit tests
 before="`git status --porcelain -b`"
 before="$before"
+# storing current working directory
+curr_dir=`pwd`
 
 run_tests() {
     # Get into a temp directory to run test from the installed scikit learn and
@@ -37,6 +39,8 @@ if [[ "$SKIP_TESTS" != "true" ]]; then
     run_tests
 fi
 
+# changing directory to stored working directory
+cd $curr_dir
 # check status and branch after running the unit tests
 # compares with $before to check for remaining files
 after="`git status --porcelain -b`"


Neeratyoy:
@mfeurer the git status directory change seems to be working.
However, the appveyor push fails for errors which I am unable to recreate locally. It all passes.

For `test_upload_task()` in `tests/test_tasks/test_clustering_task.py`:
`openml.exceptions.OpenMLServerException: https://test.openml.org/api/v1/xml/task/ returned code 614: Task already exists.`

For `test_study_attach_illegal()` in `tests/test_study/test_study_functions.py`:
`xml.parsers.expat.ExpatError: not well-formed (invalid token): line 1, column 0`

Neeratyoy:
Is there a specific reason why [this](https://github.com/openml/openml-python/blob/fix_git_status/tests/test_tasks/test_task.py#L44) and [this](https://github.com/openml/openml-python/blob/fix_git_status/tests/test_tasks/test_clustering_task.py#L30) unit test don't have any checks with `assert` and depend only on an error-free run?

Also, there is a [comment](https://github.com/openml/openml-python/blob/fix_git_status/tests/test_tasks/test_task.py#L68) anticipating the error, the test cases currently report.
@mfeurer could you please advise on what is the expectation of the test case and what exactly the `except` tries to address?

From what I understand, the concurrency on travis can cause certain issues. Which I am unable to test on my local. Hence, the confusion.

mfeurer:
> Is there a specific reason why this and this unit test don't have any checks with assert and depend only on an error-free run?

Not really, but uploading them without an error ensures that the task upload functionality works in the first place.

> Also, there is a comment anticipating the error, the test cases currently report.

A task can only exist once. Since uploading a dataset and waiting until the server is ready to create a task on it takes too long we need some strategy to figure out what task to create. Therefore, we try to create a task for a random dataset with a random target and hope that it doesn't exist yet. If it does, we just try again until we either succeed or fail with a different error. If you think the comment can be improved, please go for it.

> From what I understand, the concurrency on travis can cause certain issues. Which I am unable to test on my local. Hence, the confusion.

Only for the clustering test, because there's no try/catch which re-tries the unit test. Could you please add that (in a new pull request)?

PR #735, opened 2019-07-11
Delete files uploaded to test server during unit testing

Neeratyoy:
#### Reference Issue

Addresses #261. 

Keeps track of entities uploaded to test server and iteratively deletes them after the completion of all unit tests.

By running `pytest -s` and observing the logs.


Neeratyoy:
@mfeurer 
A few things I need advise on.

* The following IDs refuse to be deleted:
    _flow is in use by other content (it is a subflow). Can not be deleted_
    * flow, [2146](https://test.openml.org/api/v1/xml/flow/2146)
    * flow, [2378](https://test.openml.org/api/v1/xml/flow/2378)

    _Can only delete studies based on runs or studies in preparation_
    * study, [307](https://test.openml.org/api/v1/xml/study/307) (unit test benchmark suite)
* Currently the only way to manually check whether the deletion works is to capture the logs, wherein I have a bunch of print statements. However, though structured and hence visually discernible, but they will evidently populate the logs a lot lot more.

openml/testing.py
```diff
@@ -28,6 +28,9 @@ class TestBase(unittest.TestCase):
     Currently hard-codes a read-write key.
     Hopefully soon allows using a test server, not the production server.
     """
+    tracker = {}  # type: dict
```
mfeurer: Could you please make the name more descriptive and add a comment here?

openml/testing.py
```diff
@@ -28,6 +28,9 @@ class TestBase(unittest.TestCase):
...
+    test_server = None
```
mfeurer: Since this is hardcoded below, you could move the hardcoded variable assignment here and remove the line below?

openml/testing.py
```diff
@@ -135,17 +172,53 @@ def _cleanup_fixture(self):
...
         new_file_list = list(set(new_file_list) - set(old_file_list))
-        print("Files to delete in local: {}".format(new_file_list))
         for file in new_file_list:
             os.remove(file)
 
+        # Test server deletion
+        openml.config.server = TestBase.test_server
+        openml.config.apikey = TestBase.apikey
+        entity_types = list(TestBase.tracker.keys())
+        # deleting 'run' first to allow other dependent entities to be deleted
```
mfeurer: You could create an order that is always followed and just iterate over it. For example
```
for entity_type in ['study', 'suite', 'run', 'flow', 'task', 'dataset']:
    do
```

openml/testing.py
```diff
@@ -135,17 +172,53 @@ def _cleanup_fixture(self):
...
+
+        # cloning file tracker to allow deletion of entries of deleted files
+        tracker = TestBase.tracker.copy()
+        # reordering to delete sub flows later
+        if 'flow' in entity_types:
+            flows = {}
+            for entity_id, entity_name in tracker['flow']:
+                flows[entity_name] = entity_id
+            # reordering flows in descending order of their flow name lengths
+            flow_deletion_order = [flows[name] for name in sorted(list(flows.keys()),
```
mfeurer: How about 

```
flow_deletion_order = [entity_id for entity_id, _ in sorted(tracker['flow'], key=lambda x: len(x[1]), reverse=True)]
```
Saves a few lines and reads a bit easier (in my opinion).

openml/testing.py
```diff
@@ -104,9 +112,39 @@ def tearDown(self):
...
+        else:
+            TestBase.tracker[entity_type].append(entity_id)
+
+    @classmethod
+    def _delete_entity_from_tracker(self, entity_type, entity):
+        if entity_type in TestBase.tracker:
+            # delete_index handles duplicate entries
+            delete_index = []
+            for i, element in enumerate(TestBase.tracker[entity_type]):
+                if entity_type == 'flow':
```
mfeurer: Could you try to remove this separate treatment of flow?

openml/testing.py
```diff
@@ -104,9 +112,39 @@ def tearDown(self):
...
+                    id, name = element
```
mfeurer: Please do not use the variable `id`, but use something like `id_`.

openml/testing.py
```diff
@@ -104,9 +112,39 @@ def tearDown(self):
...
+                else:
+                    id = element
+                if id == entity:
+                    delete_index.append(i)
+            TestBase.tracker[entity_type] = [TestBase.tracker[entity_type][index]
```
mfeurer: This appears to be overly complex. Can't you simply do list.remove(i)?

openml/testing.py
```diff
@@ -104,9 +112,39 @@ def tearDown(self):
         openml.config.server = self.production_server
         openml.config.connection_n_retries = self.connection_n_retries
 
+    @classmethod
```
mfeurer: I'm wondering whether you need to add some synchronization as the tests are executed in parallel?

tests/test_datasets/test_dataset_functions.py
```diff
@@ -478,6 +478,8 @@ def test_publish_dataset(self):
             data_file=file_path,
         )
         dataset.publish()
+        TestBase._track_test_server_dumps('data', dataset.dataset_id)
+        print("\ncollected from {}: {}".format(__file__.split('/')[-1], dataset.dataset_id))
```
mfeurer: Could you please use a logger instead of printing?

tests/test_tasks/test_clustering_task.py
```diff
@@ -43,4 +43,5 @@ def test_upload_task(self):
         )
 
         task_id = task.publish()
+        # not tracking upload for delete since _delete_entity called end of function
```
mfeurer: Hm, could you change the behavior to be consistent?

mfeurer:
> The following IDs refuse to be deleted: flow is in use by other content (it is a subflow). Can not be deleted

I'm afraid that you need to recursively add all subflows to the tracker to make sure there are no leftovers. 

> Can only delete studies based on runs or studies in preparation study, 307 (unit test benchmark suite)

@janvanrijn is this intended behavior?

> Currently the only way to manually check whether the deletion works is to capture the logs, wherein I have a bunch of print statements. However, though structured and hence visually discernible, but they will evidently populate the logs a lot lot more.

That's fine by me.

openml/testing.py
Neeratyoy: I initially thought about it. However, felt that not having an independent list of hard-coded entities can better protect against future addition of entities.

Having said that, the API call functions do check against a hard-coded list of `entity_types`.

openml/testing.py
Neeratyoy: I agree completely. It became quite unnecessarily complex. Since for flows I am maintaining records as list of (ID, name) whereas for other entity types it is just a list of IDs. 
Tried to keep the code as lean as possible.
I'll try another design and iterate once more.

openml/testing.py
Neeratyoy: This function just adds to a dictionary and even concurrent executions will `append` a different ID, will it not? Are there any obvious edge case that can break this design? Also, does the parallel distribution (on Travis) create multiple copies of a static variable shared by the class definition?

tests/test_tasks/test_clustering_task.py
Neeratyoy: Alright, I shall then remove the `_delete_entity()` call that existed from before.

openml/testing.py
Neeratyoy: sure, will do so. didn't want to tinker too much with existing code.

openml/testing.py
Neeratyoy: This was a lot neater. Thanks a lot!

openml/testing.py
Neeratyoy: Have simplified this section a bit.
However, couldn't quite work out how to avoid the if-condition without changing how the file IDs are being tracked or recorded. That is, making the list of integers a list of a single-element tuple. Which I felt was more counter-intuitive and difficult to maintain/debug.

openml/testing.py
```diff
@@ -104,9 +114,43 @@ def tearDown(self):
...
+    def _track_test_server_dumps(self, entity_type, entity_id):
```
mfeurer: Could you please rename that function to something like `mark_entity_for_removal`?

openml/testing.py
mfeurer: Could you please add a comment to explain what the tracker is used for?

Neeratyoy:
@mfeurer I tried to debug the clustering task failure. Here are my observations.

The task being uploaded has the following attributes, as initialized by the `setUp` of class `OpenMLClusteringTaskTest`:
* `task_type_id` = 5
* `estimation_procedure` = 17
* `dataset_id` as generated by `_get_compatible_rand_dataset()`

This is attempted 100 times until the `publish()` of the task succeeds. Failure of which gives the current ValueError of _Could not create a valid task for task type ID 5_.

On the test server, the following commands:
`l = openml.tasks.list_tasks(task_type_id=5, output_format='dataframe')`
`print(set(l.did))`
result in `{2, 10, 11, 14, 22, 24, 25, 35, 36, 37, 42, 91, 92}`

If we record all the 100 `dataset_id` generated by `_get_compatible_rand_dataset()` during the test case publish attempts, that is exactly `{2, 10, 11, 14, 22, 24, 25, 35, 36, 37, 42, 91, 92}` again. 
This explains why we cannot upload a task in even 100 trials.


Using the apikey we define in class `TestBase`, I explicitly deleted one of the tasks linked with `dataset_id = 22`, and then ran the unit test. It passed as expected since the task with those attributes didn't exist. 

Are all these expected behaviour?

mfeurer:
> Are all these expected behaviour?

No, it turns out that `_get_compatible_rand_dataset` only takes classification and regression into account. Clustering tasks could be created for any dataset and therefore, the tests should not fail. Could you please extend that function?

Neeratyoy:
@mfeurer 

Most test case issue seems to be fixed barring one test which fails consistently.
Could you please explain what this does: https://github.com/openml/openml-python/blob/develop/tests/test_flows/test_flow_functions.py#L277

My local sklearn version is `0.20.2` while `flow_id=20` returns `code 181: unknown flow` from the test server. Please advise on how I can solve this.

Neeratyoy:
@mfeurer just another tiny thing.
I should be adding instructions to collect files being uploaded when adding a unit test. 
Should I be editing the [contributing.rst](https://github.com/openml/openml-python/blob/develop/doc/contributing.rst) or [contributing.md](https://github.com/openml/OpenML/blob/master/CONTRIBUTING.md) for that? Where would you like that information to be included?

mfeurer:
> Should I be editing the contributing.rst or contributing.md for that?

Both

> Where would you like that information to be included?

Actually, I think I meant this file: https://raw.githubusercontent.com/openml/openml-python/develop/PULL_REQUEST_TEMPLATE.md

Neeratyoy:
I didn't edit `contributing.rst` since that page [links](https://openml.github.io/openml-python/develop/contributing.html#contributing-code) to `contributing.md`. I also edited `pull_request_template.md` with brief instructions on the same.

mfeurer:
Okay, let me know when I should have a look at this again.

PR #738, opened 2019-07-16
Handling blank evaluations in object summary output

Neeratyoy:
Currently, if an `OpenMLRun` object hasn't got `task_evaluation_measure` and `evaluation` defined, the `__str__` call throws an error. 
This PR handles this None condition and omits the printing of that field.

Trying the [flow_and_runs_tutorial](https://github.com/openml/openml-python/blob/develop/examples/flows_and_runs_tutorial.py#L59) and printing the run object created throws the error. This PR should fix that.


Neeratyoy:
@mfeurer the task used in the example has `estimation_procedure` defined but no `evaluation_measure`. Assuming that is why the run has `fold_evaluations` while `evaluations` is __None__.

mfeurer:
This looks good and I'd be happy to accept, but I'm surprised by the high failure rate of the unit tests.

PR #740, opened 2019-07-19
Update documentation

mfeurer:
This pull request does

* dynamically generate the copyright year
* replaces `pprint` by `print`
* updates one example to use a more common task on the test server
* fix a few docstring syntax issues
* introduce stricter checks for correct XML files (I had old files on my machine, and the error message were not helpful).

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -21,16 +21,12 @@ Example
 .. code:: python
 
     import openml
-    from sklearn import preprocessing, tree, pipeline
-
-    # Set the OpenML API Key which is required to upload your runs.
-    # You can get your own API by signing up to OpenML.org.
-    openml.config.apikey = 'ABC'
+    from sklearn import impute, tree, pipeline
 
     # Define a scikit-learn classifier or pipeline
     clf = pipeline.Pipeline(
         steps=[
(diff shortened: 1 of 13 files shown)


doc/index.rst
```diff
@@ -39,10 +35,13 @@ Example
     task = openml.tasks.get_task(31)
     # Run the scikit-learn model on the task.
     run = openml.runs.run_model_on_task(clf, task)
-    # Publish the experiment on OpenML (optional, requires an API key).
+    # Publish the experiment on OpenML (optional, requires an API key.
```
Neeratyoy: Should we also include in comments that they can use `openml.config.start_using_configuration_for_example()` for trials?
Unless this is to encourage signing in and creating an API.

Neeratyoy:
Looks like a certain `test__get_cached_dataset` unit test is failing since the new _nominal values_ check in `openml/datasets/data_feature.py` is raising an error.

doc/index.rst
mfeurer: Signing up and using the own API key is definitively the better option. Also, this is the front page, so it should be as simple as possible.

PR #747, opened 2019-07-24
Add #737

sahithyaravi1493:
add a list_evaluations_setups call that merges list_evaluations with list_setups and returns a list of evaluations with hyperparameters.

evaluations.list_evaluations_setups(function=function,flow=[flow_id], sort_order=sort_order,size=size, output_format='dataframe')

#### Any other comments?
list_evaluations has been modified. This is because the existing conversion to dataframe format does not preserve  row order (even if ordered dict is passed - bug in pandas). So dataframe conversion is now performed by using a loop over all rows and collecting them in a list.


openml/evaluations/functions.py
```diff
@@ -238,3 +240,94 @@ def list_evaluation_measures() -> List[str]:
...
+        uploader: Optional[List] = None,
+        tag: Optional[str] = None,
+        per_fold: Optional[bool] = None,
+        sort_order: Optional[str] = None,
+        output_format: str = 'dataframe'
+) -> Union[Dict, pd.DataFrame]:
+    """
+    List all run-evaluation pairs matching all of the given filters
+    and their hyperparameter settings.
+
```
PGijsbers: Please either provide descriptions for each parameter explicitly (also specifying if e.g. the expected list of flows is `List[int]` or `List[OpenMLFlow]` or any, or provide a reference to `list_evaluations` in the description.

tests/test_flows/test_flow_functions.py
```diff
@@ -286,6 +286,11 @@ def test_get_flow_reinstantiate_model_no_extension(self):
     @unittest.skipIf(LooseVersion(sklearn.__version__) == "0.19.1",
                      reason="Target flow is from sklearn 0.19.1")
     def test_get_flow_reinstantiate_model_wrong_version(self):
+        openml.config.server = self.production_server
```
PGijsbers: This is code was re-introduced in your merge, please remove it.

tests/test_evaluations/test_evaluation_functions.py
```diff
@@ -142,3 +142,57 @@ def test_list_evaluation_measures(self):
         measures = openml.evaluations.list_evaluation_measures()
         self.assertEqual(isinstance(measures, list), True)
         self.assertEqual(all([isinstance(s, str) for s in measures]), True)
+
+    def test_list_evaluations_setups_filter_flow(self):
+        openml.config.server = self.production_server
+        flow_id = 405
+        size = 100
+        evals_setups = openml.evaluations.list_evaluations_setups("predictive_accuracy",
```
PGijsbers: I know we're generally not very good at code re-use, particularly in unit tests, but I'd rather change that :) Please extract the common code between `test_list_evaluations_setups_filter_flow` and `test_list_evaluations_setups_filter_task` into one function that is called from both tests (e.g. specifying either `flow=[flow_id]` or `task=[task_id]` through `**kwargs`).

tests/test_evaluations/test_evaluation_functions.py
```diff
@@ -142,3 +142,57 @@ def test_list_evaluation_measures(self):
...
+                                                                  output_format='dataframe')
+        evals = openml.evaluations.list_evaluations("predictive_accuracy",
+                                                    flow=[flow_id], size=size,
+                                                    sort_order='desc',
+                                                    output_format='dataframe')
+
+        # Check if list is non-empty
+        self.assertGreater(len(evals_setups), 0)
+        # Check if output from sort is sorted in the right order
+        self.assertTrue(sorted(list(evals_setups['value'].values), reverse=True)
```
PGijsbers: Please use `assertEqual` for equality checks.

tests/test_evaluations/test_evaluation_functions.py
```diff
@@ -142,3 +142,57 @@ def test_list_evaluation_measures(self):
...
+                        == list(evals_setups['value'].values))
+        # Check if output and order of list_evaluations is preserved
+        self.assertTrue((evals_setups['run_id'].values == evals['run_id'].values).all())
```
PGijsbers: Please use `assertSequenceEquals`, but you will also need to first use `tolist`

sahithyaravi1493:
> ted. I think we could look into adding the hyperparameters as columns if a flow filter is applied, but perhaps that's something for a different PR?

A new column to the list_evaluations call?  I am not sure if that is needed now that we have this new call, and I believe it may be better to do it in a different PR.

PGijsbers:
To the returned dataframe of `list_evaluations_setups`. If a single flow is selected, the hyperparameters are always the same, so we can have a dataframe which specifies each hyperparameter value by column, rather than one column which contains a dictionary of parameter values. But I'll ignore the idea for this PR.

PR #748, opened 2019-07-24
Making unit tests green

Neeratyoy:
Owing to the concurrent runs, combined with deletion of files at the end of unit test modules, some tests may fail. Rerunning them should ensure they pass.


tests/test_datasets/test_dataset_functions.py
```diff
@@ -1162,6 +1162,8 @@ def test___publish_fetch_ignore_attribute(self):
         self.__class__.test_publish_fetch_ignore_attribute_did = upload_did
         self.__class__.test_publish_fetch_ignore_attribute_list = ignore_attribute
 
+    # with concurrent runs this function maybe called before test__publish_fetch_ignore_attribute
+    @pytest.mark.flaky(reruns=5)
```
mfeurer: You could instead add a wait loop here to make sure this one doesn't fail due to a dataset not yet processed (like for example this one: https://github.com/openml/openml-python/blob/develop/tests/test_runs/test_run_functions.py#L67)

Neeratyoy: Is it okay if in the current while loop implementation I add a `time.sleep(10)` before the `continue`?

Neeratyoy:
@mfeurer, @PGijsbers 
If I would like to import a new package for the test cases, isn't it enough to add to `ci_scripts/install.sh` [here](https://github.com/openml/openml-python/blob/c452acdba230d7d88edbceffd321913172094708/ci_scripts/install.sh#L57).

It looks like the appveyor build failed to install it. How can I fix that?

mfeurer:
Appveyor doesn't use `install.sh`. However, I doubt that posix-ipc will work on Windows, see for example the description of the project [here](https://pypi.org/project/posix_ipc/). Probably we should discuss this offline on Monday.

PGijsbers:
I hope you can figure something out. But to answer the original question, if you want to add test dependencies, it's best to add it in [`setup.py`](https://github.com/openml/openml-python/blob/develop/setup.py#L42).

Neeratyoy:
@mfeurer 
if we delete all the files for each of the workers independently and asynchronously, `test_list_datasets_with_high_size_parameter` fails.

mfeurer:
That's unfortunate. Maybe we should execute that test on the live server?

Neeratyoy:
> That's unfortunate. Maybe we should execute that test on the live server?

I guess you mean a new instance of an OpenML server every test run? 
Will that interfere with the fetch requests to the test server, that is, by having a new `openml.config.server` value.

mfeurer:
Sorry, I don't get that. What I meant is like https://github.com/openml/openml-python/blob/develop/tests/test_datasets/test_dataset_functions.py#L213

Neeratyoy:
aah you are okay with the dataset list being fetched from the prod server
alright, I was thinking we shouldn't be touching that during test but this is a read-only operation
makes sense
thanks

PGijsbers:
We do it at several points in our unit tests. But your general intuition is right, only do it if you have to. In particular never alter the production server and think twice if you really need the production server if it's a demanding query. In this case for the reasons you pointed out it makes sense.

Neeratyoy:
@mfeurer 
all passed, except one unit test that gave: 
`openml.exceptions.OpenMLServerException: An Elastic Search Exception occured`

Neeratyoy:
Thanks!

openml/testing.py
```diff
@@ -35,11 +34,11 @@ class TestBase(unittest.TestCase):
...
-    # creating logger for unit test file deletion status
-    logger = logging.getLogger("unit_tests")
-    logger.setLevel(logging.INFO)
-    fh = logging.FileHandler('TestBase.log')
-    fh.setLevel(logging.INFO)
+    # creating logger for tracking files uploaded to test server
+    logger = logging.getLogger("unit_tests_publish")
+    logger.setLevel(logging.DEBUG)
+    fh = logging.FileHandler('UploadedFiles.log')
+    # fh.setLevel(logging.DEBUG)
```
mfeurer: Is it actually necessary to introduce a logger which logs to a file? Also, could you please remove that comment?

tests/conftest.py
```diff
@@ -0,0 +1,187 @@
...
+
+import os
+import logging
+from typing import List
+
+import openml
+from openml.testing import TestBase
+
+# creating logger for unit test file deletion status
+logger = logging.getLogger("unit_tests")
```
mfeurer: This looks a lot like a copy of the logger above. Also, I assume that the logger names should be swapped.

tests/conftest.py
```diff
@@ -0,0 +1,187 @@
...
+logger.setLevel(logging.DEBUG)
+fh = logging.FileHandler('UnitTestDeletion.log')
```
mfeurer: Same as above, do we really need to save this to disk?

tests/conftest.py
```diff
@@ -0,0 +1,187 @@
...
+def compare_delete_files(old_list, new_list) -> None:
+    '''Deletes files that are there in the new_list but not in the old_list
+
+    :param old_list: List[str]
+    :param new_list: List[str]
+    :return: None
+    '''
+    file_list = list(set(new_list) - set(old_list))
+    for file in file_list:
+        os.remove(file)
```
mfeurer: I'm not sure if that's desired behavior. The unit tests shouldn't leave any new files. Do you know the reason why there are some files left over?

tests/conftest.py
```diff
@@ -0,0 +1,187 @@
...
+    Iteratively, multiple POST requests are made to the OpenML test server using
+    openml.utils._delete_entity() to remove the entities uploaded by all the unit tests.
+
+    :param tracker: Dict
+    :return: None
+    '''
+    openml.config.server = TestBase.test_server
+    openml.config.apikey = TestBase.apikey
+
+    # legal_entities defined in openml.utils._delete_entity() - {'user'}
```
mfeurer: Are these comments still needed?

tests/conftest.py
Neeratyoy: This goes back to the original issue we were trying to address. Some files remained on git status after the unit tests were run.
If I remember correctly, some of the tests (especially in the task module) were downloading data, which were shared across unit tests. Also, the tearDown is not defined for a few of these test modules. Whereas the design of the tearDown would I believe demand the manual bookkeeping of all IDs downloaded. 
All in all, it looked like a more robust check to collect and delete all generated files from the 'master' worker.

tests/conftest.py
Neeratyoy: I agree, no.
I was tailing during the testing and this helped.
Will have them removed.

PR #755, opened 2019-08-01
Remove Py3.4 trove classifier, add python_requires

### LINKED ISSUE #741: Python version requirements inconsistent
I noticed in your `setup.py` file that you have [a requirement for Python version >= 3.5](https://github.com/openml/openml-python/blob/56fcc00cb16b102e4778c1f1dadd7d9919e9eca4/setup.py#L16), but you do not have a [`python_requires` keyword](https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires), and you have the [Python 3.4 in your trove classifiers](https://github.com/openml/openml-python/blob/56fcc00cb16b102e4778c1f1dadd7d9919e9eca4/setup.py#L82).

I suggest adding `python_requires=">= 3.5"` to your `setup()` invocation and removing Python 3.4 from your trove classifiers.

(Adding this because it's relevant for `python_requires`, not because I have any evidence you don't already know it). When uploading your distributions, make sure to use [`twine`](https://pypi.org/project/twine/), instead of `setup.py upload`, since the latter is deprecated and some versions of `setuptools` have bugs that will cause problems for some users.

PGijsbers:
Fixes #741


mfeurer:
There's one test failing due to the test `test_publish_fetch_ignore_attribute_did` not being found.

setup.py
```diff
@@ -28,6 +28,7 @@
                  version=version,
                  packages=setuptools.find_packages(),
                  package_data={'': ['*.txt', '*.md']},
+                 python_requires=">= 3.5",
```
mfeurer: I think the space before the 3.5 shouldn't be there.

PGijsbers: My bad, fixed.

Issue #175, opened 2016-09-09
Extract parameter and model description from scikit-learn docstrings


mfeurer:
and store in the description of the flow or the description of the parameters.

PGijsbers:
As far as I can tell this is not resolved yet, but most work has been done (?).
Parameter meta information can be passed to OpenMLFlow, so I think the goal is to pass this information correctly when creating a flow from a sklearn model. To me it looks like the right place to start this is [here](https://github.com/openml/openml-python/blob/develop/openml/flows/sklearn_converter.py#L360).

Neeratyoy:
@mfeurer it seems that if the str length of a flow description is > 1024, we get an `OpenMLServerException: Problem validating uploaded description file`
Is that expected?

Steps to recreate:
`task = openml.tasks.get_task(403)`
`clf = sklearn.linear_model.SGDClassifier()`
` dummy_str = '-' * 1024`
`f.description = dummy_str`
`f = f.publish()    # this works`
`openml.utils._delete_entity('flow', f.flow_id)`
` dummy_str = '-' * 1025`
`f.description = dummy_str`
`f = f.publish()    # this does not work`

Now w.r.t. this feature request, will not the flow description become excessively large given that sklearn has a good 2-3 line description for each parameter (roughly on an average). This would also entail changing the `__repr__` for OpenMLFlow.

What do you suggest?

mfeurer:
> Is that expected?
Unfortunately, yes.

I'd suggest to store the first 1024 characters of the docstring on OpenML (but always stopping before the "Read more here" or "Parameters"). I don't see a real issue with __repr__ for now. We can change that if this becomes a real issue for users?

mfeurer:
Added via #756

PR #756, opened 2019-08-06
Adding sklearn docstring to flow

Neeratyoy:
#### Reference Issue

Addresses #175.  

Adds the sklearn model description to the OpenML Flow description for a flow that uses a sklearn model. Also parses the docstring and adds the data types and parameter descriptions for the `parameters_meta_info` attribute of a flow.

`import openml`
`from sklearn.linear_model import SGDClassifier`
`openml.config.start_using_configuration_for_example()`
` task = openml.tasks.get_task(403)`
`clf = SGDClassifier()`
` r, f = openml.runs.run_model_on_task(clf, task, return_flow=True)`
` print(f.description)`
`print(f.parameters_meta_info['learning_rate']['data_type'])`
`print(f.parameters_meta_info['learning_rate']['description'])`

### DIFF
--- a/openml/extensions/sklearn/extension.py
+++ b/openml/extensions/sklearn/extension.py
@@ -480,31 +480,48 @@ class SklearnExtension(Extension):
     def _get_sklearn_description(self, model: Any, char_lim: int = 1024) -> str:
         '''Fetches the sklearn function docstring for the flow description
 
+        Retrieves the sklearn docstring available and does the following:
+        * If length of docstring <= char_lim, then returns the complete docstring
+        * Else, trims the docstring till it encounters a 'Read more in the :ref:'
+        * Or till it encounters a 'Parameters\n----------\n'
+        The final string returned is at most of length char_lim with leading and
+        trailing whitespaces removed.
+
         Parameters
         ----------
-        model: The sklearn model object
-        char_lim: int, specifying the max length of the returned string
+        model : sklearn model
+        char_lim : int
+            Specifying the max length of the returned string
             OpenML servers have a constraint of 1024 characters for the 'description' field.
 
         Returns
         -------
-        string of length <= char_lim
+        str
         '''
         def match_format(s):
             return "{}\n{}\n".format(s, len(s) * '-')
         s = inspect.getdoc(model)
         if len(s) <= char_lim:
-            return s
+            # if the fetched docstring is smaller than char_lim, no trimming required
+            return s.strip()
         try:
-            pattern = "Read more in the :ref:"  # "Parameters"
+            # trim till 'Read more'
+            pattern = "Read more in the :ref:"
             index = s.index(pattern)
         except ValueError:
+            pass
+        try:
+            # if 'Read more' doesn't exist, trim till 'Parameters'
             pattern = "Parameters"
             index = s.index(match_format(pattern))
+        except ValueError:
+            # returning full docstring
+            index = len(s)
         s = s[:index]
+        # trimming docstring to be within char_lim
         if len(s) > char_lim:
             s = "{}...".format(s[:char_lim - 3])
-        return s
+        return s.strip()
 
     def _serialize_model(self, model: Any) -> OpenMLFlow:
         """Create an OpenMLFlow.
@@ -634,38 +651,69 @@ class SklearnExtension(Extension):
                 known_sub_components.add(visitee.name)
                 to_visit_stack.extend(visitee.components.values())
 
-    def _extract_sklearn_parameter_docstring(self, model):
+    def _extract_sklearn_parameter_docstring(self, model) -> Union[None, str]:
+        '''Extracts the part of sklearn docstring containing parameter information
+
+        Fetches the entire docstring and trims just the Parameter section.
+        The assumption is that 'Parameters' is the first section in sklearn docstrings,
+        followed by other sections titled 'Attributes', 'See also', 'Note', 'References',
+        appearing in that order if defined.
+        Returns a None if no section with 'Parameters' can be found in the docstring.
+
+        Parameters
+        ----------
+        model : sklearn model
+
+        Returns
+        -------
+        str, or None
+        '''
         def match_format(s):
             return "{}\n{}\n".format(s, len(s) * '-')
         s = inspect.getdoc(model)
-        s1 = "Parameters"
-        s2 = ["Attributes", "See also", "Note", "References"]
         try:
-            index1 = s.index(match_format(s1))
+            index1 = s.index(match_format("Parameters"))
         except ValueError as e:
-            print("Parameter {}".format(e))
-            # returns the whole sklearn docstring available
-            return s
-        for h in s2:
+            # when sklearn docstring has no 'Parameters' section
+            print("{} {}".format(match_format("Parameters"), e))
+            return None
+
+        headings = ["Attributes", "See also", "Note", "References"]
+        for h in headings:
             try:
+                # to find end of Parameters section
                 index2 = s.index(match_format(h))
                 break
             except ValueError:
                 print("{} not available in docstring".format(h))
                 continue
         else:
-            # in the case only 'Parameters' exist
+            # in the case only 'Parameters' exist, trim till end of docstring
             index2 = len(s)
         s = s[index1:index2]
-        return s
+        return s.strip()
 
-    def _extract_sklearn_param_info(self, model):
+    def _extract_sklearn_param_info(self, model) -> Union[None, Dict]:
+        '''Parses parameter type and description from sklearn dosctring
+
+        Parameters
+        ----------
+        model : sklearn model
+
+        Returns
+        -------
+        Dict, or None
+        '''
         docstring = self._extract_sklearn_parameter_docstring(model)
+        if docstring is None:
+            # when sklearn docstring has no 'Parameters' section
+            return None
+
         n = re.compile("[.]*\n", flags=IGNORECASE)
         lines = n.split(docstring)
         p = re.compile("[a-z0-9_ ]+ : [a-z0-9_]+[a-z0-9_ ]*", flags=IGNORECASE)
-        parameter_docs = OrderedDict()
-        description = []
+        parameter_docs = OrderedDict()  # type: Dict
+        description = []  # type: List
 
         # collecting parameters and their descriptions
         for i, s in enumerate(lines):
@@ -681,7 +729,6 @@ class SklearnExtension(Extension):
 
         # collecting parameters and their types
         matches = p.findall(docstring)
-        parameter_docs = OrderedDict()
         for i, param in enumerate(matches):
             key, value = param.split(':')
             parameter_docs[key.strip()] = [value.strip(), description[i]]
@@ -830,9 +877,12 @@ class SklearnExtension(Extension):
                 else:
                     parameters[k] = None
 
-            data_type, description = parameters_docs[k]
-            parameters_meta_info[k] = OrderedDict((('description', description),
-                                                   ('data_type', data_type)))
+            if parameters_docs is not None:
+                data_type, description = parameters_docs[k]
+                parameters_meta_info[k] = OrderedDict((('description', description),
+                                                       ('data_type', data_type)))
+            else:
+                parameters_meta_info[k] = OrderedDict((('description', None), ('data_type', None)))
 
         return parameters, parameters_meta_info, sub_components, sub_components_explicit
 
--- a/openml/extensions/sklearn/extension.py
+++ b/openml/extensions/sklearn/extension.py
@@ -556,7 +556,7 @@ class SklearnExtension(Extension):
             index1 = s.index(match_format("Parameters"))
         except ValueError as e:
             # when sklearn docstring has no 'Parameters' section
-            logging.info("{} {}".format(match_format("Parameters"), e))
+            logging.warning("{} {}".format(match_format("Parameters"), e))
             return None
 
         headings = ["Attributes", "Notes", "See also", "Note", "References"]
@@ -566,7 +566,7 @@ class SklearnExtension(Extension):
                 index2 = s.index(match_format(h))
                 break
             except ValueError:
-                logging.info("{} not available in docstring".format(h))
+                logging.warning("{} not available in docstring".format(h))
                 continue
         else:
             # in the case only 'Parameters' exist, trim till end of docstring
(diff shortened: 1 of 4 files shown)


Neeratyoy:
@mfeurer 
Hope this is in line with the requirement of the issue.

A couple of things though:

* In the example given above, when I go to the test server page for the corresponding `flow_id` generated after `run_model_on_task`, the descriptions loads and then throws a 'Not Found` error. However, the parameter descriptions come up fine.

* Currently, I am updating the description for all sklearn flows. However, I am wondering where will this break given most of the popular sklearn models are existing as flows. I have noticed that if I set `flow_id=None` and `publish()` a flow which exists, it gives a `ValueError` for the descriptions and parameter descriptions too.

mfeurer:
>  the descriptions loads and then throws a 'Not Found` error

This is a frontend error on the test server and nothing to worry about here.

> Currently, I am updating the description for all sklearn flows

Do you mean on the test or live server? If yes, I don't think there's a reason to do so. In my opinion it's sufficient if this information is added in the future.

openml/extensions/sklearn/extension.py
```diff
@@ -476,6 +477,35 @@ def _is_sklearn_flow(cls, flow: OpenMLFlow) -> bool:
             or ',sklearn==' in flow.external_version
         )
 
+    def _get_sklearn_description(self, model: Any, char_lim: int = 1024) -> str:
+        '''Fetches the sklearn function docstring for the flow description
+
+        Parameters
+        ----------
+        model: The sklearn model object
```
mfeurer: Please have a space before the colon.

openml/extensions/sklearn/extension.py
```diff
@@ -476,6 +477,35 @@ def _is_sklearn_flow(cls, flow: OpenMLFlow) -> bool:
...
+        char_lim: int, specifying the max length of the returned string
```
mfeurer: Same here. Also, please put only the data type in the first line and put the description in the following lines.

openml/extensions/sklearn/extension.py
```diff
@@ -476,6 +477,35 @@ def _is_sklearn_flow(cls, flow: OpenMLFlow) -> bool:
...
+            OpenML servers have a constraint of 1024 characters for the 'description' field.
+
+        Returns
+        -------
+        string of length <= char_lim
```
mfeurer: I think `str` is sufficient here.

openml/extensions/sklearn/extension.py
```diff
@@ -476,6 +477,35 @@ def _is_sklearn_flow(cls, flow: OpenMLFlow) -> bool:
...
+        '''
+        def match_format(s):
+            return "{}\n{}\n".format(s, len(s) * '-')
+        s = inspect.getdoc(model)
+        if len(s) <= char_lim:
+            return s
+        try:
+            pattern = "Read more in the :ref:"  # "Parameters"
```
mfeurer: Could you briefly explain the comment here?

mfeurer: Also, could you please explain the exact behavior of the function in the docstring above?

openml/extensions/sklearn/extension.py
```diff
@@ -476,6 +477,35 @@ def _is_sklearn_flow(cls, flow: OpenMLFlow) -> bool:
...
+            OpenML servers have a constraint of 1024 characters for the 'description' field.
+
+        Returns
+        -------
+        string of length <= char_lim
+        '''
+        def match_format(s):
+            return "{}\n{}\n".format(s, len(s) * '-')
+        s = inspect.getdoc(model)
+        if len(s) <= char_lim:
```
mfeurer: I'm wondering if we should always cut the parameters and the "read more" as they are not really a description of the algorithm/flow. What do you think?

openml/extensions/sklearn/extension.py
```diff
@@ -476,6 +477,35 @@ def _is_sklearn_flow(cls, flow: OpenMLFlow) -> bool:
...
+            return s
+        try:
+            pattern = "Read more in the :ref:"  # "Parameters"
+            index = s.index(pattern)
+        except ValueError:
+            pattern = "Parameters"
```
mfeurer: What would happen if that doesn't exist either, for example due to a mistake in the sklearn docs? Maybe this should be more failure tolerant and use the whole string if the two patterns aren't part of the docstring.

openml/extensions/sklearn/extension.py
```diff
@@ -602,6 +634,60 @@ def _check_multiple_occurence_of_component_in_flow(
...
 
+    def _extract_sklearn_parameter_docstring(self, model):
+        def match_format(s):
+            return "{}\n{}\n".format(s, len(s) * '-')
+        s = inspect.getdoc(model)
+        s1 = "Parameters"
+        s2 = ["Attributes", "See also", "Note", "References"]
+        try:
+            index1 = s.index(match_format(s1))
+        except ValueError as e:
```
mfeurer: Is this triggered in practice?

openml/extensions/sklearn/extension.py
```diff
@@ -743,7 +830,9 @@ def flatten_all(list_):
                 else:
                     parameters[k] = None
 
-            parameters_meta_info[k] = OrderedDict((('description', None), ('data_type', None)))
+            data_type, description = parameters_docs[k]
```
mfeurer: Can we make this more failure tolerant if the parameter is not documented? We should then also issue a warning.

openml/extensions/sklearn/extension.py
Neeratyoy: Yeah, I am at this point trying to keep only the descriptions that are there until a "Read More". In case a "Read More" itself is missing in the docstrings, I am then trimming until "Parameters" begin. In effect, I am trying to keep only the meat of the description for the flow description.

openml/extensions/sklearn/extension.py
Neeratyoy: Are nested try...except an accepted practice?

openml/extensions/sklearn/extension.py
Neeratyoy: Not really. However, since this design is based completely on another repo's design, I thought it would be more robust to account for a failure nonetheless.
I'm certain though I am not covering all possible breaking points.

openml/extensions/sklearn/extension.py
Neeratyoy: Have added some dosctrings. Hope it is better now.

openml/extensions/sklearn/extension.py
Neeratyoy: Have added a if...else which is like communicates an all-or-nothing scenario.

openml/extensions/sklearn/extension.py
Neeratyoy: "Flat is better than nested" from [The Zen of Python](https://www.python.org/dev/peps/pep-0020/#id3) :)
Followed this...

Neeratyoy:
@mfeurer there will be quite a few errors namely `AttributeError`, `KeyError` which I need to take a look at. The problem I discussed with you should be raising `ValueError`s specifically.

mfeurer:
Hm, it appears that we need to ignore the descriptions for these unit tests.

Neeratyoy:
> Hm, it appears that we need to ignore the descriptions for these unit tests.

Which I believe tantamounts to a `get_flow()` call to compare descriptions? Or on this `ValueError`?

mfeurer:
>  Shall I edit this to exempt the check for description and parameters_meta_info?

Yes, please, but use a flag like for the other exemptions.

Regarding edge case number 1, I'm not sure why adding more characters to the regex breaks this. However, you could simply split the string into the part before and after the first colon and use these?

Regarding edge case number 2, I think we should just add no description. And yes, the whole thing is under the assumption that 3rd-party packages use the same documentation as scikit-learn. That's a bit unfortunate. Could you please add a comment to the scikit-learn extension describing this?

openml/extensions/sklearn/extension.py
mfeurer: Hey, I actually meant that you could simple drop the if statement here. Then, the "read more" would automatically be removed even if it the string was shorter than 2014 characters.

openml/extensions/sklearn/extension.py
```diff
@@ -480,31 +480,163 @@ def _is_sklearn_flow(cls, flow: OpenMLFlow) -> bool:
...
+            # trim till 'Read more'
+            pattern = "Read more in the :ref:"
             index = s.index(pattern)
+            s = s[:index]
+            # trimming docstring to be within char_lim
+            if len(s) > char_lim:
+                s = "{}...".format(s[:char_lim - 3])
+            return s.strip()
         except ValueError:
+            pass
```
mfeurer: Could you please log the error using a logger? Ignoring exceptions asks for future issues.

openml/extensions/sklearn/extension.py
```diff
@@ -480,31 +480,163 @@ def _is_sklearn_flow(cls, flow: OpenMLFlow) -> bool:
...
+        try:
+            # if 'Read more' doesn't exist, trim till 'Parameters'
             pattern = "Parameters"
```
mfeurer: Is there a way you also check that the next line is full of dashes? I'm a bit worried that Parameters is used in the docstring and we'd cut it. A different and easier check would be that the line contains something like `r"$\s*Parameters^"` (line begins, whitespace, the word Parameters, and then the end of the line).

openml/extensions/sklearn/extension.py
```diff
@@ -480,31 +480,163 @@ def _is_sklearn_flow(cls, flow: OpenMLFlow) -> bool:
...
             index = s.index(match_format(pattern))
+        except ValueError:
+            # returning full docstring
```
mfeurer: Same as above, please don't hide errors.

tests/test_extensions/test_sklearn_extension/test_sklearn_extension.py
```diff
@@ -75,7 +75,7 @@ def test_serialize_model(self):
 
             fixture_name = 'sklearn.tree.tree.DecisionTreeClassifier'
             fixture_short_name = 'sklearn.DecisionTreeClassifier'
-            fixture_description = 'Automatically created scikit-learn flow.'
+            fixture_description = self.extension._get_sklearn_description(model)
```
mfeurer: Please do not test the function(s) by calling the function it relies on itself. Is it possible that you put the actual content of the docstring here?

openml/flows/functions.py
```diff
@@ -366,6 +366,10 @@ def assert_flows_equal(flow1: OpenMLFlow, flow2: OpenMLFlow,
                                    ignore_custom_name_if_none)
         elif key == '_extension':
             continue
+        elif key == 'description':
```
mfeurer: Could you please make this a keyword argument so this can be turned on/off?

openml/flows/functions.py
```diff
@@ -397,6 +401,35 @@ def assert_flows_equal(flow1: OpenMLFlow, flow2: OpenMLFlow,
                 # Helps with backwards compatibility as `custom_name` is now auto-generated, but
                 # before it used to be `None`.
                 continue
+            elif key == 'parameters_meta_info':
+                # this value is a dictionary where each key is a parameter name, containing another
+                # dictionary with keys specifying the parameter's 'description' and 'data_type'
+                # check of descriptions can be ignored since that might change
```
mfeurer: check of -> check if

openml/extensions/sklearn/extension.py
```diff
@@ -830,9 +908,16 @@ def flatten_all(list_):
                 else:
                     parameters[k] = None
 
-            data_type, description = parameters_docs[k]
-            parameters_meta_info[k] = OrderedDict((('description', description),
-                                                   ('data_type', data_type)))
+            if parameters_docs is not None:
+                # print(type(model))
```
mfeurer: Please remove these print statements.

tests/test_extensions/test_sklearn_extension/test_sklearn_extension.py
Neeratyoy: Agreed. However, `_get_sklearn_description` is not a function being directly unit tested. 
I can always paste the 1024 character length string. 
I'll do that if you say so.

openml/extensions/sklearn/extension.py
Neeratyoy: Yeah, I agree. I too had the concern.
Since all the test cases passed, I forgot about it.
Will change this and test.

tests/test_extensions/test_sklearn_extension/test_sklearn_extension.py
mfeurer: Not directly, but it does the work under the hood, right? If yes, I'd prefer if you paste the string here.

openml/extensions/sklearn/extension.py
Neeratyoy: Oh, I myself forgot about it. On line 501, there is a function `match_format` defined which makes the string to be checked for to be "Parameters\n----------\n". Same with the other headings.

openml/extensions/sklearn/extension.py
Neeratyoy: Added.

openml/extensions/sklearn/extension.py
Neeratyoy: Added.

Neeratyoy:
@mfeurer 
The tests that are failing, I am not able to reproduce them locally. The descriptions match consistently. 
Was trying to look up an older version of docstrings but couldn't quite map it correctly. Could this be a version issue?

openml/extensions/sklearn/extension.py
mfeurer: Could you please use the warning level here? The info level is turned off in many cases.

openml/extensions/sklearn/extension.py
mfeurer: Could you make this a warning, too, please?

mfeurer:
> Could this be a version issue?

Most likely. You could do a version switch here and check the fixture for 0.21, and the output of the function (which you had before) for the others?

Neeratyoy:
@mfeurer it appears that we are testing with a really old sklearn version of 0.18.2.
Why don't we test with the latest ones, since that is probably what the users would be using?

PR #770, opened 2019-08-20
Mark unit test as flaky

mfeurer:
This PR marks one constantly failing unit test as flaky which often fails under high test server load (i.e. multiple travis runs at the same time). Also, it install the flaky package which is a plugin to pytest to actually rerun the marked unit test.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -48,7 +48,8 @@ setuptools.setup(name="openml",
                          'pytest-xdist',
                          'pytest-timeout',
                          'nbformat',
-                         'oslo.concurrency'
+                         'oslo.concurrency',
+                         'flaky',
                      ],
                      'examples': [
                          'matplotlib',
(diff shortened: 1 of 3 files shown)


PGijsbers:
I'm sorry, I am confused about this.. The test was already marked as flaky? What does it mean to use the default value for `rerun`? I don't see a reference in the [flakey readme](https://github.com/box/flaky).

mfeurer:
> The test was already marked as flaky?

One of them, the other one is newly marked. I think nothing happened though because the flaky plugin was not yet installed

> What does it mean to use the default value for rerun

There exists no `rerun` attribute. There's only something called `max_runs`, which by default is set to two.

All in all, I think this PR is an improvement over the previous state. If the unit tests continue to fail, we can increase the number of retries even further.

PGijsbers:
To make sure I understand it right, what you are saying is that the previous `@pytest.mark.flaky(reruns=3)` had no effect since the package was not installed, and failed silently somehow?

mfeurer:
From my understanding of the [pytest docs](https://docs.pytest.org/en/latest/flaky.html), pytest by itself does not provide functionality to rerun flaky tests by itself. Therefore, I concluded that it did not do anything and we need the flaky plugin.

PGijsbers:
I'll trust your judgement on that. Seems weird but also probably correct.

PR #772, opened 2019-08-22
Add #737

### LINKED ISSUE #737: New evaluation listing call that includes hyperparameters
Currently, if one wants to download all hyperparameters and evaluations of a specific flow, they need to first download the list of evaluations (fast), and then download runs one by one to get the hyperparameters (slow).

This is a problem for the current Frontend, plus it also complicates downloading meta-data from the server (last time I tried this it worked but took a very long time).

Since this is such a common use case, should we add an API call for that? Basically something 
returning something like:
```
{
  "evaluations": {
    "evaluation": [
      {
        "run_id": "1",
        "task_id": "68",
        "flow_id": "61",
        "function": "area_under_roc_curve",
        "upload_time": "2014-04-06 23:30:40",
        "value": "0.839359",
        "array_data": "[0,0.99113,0.898048,0.874862,0.791282,0.807343,0.820674]",
        "setup_parameters": "parameter": [
            {
              "full_name": "weka.JRip(1)_F",
              "parameter_name": "F",
              "value": "3"
            },
           {
             "full_name": "weka.JRip(1)_N",
             "parameter_name": "N",
             "value": "2.0"
           },
           ...]
      }
```

Different options:
* Add it as an option `return_setups=true` to the current evaluation list call. It would only work if there is a specific flow and function, otherwise return an error.
* Add it as a new call `list_evaluations_and_setup(flow_id,function)
* Use the current setup/id call, but that has the same issue that you need to ask for the setups one by one, plus the evaluation list does not return the setup id. 

Thoughts?

sahithyaravi1493:
Changes to #737  from parameter_name to full name and added option to get parameters in separate columns in case single flow is supplied

#### How should this PR be tested

`evaluations.list_evaluations_setups(function=function,flow=[flow_id], sort_order=sort_order,size=size, output_format='dataframe')`

`evaluations.list_evaluations_setups(function=function,flow=[flow_id], sort_order=sort_order,size=size, output_format='dataframe', parameters_in_separate_columns = True)` 

#### Any other comments?
Previous PR for #737 - https://github.com/openml/openml-python/pull/747

### DIFF
--- a/examples/fetch_evaluations_tutorial.py
+++ b/examples/fetch_evaluations_tutorial.py
@@ -149,12 +149,13 @@ flow_names = evals.flow_name.unique()[:top_n]
 for i in range(top_n):
     print((flow_ids[i], flow_names[i]))
 
-###############################################################################
+#############################################################################
 # Obtaining evaluations with hyperparameter settings
-# ==========================================
+# ==================================================
 # We'll now obtain the evaluations of a task and a flow with the hyperparameters
 
-# List evaluations in desc order based on predictive_accuracy with hyperparameters
+# List evaluations in descending order based on predictive_accuracy with
+# hyperparameters
 evals_setups = openml.evaluations.list_evaluations_setups(function='predictive_accuracy', task=[31],
                                                           size=100, sort_order='desc')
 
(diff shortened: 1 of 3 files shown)


examples/fetch_evaluations_tutorial.py
```diff
@@ -147,3 +148,28 @@ def plot_flow_compare(evaluations, top_n=10, metric='predictive_accuracy'):
 flow_names = evals.flow_name.unique()[:top_n]
 for i in range(top_n):
     print((flow_ids[i], flow_names[i]))
+
+###############################################################################
+# Obtaining evaluations with hyperparameter settings
+# ==========================================
```
mfeurer: The number of equal characters needs to equal the number of characters in the header.

examples/fetch_evaluations_tutorial.py
```diff
@@ -147,3 +148,28 @@ def plot_flow_compare(evaluations, top_n=10, metric='predictive_accuracy'):
...
+# We'll now obtain the evaluations of a task and a flow with the hyperparameters
+
+# List evaluations in desc order based on predictive_accuracy with hyperparameters
```
mfeurer: Please do not abbreviate here but write `descending`.

Issue #773, opened 2019-08-25
Columntransformer causes openml to open new setup for otherwise identical pipelines 


hp2500:
Hi there, 
i am running experiments where it is important that I can compare the performance of the same setup across different tasks. Generally I have no problems doing this, but whenever I am running a task that has mixed data types in its feature matrix I run into issues. 

Specifically, I am using the columntransformer to run different preprocessing steps on categorical and numerical features. I am running the same pipeline on each one of the tasks, but openml creates a new setup for each of these runs, making it impossible to find the runs belonging to each of my pipelines / parameter configurations.

The only thing I can think of that changes from run to run are the columns of the numeric and categorical features, hence the indicator mask that I am passing to the columntransformer is different from run to run. My suspicion is that this is causing openml to treat each pipeline as a different pipeline, creating a new setup for each run, even though the pipeline is technically the same.

Here is a simple example: 
```python
import openml
from sklearn.svm import SVC 
from sklearn.compose import ColumnTransformer
from sklearn.pipeline import Pipeline, make_pipeline
from sklearn.preprocessing import StandardScaler, OneHotEncoder 

# get task
task = openml.tasks.get_task(3022)

# get dataset object 
data = openml.datasets.get_dataset(task.dataset_id)

# get relevant info from dataset object
X, y, categorical_indicator, attribute_names = data.get_data(dataset_format='array',
                                                            target=data.default_target_attribute)
# make indicator masks
cat = categorical_indicator
num = [not k for k in categorical_indicator]

# make columntransformer
numeric_transformer = make_pipeline(StandardScaler())
categorical_transformer = make_pipeline(OneHotEncoder(handle_unknown='ignore'))
preprocessor = ColumnTransformer(
transformers=[('num', numeric_transformer, num), ('cat', categorical_transformer, cat)])

# make pipeline
clf = SVC(gamma = 'scale', random_state=1)
pipe = make_pipeline(preprocessor, clf)

# run task
run = openml.runs.run_model_on_task(pipe, task, avoid_duplicate_runs=True)
run.publish()

# check setup 
openml.runs.get_run(run.run_id).setup_id
```
8231820

```python
# get task
task = openml.tasks.get_task(23)

# get dataset object 
data = openml.datasets.get_dataset(task.dataset_id)

# get relevant info from dataset object
X, y, categorical_indicator, attribute_names = data.get_data(dataset_format='array',
                                                            target=data.default_target_attribute)
# make indicator masks
cat = categorical_indicator
[...]

PR #777, opened 2019-09-03
Dataframe run on task

amueller:
This tries to solve #773.
Still probably needs some handling for sparse datasets.


amueller:
@mfeurer can you explain the restriction on the third part of the subcomponent?

mfeurer:
Please see my answer in #773

Neeratyoy:
@amueller shall I try taking this PR up?

Neeratyoy:
@mfeurer about 4/5 unit tests throw the following error when running on the latest sklearn version locally
`openml.exceptions.OpenMLServerException: Problem validating uploaded description file - XML does not correspond to XSD schema`

Also, I couldn't understand why for `task.class_labels` which is annotated as `Optional[List[str]]`, it gives a flake error if I index it `task.class_labels[0]`.
Flake ` error: Value of type "Optional[List[str]]" is not indexable`

mfeurer:
> Also, I couldn't understand why for task.class_labels which is annotated as Optional[List[str]], it gives a flake error if I index it task.class_labels[0].

As it is typed optional it can actually be `None`, which would not be indexable. You probably need to use `typing.cast` to please mypy.

> openml.exceptions.OpenMLServerException: Problem validating uploaded description file - XML does not correspond to XSD schema

Hm, apparently, the dummy imputor needs a docstring, otherwise the description uploaded to OpenML will be emptly.

Neeratyoy:
@mfeurer all test cases appear to be passing now, barring one, for which I am unable to debug the cause and have raised an issue [here](https://github.com/openml/OpenML/issues/1045).
Not sure if [this](https://travis-ci.org/github/openml/openml-python/jobs/699061341#L1967) also should be raised as an issue there or I can fix it.

examples/30_extended/datasets_tutorial.py
```diff
@@ -68,7 +68,7 @@
 # Get the actual data.
 #
 # The dataset can be returned in 2 possible formats: as a NumPy array, a SciPy
```
mfeurer, line 70: Could you please change this to say '3 possible formats'?

openml/__init__.py
```diff
@@ -114,7 +114,6 @@ def populate_cache(task_ids=None, dataset_ids=None, flow_ids=None,
     'study',
     'utils',
     '_api_calls',
-    '__version__',
```
mfeurer, line 117: What's the reason to drop this one?

openml/extensions/sklearn/extension.py
```diff
@@ -1512,10 +1515,11 @@ def _prediction_to_probabilities(y: np.ndarray, classes: List[Any]) -> np.ndarra
             if not isinstance(classes, list):
                 raise ValueError('please convert model classes to list prior to '
                                  'calling this fn')
-            result = np.zeros((len(y), len(classes)), dtype=np.float32)
-            for obs, prediction_idx in enumerate(y):
-                result[obs][prediction_idx] = 1.0
-            return result
+            # DataFrame allows more accurate mapping of classes as column names
```
mfeurer, line 1518: I don't understand this change. What means more accurate here?

tests/test_extensions/test_sklearn_extension/test_sklearn_extension.py
```diff
@@ -1337,12 +1337,13 @@ def setUp(self):
     # Test methods for performing runs with this extension module
 
     def test_run_model_on_task(self):
-        class MyPipe(sklearn.pipeline.Pipeline):
-            pass
+        # class MyPipe(sklearn.pipeline.Pipeline):
```
mfeurer, line 1340: If this is no longer necessary, could you please drop it?

openml/extensions/sklearn/extension.py
```diff
@@ -1573,6 +1577,16 @@ def _prediction_to_probabilities(y: np.ndarray, classes: List[Any]) -> np.ndarra
             else:
                 model_classes = used_estimator.classes_
 
+            if not isinstance(model_classes, list):
+                model_classes = model_classes.tolist()
+
+            # to handle the case when dataset is numpy and categories are encoded
+            # however the class labels stored in task are still categories
+            if isinstance(y_train, np.ndarray) and \
+                    isinstance(typing.cast(List[str], task.class_labels)[0], str):
```
mfeurer, line 1586: I believe this should only be `typing.cast[List]` as we don't know whether it's a string, right?

openml/extensions/sklearn/extension.py
```diff
@@ -1573,6 +1577,16 @@ def _prediction_to_probabilities(y: np.ndarray, classes: List[Any]) -> np.ndarra
...
+                model_classes = [typing.cast(List[str], task.class_labels)[i]
```
mfeurer, line 1587: Do we actually have a unit test for this one?

openml/extensions/sklearn/extension.py
```diff
@@ -1601,9 +1615,16 @@ def _prediction_to_probabilities(y: np.ndarray, classes: List[Any]) -> np.ndarra
...
             try:
                 proba_y = model_copy.predict_proba(X_test)
-            except AttributeError:
+            except AttributeError:  # predict_proba is not available when probability=False
                 if task.class_labels is not None:
-                    proba_y = _prediction_to_probabilities(pred_y, list(task.class_labels))
+                    if isinstance(y_train, np.ndarray) and isinstance(task.class_labels[0], str):
+                        # mapping (decoding) the predictions to the categories
+                        # creating a separate copy to not change the expected pred_y type
+                        preds = [task.class_labels[pred] for pred in pred_y]
```
mfeurer, line 1623: Can this be moved into the `_prediction_to_probabilities` method?

openml/extensions/sklearn/extension.py
```diff
@@ -1763,7 +1787,8 @@ def is_subcomponent_specification(values):
                             }
                         }
                         if len(subcomponent) == 3:
-                            if not isinstance(subcomponent[2], list):
+                            if not isinstance(subcomponent[2], list) and \
+                                    not isinstance(subcomponent[2], OrderedDict):
```
mfeurer, line 1791: The type error needs an update, too.

openml/runs/functions.py
```diff
@@ -161,6 +167,9 @@ def run_flow_on_task(
     upload_flow : bool (default=False)
         If True, upload the flow to OpenML if it does not exist yet.
         If False, do not upload the flow to OpenML.
+    dataset_format : str (default='dataframe')
+        If 'array', the dataset is read as a numpy array
```
mfeurer, line 171: Could you please re-phrase as:
If 'array', the dataset is passed to the model as a numpy array.
(and the same for the dataframe, too).

tests/test_runs/test_run_functions.py
```diff
@@ -342,7 +341,9 @@ def test_run_regression_on_classif_task(self):
 
         clf = LinearRegression()
         task = openml.tasks.get_task(task_id)
-        with self.assertRaises(AttributeError):
+        # internally dataframe is loaded and targets are categorical
+        # which LinearRegression() cannot handle
+        with self.assertRaises(ValueError):
```
mfeurer, line 346: Is it possible that you use a different task which has no categories/categoricals to keep the original behavior? Could you then please also make this `self.assertRaisesRegex` so that we know the error that's happening?

tests/test_runs/test_run_functions.py
```diff
@@ -701,9 +715,11 @@ def test_initialize_cv_from_run(self):
             cv=StratifiedKFold(n_splits=2, shuffle=True),
             n_iter=2)
 
+        clf = make_pipeline(OneHotEncoder(handle_unknown='ignore'), randomsearch)
```
mfeurer, line 718: Can you please put the one hot encoder inside a pipeline that's inside random search?

openml/extensions/sklearn/extension.py
Neeratyoy: Some target columns were being loaded as the class labels themselves. With a dataframe, having the class labels as column names allows an easier access using `loc`. With _accurate_ I meant the more foolproof access that is possible with dataframes.

tests/test_runs/test_run_functions.py
Neeratyoy: I am not sure if I understood the requirement completely. However, made a change based on how I understood it.
Kindly let me know if this is not what was meant.

openml/extensions/sklearn/extension.py
Neeratyoy: No, don't think so. Should I add a unit test in this PR or add a new PR?

openml/extensions/sklearn/extension.py
Neeratyoy: Moved it. Testing if all passes.

openml/extensions/sklearn/extension.py
Neeratyoy: Edited the message.

tests/test_runs/test_run_functions.py
Neeratyoy: Just changed the assert to `assertRaisesRegex`.
From what I am gathering, this test is designed to fail when running `LinearRegression` on a classification task since the target variables for a classification task should be categorical. I am not sure how we can take a classification task without categorical targets. Or do you mean, finding a dataset where the categories are numeric and represented so as well? In that case, do we expect or want the test to pass?

openml/extensions/sklearn/extension.py
mfeurer: Sorry, I don't understand the original problem. Could you please give an example?

tests/test_runs/test_run_functions.py
```diff
@@ -343,7 +343,8 @@ def test_run_regression_on_classif_task(self):
         task = openml.tasks.get_task(task_id)
         # internally dataframe is loaded and targets are categorical
         # which LinearRegression() cannot handle
-        with self.assertRaises(ValueError):
+        print("\nRegex check coming in \n")
```
mfeurer, line 346: Could you please drop the print statement?

tests/test_runs/test_run_functions.py
mfeurer: The original error was that it was an `AttributeError`, and it would be great if we keep that (I just don't remember what attribute failed). By using a different task I was hoping that the target would be encoded as a numpy array with integers/floats. Is this somehow possible?

tests/test_runs/test_run_functions.py
mfeurer: Yes, that's exactly how it should look.

tests/test_runs/test_run_functions.py
Neeratyoy: @mfeurer 
enforced a numpy array load of the dataset in the task run, and we get the `AttributeError`. So I updated the test case to catch an `AttributeError` and it passes.
Hope this suffices for now for this PR.

The rest of the failing test cases are I think dependent on other issues that need fixing. The changes this PR intends to implement, in my opinion, looks fine for now.

tests/test_runs/test_run_functions.py
```diff
@@ -597,31 +597,21 @@ def get_ct_cf(nominal_indices, numeric_indices):
...
-                        DecisionTreeClassifier(),
-                        {
-                            "min_samples_split": [2 ** x for x in range(1, 8)],
-                            "min_samples_leaf": [2 ** x for x in range(0, 7)],
-                        },
-                        cv=3,
-                        n_iter=10,
-                    ),
-                ),
+                ("Estimator", RandomizedSearchCV(KNeighborsClassifier(), {}, cv=3, n_iter=10,),),
```
mfeurer, line 614: Hey, shouldn't there be a parameter to search over?

Neeratyoy: My bad. Made the change and pushed.

openml/extensions/sklearn/extension.py
```diff
@@ -1531,15 +1534,31 @@ def _prediction_to_probabilities(y: np.ndarray, classes: List[Any]) -> np.ndarra
             -------
             np.ndarray
             """
+
+            if isinstance(task, (OpenMLClassificationTask, OpenMLLearningCurveTask)):
+                print("\n\nInside _prediction_to_probabilities\n\n")
```
PGijsbers, line 1539: Don't leave print statements in the code. Use `logging` module.

mfeurer: Thanks for catching that @PGijsbers. Actually, please don't log here at all. @PGijsbers can you think of a reason to use a print statement in regular code? Maybe that can be checked by black, too?

openml/extensions/sklearn/extension.py
```diff
@@ -1531,15 +1534,31 @@ def _prediction_to_probabilities(y: np.ndarray, classes: List[Any]) -> np.ndarra
...
+                if task.class_labels is not None:
+                    if isinstance(y_train, np.ndarray) and isinstance(task.class_labels[0], str):
+                        # mapping (decoding) the predictions to the categories
+                        # creating a separate copy to not change the expected pred_y type
```
PGijsbers, line 1543: I am not entirely sure why you don't just reassign to `y` directly?

openml/extensions/sklearn/extension.py
```diff
@@ -1531,15 +1534,31 @@ def _prediction_to_probabilities(y: np.ndarray, classes: List[Any]) -> np.ndarra
             -------
```
PGijsbers, line 1534: I can't comment on line L#1527 as it is not changed this commit. But I noticed that the parameter name in the docstring is different from the parameter name in the signature. Can you fix this?

openml/extensions/sklearn/extension.py
```diff
@@ -1531,15 +1534,31 @@ def _prediction_to_probabilities(y: np.ndarray, classes: List[Any]) -> np.ndarra
...
+                    raise ValueError("The task has no class labels")
+                classes = model_classes
+            else:
+                return None
+
             # y: list or numpy array of predictions
             # model_classes: sklearn classifier mapping from original array id to
             # prediction index id
             if not isinstance(classes, list):
                 raise ValueError("please convert model classes to list prior to " "calling this fn")
```
PGijsbers, line 1556: Looks like the initial black refactor de-wrapped this line wrongly (please remove the redundant space and quotes).

openml/extensions/sklearn/extension.py
```diff
@@ -1531,15 +1534,31 @@ def _prediction_to_probabilities(y: np.ndarray, classes: List[Any]) -> np.ndarra
...
+                        preds = [task.class_labels[pred] for pred in y]
+                        y = preds
+                else:
+                    raise ValueError("The task has no class labels")
+                classes = model_classes
+            else:
+                return None
+
             # y: list or numpy array of predictions
             # model_classes: sklearn classifier mapping from original array id to
```
PGijsbers, line 1553: Update comments depending on whether or not the variable gets renamed.

openml/extensions/sklearn/extension.py
```diff
@@ -1597,6 +1616,18 @@ def _prediction_to_probabilities(y: np.ndarray, classes: List[Any]) -> np.ndarra
             else:
                 model_classes = used_estimator.classes_
 
+            if not isinstance(model_classes, list):
+                model_classes = model_classes.tolist()
+
+            # to handle the case when dataset is numpy and categories are encoded
+            # however the class labels stored in task are still categories
+            if isinstance(y_train, np.ndarray) and isinstance(
+                typing.cast(List, task.class_labels)[0], str
```
PGijsbers, line 1625: This requires `task.class_labels` to be not None. I don't think this is checked at this point?

openml/extensions/sklearn/extension.py
```diff
@@ -1646,10 +1674,17 @@ def _prediction_to_probabilities(y: np.ndarray, classes: List[Any]) -> np.ndarra
...
+                    proba_y_new = pd.DataFrame(
+                        0,
+                        index=np.arange(proba_y.shape[0]),
+                        columns=task.class_labels,
+                        dtype=np.float32,
+                    )
                     for idx, model_class in enumerate(model_classes):
-                        proba_y_new[:, model_class] = proba_y[:, idx]
-                    proba_y = proba_y_new
+                        proba_y_new.loc[:, model_class] = proba_y[:, idx]
```
PGijsbers, line 1686: Would it perhaps not be clear to:
 - keep the result of `prediction_to_probability` as a dataframe.
 - convert proba_y to a dataframe if `perdict_proba` was called:
    ```python
    proba_y = pd.DataFrame(proba_y, columns=model_classes)
    ```
 - then add missing columns
 - then reorder the dataframe:
   ```python
   proba_y = proba_y[task.class_labels]
   ```

Advantages:
 - Don't produce a whole copy in memory.
 - Keep indexing by class label only, never integers.
 - Can make (and keep) `_prediction_to_proba` store the probabilities as e.g. `uint8` for much smaller footprint (though conversion might be required later anyway, so I don't know if it is a win).

openml/extensions/sklearn/extension.py
```diff
@@ -1646,10 +1674,17 @@ def _prediction_to_probabilities(y: np.ndarray, classes: List[Any]) -> np.ndarra
...
+                    proba_y = proba_y_new.to_numpy()
 
                 if proba_y.shape[1] != len(task.class_labels):
```
PGijsbers, line 1689: This is a repeated clause from Line#1666.
If it was not true there (they are equal), then it is not true here.
If it was true there (they were unequal), then Line#1682 specifically creates `proba_y_new` to have the columns `task.class_labels`, and `proba_y` is later assigned the value of `proba_y_new`.
So I see no scenario where this statement is true (i.e. at this point `proba_y` should always a column for each label in `task.class_labels`. Am I missing something? Or is it forgotten code?

openml/runs/functions.py
```diff
@@ -483,8 +503,11 @@ def _calculate_local_measure(sklearn_fn, openml_name):
                     for j, class_label in enumerate(task.class_labels):
```
PGijsbers, line 503: I know this code was there before too. But could you change `class_label` to `_` since the variable is unused?

openml/runs/functions.py
```diff
@@ -498,7 +521,7 @@ def _calculate_local_measure(sklearn_fn, openml_name):
         elif isinstance(task, OpenMLRegressionTask):
 
             for i in range(0, len(test_indices)):
```
PGijsbers, line 523: Use `enumerate`

openml/extensions/sklearn/extension.py
PGijsbers: I don't Black will do that, Black only formats and checks that the resulting syntax trees are identical (if I remember the right terminology). They wouldn't be if expressions are removed.
However I did find a Flake8 plugin [flake8-print](https://pypi.org/project/flake8-print/) which will warn you about print statements. It's not automatic removal - but that's probably a good thing considering in the worst case you call a function with side effects from within your print statement, and the print statement needs to be removed with some care.

mfeurer: That sounds great to have.

openml/extensions/sklearn/extension.py
Neeratyoy: Yes, you're right. I was scratching my head too. I think it makes sense to remove the redundant `if` but retain the warning message that is being logged.

openml/extensions/sklearn/extension.py
Neeratyoy: That's a good spot. Thanks.
I made some changes to not have the extra copy and retain it as dataframe whenever applicable, without a conversion back and forth.

However, right at the end, needed to convert the probability array to numpy since [_run_task_get_arffcontent](https://github.com/openml/openml-python/blob/15743ee9f1e1cd5bfc772b7d67aec8d33724740d/openml/runs/functions.py#L504) appears to require a numpy as the probability matrix.

openml/extensions/sklearn/extension.py
Neeratyoy: Not directly, but this check is under the scope of an if-statement checking for `OpenMLClassificationTask` and `OpenMLLearningTask` which should imply `task.class_labels` is not `None` for that task.
Moreover, the second half of this if-statement checks if `task.class_labels` is a list of strings, which again is indirectly checking `task.class_labels` to be not `None`.

Do we then need to check for it?
I can still add an explicit check for None and raise an error if you deem necessary.

This though appears to be related to the `Optional` status of `task.class_labels` as you mention in a comment below.

mfeurer:
The reason why it is marked as optional is because it is a field populated by the server and not allowed for upload in the [task XSD](https://github.com/openml/OpenML/blob/develop/openml_OS/views/pages/api_new/v1/xsd/openml.task.upload.xsd). Thus, when creating a task, we don't have a value for the class labels and the field can be None.

openml/extensions/sklearn/extension.py
mfeurer: > However, right at the end, needed to convert the probability array to numpy since _run_task_get_arffcontent appears to require a numpy as the probability matrix.

You could also change that function to accept a dataframe - it appears easier and safer to work with.

openml/extensions/sklearn/extension.py
PGijsbers: Sounds good to me.

PGijsbers:
Do you think it might make sense to change the behavior of [create_task](https://github.com/openml/openml-python/blob/develop/openml/tasks/functions.py#L461) to also publish the task (and retrieve the new version with server added fields)?
A non-synced task seems useless to me (e.g. does not have split data), so it's not like there is real offline functionality to preserve. I don't see the xsd supporting custom splits, unless I am missing something.

mfeurer:
> Do you think it might make sense to change the behavior of create_task to also publish the task (and retrieve the new version with server added fields)?

Yes and no. I believe we have this pattern of `Optional` in quite a few places so we should the maybe change all of them at once? Also, it might be good to have the chance to look at the created object before uploading it?

Another option would be to simply pass an empty list when creating a classification task, and also add a check to run_flow_on_task to check whether the task has an ID.

PGijsbers:
I don't really like using an empty list, I think I would prefer the current situation over that.
But I understand immediately uploading the task is also not ideal.
For now I would be happy with just the centralized checks, i.e. at the top of each function once, or only in `run_flow_on_task` and remove further checks in functions which are only called through that.

mfeurer:
> For now I would be happy with just the centralized checks, i.e. at the top of each function once, or only in run_flow_on_task and remove further checks in functions which are only called through that.

That sounds good to me. However, that check is most likely only necessary for mypy, so I think there should be a comment that this check is only to please mypy.

PGijsbers:
> I think there should be a comment that this check is only to please mypy

The structure of the code (having else-clause that raises an exception) implies that it's not mypy only.
Keeping the else-clause which raises an error (just one instead of many), makes sense if that catches users attempting to use this functionality on newly created but not yet uploaded tasks.

openml/extensions/sklearn/extension.py
```diff
@@ -1674,27 +1673,28 @@ def _prediction_to_probabilities(y: np.ndarray, classes: List[Any]) -> np.ndarra
...
                     message = "Estimator only predicted for {}/{} classes!".format(
                         proba_y.shape[1], len(task.class_labels),
                     )
                     warnings.warn(message)
                     openml.config.logger.warn(message)
+
+                    # True if predict_proba is successfully called with X_test as numpy
+                    # if X_test is dataframe, proba_y will be dataframe with model_classes columns
+                    if isinstance(proba_y, np.ndarray):
+                        proba_y = pd.DataFrame(proba_y, columns=model_classes)
```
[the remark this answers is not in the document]
mfeurer: (or maybe scikit-learn doesn't support that yet...)

openml/extensions/sklearn/extension.py
```diff
@@ -1646,20 +1673,28 @@ def _prediction_to_probabilities(y: np.ndarray, classes: List[Any]) -> np.ndarra
...
+
+                    for i, col in enumerate(task.class_labels):
+                        # adding missing columns with 0 probability
+                        if col not in model_classes:
+                            proba_y[col] = 0
```
mfeurer, line 1690: Don't you have to re-order the columns if you add a new column to the dataframe?

openml/extensions/sklearn/extension.py
```diff
@@ -1646,20 +1673,28 @@ def _prediction_to_probabilities(y: np.ndarray, classes: List[Any]) -> np.ndarra
...
+                    proba_y = proba_y[task.class_labels]
             else:
                 raise ValueError("The task has no class labels")
 
+            if not isinstance(proba_y, pd.DataFrame):
+                proba_y = pd.DataFrame(proba_y, columns=task.class_labels)
```
mfeurer, line 1696: Could you please add an elif here that checks that the columns of the dataframe are equal to the class labels?

Neeratyoy: Sure, certainly.

openml/extensions/sklearn/extension.py
Neeratyoy: Doesn't the next line, at the same scope as the `for` do that?

openml/extensions/sklearn/extension.py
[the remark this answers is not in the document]
Neeratyoy: If `proba_y` is a matrix of prediction probabilities, shouldn't that remain numeric, while being cast from numpy to dataframe?
Or I didn't understand what you meant here.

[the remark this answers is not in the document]
mfeurer: No, you're absolutely right. Please ignore my comment here.

Neeratyoy:
@mfeurer 
added a new [unit test](https://github.com/openml/openml-python/blob/fbdcd6f87808e2d2890342ff308c5591733cd857/tests/test_extensions/test_sklearn_extension/test_sklearn_extension.py#L1939) as discussed
however, still unsure of how to create a model that is serialized and handles mixed columns for input dataframes

openml/extensions/sklearn/extension.py
Neeratyoy: Added [here](https://github.com/openml/openml-python/blob/c06eb0da24498c08c50edcf579075c7e500a5301/openml/extensions/sklearn/extension.py#L1697).

mfeurer: Not sure if it's necessary, but we could also check that the order stays the same

openml/extensions/sklearn/extension.py
mfeurer: My bad, of course it does.

openml/extensions/sklearn/extension.py
mfeurer: Hey, I just checked but I think you did not yet update the return value type annotation for the functions `openml.extensions.sklearn.extension._prediction_to_probabilities` and `openml.extensions.sklearn.extension._run_model_on_fold`.

Neeratyoy:
@mfeurer testing the docpush with this [unit test](https://github.com/amueller/openml-python/blob/dataframe_run_on_task/.travis.yml#L18) trigger s.t. the sklearn version matches and we can observe what other changes we might need for docs on a higher sklearn version

tests/test_extensions/test_sklearn_extension/test_sklearn_extension.py
```diff
@@ -2014,6 +2014,10 @@ def test_sklearn_serialization_with_none_step(self):
         with self.assertRaisesRegex(ValueError, msg):
             self.extension.model_to_flow(clf)
 
+    @unittest.skipIf(
+        LooseVersion(sklearn.__version__) < "0.20",
+        reason="columntransformer introduction in 0.20.0",
```
mfeurer, line 2019: There's no column transformer happening here, right?

examples/30_extended/study_tutorial.py
```diff
@@ -78,18 +80,40 @@
...
+        ("transform", ct),
+        ("estimator", HistGradientBoostingClassifier()),
     ]
 )
 
 suite = openml.study.get_suite(1)
 # We'll create a study with one run on three random datasets each
 tasks = np.random.choice(suite.tasks, size=3, replace=False)
 run_ids = []
+# 391 is all numeric
```
mfeurer, line 112: Could you please drop those comments again?

tests/test_extensions/test_sklearn_extension/test_sklearn_extension.py
Neeratyoy: We do have it [here](https://github.com/openml/openml-python/blob/f0ff562d9ca428fc6bae0c52ee6694e2f0a9fdec/tests/test_extensions/test_sklearn_extension/test_sklearn_extension.py#L2049), right?

examples/30_extended/study_tutorial.py
Neeratyoy: yes!

tests/test_extensions/test_sklearn_extension/test_sklearn_extension.py
mfeurer: Ha, I didn't see that the unit test continues, sorry

mfeurer:
Currently blocking on https://github.com/scikit-learn/scikit-learn/issues/18272

mfeurer (commented):
This looks good to me by now :) I have added a few more comments, and I think there are few comments still open.

examples/30_extended/flow_id_tutorial.py
```diff
@@ -15,6 +15,12 @@
 
 import openml
 
+
+print(openml.config.server)
```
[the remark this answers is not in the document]
Neeratyoy: Not sure why it was there...tested the script and as expected works fine on the live server too!
Removing the switch to test server.

examples/30_extended/run_setup_tutorial.py
```diff
@@ -65,45 +65,35 @@
...
-cat_imp = make_pipeline(SimpleImputer(strategy="most_frequent"), OneHotEncoder(handle_unknown="ignore", sparse=False))
-ct = ColumnTransformer([
-    ("cat", cat_imp, cat),
-    ("cont", FunctionTransformer(lambda x: x, validate=False), cont)
-])
+cat_imp = make_pipeline(
+    SimpleImputer(strategy="most_frequent"), OneHotEncoder(handle_unknown="ignore", sparse=False)
+)
+ct = ColumnTransformer(
+    [("cat", cat_imp, cat), ("cont", FunctionTransformer(lambda x: x, validate=False), cont)]
```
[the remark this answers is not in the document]
Neeratyoy: This is really confusing to me right now. The copy of the source code I have is this [here](https://github.com/amueller/openml-python/blob/dataframe_run_on_task/examples/30_extended/run_setup_tutorial.py#L79).
Did I mess up some merge?

.travis.yml
```diff
@@ -16,15 +16,12 @@ env:
   - MODULE=openml
   matrix:
   - DISTRIB="conda" PYTHON_VERSION="3.6" SKLEARN_VERSION="0.23.1" DOCPUSH="true" SKIP_TESTS="true"
-  - DISTRIB="conda" PYTHON_VERSION="3.7" SKLEARN_VERSION="0.21.2" RUN_FLAKE8="true" SKIP_TESTS="true"
-  - DISTRIB="conda" PYTHON_VERSION="3.7" SKLEARN_VERSION="0.21.2" COVERAGE="true" DOCPUSH="true"
+  - DISTRIB="conda" PYTHON_VERSION="3.7" SKLEARN_VERSION="0.23.1" RUN_FLAKE8="true" SKIP_TESTS="true"
```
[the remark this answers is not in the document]
Neeratyoy: Added coverage, as was in 'develop'.
Probably got omitted during the merge.

.travis.yml
```diff
@@ -16,15 +16,12 @@ env:
...
   - DISTRIB="conda" PYTHON_VERSION="3.7" SKLEARN_VERSION="0.23.1" TEST_DIST="true"
   - DISTRIB="conda" PYTHON_VERSION="3.6" SKLEARN_VERSION="0.23.1" TEST_DIST="true"
   - DISTRIB="conda" PYTHON_VERSION="3.7" SKLEARN_VERSION="0.22.2" TEST_DIST="true"
   - DISTRIB="conda" PYTHON_VERSION="3.6" SKLEARN_VERSION="0.22.2" TEST_DIST="true"
   - DISTRIB="conda" PYTHON_VERSION="3.6" SKLEARN_VERSION="0.21.2" TEST_DIST="true"
-  - DISTRIB="conda" PYTHON_VERSION="3.7" SKLEARN_VERSION="0.21.2" RUN_FLAKE8="true" SKIP_TESTS="true"
```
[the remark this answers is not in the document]
Neeratyoy: Probably was an error by me during the merge.

examples/30_extended/run_setup_tutorial.py
```diff
@@ -52,22 +56,40 @@
...
+def cont(X):
+    return X.dtypes != "category"
+
+
+def cat(X):
+    return X.dtypes == "category"
+
+
+cat_imp = make_pipeline(
+    SimpleImputer(strategy="most_frequent"),
```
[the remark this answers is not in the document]
Neeratyoy: so for now, we let this pass?

openml/exceptions.py
```diff
@@ -27,6 +27,9 @@ def __init__(self, message: str, code: int = None, url: str = None):
         self.url = url
         super().__init__(message)
 
+    def __str__(self):
```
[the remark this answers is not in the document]
Neeratyoy: Removed it.

examples/30_extended/run_setup_tutorial.py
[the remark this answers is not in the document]
mfeurer: I don't know about that, but I guess you can then remove the function transformer.

examples/30_extended/run_setup_tutorial.py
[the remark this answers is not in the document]
mfeurer: for now we can't change anything.

mfeurer:
Thanks for addressing my change requests, that all looks good now. There are a few old comments though, which are collapsed and you need to press 'show more to see them, that would be great if you could have a look at them. We can then ask @PGijsbers for another look and finally merge this :)

openml/extensions/sklearn/extension.py
Neeratyoy: Updated it, thanks for pointing it out.
However, not sure how to edit the docstrings since they don't appear to bear semblance to what the function is actually returning.

openml/runs/functions.py
```diff
@@ -561,7 +559,8 @@ def _calculate_local_measure(sklearn_fn, openml_name):
                 )
 
         elif isinstance(task, OpenMLClusteringTask):
-            for i in range(0, len(test_indices)):
+
+            for i, test_idx in enumerate(test_indices):
```
mfeurer, line 563: Please use `for i, _ in enumerate(test_indices)` if you don't use `test_idx`

openml/extensions/sklearn/extension.py
mfeurer: That appears to be a wrong since two years, you can just update the docstring.

openml/runs/functions.py
```diff
@@ -499,7 +499,7 @@ def _calculate_local_measure(sklearn_fn, openml_name):
 
         if isinstance(task, (OpenMLClassificationTask, OpenMLLearningCurveTask)):
 
-            for i, tst_idx in enumerate(test_indices):
+            for i, _ in enumerate(test_indices):
```
mfeurer, line 502: Here you can (and should) keep the `tst_idx` as you're using it below.

openml/extensions/sklearn/extension.py
mfeurer: Thanks for the update. I think Pieter's original suggestion is not yet addressed - do you think you could have a look whether that's still possible?

openml/runs/functions.py
Neeratyoy: I [replaced](https://github.com/openml/openml-python/blob/da8dbb921e86c5ad5058d62dc12ad9d45fbfcd47/openml/runs/functions.py#L526) the usage of `tst_idx` with `test_indices[i]` to be consistent with the subsequent `elif` blocks where we are enumerating in a similar fashion of `for i, _ in enumerate(test_indices)`.

Is that alright? Or shall I replace it with `tst_idx` as it was, and push again?

mfeurer: > Or shall I replace it with tst_idx as it was, and push again?

Yes, please. Please use `_` only if the variable is not used later on. Here it's used, so no need for an underscore. In the other two cases it's not used, so it should be `_`.

Neeratyoy: Okay, done!

mfeurer (approved):

openml/extensions/sklearn/extension.py
Neeratyoy: is this resolved now then?

examples/30_extended/run_setup_tutorial.py
```diff
@@ -52,22 +57,39 @@
 # we will create a fairly complex model, with many preprocessing components and
 # many potential hyperparameters. Of course, the model can be as complex and as
 # easy as you want it to be
-model_original = sklearn.pipeline.make_pipeline(
-    sklearn.impute.SimpleImputer(), sklearn.ensemble.RandomForestClassifier()
-)
 
+from sklearn.ensemble import HistGradientBoostingClassifier
```
PGijsbers, line 61: What's the reason we are using the experimental `HistGradientBoostingClassifier` instead of a `RandomForestClassifier`?

examples/30_extended/run_setup_tutorial.py
```diff
@@ -37,6 +37,11 @@
 import sklearn.ensemble
 import sklearn.impute
 import sklearn.preprocessing
```
PGijsbers, line 39: I don't think we need to import these separately as their used components get imported below?

examples/30_extended/study_tutorial.py
```diff
@@ -68,7 +71,7 @@
 )
 print(evaluations.head())
 
-############################################################################
+###########################################################from openml.testing import cat, cont#################
```
PGijsbers, line 74: The `from openml.testing import cat, cont` looks misplaced?

examples/30_extended/study_tutorial.py
```diff
@@ -17,8 +17,11 @@
 
 import numpy as np
 import sklearn.tree
```
PGijsbers, line 19: I don't think `tree` is used anymore?

openml/datasets/functions.py
```diff
@@ -815,12 +815,12 @@ def edit_dataset(
...
       Edits an OpenMLDataset.
-      Specify atleast one field to edit, apart from data_id
+      Specify at least one field to edit, apart from data_id
        - For certain fields, a new dataset version is created : attributes, data,
        default_target_attribute, ignore_attribute, row_id_attribute.
 
-       - For other fields, the uploader can edit the exisiting version.
-        Noone except the uploader can edit the exisitng version.
+       - For other fields, the uploader can edit the existing version.
+        No one except the uploader can edit the existing version.
```
PGijsbers, line 823: Perhaps we should exclude these changes so there won't be a merge conflict from both this and #944 (these changes should be overwritten by #944).

PGijsbers (changes requested):
Before I dive into the meat of this PR (which I will, with additional comments), I figured I would already add change requests/clarifications to get things moving :)

openml/extensions/sklearn/extension.py
```diff
@@ -1579,24 +1579,21 @@ def _run_model_on_fold(
...
-        arff_tracecontent :  List[List]
-            Arff representation (list of lists) of the trace data that was generated by this
-            fold
-            (will be used to populate trace.arff, leave it empty if the model did not perform
-            any
-            hyperparameter optimization).
+        pred_y : np.ndarray
+            Predictions on the training/test set, depending on the task type.
+            For supervised tasks, predicitons are on the test set.
+            For unsupervised tasks, predicitons are on the training set.
```
PGijsbers, line 1585:
```suggestion
            For supervised tasks, predictions are on the test set.
            For unsupervised tasks, predicitions are on the training set.
```

openml/extensions/sklearn/extension.py
```diff
@@ -1546,7 +1546,7 @@ def _run_model_on_fold(
         fold_no: int,
         y_train: Optional[np.ndarray] = None,
         X_test: Optional[Union[np.ndarray, scipy.sparse.spmatrix, pd.DataFrame]] = None,
-    ) -> Tuple[np.ndarray, np.ndarray, "OrderedDict[str, float]", Optional[OpenMLRunTrace]]:
+    ) -> Tuple[np.ndarray, pd.DataFrame, "OrderedDict[str, float]", Optional[OpenMLRunTrace]]:
```
PGijsbers, line 1549:
```suggestion
    ) -> Tuple[np.ndarray, Optional[pd.DataFrame], "OrderedDict[str, float]", Optional[OpenMLRunTrace]]:
```
According to the docstring, the second return value may be `None` if it is not a classification or learning curve task.

openml/extensions/sklearn/extension.py
```diff
@@ -1579,24 +1579,21 @@ def _run_model_on_fold(
...
-            Arff representation (list of lists) of the trace data that was generated by this
-            fold
-            (will be used to populate trace.arff, leave it empty if the model did not perform
-            any
-            hyperparameter optimization).
+        pred_y : np.ndarray
+            Predictions on the training/test set, depending on the task type.
+            For supervised tasks, predicitons are on the test set.
+            For unsupervised tasks, predicitons are on the training set.
+        proba_y : pd.DataFrame
```
PGijsbers, line 1586:
```suggestion
        proba_y : pd.DataFrame, optional
```

openml/extensions/sklearn/extension.py
```diff
@@ -1579,24 +1579,21 @@ def _run_model_on_fold(
...
+            Predicted probabilities for the test set.
+            None, if task is not Classification or Learning Curve prediction.
         user_defined_measures : OrderedDict[str, float]
             User defined measures that were generated on this fold
-        model : Any
-            The model trained on this repeat,fold,subsample triple. Will be used to generate
-            trace
-            information later on (in ``obtain_arff_trace``).
+        trace : Optional[OpenMLRunTrace]]
```
PGijsbers, line 1591:
```suggestion
        trace : OpenMLRunTrace, optional
```

I think? That's at least how it is typical in doc string parameters, I'm not sure why return values would go by different rules.

openml/extensions/sklearn/extension.py
```diff
@@ -1609,16 +1606,31 @@ def _prediction_to_probabilities(y: np.ndarray, classes: List[Any]) -> np.ndarra
...
+                raise ValueError("please convert model classes to list prior to calling this fn")
+            # DataFrame allows more accurate mapping of classes as column names
+            result = pd.DataFrame(
+                0, index=np.arange(len(y)), columns=model_classes, dtype=np.float32
+            )
+            for obs, prediction in enumerate(y):
+                result.loc[obs, prediction] = 1.0
             return result
 
         if isinstance(task, OpenMLSupervisedTask):
```
PGijsbers, line 1636: Sorry for doing it this way. But a few lines below there is a `TODO`  in comments in the code, I can't mark it on Github. Can you remove the TODO and create a Github issue for it?

openml/extensions/sklearn/extension.py
```diff
@@ -1677,6 +1689,16 @@ def _prediction_to_probabilities(y: np.ndarray, classes: List[Any]) -> np.ndarra
...
 
+            if not isinstance(model_classes, list):
+                model_classes = model_classes.tolist()
+
+            # to handle the case when dataset is numpy and categories are encoded
+            # however the class labels stored in task are still categories
+            if isinstance(y_train, np.ndarray) and isinstance(
+                cast(List, task.class_labels)[0], str
+            ):
+                model_classes = [cast(List[str], task.class_labels)[i] for i in model_classes]
```
PGijsbers, line 1700: What guarantees this mapping is correct?

openml/extensions/sklearn/extension.py
```diff
@@ -1726,20 +1749,24 @@ def _prediction_to_probabilities(y: np.ndarray, classes: List[Any]) -> np.ndarra
...
                         proba_y.shape[1], len(task.class_labels),
                     )
                     warnings.warn(message)
                     openml.config.logger.warn(message)
+
+                    for i, col in enumerate(task.class_labels):
+                        # adding missing columns with 0 probability
+                        if col not in model_classes:
+                            proba_y[col] = 0
+                    proba_y = proba_y[task.class_labels]
```
PGijsbers, line 1762: I'd add a comment that we do this to re-order the columns.

openml/extensions/sklearn/extension.py
PGijsbers: I think I understand it now. I assume it's us who are doing the encoding.

openml/extensions/sklearn/extension.py
PGijsbers:
```suggestion
                    # We re-order the columns to move possibly added missing columns into place.
                    proba_y = proba_y[task.class_labels]
```

openml/extensions/sklearn/extension.py
```diff
@@ -1609,16 +1606,31 @@ def _prediction_to_probabilities(y: np.ndarray, classes: List[Any]) -> np.ndarra
 
             Returns
             -------
-            np.ndarray
+            pd.DataFrame
             """
+
+            if isinstance(task, (OpenMLClassificationTask, OpenMLLearningCurveTask)):
```
PGijsbers, line 1612: As far as I can tell this function is only called both of the first conditions are met (correct task type and has `class_labels`. It's a little hard to tell from Github though. But if that's correct, you can greatly simplify the flow of this function (and I would encourage you to do so).

PGijsbers (approved):
I'll approve this PR so we can push a release before the workshop. Please open a new Github issue which links back to this PR so we an easy reference to some further improvements that can be done in these modules/examples.

openml/extensions/sklearn/extension.py
Neeratyoy: Done.
#968

examples/30_extended/study_tutorial.py
Neeratyoy: Done.

examples/30_extended/study_tutorial.py
Neeratyoy: Fixed.

openml/extensions/sklearn/extension.py
Neeratyoy: Fixed.

openml/extensions/sklearn/extension.py
Neeratyoy: Done.

examples/30_extended/run_setup_tutorial.py
Neeratyoy: Done.

openml/datasets/functions.py
Neeratyoy: This is no longer relevant given the merge has been done right?

mfeurer: I think so

examples/30_extended/run_setup_tutorial.py
Neeratyoy: Addressed in #967.

Issue #730, opened 2019-07-09
Guidelines for publishing OpenML-python extensions


joaquinvanschoren:
#### Description
OpenML has an extension API, and you can use it to connect other libraries to OpenML. It is however unclear how one should later up 'publish' this extension so that other people can use it.

#### Steps/Code to Reproduce
Here is an example of an extension that adds full support for Keras, MXnet, PyTorch and ONNX:
https://github.com/openml/openml-deeplearning/

It is currently a clone of the pull python API with the extension added in separate folders: https://github.com/openml/openml-deeplearning/tree/master/openml/extensions

Each folder has an extensions.py file and usually a config.py file with additional configuration functionality.

It also has documentation in the `examples` folder and test in the `tests` folder.

It is currently installable as a standalone package through pip. However, this may lead to divergence between this package and the core python API.

#### Expected Results
Clear guidelines on how extensions can be published and imported. It should allow independent testing and documentation.

If this makes sense, maybe also define a way to 'register' extensions so it is easier to find and install them.

Thanks!

joaquinvanschoren:
Basically:

* Rip out everything except the extension class which extends the interface
* Register the extension: https://openml.github.io/openml-python/master/generated/openml.extensions.register_extension.html#openml.extensions.register_extension
* Set up openml-python extensions organization
* Unit tests: every new repo need to set up own CI
* Documentation: also local?

janvanrijn:
> Rip out everything except the extension class which extends the interface

Except this step. It should be possible to build an extension from scratch, by including the `openml-python` package.

mfeurer:
Added via #778.

PR #778, opened 2019-09-05
Instructions to publish new extensions

Neeratyoy:
#### Reference Issue

Addresses #730. 

Adds a section _Connecting new machine learning libraries_ to the _Contributing_ section of the documentation page giving guidelines on how a new extension can be published, used and maintained.

### DIFF
--- a/doc/contributing.rst
+++ b/doc/contributing.rst
@@ -158,5 +158,25 @@ Happy testing!
 Connecting new machine learning libraries
 =========================================
 
-Coming soon - please stay tuned!
+Each extension created should be a stand-alone repository, compatible with the
+`OpenML repository <https://github.com/openml/openml-python>`_.
+The extension repository should work off-the-shelf with openml-python installed.
+It should follow the structure:
 
+::
+
+| openml
+|   |-- extensions
+|   |    |-- [extension name]
+|   |    |    |-- __init__.py
+|   |    |    |-- extension.py
+|   |    |    |-- config.py (optionally)
+|   |    |-- sklearn
+|   |    |    |-- __init__.py
+|   |    |    |-- extension.py
+|   |    |-- keras
+|   |    |    |-- __init__.py
+|   |    |    |-- extension.py
+
+The `sklearn extension <https://github.com/openml/openml-python/tree/develop/openml/extensions>`_ packaged with the OpenML repository can be used as a template/benchmark
+to build the extension.
--- a/doc/contributing.rst
+++ b/doc/contributing.rst
@@ -166,11 +166,11 @@ machine learning libraries can be beneficial for the community, please refer
 `here <https://openml.github.io/OpenML/>`_.
 
 Various `examples <https://openml.github.io/openml-python/develop/examples/index.html>`_
-can show how OpenML works in
+can show how OpenML-Python works in
 practice.
 Moreover `this example <https://openml.github.io/openml-python/develop/examples/
 flows_and_runs_tutorial.html>`_
-shows how scikit-learn currently works with OpenMl as an extension. The
+shows how scikit-learn currently works with OpenML-Python as an extension. The
 `sklearn extension <https://github.com/openml/openml-python/tree/develop/openml/extensions>`_
 packaged with the `openml-python <https://github.com/openml/openml-python>`_
 repository can be used as a template/benchmark to build the extension.
@@ -179,7 +179,7 @@ repository can be used as a template/benchmark to build the extension.
 API
 +++
 * The extension scripts should be importing `openml` and be able to execute any
-  function from the OpenML `API <https://openml.github.io/openml-python/develop/api.html#>`_.
+  function from the OpenML-Python `API <https://openml.github.io/openml-python/develop/api.html#>`_.
 * The extension class defined here in the extension script should inherit the `openml extension
   class <https://openml.github.io/openml-python/develop/generated/openml.extensions.
   Extension.html#openml.extensions.Extension>`_.
@@ -190,11 +190,11 @@ API
   is a good benchmark to follow.
 
 
-Interfacing with OpenML
-+++++++++++++++++++++++
-Once the new extension class has been defined, the openml module to `register extensions
+Interfacing with OpenML-Python
+++++++++++++++++++++++++++++++
+Once the new extension class has been defined, the openml-python module to `register extensions
 <https://openml.github.io/openml-python/develop/generated/openml.extensions.register_
-extension.html#openml.extensions.register_extension>`_ must be called to allow OpenML to
+extension.html#openml.extensions.register_extension>`_ must be called to allow OpenML-Python to
 interface the new extension.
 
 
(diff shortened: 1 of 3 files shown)


doc/contributing.rst
```diff
@@ -158,5 +158,81 @@ Happy testing!
...
+`OpenML repository <https://github.com/openml/openml-python>`_.
+The extension repository should work off-the-shelf with openml-python installed.
+
+Create a `public Github repo <https://help.github.com/en/articles/create-a-repo>`_ with
+the following directory structure:
+
+::
+
+| [repo name]
+|   |-- extensions
```
mfeurer: We don't need this directory level. In fact, it might rather been hurtful as the user then cannot install the extension as a package if it is called `extensions`.

doc/contributing.rst
```diff
@@ -158,5 +158,81 @@ Happy testing!
...
++++
+* The extension scripts should be importing `openml` and be able to execute any
+  function from the OpenML `API <https://openml.github.io/openml-python/develop/api.html#>`_.
+* The extension class defined here should inherit the `openml extension
+  class <https://openml.github.io/openml-python/develop/generated/openml.extensions.
+  Extension.html#openml.extensions.Extension>`_.
+* The class should have the functions from `class Extension` overloaded as required.
+  The `Sklearn Extension API <https://openml.github.io/openml-python/develop/generated/
+  openml.extensions.sklearn.SklearnExtension.html#openml.extensions.sklearn.SklearnExtension>`_
+  is a good checklist to follow.
```
mfeurer: checklist -> example ?

doc/contributing.rst
```diff
@@ -158,5 +158,81 @@ Happy testing!
...
+
+
+Interfacing with OpenML
++++++++++++++++++++++++
+Once the new extension class has been defined, the openml module to `register extensions
+<https://openml.github.io/openml-python/develop/generated/openml.extensions.register_
+extension.html#openml.extensions.register_extension>`_ can be called to allow OpenML
+interface the new extension. For example, the `sklearn extension registering
```
mfeurer: interface -> interfacing

doc/contributing.rst
```diff
@@ -158,5 +158,81 @@ Happy testing!
...
+<https://github.com/openml/openml-python/blob/develop/openml/extensions/sklearn/
+extension.py#L1744>`_.
```
mfeurer: Could you please not refer to specific lines? They might change in the future and make the docs wrong. Could you instead please try to move the registring of the extension to `__init__.zy` and simply link that file?

doc/contributing.rst
```diff
@@ -158,5 +158,81 @@ Happy testing!
 Connecting new machine learning libraries
 =========================================
 
-Coming soon - please stay tuned!
-
+Hosting the library
+~~~~~~~~~~~~~~~~~~~
+
+Each extension created should be a stand-alone repository, compatible with the
+`OpenML repository <https://github.com/openml/openml-python>`_.
```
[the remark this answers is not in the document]
Neeratyoy: Yes. Will change it.

doc/contributing.rst
```diff
@@ -158,5 +158,81 @@ Happy testing!
...
+
+::
+
+| [repo name]
+|   |-- extensions
+|   |    |-- [extension name]
+|   |    |    |-- __init__.py
+|   |    |    |-- extension.py
+|   |    |    |-- config.py (optionally)
+|   |    |-- sklearn
```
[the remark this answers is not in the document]
Neeratyoy: I meant it as an example that one repo to be published as an OpenML extension, could house multiple extensions supporting different libraries.
Meant to portray [this](https://github.com/openml/openml-deeplearning/tree/master/openml/extensions).

doc/contributing.rst
Neeratyoy: In that case, should I include a setup.py in the directory tree, to indicate or reinforce that we encourage them to publish as a separate package?

doc/contributing.rst
Neeratyoy: I wanted to say that when defining the API for the new extension, one could use the existing [sklearn API](https://openml.github.io/openml-python/develop/generated/openml.extensions.sklearn.SklearnExtension.html#openml.extensions.sklearn.SklearnExtension) to use as a checklist for which functions need to be overloaded from the base class Extension.

doc/contributing.rst
Neeratyoy: Are you referring to [this](https://github.com/openml/openml-python/blob/develop/openml/extensions/sklearn/__init__.py) file?
So in this file, I'll import `register_extension` and call `register_extension(SklearnExtension)`?
And remove `register_extension(SklearnExtension)` from the current file it is in?

doc/contributing.rst
```diff
@@ -158,5 +158,81 @@ Happy testing!
...
 =========================================
 
-Coming soon - please stay tuned!
-
+Hosting the library
+~~~~~~~~~~~~~~~~~~~
+
+Each extension created should be a stand-alone repository, compatible with the
+`OpenML repository <https://github.com/openml/openml-python>`_.
+The extension repository should work off-the-shelf with openml-python installed.
```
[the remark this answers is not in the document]
Neeratyoy: I'm sorry. Couldn't quite grasp the requirement.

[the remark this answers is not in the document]
mfeurer: You're right, that's actually not necessary. But it would be great if you could write OpenML-Python (camel case).

doc/contributing.rst
```diff
@@ -158,5 +158,73 @@ Happy testing!
...
+To get an understanding of what OpenML does and why support of the latest
+machine learning libraries can be beneficial for the community, please refer
+`here <https://openml.github.io/OpenML/>`_.
+
+Various `examples <https://openml.github.io/openml-python/develop/examples/index.html>`_
+can show how OpenML works in
+practice.
+Moreover `this example <https://openml.github.io/openml-python/develop/examples/
+flows_and_runs_tutorial.html>`_
+shows how scikit-learn currently works with OpenMl as an extension. The
```
mfeurer: Do you mean OpenML-Python instead of OpenML here? Anyway, this should be OpenML instead of OpenMl (camelcase of the l).

doc/contributing.rst
```diff
@@ -158,5 +158,73 @@ Happy testing!
...
+|    |    |-- __init__.py
+|    |    |-- extension.py
+|    |    |-- config.py (optionally)
+
+
+
+Recommended
+~~~~~~~~~~~
+* Test cases to keep the extension up to date with the `openml-python` upstream changes.
+* Documentation of the extension API, especially if any new functionality added to OpenML's
```
mfeurer: These also should be OpenML-Python instead of OpenML, right?

doc/contributing.rst
mfeurer: Yes, thank you. I'm actually only seeing this now. You're using the full URL, but you can simply specify to which part of the documentation a link should go, see for example here: https://github.com/openml/openml-python/blob/develop/doc/index.rst

Then the exact links are auto-generated and automatically point to the right branch (develop or master).

doc/contributing.rst
mfeurer: I think this applies to almost the whole PR.

doc/contributing.rst
Neeratyoy: Hi,
I guess I understand what you mean here. However, in this section [here](https://github.com/openml/openml-python/blob/develop/doc/index.rst#example) where the examples gallery is linked, the generated URL links to [this](https://github.com/openml/openml-python/blob/develop/doc/examples/index.html). 
Whereas, what we would want is probably [this](https://openml.github.io/openml-python/develop/examples/index.html). 
Is that so?

PR #787, opened 2019-09-27
Fix 779

### LINKED ISSUE #779: openml-python does not recognize corrupt dataset.pkl.py3 file
People are experiencing issues downloading dataset `41081` [here](https://github.com/ML-course/ML-course/issues/19).
Investigation is on-going, but it looks like the Python process gets killed when pickling the dataset to disk.
Subsequent loads of the cached data fail because the `dataset.pkl.py3` file is created (and sits at 0 bytes). Because the file exists it is attempted to be loaded without further sanity checks. I think the correct behavior would to check for this faulty file and if it exists either (1) raise an error or (2) delete the file, attempt to pickle is again, and raise an error only if this fails. I would prefer option 2.

edit: to clarify, this issue is about trying to mitigate the effect of corrupt files in cache. Not about the fact that the process gets killed (possibly during pickling). That warrants a separate issue which I will open after we have some more information from the linked issue.

PGijsbers:
Fixes #779. If the pickle file is corrupt, data will be loaded from arff instead.

Also:
 - refactor some of the data loading flow
 - prevent `_convert_array_format` from issuing warnings on `np.ndarray -> np.ndarray` and `pd.DataFrame -> pd.DataFrame`  conversions.

### DIFF
--- a/openml/datasets/dataset.py
+++ b/openml/datasets/dataset.py
@@ -414,7 +414,7 @@ class OpenMLDataset(object):
         return data_pickle_file
 
     def _load_data(self):
-        """ Load data from pickle or, if needed, arff. Download data first if not present on disk. """
+        """ Load data from pickle or arff. Download data first if not present on disk. """
         if self.data_pickle_file is None:
             if self.data_file is None:
                 self._download_data()
--- a/openml/datasets/dataset.py
+++ b/openml/datasets/dataset.py
@@ -4,7 +4,7 @@ import io
 import logging
 import os
 import pickle
-from typing import List, Optional, Union, Tuple, Iterable
+from typing import List, Optional, Union, Tuple, Iterable, Dict
 
 import arff
 import numpy as np
@@ -168,7 +168,7 @@ class OpenMLDataset(object):
         self.qualities = _check_qualities(qualities)
 
         if data_file is not None:
-            self.data_pickle_file = self._data_arff_to_pickle(data_file)
+            self.data_pickle_file = self._create_pickle_in_cache(data_file)
         else:
             self.data_pickle_file = None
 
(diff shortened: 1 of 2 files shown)


openml/datasets/dataset.py
```diff
@@ -295,17 +372,73 @@ def _data_arff_to_pickle(self, data_file):
...
+        """ Parse the arff and pickle the result. Update any old pickle objects. """
+        data_pickle_file = data_file.replace('.arff', '.pkl.py3')
+        if os.path.exists(data_pickle_file):
+            # Load the data to check if the pickle file is outdated (i.e. contains numpy array)
+            with open(data_pickle_file, "rb") as fh:
+                try:
+                    data, categorical, attribute_names = pickle.load(fh)
+                except EOFError:
+                    # The file is likely corrupt, see #780.
+                    # We deal with this when loading the data in `_load_data_from_disk`.
```
mfeurer: Do you mean `_load_data`?

openml/datasets/dataset.py
```diff
@@ -295,17 +372,73 @@ def _data_arff_to_pickle(self, data_file):
...
+        if self.data_pickle_file is None:
+            if self.data_file is None:
+                self._download_data()
+            self.data_pickle_file = self._create_pickle_in_cache(self.data_file)
+
+        try:
+            with open(self.data_pickle_file, "rb") as fh:
+                data, categorical, attribute_names = pickle.load(fh)
+        except EOFError:
+            warn(
```
mfeurer: Could you please use the logger here?

openml/datasets/dataset.py
```diff
@@ -202,26 +202,104 @@ def __repr__(self):
...
+            with io.open(filename, encoding='utf8') as fh:
+                return decode_arff(fh)
+
+    def _parse_data_from_arff(
+            self,
+            arff_file_path: str
+    ) -> Tuple[Union[pd.DataFrame, scipy.sparse.csr_matrix], List[bool], List[str]]:
+        """ Parse all required data from arff file.
+
+        :param arff_file_path: str
```
mfeurer: Could you please format docstrings according to the numpydoc format? We're currently using this and it would be great if we stay consistent.

openml/datasets/dataset.py
```diff
@@ -202,26 +202,104 @@ def __repr__(self):
...
+            DataFrame or csr_matrix: dataset
+            List[bool]: List indicating which columns contain categorical variables.
+            List[str]: List of column names.
+        """
         try:
             data = self._get_arff(self.format)
         except OSError as e:
-            logger.critical("Please check that the data file %s is "
-                            "there and can be read.", data_file)
+            logger.critical("Please check that the data file {} is "
```
mfeurer: I'm not sure if we want to interpolate the strings ourselves for log messages (see this https://www.reddit.com/r/Python/comments/387oog/when_will_logging_use_newstyle_string_formatting/ for reasons). What do you think?

openml/datasets/dataset.py
```diff
@@ -202,26 +202,104 @@ def __repr__(self):
...
+        # import required here to avoid circular import.
+        from .functions import _get_dataset_arff
+        self.data_file = _get_dataset_arff(self)
+
+    def _get_arff(self, format: str) -> Dict:
+        """Read ARFF file and return decoded arff.
+
+        Reads the file referenced in self.data_file.
+
+        :param format: str
```
mfeurer: Could you please use the numpydoc format here?

tests/test_datasets/test_dataset.py
```diff
@@ -190,6 +191,20 @@ def test_get_data_with_nonexisting_class(self):
...
         self.assertEqual(np.sum(y == 3), 0)
 
+    def test_get_data_corrupt_pickle(self):
+        # lazy loaded dataset, populate cache
+        self.iris.get_data()
+        # corrupt pickle file, overwrite as empty.
+        with open(self.iris.data_pickle_file, "w") as fh:
+            fh.write("")
+
+        with catch_warnings():
```
mfeurer: Could you please also check for the error message? I find it way easier to to parse the code then when I know what error message is tested for.

openml/datasets/dataset.py
PGijsbers: Yep, PyCharm did not rename the comment.

openml/datasets/dataset.py
PGijsbers: Had that initially but I saw `warn` used in the format conversion function, and figured this is more important, so adopted `warn`. Will change to log.

tests/test_datasets/test_dataset.py
PGijsbers: I removed this check since we now use the logger instead of raising an actual warning.

PGijsbers:
I reverted the change to shortcut the conversion warning.
The warning is now written to log instead of raised.
The problem is that it is also called in dataset#648 to convert the `y` data from pd.Series to pd.DataFrame, which it wouldn't do.

PR #789, opened 2019-09-30
Improve the usage of dataframes in examples

mfeurer:
By making use of automatic conversion unless there are pedagogic reasons to perform the cast manually.

### DIFF
--- a/examples/30_extended/tasks_tutorial.py
+++ b/examples/30_extended/tasks_tutorial.py
@@ -73,24 +73,21 @@ print(len(filtered_tasks))
 #
 # Similar to listing tasks by task type, we can list tasks by tags:
 
-tasks = openml.tasks.list_tasks(tag='OpenML100')
-tasks = pd.DataFrame.from_dict(tasks, orient='index')
+tasks = openml.tasks.list_tasks(tag='OpenML100', output_format='dataframe')
 print("First 5 of %s tasks:" % len(tasks))
 print(tasks.head())
 
 ############################################################################
 # Furthermore, we can list tasks based on the dataset id:
 
-tasks = openml.tasks.list_tasks(data_id=1471)
-tasks = pd.DataFrame.from_dict(tasks, orient='index')
+tasks = openml.tasks.list_tasks(data_id=1471, output_format='dataframe')
 print("First 5 of %s tasks:" % len(tasks))
 print(tasks.head())
 
 ############################################################################
 # In addition, a size limit and an offset can be applied both separately and simultaneously:
 
-tasks = openml.tasks.list_tasks(size=10, offset=50)
-tasks = pd.DataFrame.from_dict(tasks, orient='index')
+tasks = openml.tasks.list_tasks(size=10, offset=50, output_format='dataframe')
 print(tasks)
 
 ############################################################################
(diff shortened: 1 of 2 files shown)


examples/30_extended/tasks_tutorial.py
```diff
@@ -106,8 +103,7 @@
 # Finally, it is also possible to list all tasks on OpenML with:
 
 ############################################################################
-tasks = openml.tasks.list_tasks()
-tasks = pd.DataFrame.from_dict(tasks, orient='index')
+tasks = openml.tasks.list_tasks(output_format='dataframe')
 print(len(tasks))
```
janvanrijn: `logging.info()`?

examples/30_extended/tasks_tutorial.py
```diff
@@ -106,8 +103,7 @@
 # Finally, it is also possible to list all tasks on OpenML with:
 
 ############################################################################
-tasks = openml.tasks.list_tasks()
-tasks = pd.DataFrame.from_dict(tasks, orient='index')
+tasks = openml.tasks.list_tasks(output_format='dataframe')
```
janvanrijn: didn't know this new feature. Are examples somehow tested?

mfeurer: Yeah, that's pretty handy. Once we have closed all more important issues we should also make this the default behavior ;)

Examples are tested by running while building the documentation. They should make the unit test fail if they fail themselves.

examples/30_extended/tasks_tutorial.py
mfeurer: I disagree for the same reason as stated [here](https://github.com/openml/openml-python/pull/791/files#r330119323).

examples/30_extended/tasks_tutorial.py
```diff
@@ -73,24 +73,21 @@
 #
 # Similar to listing tasks by task type, we can list tasks by tags:
 
-tasks = openml.tasks.list_tasks(tag='OpenML100')
-tasks = pd.DataFrame.from_dict(tasks, orient='index')
+tasks = openml.tasks.list_tasks(tag='OpenML100', output_format='dataframe')
```
ArlindKadra: I think we should add this in the `tasks_tutorial.py`:
```
#   * Use the output_format parameter to select output type
#   * Default gives 'dict' (other option: 'dataframe', see below)
```

https://github.com/openml/openml-python/blob/6582cbd52e84c10811c6eb303401c9101a252e6b/examples/30_extended/tasks_tutorial.py#L34-L36

The above makes it seem like there is only one return type (a dictionary of dictionaries) while the return type is controlled by the `output_format`.

mfeurer: I rephrased the example. I hope this makes it a bit better.

mfeurer: (the actual information you wanted was already there, but just not as easily visible as for the dataset tutorial).

ArlindKadra: > (the actual information you wanted was already there, but just not as easily visible as for the dataset tutorial).

No, I found it, however I still thought it would be confusing for a new user.

ArlindKadra: > I rephrased the example. I hope this makes it a bit better.

Yes, it looks way better to me now.

PR #791, opened 2019-10-01
Add example

mfeurer:
* adds example for Feurer et al. (2015)
* removes the stub for Fusi et al. (2018) as they actually perform the
  same task. I can't create an example, though, as they used regression
  datasets for classification (and OpenML by now forbids creating such
  tasks).


examples/40_paper/2015_neurips_feurer_example.py
```diff
@@ -10,9 +10,81 @@
...
+    803, 806, 807, 813, 816, 819, 821, 822, 823, 833, 837, 843, 845, 846, 847,
+    849, 866, 871, 881, 897, 901, 903, 904, 910, 912, 913, 914, 917, 923, 930,
+    934, 953, 958, 959, 962, 966, 971, 976, 977, 978, 979, 980, 991, 993, 995,
+    1000, 1002, 1018, 1019, 1020, 1021, 1036, 1040, 1041, 1049, 1050, 1053,
+    1056, 1067, 1068, 1069, 1111, 1112, 1114, 1116, 1119, 1120, 1128, 1130,
+    1134, 1138, 1139, 1142, 1146, 1161, 1166,
+]
+
+####################################################################################################
+# The dataset IDs could be used directly to load the dataset and split the data into a training
```
janvanrijn: Are you sure you want to start with the dataset ids, rather than the task ids? 

If the answer is yes, this clearly signals that we do not have any good procedures for "getting tasks that belong to a given set of datasets". We should either extend the API to support this better or provide the functions below as convenience function (or combination of both)

examples/40_paper/2015_neurips_feurer_example.py
```diff
@@ -10,9 +10,81 @@
...
+    output_format='dataframe',
+)
+
+# query only those with holdout as the resampling startegy.
+tasks_a = tasks_a[(tasks_a.estimation_procedure == "33% Holdout set")]
+
+# deactivated tasks
+tasks_d = openml.tasks.list_tasks(
+    task_type_id=1,
+    status='deactivated',
```
janvanrijn: why not search for status "all" ?

examples/40_paper/2015_neurips_feurer_example.py
```diff
@@ -10,9 +10,81 @@
...
+    #         (task.target_name, dataset.default_target_attribute)
+    #     )
+
+    task_ids.append(task_id)
+
+assert len(task_ids) == 140
+task_ids.sort()
+
+# These are the tasks to work with:
+print(task_ids)
```
janvanrijn: `logging.info`?

examples/40_paper/2015_neurips_feurer_example.py
mfeurer: The reasoning here is to stay close to the Auto-sklearn paper, where only dataset IDs are given. What kind of convenience function would you like to have? Something like:
```
def get_tasks_for_dataset(
    dataset_id: int,
    task_type_id: int,
    estimation_procedure: str,
    status: str,
    check_target_attribute: bool,
) -> List:
    pass
```

examples/40_paper/2015_neurips_feurer_example.py
mfeurer: Lack of knowledge, I'll update the example.

examples/40_paper/2015_neurips_feurer_example.py
mfeurer: I will also make the note more drastic.

examples/40_paper/2015_neurips_feurer_example.py
mfeurer: I think print is fine for examples. logging is only important for the library itself to make the amount of output controllable.

mfeurer:
Thanks for the review, I hope I could address your comments.

examples/40_paper/2015_neurips_feurer_example.py
```diff
@@ -10,9 +10,80 @@
 ~~~~~~~~~~~
 
 | Efficient and Robust Automated Machine Learning
-| Matthias Feurer, Aaron Klein, Katharina Eggensperger, Jost Springenberg, Manuel Blum and Frank Hutter
+| Matthias Feurer, Aaron Klein, Katharina Eggensperger, Jost Springenberg, Manuel Blum and Frank Hutter  # noqa F401
```
ArlindKadra: @mfeurer wrong usage of # noqa F401 in the text ? It is not interpreted as a comment.
Maybe you meant here:
https://github.com/openml/openml-python/blob/5a2830cc494dbffe93fb324aa7eb98b8bf3f0b33/examples/40_paper/2015_neurips_feurer_example.py#L18

examples/40_paper/2015_neurips_feurer_example.py
```diff
@@ -10,9 +10,80 @@
...
+    803, 806, 807, 813, 816, 819, 821, 822, 823, 833, 837, 843, 845, 846, 847,
+    849, 866, 871, 881, 897, 901, 903, 904, 910, 912, 913, 914, 917, 923, 930,
+    934, 953, 958, 959, 962, 966, 971, 976, 977, 978, 979, 980, 991, 993, 995,
+    1000, 1002, 1018, 1019, 1020, 1021, 1036, 1040, 1041, 1049, 1050, 1053,
+    1056, 1067, 1068, 1069, 1111, 1112, 1114, 1116, 1119, 1120, 1128, 1130,
+    1134, 1138, 1139, 1142, 1146, 1161, 1166,
+]
+
+####################################################################################################
+# The dataset IDs could be used directly to load the dataset and split the data into a training
```
ArlindKadra: *training set

examples/40_paper/2015_neurips_feurer_example.py
```diff
@@ -10,9 +10,80 @@
...
+# and a test set. However, to be reproducible, we will first obtain the respective tasks from
+# OpenML, which define both the target feature and the train/test split.
+#
+# .. note::
+#    It is discouraged to work directly on datasets and only provide dataset IDs in a paper as
+#    this does not allow reproducibility (unclear splitting). Please do not use datasets but the
+#    respective tasks as basis for a paper and publish task IDS. This example is only given to
+#    showcase the use OpenML-Python for a published paper and as a warning on how not to do it.
```
ArlindKadra: the use of OpenML-Python*

PR #792, opened 2019-10-01
Add new convenience function get_flow_id

mfeurer:
## This adds:

* function `get_flow_id`
* a unit test
* an example/tutorial showcasing the function's usage

## Reason

Replaces:

```openml.extensions.get_extension_by_model(clf).model_to_flow(clf).publish().flow_id```

by

```openml.flows.get_flow_id(model=clf)```

and also allows to query for all flows of a given model across all external versions:

```openml.flows.get_flow_id(model=clf, exact_version=False)```

### DIFF
--- a/examples/30_extended/flow_id_tutorial.py
+++ b/examples/30_extended/flow_id_tutorial.py
@@ -48,7 +48,6 @@ print(flow_id)
 # flows with the same name for different versions of a software. This might be necessary if an
 # algorithm or implementation introduces, renames or drop hyperparameters over time.
 
-flow = openml.extensions.get_extension_by_model(clf).model_to_flow(clf)
 print(flow.name, flow.external_version)
 
 ####################################################################################################
--- a/examples/30_extended/flow_id_tutorial.py
+++ b/examples/30_extended/flow_id_tutorial.py
@@ -59,7 +59,7 @@ print(flow_id)
 
 ####################################################################################################
 # We can also retrieve all flows for a given name:
-flow_ids = openml.flows.get_flow_id(name=flow.name, exact_version=False)
+flow_ids = openml.flows.get_flow_id(name=flow.name)
 print(flow_ids)
 
 ####################################################################################################
--- a/examples/30_extended/flow_id_tutorial.py
+++ b/examples/30_extended/flow_id_tutorial.py
@@ -1,0 +1,69 @@
+"""
+==================
+Obtaining Flow IDs
+==================
+
+This tutorial discusses different ways to obtain the ID of a flow in order to perform further
+analysis.
+"""
+
+####################################################################################################
+import sklearn.tree
+
+import openml
+
+clf = sklearn.tree.DecisionTreeClassifier()
+
+####################################################################################################
+# 1. Obtaining a flow given a classifier
+# ======================================
+#
+
+flow = openml.extensions.get_extension_by_model(clf).model_to_flow(clf).publish()
+flow_id = flow.flow_id
+print(flow_id)
+
+####################################################################################################
+# This piece of code is rather involved. First, it retrieves an
+# :class:`~openml.extensions.Extension` which is registered and can handle the given model,
+# in our case it is :class:`openml.extensions.sklearn.SklearnExtension`. Second, the extension
+# converts the classifier into an instance of :class:`openml.flow.OpenMLFlow`. Third and finally,
+# the publish method checks whether the current flow is already present on OpenML. If not,
+# it uploads the flow, otherwise, it updates the current instance with all information computed
+# by the server (which is obviously also done when uploading/publishing a flow).
+#
+# To simplify the usage we have created a helper function which automates all these steps:
+
+flow_id = openml.flows.get_flow_id(model=clf)
+print(flow_id)
+
+####################################################################################################
+# 2. Obtaining a flow given its name
+# ==================================
+# The schema of a flow is given in XSD (`here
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: examples/30_extended/flow_id_tutorial.py [hunk 1]
```diff
@@ -0,0 +1,69 @@
...
+####################################################################################################
+# The name and external version are automatically added to a flow when constructing it from a
+# model. We can then use them to retrieve the flow id as follows:
+
+flow_id = openml.flows.flow_exists(name=flow.name, external_version=flow.external_version)
+print(flow_id)
+
+####################################################################################################
+# We can also retrieve all flows for a given name:
+flow_ids = openml.flows.get_flow_id(name=flow.name, exact_version=False)
```
ArlindKadra: I understand that you want to emphasize the fact that we get multiple flow ids and there is no `exact_version` when given a flow name. However, in the code, the exact_version is always set to `False`, no matter what the user gives.

tests/test_flows/test_flow_functions.py
```diff
@@ -303,3 +303,21 @@ def test_get_flow_reinstantiate_model_wrong_version(self):
...
+        self.assertGreater(len(flow_ids), 2)
+
+        flow_ids_exact_version_True = openml.flows.get_flow_id(name=flow.name, exact_version=True)
+        flow_ids_exact_version_False = openml.flows.get_flow_id(
+            name=flow.name,
+            exact_version=False,
+        )
+        self.assertEqual(flow_ids_exact_version_True, flow_ids_exact_version_False)
+        self.assertIn(flow.flow_id, flow_ids_exact_version_True)
+        self.assertGreater(len(flow_ids_exact_version_True), 2)
```
ArlindKadra: Won´t this fail ? Maybe you meant `assertLess`

tests/test_flows/test_flow_functions.py
```diff
@@ -303,3 +303,21 @@ def test_get_flow_reinstantiate_model_wrong_version(self):
...
+        flow_ids = openml.flows.get_flow_id(model=clf, exact_version=False)
+        self.assertIn(flow.flow_id, flow_ids)
+        self.assertGreater(len(flow_ids), 2)
+
+        flow_ids_exact_version_True = openml.flows.get_flow_id(name=flow.name, exact_version=True)
+        flow_ids_exact_version_False = openml.flows.get_flow_id(
+            name=flow.name,
+            exact_version=False,
+        )
+        self.assertEqual(flow_ids_exact_version_True, flow_ids_exact_version_False)
```
ArlindKadra: Is this not the same exact call ?
The value of `exact_version` is always ignored and set to False.

openml/flows/functions.py
```diff
@@ -268,6 +268,67 @@ def flow_exists(name: str, external_version: str) -> Union[int, bool]:
...
+    int or bool, List
+        flow id iff exists, ``False`` otherwise, List if exact_version is ``False``
+    """
+    if model is None and name is None:
+        raise ValueError(
+            'Need to provide either argument `model` or argument `name`, but both are `None`.'
+        )
+    elif model is not None and name is not None:
+        raise ValueError(
+            'Must provide either argument `model` or argument `name`, but both both.'
```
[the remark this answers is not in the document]
mfeurer: Thanks, will fix.

examples/30_extended/flow_id_tutorial.py
```diff
@@ -0,0 +1,69 @@
...
+# ==================================
+# The schema of a flow is given in XSD (`here
+# <https://github.com/openml/OpenML/blob/master/openml_OS/views/pages/api_new/v1/xsd/openml.implementation.upload.xsd>`_).  # noqa E501
+# Only two fields are required, a unique name, and an external version. While it should be pretty
+# obvious why we need a name, the need for the additional external version information might not
+# be immediately clear. However, this information is very important as it allows to have multiple
+# flows with the same name for different versions of a software. This might be necessary if an
+# algorithm or implementation introduces, renames or drop hyperparameters over time.
+
+flow = openml.extensions.get_extension_by_model(clf).model_to_flow(clf)
```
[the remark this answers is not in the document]
mfeurer: Makes sense, will change this.

examples/30_extended/flow_id_tutorial.py
mfeurer: No, the user can give a model, and then `exact_version` makes a difference in the code (code = function `get_flow_id`). Are you talking about the example or the package code?

tests/test_flows/test_flow_functions.py
mfeurer: Nope, this works. If only a name is given, exact version does not matter and is ignored.

tests/test_flows/test_flow_functions.py
mfeurer: Yep, and the test is checking this. I'll add a comment.

examples/30_extended/flow_id_tutorial.py
ArlindKadra: The example, `exact_version` does not add or remove anything when giving a flow name as above. It is always replace by `False` in the `get_flow_id` method.
It was a remark, I understand that this way it is clearer to a new user.

tests/test_flows/test_flow_functions.py
ArlindKadra: Ups, missed this one.

PR #796, opened 2019-10-02
Rename argument to be more intuitive

mfeurer:
### What does this PR implement/fix? Explain your changes.

Renames argument `id` of function `openml.evaluations.list_evaluations` to `run` to make it clearer that the IDs are for the run.

There is an existing unit test for this functionality. I updated at to reflect the changed argument name.

### DIFF
--- a/openml/evaluations/functions.py
+++ b/openml/evaluations/functions.py
@@ -15,10 +15,10 @@ def list_evaluations(
     function: str,
     offset: Optional[int] = None,
     size: Optional[int] = None,
-    id: Optional[List] = None,
     task: Optional[List] = None,
     setup: Optional[List] = None,
     flow: Optional[List] = None,
+    run: Optional[List] = None,
     uploader: Optional[List] = None,
     tag: Optional[str] = None,
     per_fold: Optional[bool] = None,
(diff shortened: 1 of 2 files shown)


openml/evaluations/functions.py
```diff
@@ -15,10 +15,10 @@ def list_evaluations(
     function: str,
     offset: Optional[int] = None,
     size: Optional[int] = None,
-    id: Optional[List] = None,
     task: Optional[List] = None,
     setup: Optional[List] = None,
     flow: Optional[List] = None,
+    run: Optional[List] = None,
     uploader: Optional[List] = None,
     tag: Optional[str] = None,
```
PGijsbers: If we're going for more intuitive names, should be change them to be plural, as they expect a list? Or perhaps event better, plural and `_id` suffix (e.g. `run_ids`), to indicate we are looking for something that parses to an id and not actual `OpenMLX` (e.g. `OpenMLRun`) objects instead?

Do we also want to change the typehint to `Optionl[List[Union[str, int]]]`? It's pretty verbose, but more complete... Maybe it would be enough to include this in the docstrings.

mfeurer: I'd be happy to change those, but I think this is for a future PR while this is about renaming an argument from `id` to `run`. Shall I open an issue so we don't forget about this?

PGijsbers: Just did this so I can merge & close PR. See #804.

PGijsbers:
I didn't find a related open issue that should be closed by this PR.

mfeurer:
> I didn't find a related open issue that should be closed by this PR.

There is none. I tried to create an example and found the resulting code unintuitive so I just went ahead.

Thanks for merging!

PR #803, opened 2019-10-07
Add example rijn

janvanrijn:


examples/40_paper/2018_kdd_rijn_example.py
```diff
@@ -8,9 +8,116 @@
...
+##############################################################################
+# With the advent of automated machine learning, automated hyperparameter
+# optimization methods are by now routinely used in data mining. However, this
+# progress is not yet matched by equal progress on automatic analyses that
+# yield information beyond performance-optimizing hyperparameter settings.
+# In this example, we aim to answer the following two questions: Given an
+# algorithm, what are generally its most important hyperparameters?
+#
+# This work is carried out on the OpenML-100 benchmark suite, which can be
+# obtained by `openml.study.get_suite('OpenML100')`. In this example, we
```
mfeurer: Unfortunately, this isn't markdown but rst and you need to use two backticks to get code rendering.

examples/40_paper/2018_kdd_rijn_example.py
```diff
@@ -8,9 +8,116 @@
...
+limit_per_task = 500
+n_trees = 16
+
+fanova_results = []
+# we will obtain all results from OpenML per task. Practice has shown that this places the bottleneck on the
+# communication with OpenML, and for iterated experimenting it is better to cache the results in a local file.
+for idx, task_id in enumerate(suite.tasks):
+    logging.info('Starting with task %d (%d/%d)' % (task_id, idx+1, len(suite.tasks)))
+    evals = openml.evaluations.list_evaluations_setups(
+        evaluation_measure, flow=[flow_id], task=[task_id], size=limit_per_task, output_format='dataframe')
```
mfeurer: Don't you also have to limit by the uploader or study?

examples/40_paper/2018_kdd_rijn_example.py
```diff
@@ -8,9 +8,116 @@
...
+    evaluator = fanova.fanova.fANOVA(
+        X=setups_evals[parameter_names].values, Y=setups_evals[performance_column].values, n_trees=n_trees)
+    for idx, pname in enumerate(parameter_names):
+        try:
+            fanova_results.append({
+                'hyperparameter': pname if len(pname) < 35 else '[...] %s' % pname[-30:],
+                'fanova': evaluator.quantify_importance([idx])[(idx,)]['individual importance']
+            })
+        except RuntimeError as e:
+            # functional ANOVA sometimes crashes with a RuntimeError, e.g., on tasks where the performance is constant
```
mfeurer: What's the effect of this? Is there any adverse effect of this?

examples/40_paper/2018_kdd_rijn_example.py
```diff
@@ -8,9 +8,116 @@
...
+parameter_filters = {
+    'sklearn.svm.classes.SVC(17)_kernel': 'sigmoid'
+}
+evaluation_measure = 'predictive_accuracy'
+limit_per_task = 500
+n_trees = 16
+
+fanova_results = []
+# we will obtain all results from OpenML per task. Practice has shown that this places the bottleneck on the
+# communication with OpenML, and for iterated experimenting it is better to cache the results in a local file.
```
mfeurer: Is there any way for speeding this up? Otherwise, would it be possible to only conduct a part of the evaluation to limit the running time of this example?

examples/40_paper/2018_kdd_rijn_example.py
```diff
@@ -8,9 +8,116 @@
...
+    # in this simplified example, we only display numerical and float hyperparameters. For categorical hyperparameters,
+    # the fanova library needs to be informed by using a configspace object.
+    setups_evals = setups_evals.select_dtypes(include=['int64', 'float64'])
+    # drop rows with unique values. These are by definition not an interesting hyperparameter, e.g., `axis`, `verbose`
+    setups_evals = setups_evals[[c for c in list(setups_evals)
+                                 if len(setups_evals[c].unique()) > 1 or c == performance_column]]
+    # determine x values to pass to fanova library
+    parameter_names = [pname for pname in setups_evals.columns.values if pname != performance_column]
+    evaluator = fanova.fanova.fANOVA(
+        X=setups_evals[parameter_names].values, Y=setups_evals[performance_column].values, n_trees=n_trees)
```
mfeurer: Could you please use https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.to_numpy.html instead of `dataframe.values`?

examples/40_paper/2018_kdd_rijn_example.py
```diff
@@ -8,9 +8,116 @@
...
+    for idx, pname in enumerate(parameter_names):
```
mfeurer: The final output includes the random state as a contributor of variance, is this intended?

examples/40_paper/2018_kdd_rijn_example.py
```diff
@@ -8,9 +8,116 @@
...
 
 | Hyperparameter importance across datasets
-| Jan van Rijn and Frank Hutter
+| Jan N. van Rijn and Frank Hutter
 | In *Proceedings of the 24th ACM SIGKDD International Conference on Knowledge Discovery & Data Mining*, 2018
 | Available at https://dl.acm.org/citation.cfm?id=3220058
-
-This is currently a placeholder.
 """
+import fanova
```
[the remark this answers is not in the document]
janvanrijn: ouch, that might be hard, as `pyrfr` is a direct requirement of this, which can not be installed through pip (as it depends on `swig`). Any ideas?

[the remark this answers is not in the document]
janvanrijn:
```
import json
import logging

import fanova
import matplotlib.pyplot as plt
import pandas as pd
import seaborn as sns

import openml
```
would be about right?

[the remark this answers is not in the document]
mfeurer: You do `conda install gxx_linux-64 gcc_linux-64 swig` around line 40 in `ci_scripts/install.sh` (under the if clause `"$EXAMPLES" == "true"`). Then you can do `pip install pyrfr`.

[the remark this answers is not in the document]
mfeurer: yep, that's perfect

examples/40_paper/2018_kdd_rijn_example.py
janvanrijn: the filtering for study is implicit, (as we enumerate over tasks from the suite). 

as openml is a collaborative platform, we should in  theory not care about who uploaded the results. if we did our job right, the results will be consistent

examples/40_paper/2018_kdd_rijn_example.py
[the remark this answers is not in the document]
janvanrijn: cool, thanks for your help! 

> pip install pyrfr

I don't think that would be necessary, should be handled in the `setup.py` with the fanova dependency, right?

examples/40_paper/2018_kdd_rijn_example.py
janvanrijn: Yes. I will add it somewhere in the comments

examples/40_paper/2018_kdd_rijn_example.py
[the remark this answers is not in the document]
janvanrijn: btw, we also need this under the doc test line? (currently the CI with doc_push seems to fail)

Also Appveyor does not like fanova (not sure if you can install `pyrfr` on windows machines)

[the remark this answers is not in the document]
mfeurer: > I don't think that would be necessary, should be handled in the setup.py with the fanova dependency, right?

You're right.

> Also Appveyor does not like fanova

Interesting, I did not know that we try to install the dependencies for the examples under Windows, sorry for that. I also don't know why we do so, and I don't think it is necessary, so you could remove this test-wise. @PGijsbers do you have any idea why this is the case?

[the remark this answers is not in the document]
mfeurer: Also, it seems that there's an issue in install.py as that the package actually isn't installed under examples any more, and therefore also not the dependencies (that's why fANOVA isn't there). It seems like you need to move the installation into the different if statements.

examples/40_paper/2018_kdd_rijn_example.py
```diff
@@ -42,8 +44,8 @@
 # Note that we simplify the example in several ways
 # 1) We only consider numerical hyperparameters
 # 2) We consider all hyperparameters that are numerical (in reality, some
-#    hyperparameters might be inactive (e.g., `degree`) or irrelevant (e.g.,
-#    `random_state`)
+#    hyperparameters might be inactive (e.g., ``degree``) or irrelevant
```
mfeurer: There's a formatting issue here. Could you please add a newline before `1)` to make the ordered list start?

examples/40_paper/2018_kdd_rijn_example.py
```diff
@@ -8,9 +8,132 @@
...
+# Note that we simplify the example in several ways
+# 1) We only consider numerical hyperparameters
+# 2) We consider all hyperparameters that are numerical (in reality, some
+#    hyperparameters might be inactive (e.g., ``degree``) or irrelevant
+#    (e.g., ``random_state``)
+# 3) We assume all hyperparameters to be on uniform scale
+#
+# Any difference in conclusion between the actual paper and the presented
+# results is most likely due to one of these simplifications. For example,
+# # the hyperparameter C looks rather insignificant, whereas it is quite
```
mfeurer: The 2nd hash isn't necessary here.

examples/40_paper/2018_kdd_rijn_example.py
```diff
@@ -8,9 +8,132 @@
...
+# important when it is put on a log-scale. All these simplifications can be
+# addressed by defining a ConfigSpace. For a more elaborated example that uses
+# this, please see:
+# https://github.com/janvanrijn/openml-pimp/blob/d0a14f3eb480f2a90008889f00041bdccc7b9265/examples/plot/plot_fanova_aggregates.py # noqa F401
```
mfeurer: The noqa is rendered to the example. Do you have any idea on how to avoid this?

examples/40_paper/2018_kdd_rijn_example.py
mfeurer: True, do you think it would make sense to specifically note this in the example?

examples/40_paper/2018_kdd_rijn_example.py
janvanrijn: how can I see the examples in this branch btw? I have no clue how to manually inspect the examples before merging into master, which might take a while

examples/40_paper/2018_kdd_rijn_example.py
janvanrijn: nope, I copied this from your examples. Shall I just drop it or will flake8 hate us forever?

examples/40_paper/2018_kdd_rijn_example.py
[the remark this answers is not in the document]
janvanrijn: apparently I needed to add the if-statement regarding  `$DOCPUSH` (assuming that this was the one that uses the examples, rather than coverage, correct me if I am wrong)

Regarding this statement: `conda install gxx_linux-64 gcc_linux-64 swig`, why doesn't there need to be a `--yes` in there? For some runs, the `--yes` seems not necessary, for others it is. (in particular, the one in the examples if-else is fine, whereas the one in docpush crashes otherwise)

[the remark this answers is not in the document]
janvanrijn: I think the answer to the last question is: the examples variable is nowhere used. 

I have added the --yes flag there as well, for consistency and competency. We should probably investigate.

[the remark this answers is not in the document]
janvanrijn: `swig` now installs, but `fanova` still doesn't. It currently crashes. I suspect this has to do with the missing "examples" flag.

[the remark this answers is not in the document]
PGijsbers: On installing example dependencies on Appveyor. It looks like I added the installation of example dependencies in commit c61064a. I'd say this is probably to keep the installed packages consistent with what was installed before the commit (as some notebook packages got installed). As we explicitly ignore unit tests for examples on Windows, we could forgo installing the example dependencies. 

But why are we ignoring examples on Windows? If we support Windows, shouldn't we test the examples (or at least those that only rely on straight forward installable packages)? I think it's reasonable for a Windows user to install `openml-python` with current examples and run them - not get some error because installing `pyrfr` can be problematic out of the box. 

Of course we could set up our Appveyor scripts to test these examples with more difficult dependencies. However, end-users will then have to resolve any installation issues too, which I think is messy (and probably will create additional issues on our repo for users being unable to install e.g. `pyrfr`). For that reason I would propose something along the lines of dividing examples in two: examples which install easily out of the box cross-platform, and examples which don't. The messiness of having two example categories is in my opinion still cleaner than the messiness of effectively being unable to run `pip install openml-python[examples]` on Windows.

examples/40_paper/2018_kdd_rijn_example.py
mfeurer: Examples in this branch are not deployed online, you need to build them locally using `make html` in the `doc` directory.

examples/40_paper/2018_kdd_rijn_example.py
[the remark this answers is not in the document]
mfeurer: > doesn't there need to be a `--yes` in there

Thinking about this again, I'd also say you need a `--yes` there.

> apparently I needed to add the if-statement regarding $DOCPUSH

I just realized that, too. Sorry for not noticing this earlier. You can copy the code from my PR #806 to make it clearer.

> not get some error because installing `pyrfr` can be problematic out of the box.

I don't know if that will be possible at all. Should we use two categories `[examples-windows]` and `[examples-unix]` to differentiate here? Then we'd only have to add a note to the fanova example to highlight that it won't work under windows.

examples/40_paper/2018_kdd_rijn_example.py
mfeurer: You can drop. Arlind temporarily disabled the flake8 checks for the examples.

examples/40_paper/2018_kdd_rijn_example.py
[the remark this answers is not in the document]
PGijsbers: > I don't know if that will be possible at all. Should we use two categories [examples-windows] and [examples-unix] to differentiate here? Then we'd only have to add a note to the fanova example to highlight that it won't work under windows.

Yes, that is exactly what I had in mind. I'm all for cross-platform support, but I do not think it warrants the time to include the fanova in Appveyor CI or to try and support Windows users from running the fanova experiment.  Having a clean demarkation of which examples are not supported on Windows (for now just fanova) and making sure it's easy the install dependencies for just the examples that work on your platform seems like the way to go to me.

janvanrijn:
Everything excepts for a random failure at appveyor:

`FAILED tests/test_tasks/test_split.py::OpenMLSplitTest::test_eq `

ci_scripts/install.sh
```diff
@@ -37,11 +37,18 @@ python -c "import numpy; print('numpy %s' % numpy.__version__)"
 python -c "import scipy; print('scipy %s' % scipy.__version__)"
 
 if [[ "$EXAMPLES" == "true" ]]; then
```
mfeurer: I think you can remove the examples if clause as it's not used at all.

ci_scripts/install.sh
```diff
@@ -37,11 +37,18 @@ python -c "import numpy; print('numpy %s' % numpy.__version__)"
...
     pip install -e '.[examples]'
+    pip install -e '.[examples_unix]'
 fi
 if [[ "$DOCTEST" == "true" ]]; then
     pip install sphinx_bootstrap_theme
 fi
+if [[ "$DOCPUSH" == "true" ]]; then
+    conda install --yes gxx_linux-64 gcc_linux-64 swig
+    pip install -e '.[examples]'
+    pip install -e '.[examples_unix]'
```
mfeurer: You can save a line by having both in a single statement: `pip install -e '.[examples,examples_unix]'`.

PR #807, opened 2019-10-11
extended

janvanrijn:


examples/40_paper/2018_ida_strang_example.py
```diff
@@ -98,3 +98,15 @@ def determine_class(val_lin, val_nonlin):
 ax_scatter.set_xscale('log')
 ax_scatter.set_yscale('log')
 plt.show()
+
+# makes a scatter plot where each data point represents the performance of the two algorithms on various axis
```
mfeurer: Cool idea. Could you please add a line of hashes above this comment so that the documentation renders this in a new cell?

PR #811, opened 2019-10-11
Add additional part of OpenML error message to exception message

mfeurer:
This allows us to see the error message given by OpenML to simplify debugging.


PGijsbers:
It looks like [this](https://github.com/openml/openml-python/blob/update_error_handling/tests/test_runs/test_run_functions.py#L423) unit test still refers to the now-missing attribute. Please update and do a quick search through all files to see if there is other code attempting to directly set the attribute.

PR #815, opened 2019-10-14
make output of rijn example a bit nicer

amueller:
Slightly nicer output for 2018 kkd rijn example


mfeurer:
I think @janvanrijn should approve this.

janvanrijn:
Not sure if this is an improvement, there might be a lot of issues with tailing brackets?

amueller:
Can you explain? This hides the package and module name (and a repeated type output that adds lots of output to the example and doesn't really say anything).

PR #818, opened 2019-10-14
fix issue #305 by not requiring external version in the flow xml

### LINKED ISSUE #305: API call to flow/1 raises an error (status_code: 500)
Hi,

An API call flows.get_flow(1) or REST API (flow/1) raises an error. (Attached is the error message.)

screen shot 2017-09-01 at 15 46 26

I would like to fetch every flow description, and I wondered what is the problem here.

Thanks!
Minori

mfeurer:
Closes #305 

Apparently, the `external_version` is not a required field of a flow: https://github.com/openml/OpenML/blob/master/openml_OS/views/pages/api_new/v1/xsd/openml.implementation.upload.xsd This PR removes this assumption from the Python package and thereby allows loading flow number 1.

### DIFF
--- a/tests/test_flows/test_flow_functions.py
+++ b/tests/test_flows/test_flow_functions.py
@@ -264,9 +264,11 @@ class TestFlowFunctions(TestBase):
         self.assertEqual(server_flow.model.categories, flow.model.categories)
 
     def test_get_flow1(self):
-        # Make sure that issue #305 doesn't pop up any more
+        # Regression test for issue #305
+        # Basically, this checks that a flow without an external version can be loaded
         openml.config.server = self.production_server
-        openml.flows.get_flow(1)
+        flow = openml.flows.get_flow(1)
+        self.assertIsNone(flow.external_version)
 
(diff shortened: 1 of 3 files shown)


tests/test_flows/test_flow_functions.py
```diff
@@ -263,6 +263,11 @@ def test_sklearn_to_flow_list_of_lists(self):
         self.assertEqual(server_flow.parameters['categories'], '[[0, 1], [0, 1]]')
         self.assertEqual(server_flow.model.categories, flow.model.categories)
 
+    def test_get_flow1(self):
+        # Make sure that issue #305 doesn't pop up any more
```
PGijsbers: While linking to Github issues is very useful for full context/explanation, I think it's good to accompany it where reasonable with a very brief explanation in the file. That way one has a slight idea of what's going on without leaving the editor, e.g.:

> `# Fetching a flow with no external version is allowed (see #305)`.

You can commit with `[skip ci]` to avoid a rerun for this change.

Issue #821, opened 2019-10-14
Validate datasets against XSD prior to upload


amueller:
It would be nice we could validate some fields in the dataset prior to uploading. I just got an error in the parsing after uploading a large file because there was an umlaut in the citation.

We could either create the XML and run XSD validation on it with our schema. Or we could try to reimplement individual checks (this one would have been very easy to check in Python).

PR #822, opened 2019-10-14
add validation for strings in datasets

amueller:
Add some basic validation for strings in datasets.
Adding tests now.


janvanrijn:
Wouldn't it be better to locally check the XML on the XSD schema (available on line)? That keeps the burden to maintain the schema in one place.

mfeurer:
Yeah, Andy opened an issue for that, too (#821), but I think for the time being this one was easier to implement.

PR #823, opened 2019-10-14
Fixing fetching of categorical sparse data

### LINKED ISSUE #758: Cannot load sparse datasets with categorical variables any more
Can be reproduced with:

```
import openml

openml.datasets.get_dataset(395)
```

Neeratyoy:
#### Reference Issue

Fixes #758. 

There was a try block checking for a numpy based conversion. Converted that to pandas categorical encoding.

`import openml`
`openml.datasets.get_dataset(395)`

### DIFF
--- a/openml/datasets/dataset.py
+++ b/openml/datasets/dataset.py
@@ -245,6 +245,7 @@ class OpenMLDataset(object):
             when converted to lower case.
 
 
+
         Returns
         -------
         dict
@@ -319,13 +320,15 @@ class OpenMLDataset(object):
         attribute_names = []
         categories_names = {}
         categorical = []
-        for name, type_ in data['attributes']:
+        for i, (name, type_) in enumerate(data['attributes']):
             # if the feature is nominal and the a sparse matrix is
             # requested, the categories need to be numeric
             if (isinstance(type_, list)
                     and self.format.lower() == 'sparse_arff'):
                 try:
-                    np.array(type_, dtype=np.float32)
+                    # checks if the strings which should be the class labels
+                    # can be encoded into integers
+                    pd.factorize(type_)[0]
                 except ValueError:
                     raise ValueError(
                         "Categorical data needs to be numeric when "
--- a/tests/test_datasets/test_dataset.py
+++ b/tests/test_datasets/test_dataset.py
@@ -322,7 +322,7 @@ class OpenMLDatasetTestSparse(TestBase):
         self.assertEqual(y.shape, (600, ))
 
     def test_get_sparse_categorical_data_id_395(self):
-        dataset = openml.datasets.get_dataset(395, download_data=False)
+        dataset = openml.datasets.get_dataset(395, download_data=True)
         feature = dataset.features[3758]
         self.assertTrue(isinstance(dataset, OpenMLDataset))
         self.assertTrue(isinstance(feature, OpenMLDataFeature))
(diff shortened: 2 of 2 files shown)


tests/test_datasets/test_dataset.py
```diff
@@ -320,6 +320,14 @@ def test_get_sparse_dataset_rowid_and_ignore_and_target(self):
         self.assertListEqual(categorical, [False] * 19998)
         self.assertEqual(y.shape, (600, ))
 
+    def test_get_sparse_categorical_data_id_395(self):
+        dataset = openml.datasets.get_dataset(395, download_data=False)
+        feature = dataset.features[3758]
+        self.assertEqual(dataset.name, 're1.wc')
+        self.assertEqual(feature.name, 'CLASS_LABEL')
+        self.assertEqual(feature.data_type, 'nominal')
+        self.assertEqual(len(feature.nominal_values), 25)
```
mfeurer: Could you please add a check for the type of the output value?

tests/test_datasets/test_dataset.py
```diff
@@ -320,6 +320,14 @@ def test_get_sparse_dataset_rowid_and_ignore_and_target(self):
         self.assertListEqual(categorical, [False] * 19998)
         self.assertEqual(y.shape, (600, ))
 
+    def test_get_sparse_categorical_data_id_395(self):
```
mfeurer: This test does not explicitly load the data, so is the function in question actually called?

openml/datasets/dataset.py
```diff
@@ -319,13 +320,15 @@ def _parse_data_from_arff(
...
+        for i, (name, type_) in enumerate(data['attributes']):
             # if the feature is nominal and the a sparse matrix is
             # requested, the categories need to be numeric
             if (isinstance(type_, list)
                     and self.format.lower() == 'sparse_arff'):
                 try:
-                    np.array(type_, dtype=np.float32)
+                    # checks if the strings which should be the class labels
+                    # can be encoded into integers
+                    pd.factorize(type_)[0]
```
mfeurer: Didn't you mention in person that you need to assign the value of this function call?

tests/test_datasets/test_dataset.py
Neeratyoy: Added.

tests/test_datasets/test_dataset.py
Neeratyoy: Yeah, my bad. Made it **True**.

openml/datasets/dataset.py
Neeratyoy: Yes, but with further testing realized that assignment doesn't make sense here since this loop iterates over the _attributes_. Whereas, if anything needs to be checked, we should check the _data_. Which is not seemingly throwing any issue.

I checked this chunk of code with both a _Sparse_Arff_ and _Arff_ data formats, the _type__ receives exactly the same type and structure of the output. I don't know why the attribute list is being checked for type whereas the `arff.ArffDecoder.decode()` seems to return the target feature as a list of the classes. Don't know why a sparse format requires numeric encoding of that attribute list.

Hence, I replaced the numpy check with the pandas categorical encoding.

tests/test_datasets/test_dataset.py
mfeurer: Sorry for not being clear enough here. Could you please load X and y and check their type, dtype and shape?

Neeratyoy: As discussed, I'm refraining from changing this test now. Have created an [issue](https://github.com/openml/openml-python/issues/827) to take care of such checks independently.

PR #828, opened 2019-10-15
Create OpenMLBase, have most OpenML objects derive from it

### LINKED ISSUE #433: Add open in browser function for each entity
To allow code like `run.print_url()` or `run.open_in_browser()`.

PGijsbers:
Several OpenML entities are sharing a lot of code in theory, but have copy paste duplicate code in practice. This PR introduces a base class from which OpenML entities can derive and share common logic. In this draft some bare basics are implemented and OpenMLFlow derives from it.

This will be further extended to make the other main entities (Dataset, Task, Run and Study) derive from the base class.
I will take a look at a few other functions shared between some or all of those types such as `_to_xml` and `publish`.

edit: looks like unit tests fail due to illegal type annotation for Py3.6, please ignore. The Flow unit tests worked locally, will fix it to be Py3.6 compatible.

edit: Also fixes #433

### DIFF
--- a/openml/config.py
+++ b/openml/config.py
@@ -28,7 +28,8 @@ config_file = os.path.expanduser(os.path.join('~', '.openml', 'config'))
 
 # Default values are actually added here in the _setup() function which is
 # called at the end of this module
-server = _defaults['server']
+server = str(_defaults['server'])  # so mypy knows it is a string
+server_base_url = server[:-len('/api/v1/xml')]
 apikey = _defaults['apikey']
 # The current cache directory (without the server name)
 cache_directory = _defaults['cachedir']
--- a/openml/base.py
+++ b/openml/base.py
@@ -4,7 +4,7 @@ from typing import Optional, List, Tuple
 import webbrowser
 
 import openml.config
-from .utils import _tag_entity
+from .utils import _tag_openml_base
 
 
 class OpenMLBase(ABC):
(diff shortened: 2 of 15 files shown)


PGijsbers:
For the `__repr__` header I now parse the class names, which gives nearly identical output with what was before, but is generic and shared between all child classes. For tasks there is now also more information in the header, instead of "OpenML Task" it now says the full task type, e.g.:
```
OpenML Classification Task
==========================
Task Type...........: Supervised Classification
Task ID.............: 59
Task URL............: https://www.openml.org/t/59
Estimation Procedure: crossvalidation
Evaluation Measure..: predictive_accuracy
Target Feature......: class
# of Classes........: 3
Cost Matrix.........: Available
```
I propose to remove the 'Task Type' row from the body of the `__repr__`.

openml/config.py
```diff
@@ -28,7 +28,8 @@
 
 # Default values are actually added here in the _setup() function which is
 # called at the end of this module
-server = _defaults['server']
+server = str(_defaults['server'])  # so mypy knows it is a string
```
mfeurer: you could also use `mypy.cast` to not have to use a comment here

PGijsbers: I generally prefer this for simple types, as it is clear what is happening (whereas you have to be familiar with `mypy.cast` to immediately know what's being done and why).

PGijsbers:
I fixed the mypy errors by using `getattr` as mypy can't know the attributes exist in child classes. I think this is probably the cleanest way to do this.

PGijsbers:
Looks like my local mypy reported different errors than CI. Great.

PR #832, opened 2019-10-15
Adding Perrone example for building surrogate

Neeratyoy:
Shows an example of Perrone et al's use of previous OpenML runs to build a surrogate model.


Neeratyoy:
@mfeurer might need some more textual descriptions, but I presume that the person going through this example already is aware of the paper.

mfeurer:
@Neeratyoy could you please have a look at my changes? Apparently, the flake8 checks only fail on the push, not the pr check, so we should be fine. Please feel free to merge if you agree with my changes.

PR #833, opened 2019-10-15
Adding option to print logs during an api call

Neeratyoy:
Adds a `logger.info` which can be used to track what is happening during wait times for a long api call.

### DIFF
--- a/openml/_api_calls.py
+++ b/openml/_api_calls.py
@@ -53,9 +53,9 @@ def _perform_api_call(call, request_method, data=None, file_elements=None):
     else:
         response = _read_url(url, request_method, data)
     logging.info(
-        '%.7fs taken for [%s] request for the URL %s', 
-        time.time() - start, 
-        request_method, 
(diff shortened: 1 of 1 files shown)


openml/_api_calls.py
```diff
@@ -42,7 +43,7 @@ def _perform_api_call(call, request_method, data=None, file_elements=None):
     url += call
 
     url = url.replace('=', '%3d')
-
+    logging.info('[%s] %s' % (request_method, url))
```
mfeurer: Hm, I think it would make a lot of sense if you also add a `logging.info` for when the call is done, which then also states how much time was spent in that function.

mfeurer: Please don't use `%` for string interpolation in logging, but pass the formatting arguments as regular arguments to the function as `logging.info('[%s] %s', request_method, url)`. Then, they're only interpolated if the logging level requires the program to do so.

PR #835, opened 2019-10-15
Add automatic type conversion prototype

mfeurer:
@janvanrijn @mllg these are some simple changes necessary to work with the type information added to the dataset today: `<oml:id type="xs:integer">2</oml:id>`. Will we keep it that way or will this be reverted?


mfeurer:
The proposed style is XSD, not XML, and will therefore not be integrated into the OpenML API.

PR #849, opened 2019-10-17
Put shared logic of Publish into OpenMLBase

PGijsbers:
Also unifies the return value of OpenMLBase.publish to return the OpenMLBase object (before it was either that *or* id)


PGijsbers:
Anyone has an idea what this halting can be about? https://travis-ci.org/openml/openml-python/jobs/599521410

mfeurer:
Nope, but I wouldn't worry about that as it just happened once. I just restarted the test.

Issue #852, opened 2019-10-18
Peronne example stalls on CI sometimes


PGijsbers:
The Peronne example seems to stall on CI sometimes. It could be server issues, but it would be nice to verify that. See e.g.: https://travis-ci.org/openml/openml-python/jobs/599553086

mfeurer:
How about wrapping a context manager into this example to have some output so that it doesn't stall?

(We have to download quite a few runs with `get_runs` which iteratively calls `get_run`. Instead, we could use `list_evaluations_hyperparameters`).

mfeurer:
Solved via #853 by using `list_evaluations_hyperparameters`.

Issue #805, opened 2019-10-08
ensure examples are unit tested


janvanrijn:
i think they might be, since one of the travis runs crashes when they are not properly handled. 

The travis script has several variables, including "examples". None of the travis runs invokes this variable. Is this good?

janvanrijn:
side-question. what happens if we execute first
`pip install -e '.test'`
and subsequently
`pip install -e '.examples'`

Will that include the installation of all dependencies of both commands and keep the package as is?

mfeurer:
> i think they might be, since one of the travis runs crashes when they are not properly handled.

Yes, they're actually executed as part of building the documentation: https://github.com/openml/openml-python/blob/develop/ci_scripts/create_doc.sh#L24

> The travis script has several variables, including "examples". None of the travis runs invokes this variable. Is this good?

This is not as bad as it sounds as there are no more "examples": https://github.com/openml/openml-python/blob/develop/ci_scripts/test.sh#L19 I'll remove those lines.

Regarding your side question: I'm not sure, but I'd expect that the installation is skipped the second time. (and therefore the dependencies will not be installed).

mfeurer:
I just created #806 to remove these unnecessary variables.

mfeurer:
Closing this as we test the examples (they constantly for one PR so I'm really sure about this) and removed the unused variable from the travis scripts. Please reopen if there's something else to this issue.

PR #853, opened 2019-10-19
Optimizing Perrone example

Neeratyoy:
#### Reference Issue

Addresses #852. 

Removes the repeated calls to `openml.runs.get_runs` and replaces it with one single API call using `openml.evaluations.list_evaluations_setups`.
Also, added a unit test for this example.


Neeratyoy:
> However, could you please explain why you need an extra test given that the example is executed by travis-ci?

I came across the issue #805 and thought this was a good opportunity to incorporate it. Also, I came across a unit test for the SVM example. Hence, I thought this was the protocol.
Having the example as a unit test also seemed to be a good check since from what I understand, we won't know if the example script failed, in terms of an error from travis-ci, will we? That is, we might still get all greens.

PR #857, opened 2019-10-28
add task_type to list_runs

sahithyaravi1493:
API call to list runs https://www.openml.org/api/v1/run/list/limit/10 - returns task type, while
python API does not.

added task type to list_runs call in python API

runs.list_runs(size=10)


doc/progress.rst
```diff
@@ -8,6 +8,7 @@ Changelog
 
 0.10.1
 ~~~~~~
+* ADD #857: Adds task type ID to list_runs
```
mfeurer: This is actually my mistake (as I didn't update this yet), but since we made a new release, you need to put this under a new header.

PR #859, opened 2019-10-29
Remove assertion on the test server

### LINKED ISSUE #858: `test__name_to_id_with_multiple_active_error` assumes there are multiple active public iris datasets on test server
this assumption is incorrect, as whenever the test server is resetted, there is only 1

mfeurer:
Closes #858 by using the live instead of the test server.

### DIFF
--- a/tests/test_datasets/test_dataset_functions.py
+++ b/tests/test_datasets/test_dataset_functions.py
@@ -265,6 +265,7 @@ class TestOpenMLDataset(TestBase):
 
     def test__name_to_id_with_multiple_active_error(self):
         """ With multiple active datasets, retrieve the least recent active. """
+        openml.config.server = self.production_server
         self.assertRaisesRegex(
             ValueError,
             "Multiple active datasets exist with name iris",


PGijsbers:
Looks like tests fail due to server issues, otherwise the change looks good to me.

mfeurer:
Yes, trying to fix them together with @janvanrijn (this PR is part of it).

janvanrijn:
> Looks like tests fail due to server issues, otherwise the change looks good to me.

I disagree. 

The unit tests seem to assert that there are 2 public versions of the iris dataset on the test server, which is an incorrect assumption, especially after we reset the testserver.

mfeurer:
> The unit tests seem to assert that there are 2 public versions of the iris dataset on the test server

This PR changes this. The unit test now assumes that there are 2 public versions of the iris dataset on the live server.

janvanrijn:
That seems perfectly reasonable to me. This solves issue #858.

PGijsbers:
I meant the CI results on this PR came back negative, but those were due to server issues. This was unrelated to the change made for this PR, which was not a server issue.

janvanrijn:
can you point out specifically which failing unit tests you refer to?

Some might fail due to recent reset of test server.

PGijsbers:
Errors like [this](https://travis-ci.org/openml/openml-python/jobs/604334447#L1671). I don't think they are persistent bugs, and might well have to do with the recent reset - I'm not saying the server is broken. But for the purpose of an `openml-python` PR I leave a comment indicating them as 'server issue' for later reference, indicating that we do see CI failures but don't expect them to be caused by the changes to `openml-python`.

PR #860, opened 2019-10-29
Add debug output

mfeurer:
to find the cause of previously failing unit tests.

### DIFF
--- a/examples/30_extended/tasks_tutorial.py
+++ b/examples/30_extended/tasks_tutorial.py
@@ -196,11 +196,11 @@ except openml.exceptions.OpenMLServerException as e:
     # Error code for 'task already exists'
     if e.code == 614:
         # Lookup task
-        tasks = openml.tasks.list_tasks(data_id=128, output_format='dataframe').to_numpy()
-        tasks = tasks[tasks[:, 4] == "Supervised Classification"]
-        tasks = tasks[tasks[:, 6] == "10-fold Crossvalidation"]
-        tasks = tasks[tasks[:, 19] == "predictive_accuracy"]
-        task_id = tasks[0][0]
+        tasks = openml.tasks.list_tasks(data_id=128, output_format='dataframe') #.to_numpy()
+        tasks = tasks[tasks.loc[:, "task_type"] == "Supervised Classification"]
+        tasks = tasks[tasks.loc[:, "estimation_procedure"] == "10-fold Crossvalidation"]
+        tasks = tasks[tasks.loc[:, "evaluation_measures"] == "predictive_accuracy"]
+        task_id = tasks.loc[:, "tid"].values[0]
         print("Task already exists. Task ID is", task_id)
 
 # reverting to prod server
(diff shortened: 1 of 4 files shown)


tests/test_runs/test_run_functions.py
```diff
@@ -1121,7 +1121,7 @@ def test_get_run(self):
 
     def _check_run(self, run):
         self.assertIsInstance(run, dict)
-        self.assertEqual(len(run), 7)
+        assert len(run) == 7, str(run)
```
janvanrijn: can you please explain in comments why 7? is not clear from the (direct) context

mfeurer: That's the amount of entries per run returned by the API: https://openml.org/api/v1/xml/run/list/flow/1154

mfeurer: I just pushed a commit explaining this. I will merge once the tests are through.

janvanrijn: good. I am not sure if we should really check this, as this is flexible for extension. But in those cases we can adjust the unit tests as well

examples/30_extended/tasks_tutorial.py
```diff
@@ -196,11 +196,11 @@
...
-        tasks = openml.tasks.list_tasks(data_id=128, output_format='dataframe').to_numpy()
-        tasks = tasks[tasks[:, 4] == "Supervised Classification"]
-        tasks = tasks[tasks[:, 6] == "10-fold Crossvalidation"]
-        tasks = tasks[tasks[:, 19] == "predictive_accuracy"]
-        task_id = tasks[0][0]
+        tasks = openml.tasks.list_tasks(data_id=128, output_format='dataframe') #.to_numpy()
+        tasks = tasks[tasks.loc[:, "task_type"] == "Supervised Classification"]
+        tasks = tasks[tasks.loc[:, "estimation_procedure"] == "10-fold Crossvalidation"]
+        tasks = tasks[tasks.loc[:, "evaluation_measures"] == "predictive_accuracy"]
+        task_id = tasks.loc[:, "tid"].values[0]
```
PGijsbers: You can access columns directly as if they were attributes (if the naming follows legal attribute naming). E.g. `tasks.task_type` instead of `tasks.loc[:, "task_type"]`. As I understand this should be identical, but (in my opinion) at lot easier to parse?

Or use panda's query functionality:
```
tasks.query('task_type == "Supervised Classification" '
            'and estimation_procedure == "10-fold Crossvalidation" '
            'and evaluation_measures == "predictive_accuracy"')
```

Neeratyoy: In a recent push I made I received feedback that pandas might do away with the support for named access to a column attribute such as `task.task_type`. Hence, I am using `loc`. 
Though I do agree that direct column access makes the code readable.
Also, I too wanted to use a query but kept the structure that was already there wrt filtering hierarchically. 
I'll make it into a query and push again.

mfeurer: I think this could be simplified to `task_id = tasks['tid'][0]`. However, I think we should rather try to get the new version out than doing minor simplifications.

PR #861, opened 2019-10-30
Fix736

### LINKED ISSUE #736: openml registers handler on root logger
`openml` registers a streamhandler with the root logger, because [it uses `logging.basicConfig`](https://github.com/openml/openml-python/blob/develop/openml/config.py). That functions registers a `logging.StreamHandler` with the root logger if it does not yet have a handler. This means that logging of other packages may automatically be processed by this streamhandler. An example to demonstrate:

```python
import logging

if __name__ == '__main__':
    #  Commenting out the line below prevents `basicConfig` from registering a streamhandler
    #  logging.getLogger().addHandler(logging.NullHandler())
    
    log = logging.getLogger('mypackage')
    log.setLevel(logging.DEBUG)

    log.info("Message one")  # first call, no log handler registered
    import openml
    log.info("Message two")  # second call, log handler registered by openml
```
output:
```
[INFO] [14:47:46:mypackage] Message two
```

Due to the `logging.basicConfig` behavior, if a handler already is registered (such as is done in the commented out line), it will not add a `logging.StreamHandler`. So there is no output if the relevant line of code is commented out.

I also see a verbosity in the config defaults, but I don't see it used anywhere. Am I missing something?

I think adding a `logging.StreamHandler` to the root logger is undesirable behavior. Developers should have full autonomy over their logging handlers without resorting to adding `logging.NullHandler`. I think we should avoid using `logging.basicConfig` to set our streamhandler. Instead, we should register a streamhandler to an openml specific logger instead of the root (i.e. `log = logging.getLogger('openml')` or `log = logging.getLogger(__name__)`). We should make sure to take into account the verbosity level as defined in the config when doing so.

PGijsbers:
Fixes #736.

 - All openml package logging is logged through the 'openml' logger (this fixes side-effects of working with the root logger that are demonstrated in the issue and code below).
 - All messages at `logging.WARNING` and up are printed to console (by default). This is in line with old behavior.
 - A file log of messages at `logging.DEBUG` and up (i.e. all of them) are logged to a file in the cache directory (by default). A new log file is created if the old log file reaches a size of 1 MB. At most one current log file and and old log file will be kept on disk at a time (i.e. at most 2 MB of openml-python logs are kept).
 - The log level defaults can be changed through the configuration file, for both channels individually. A new `file_verbosity` is introduced alongside the regular `verbosity`. 

The difference can be seen when running the following lines of code:
```
import logging

if __name__ == '__main__':
    log = logging.getLogger('mypackage')
    log.setLevel(logging.DEBUG)

    log.info("Message one")
    import openml
    log.info("Message two") 

    from openml.flows import get_flow
    f = get_flow(123)
    f.description = ''
    f._to_dict()  # logging.warning will be called due to empty description.

    from openml.datasets import get_dataset
    iris = get_dataset(61)  # since I have it in cache, a logging.debug statement is called.
```
in current `develop` we will see:
```
[INFO] [14:11:03:mypackage] Message two
[DEBUG] [14:11:03:mypackage] Message three
[WARNING] [14:11:03:openml.flows.flow] Flow 'weka.LinearRegression's empty description
```
printed in console, with no (semi-)permanent of the log is kept automatically.

With this PR, the console output becomes:
```
[WARNING] [14:12:42:openml.flows.flow] Flow 'weka.LinearRegression's empty description
```
and in my cache directory I find a file 'openml_python.log' with the following content:
```
[WARNING] [14:12:59:openml.flows.flow] Flow 'weka.LinearRegression's empty description
[DEBUG] [14:12:59:openml.datasets.dataset] Data pickle file already exists and is up to date.
```

I don't know how to unit test this, if at all.

----
A note on examples. I noticed some examples use logging (run setup and the fanova example). The generated pages do not show any logging output, so I was wondering what the purpose for the logging is. Perhaps @janvanrijn can help me here, as he seems to be the author.

### DIFF
--- a/examples/40_paper/2018_kdd_rijn_example.py
+++ b/examples/40_paper/2018_kdd_rijn_example.py
@@ -15,16 +15,10 @@ Publication
 | In *Proceedings of the 24th ACM SIGKDD International Conference on Knowledge Discovery & Data Mining*, 2018
 | Available at https://dl.acm.org/citation.cfm?id=3220058
 """
-import logging
 import sys
 
-logger = logging.getLogger(__name__)
-console_output = logging.StreamHandler()
-console_output.setLevel(logging.INFO)
-logger.addHandler(console_output)
-
 if sys.platform == 'win32':  # noqa
-    logger.critical('The pyrfr library (requirement of fanova) can currently not be installed on Windows systems')
+    print('The pyrfr library (requirement of fanova) can currently not be installed on Windows systems')
     exit()
 
 import json
@@ -84,8 +78,8 @@ fanova_results = []
 for idx, task_id in enumerate(suite.tasks):
     if limit_nr_tasks is not None and idx >= limit_nr_tasks:
         continue
-    logger.info('Starting with task %d (%d/%d)' % (task_id, idx+1,
-                                                    len(suite.tasks) if limit_nr_tasks is None else limit_nr_tasks))
+    print('Starting with task %d (%d/%d)'
+          % (task_id, idx+1, len(suite.tasks) if limit_nr_tasks is None else limit_nr_tasks))
     # note that we explicitly only include tasks from the benchmark suite that was specified (as per the for-loop)
     evals = openml.evaluations.list_evaluations_setups(
         evaluation_measure, flow=[flow_id], task=[task_id], size=limit_per_task, output_format='dataframe')
@@ -102,7 +96,7 @@ for idx, task_id in enumerate(suite.tasks):
                                           **{performance_column: setup[performance_column]})
                                      for _, setup in evals.iterrows()])
     except json.decoder.JSONDecodeError as e:
-        logger.warning('Task %d error: %s' % (task_id, e))
+        print('Task %d error: %s' % (task_id, e))
         continue
     # apply our filters, to have only the setups that comply to the hyperparameters we want
     for filter_key, filter_value in parameter_filters.items():
@@ -131,7 +125,7 @@ for idx, task_id in enumerate(suite.tasks):
             # functional ANOVA sometimes crashes with a RuntimeError, e.g., on tasks where the performance is constant
             # for all configurations (there is no variance). We will skip these tasks (like the authors did in the
             # paper).
-            logger.warning('Task %d error: %s' % (task_id, e))
+            print('Task %d error: %s' % (task_id, e))
             continue
 
 # transform ``fanova_results`` from a list of dicts into a DataFrame
--- a/examples/40_paper/2018_kdd_rijn_example.py
+++ b/examples/40_paper/2018_kdd_rijn_example.py
@@ -24,7 +24,7 @@ console_output.setLevel(logging.INFO)
 logger.addHandler(console_output)
 
 if sys.platform == 'win32':  # noqa
-    logger.warning('The pyrfr library (requirement of fanova) can currently not be installed on Windows systems')
+    logger.critical('The pyrfr library (requirement of fanova) can currently not be installed on Windows systems')
     exit()
 
 import json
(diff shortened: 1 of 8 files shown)


examples/40_paper/2018_kdd_rijn_example.py
```diff
@@ -15,22 +15,26 @@
...
 import sys
 
+logger = logging.getLogger(__name__)
+console_output = logging.StreamHandler()
+console_output.setLevel(logging.INFO)
+logger.addHandler(console_output)
+
 if sys.platform == 'win32':  # noqa
-    logging.warning('The pyrfr library (requirement of fanova) can currently not be installed on Windows systems')
+    logger.warning('The pyrfr library (requirement of fanova) can currently not be installed on Windows systems')
```
mfeurer: This should actually be `critical`.

openml/config.py
```diff
@@ -2,23 +2,44 @@
...
+    openml_logger.addHandler(console_stream)
+    openml_logger.addHandler(file_stream)
+
 
 # Default values!
 _defaults = {
     'apikey': None,
     'server': "https://www.openml.org/api/v1/xml",
-    'verbosity': 0,
+    'verbosity': logging.WARNING,
```
mfeurer: Not sure if (and how) we should follow the standards we once set up for this: https://github.com/openml/OpenML/wiki/Client-API-Standards

openml/config.py
```diff
@@ -101,19 +122,28 @@ def _setup():
...
     server = config.get('FAKE_SECTION', 'server')
 
     short_cache_dir = config.get('FAKE_SECTION', 'cachedir')
     cache_directory = os.path.expanduser(short_cache_dir)
 
+    # create the cache subdirectory
+    try:
+        os.mkdir(cache_directory)
+    except (IOError, OSError):
+        # TODO add debug information
```
mfeurer: Could you please do so?

mfeurer:
> The generated pages do not show any logging output, so I was wondering what the purpose for the logging is. Perhaps @janvanrijn can help me here, as he seems to be the author.

I guess we should move to print then. Could you please change that?

mfeurer:
This looks good besides my (small comments). Could you please also add this to the changelog?

openml/config.py
PGijsbers: I was unaware of this document. I propose to map the verbosity for console logging (naturally) as:
 - 0, normal: `logging.WARNING`
 - 1, info: `logging.INFO`
 - 2 debug: `logging.DEBUG`

I suppose it makes sense to do this conversion in Python only, so that a user that uses both R and Python can share its configuration file across platforms (as was intended). It's only a minor inconvenience at our part to map the numbers.

The file specifies a *minimal* standard. So to me that implicitly reads that it may be expanded, and I think adding the `file_verbosity` is fine (but for consistency, probably also should follow 0/1/2 log levels).

mfeurer: I agree.

openml/config.py
```diff
@@ -123,6 +153,13 @@ def _setup():
             'server load reasonable'
         )
 
+    # Verbosity levels as defined (https://github.com/openml/OpenML/wiki/Client-API-Standards)
+    # don't match Python values directly:
+    verbosity_map = {0: logging.WARNING, 1: logging.INFO, 2: logging.DEBUG}
+    console_log_level = verbosity_map[_defaults['verbosity']]
```
mfeurer: I'm afraid that you need to change this mapping here...

mfeurer:
I know it wasn't part of my original requests, but what do you think about having an example in which we show how one can easily change the log level? That might be good for folks who do ML, but do not care too much about how logging works in python.

openml/config.py
PGijsbers: Oops. Sorry. Though I shouldn't change the mapping here, but rather change the default values. At least, I assume we want to use the API Client defined levels instead of the Python levels (and values)?

PR #862, opened 2019-11-05
Adding license to each source file

Neeratyoy:
This is w.r.t. the [JMLR guidelines](http://www.jmlr.org/mloss/mloss-info.html) requiring the license to be mentioned in each and every source file of the package.

#### Any other comments?
The sklearn repo was used as a reference. Random examples below:
* [Source code](https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/svm/__init__.py#L11)
* [Source code](https://github.com/scikit-learn/scikit-learn/blob/a311ec0d8bd6d6cb9d926448cfd9cd128a0e5847/sklearn/gaussian_process/kernels.py#L17)
* [Example](https://github.com/scikit-learn/scikit-learn/blob/a311ec0d8bd6d6cb9d926448cfd9cd128a0e5847/sklearn/gaussian_process/kernels.py#L17)

If I understood the __Submission Checklist__ correctly, they do not expect the package zip to contain other auxiliary files such as the rst files for documentation, testing files, examples, etc.


Neeratyoy:
Currently, have added license to only the source files under `openml.datasets` and an example (which I checked by rendering the docs).

mfeurer (commented):
That looks alright. Great that you already checked the example. I think you can then annotate the whole package.

Could you please update the contribution guides, the changelog and the PR template with this information, too?

Neeratyoy:
@mfeurer I shall make the changes and create a PR.
Just to confirm, were my assumptions correct? That is, should I not touch other directories and edit just the `openml` and `examples` directories?

mfeurer:
I guess it would be great if for consistency we just annotate all files where possible. What do you think about that?

Neeratyoy:
Makes sense. I'll see where it is not possible. Especially in the rst files.

mfeurer (approved):

Issue #825, opened 2019-10-15
Can't create flow from ColumnTransformer


amueller:
Trying to write a general ColumnTransformer to run a decision stump.
I tried passing the feature masks as boolean arrays, which broke. Converting to lists of integers works a bit further but also breaks.

```python
import openml
import numpy as np
import pandas as pd
from sklearn.pipeline import make_pipeline
from sklearn.compose import ColumnTransformer
from sklearn.preprocessing import OneHotEncoder, StandardScaler
from sklearn.impute import SimpleImputer
from sklearn.tree import DecisionTreeClassifier

def get_frequent_task(did):
    tasks = openml.tasks.list_tasks(task_type_id=1, data_id=did)
    most_frequent_task = pd.Series({task: len(openml.runs.list_runs(task=[task])) for task in tasks}).index[0]
    return (most_frequent_task, tasks[most_frequent_task]['NumberOfInstances'])

dataset_id = 61
dataset = openml.datasets.get_dataset(dataset_id)

X, y, categorical_ind, feature_names = dataset.get_data(target=dataset.default_target_attribute, dataset_format='array')
categorical_ind = np.array(categorical_ind)
cat_idx, = np.where(categorical_ind)
cont_idx, = np.where(~categorical_ind)

clf = make_pipeline(ColumnTransformer([('cat', make_pipeline(SimpleImputer(strategy='most_frequent'), OneHotEncoder()), cat_idx.tolist()),
                                       ('cont', make_pipeline(SimpleImputer(strategy='median'), StandardScaler()), cont_idx.tolist())])          
                , DecisionTreeClassifier(max_depth=1))
if not categorical_ind.any():
    clf[0].set_params(cat='drop')
if not (~categorical_ind).any():
    clf[0].set_params(cont='drop')
hotencoded = clf[:-1].fit_transform(X)

bla = openml.runs.run_model_on_task(model=clf, task=get_frequent_task(dataset_id), return_flow=True)
```
TypeError: Second item of tuple does not match assumptions. Expected OpenMLFlow, got <class 'str'>

I guess having a "drop" transformer in ColumnTransformer doesn't work?

amueller:
Hm I think supporting ``'drop'`` correctly will be tricky. This will be different flows depending on whether a transformer is ``'drop'`` or an actual estimator, right? But this is a parameter setting according to sklearn....

amueller:
This is closely related to #480.

Issue #480, opened 2018-06-19
Cope with empty steps in scikit-learn


mfeurer:
Reproduce with:
```python
import openml

import sklearn.pipeline
import sklearn.preprocessing
import sklearn.feature_selection
import sklearn.svm

#flow = openml.flows.get_flow(7103)

flow = sklearn.pipeline.Pipeline(
    [
        ('imputation', sklearn.preprocessing.Imputer()),
        ('onehot', sklearn.preprocessing.OneHotEncoder(handle_unknown='ignore',
                                                       categorical_features=[])),
        ('variancethreshold', None),
        ('classifier', sklearn.svm.SVC())
    ]
)

task = openml.tasks.get_task(3)
#run = openml.runs.run_flow_on_task(task, flow)
run = openml.runs.run_model_on_task(task, flow)
```
follow up of issue #479

janvanrijn:
**Current Outcome:**

```
Traceback (most recent call last):
  File "/home/janvanrijn/projects/openml-python-contrib/openmlcontrib/testjan.py", line 24, in <module>
    run = openml.runs.run_model_on_task(task, flow)
  File "/home/janvanrijn/projects/openml-python/openml/runs/functions.py", line 50, in run_model_on_task
    add_local_measures=add_local_measures)
  File "/home/janvanrijn/projects/openml-python/openml/runs/functions.py", line 129, in run_flow_on_task
    run.parameter_settings = OpenMLRun._parse_parameters(flow)
  File "/home/janvanrijn/projects/openml-python/openml/runs/run.py", line 455, in _parse_parameters
    True, flow.flow_id)
  File "/home/janvanrijn/projects/openml-python/openml/runs/run.py", line 402, in extract_parameters
    sorted(expected_parameters| expected_components), sorted(model_parameters)))
ValueError: Parameters of the model do not match the parameters expected by the flow:
expected flow parameters: ['classifier', 'imputation', 'memory', 'onehot', 'steps']
model parameters: ['classifier', 'imputation', 'memory', 'onehot', 'steps', 'variancethreshold']
```

What would be the expected outcome?

mfeurer:
The `variancethreshold` should be part of the expected flow parameters.

Neeratyoy:
Using sklearn version `0.21.3`, using the modified exampled from above as mentioned below, the error doesn't seem to occur. 

```python
from sklearn.impute import SimpleImputer 
import sklearn.pipeline
import sklearn.svm

import openml

[...]

PR #866, opened 2019-11-06
To handle non-actionable steps in sklearn

Neeratyoy:
#### Reference Issue

Addresses #825 and #480. 

The serialization expects each step to be a sklearn module and eventually an OpenMLFlow or a None. For ColumnTransform operations, a _drop_ or _passthrough_ can be specified to include a step in a pipeline but not execute it. The changes in this PR incorporates this additional string handling of the keywords _drop_ and _passthrough_ in a ColumnTransformer call.

```python
import openml
import numpy as np
import pandas as pd
from sklearn.pipeline import make_pipeline
from sklearn.compose import ColumnTransformer
from sklearn.preprocessing import OneHotEncoder, StandardScaler
from sklearn.impute import SimpleImputer
from sklearn.tree import DecisionTreeClassifier

dataset_id = 61
dataset = openml.datasets.get_dataset(dataset_id)

task = openml.tasks.get_task(59)

X, y, categorical_ind, feature_names = dataset.get_data(target=dataset.default_target_attribute, dataset_format='array')
categorical_ind = np.array(categorical_ind)
cat_idx, = np.where(categorical_ind)
cont_idx, = np.where(~categorical_ind)

clf = make_pipeline(ColumnTransformer([('cat', make_pipeline(SimpleImputer(strategy='most_frequent'), OneHotEncoder()), cat_idx.tolist()),
                                       ('cont', make_pipeline(SimpleImputer(strategy='median'), StandardScaler()), cont_idx.tolist())])          
                , DecisionTreeClassifier(max_depth=1))
if not categorical_ind.any():
    clf[0].set_params(cat='drop')
if not (~categorical_ind).any():
    clf[0].set_params(cont='drop')
hotencoded = clf[:-1].fit_transform(X)

bla = openml.runs.run_model_on_task(model=clf, task=task, return_flow=True)
```

#### Any other comments?
This PR handles the specific case of a ColumnTransform. I'm not sure if there are other such keywords which should be handled equivalently. We can then maintain a keyword listing, based on the sklearn version under extension.py and handle them with a similar logic. 
If there are more diverse cases that may arise, some examples would help in designing a generalized solution in that case.
Would need to accordingly design relevant unit tests too.

### DIFF
--- a/openml/extensions/sklearn/extension.py
+++ b/openml/extensions/sklearn/extension.py
@@ -754,7 +754,7 @@ class SklearnExtension(Extension):
     def _get_external_version_string(
         self,
         model: Any,
-        sub_components: Dict[str, Union[OpenMLFlow, str, None]],
+        sub_components: Dict[str, OpenMLFlow],
     ) -> str:
         # Create external version string for a flow, given the model and the
         # already parsed dictionary of sub_components. Retrieves the external
@@ -785,9 +785,9 @@ class SklearnExtension(Extension):
     def _check_multiple_occurence_of_component_in_flow(
         self,
         model: Any,
-        sub_components: Dict[str, Union[OpenMLFlow, str, None]],
+        sub_components: Dict[str, OpenMLFlow],
     ) -> None:
-        to_visit_stack = []  # type: List[Union[OpenMLFlow, str, None]]
+        to_visit_stack = []  # type: List[OpenMLFlow]
         to_visit_stack.extend(sub_components.values())
         known_sub_components = set()  # type: Set[str]
 
@@ -810,7 +810,7 @@ class SklearnExtension(Extension):
     ) -> Tuple[
         'OrderedDict[str, Optional[str]]',
         'OrderedDict[str, Optional[Dict]]',
-        'OrderedDict[str, Union[OpenMLFlow, str, None]]',
+        'OrderedDict[str, OpenMLFlow]',
         Set,
     ]:
         # This function contains four "global" states and is quite long and
@@ -820,7 +820,7 @@ class SklearnExtension(Extension):
         # separate class methods
 
         # stores all entities that should become subcomponents
-        sub_components = OrderedDict()  # type: OrderedDict[str, Union[OpenMLFlow, str, None]]
+        sub_components = OrderedDict()  # type: OrderedDict[str, OpenMLFlow]
         # stores the keys of all subcomponents that should become
         sub_components_explicit = set()
         parameters = OrderedDict()  # type: OrderedDict[str, Optional[str]]
--- a/openml/extensions/sklearn/extension.py
+++ b/openml/extensions/sklearn/extension.py
@@ -880,8 +880,8 @@ class SklearnExtension(Extension):
                         else:
                             pass
                     elif isinstance(sub_component, type(None)):
-                        msg = 'Cannot serialize objects of None type. Please replace with a '\
-                              'relevant placeholder. Note that empty sklearn estimators can be '\
+                        msg = 'Cannot serialize objects of None type. Please use a valid ' \
+                              'placeholder for None. Note that empty sklearn estimators can be '\
                               'replaced with \'drop\' or \'passthrough\'.'
                         raise ValueError(msg)
                     elif not isinstance(sub_component, OpenMLFlow):
--- a/openml/extensions/sklearn/extension.py
+++ b/openml/extensions/sklearn/extension.py
@@ -698,8 +698,6 @@ class SklearnExtension(Extension):
                 name = subcomponents[key].name
             elif isinstance(subcomponents[key], str):  # 'drop', 'passthrough' can be passed
                 name = subcomponents[key]
-            elif subcomponents[key] is None:
-                name = "None"
             if key in subcomponents_explicit:
                 sub_components_names += "," + key + "=" + name
             else:
@@ -776,7 +774,7 @@ class SklearnExtension(Extension):
         external_versions.add(sklearn_version)
         for visitee in sub_components.values():
             # 'drop', 'passthrough', None can be passed as estimators
-            if isinstance(visitee, str) or visitee is None:
+            if isinstance(visitee, str):
                 continue
             for external_version in visitee.external_version.split(','):
                 external_versions.add(external_version)
@@ -795,8 +793,6 @@ class SklearnExtension(Extension):
             visitee = to_visit_stack.pop()
             if isinstance(visitee, str):  # 'drop', 'passthrough' can be passed as estimators
                 known_sub_components.add(visitee)
-            elif visitee is None:  # a None step can be included in a Pipeline
-                known_sub_components.add(str(visitee))
             elif visitee.name in known_sub_components:
                 raise ValueError('Found a second occurence of component %s when '
                                  'trying to serialize %s.' % (visitee.name, model))
@@ -883,7 +879,12 @@ class SklearnExtension(Extension):
                             raise ValueError(msg)
                         else:
                             pass
-                    elif not isinstance(sub_component, (OpenMLFlow, type(None))):
+                    elif isinstance(sub_component, type(None)):
+                        msg = 'Cannot serialize objects of None type. Please replace with a '\
+                              'relevant placeholder. Note that empty sklearn estimators can be '\
+                              'replaced with \'drop\' or \'passthrough\'.'
+                        raise ValueError(msg)
+                    elif not isinstance(sub_component, OpenMLFlow):
                         msg = 'Second item of tuple does not match assumptions. ' \
                               'Expected OpenMLFlow, got %s' % type(sub_component)
                         raise TypeError(msg)
--- a/openml/extensions/sklearn/extension.py
+++ b/openml/extensions/sklearn/extension.py
@@ -694,10 +694,14 @@ class SklearnExtension(Extension):
         # will be part of the name (in brackets)
         sub_components_names = ""
         for key in subcomponents:
-            if key in subcomponents_explicit:
-                sub_components_names += "," + key + "=" + subcomponents[key].name
+            if isinstance(subcomponents[key], str):
+                name = subcomponents[key]
             else:
-                sub_components_names += "," + subcomponents[key].name
+                name = subcomponents[key].name
+            if key in subcomponents_explicit:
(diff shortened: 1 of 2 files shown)


openml/extensions/sklearn/extension.py
```diff
@@ -694,10 +694,14 @@ def _serialize_model(self, model: Any) -> OpenMLFlow:
         # will be part of the name (in brackets)
         sub_components_names = ""
         for key in subcomponents:
+            if isinstance(subcomponents[key], str):
```
mfeurer: Could you please add a comment why this is necessary?

openml/extensions/sklearn/extension.py
```diff
@@ -769,21 +773,25 @@ def _get_external_version_string(
         external_versions.add(openml_version)
         external_versions.add(sklearn_version)
         for visitee in sub_components.values():
+            if isinstance(visitee, str):
```
mfeurer: Same as above, why could this be a string?

openml/extensions/sklearn/extension.py
```diff
@@ -769,21 +773,25 @@ def _get_external_version_string(
...
+                continue
             for external_version in visitee.external_version.split(','):
                 external_versions.add(external_version)
         return ','.join(list(sorted(external_versions)))
 
     def _check_multiple_occurence_of_component_in_flow(
         self,
         model: Any,
-        sub_components: Dict[str, OpenMLFlow],
+        sub_components: Dict[str, Any],
```
mfeurer: Hm, why can this map to `Any` now?

openml/extensions/sklearn/extension.py
```diff
@@ -769,21 +773,25 @@ def _get_external_version_string(
...
     ) -> None:
         to_visit_stack = []  # type: List[OpenMLFlow]
         to_visit_stack.extend(sub_components.values())
         known_sub_components = set()  # type: Set[str]
         while len(to_visit_stack) > 0:
             visitee = to_visit_stack.pop()
-            if visitee.name in known_sub_components:
+            if isinstance(visitee, str):
```
mfeurer: Same as above.

openml/extensions/sklearn/extension.py
```diff
@@ -796,7 +804,7 @@ def _extract_information_from_model(
     ) -> Tuple[
         'OrderedDict[str, Optional[str]]',
         'OrderedDict[str, Optional[Dict]]',
-        'OrderedDict[str, OpenMLFlow]',
+        'OrderedDict[str, Any]',
```
mfeurer: Same as above.

openml/extensions/sklearn/extension.py
```diff
@@ -860,7 +868,16 @@ def flatten_all(list_):
                         # length 3 is for ColumnTransformer
                         msg = 'Length of tuple does not match assumptions'
                         raise ValueError(msg)
-                    if not isinstance(sub_component, (OpenMLFlow, type(None))):
+
+                    if isinstance(sub_component, str):
```
mfeurer: This makes sense, but could you please add an explanation of which scikit-learn model these support?

Neeratyoy: I shall add the explanations based on the example I know. However, as I mentioned in the PR comment, I am not sure if these changes/assumptions can generalize to ALL such sklearn models.
I don't think we need to handle cases wherein sklearn cannot process a user input, that is, a str in place of an estimator. For a `ColumnTransform` sklearn handles the keywords _drop_ and _passthrough_ by treating them as estimators that are not evaluated. Hence, I am explicitly handling this case.
Should I forage the sklearn docs a bit more to see if any such models exist?

mfeurer: I don't know any other examples myself, so testing the example from the issue plus maybe one or two examples from the scikit-learn docs should be sufficient. I don't think you need to search for other scikit-learn models which allow strings, though. Maybe @amueller has some additional input?

openml/extensions/sklearn/extension.py
Neeratyoy: The annotations used earlier expected the _subcomponent_ to be OpenML, however, we now need to handle _str_ too. I had the type annotations as `Dict[str, Union[str, OpenMLFlow]]`. However, flake8 was not able to decipher the rest of the if...else bodies and was raising potential errors, which I am handling with exactly those if...else cases. 
I agree that having *Any* is loosening the type checks but I wanted to see if these changes affect test cases on Travis, so I went ahead with Any in this draft PR since I was unable to think of a better solution.

mfeurer: Could you please paste the error you got for the annotation `Dict[str, Union[str, OpenMLFlow]]`?

Neeratyoy: There were a dozen or more :neutral_face: 
One of them, for instance, was for [this](https://github.com/openml/openml-python/blob/78b163e8a3f7fe2c977a26912f7a7a4ae2742f2a/openml/extensions/sklearn/extension.py#L700).
_str has got no 'name' attribute'_

openml/extensions/sklearn/extension.py
Neeratyoy: Sure, will do so.

openml/extensions/sklearn/extension.py
mfeurer: I think the solution is to use an if/elif/else statement here, where the access to `name` is after a check if the instance is of type `OpenMLFlow`. I'm not really happy about this as it bloat up the code, but it will allow us to keep the annotations as meaningful as possible.

Neeratyoy: Though test cases passed locally, it might fail on travis since I still get these two flake issues.

`openml/extensions/sklearn/extension.py:698: error: Item "str" of "Union[OpenMLFlow, str, None]" has no attribute "name"`
`openml/extensions/sklearn/extension.py:698: error: Item "None" of "Union[OpenMLFlow, str, None]" has no attribute "name"`

I am unable to think of how I can bypass how flake is parsing the if-conditions.
It doesn't make sense, for multiple types.

mfeurer: That's annoying. You could try [cast](https://docs.python.org/3/library/typing.html#typing.cast).

openml/extensions/sklearn/extension.py
```diff
@@ -782,15 +785,18 @@ def _get_external_version_string(
...
+        to_visit_stack = []  # type: List[Union[OpenMLFlow, str, None]]
         to_visit_stack.extend(sub_components.values())
         known_sub_components = set()  # type: Set[str]
+
         while len(to_visit_stack) > 0:
             visitee = to_visit_stack.pop()
-            if isinstance(visitee, str):
+            if isinstance(visitee, str):  # 'drop', 'passthrough' can be passed as estimators
                 known_sub_components.add(visitee)
+            elif visitee is None:  # a None step can be included in a Pipeline
```
mfeurer: Can we maybe simplify the code and only support `passthrough` instead of `None`? It appears to me that `passthrough` can be used in the same places and is much clearer to handle. Maybe raise an error message if a user puts in a `None` step?

Neeratyoy: On checking further, [FeatureUnion supports _drop_ and _None_](https://github.com/openml/openml-python/blob/cbf0d6a749a4f23245f15668a548f235c61b82f6/tests/test_extensions/test_sklearn_extension/test_sklearn_extension.py#L678) while [Pipeline supports _passthrough_ and _None_](https://github.com/openml/openml-python/blob/cbf0d6a749a4f23245f15668a548f235c61b82f6/tests/test_extensions/test_sklearn_extension/test_sklearn_extension.py#L1804).

Should we still _not_ support None and raise an Exception along the lines of: 
`None not allowed, please use 'drop' or 'passthrough'`?

However, when creating the sklearn Pipeline object, using _drop_ flags an error.
Whereas for FeatureUnion, using _passthrough_ is not flagged until runtime, i.e., calling _fit_.

Not sure what would be the best design here, especially wrt the Exception message, if we do not allow None.

Neeratyoy:
@mfeurer one of the old unit tests that I had to change `test_serialize_feature_union` fails for versions `0.18.2` and `0.19.2`.
It appears that the manner in which we are serializing, the behaviour of one of the intermediate stages have changed since these two versions. To be more specific, the parameter metadata extraction.

Should I then try to change the code to see how this can be handled or skip that particular unit test?

tests/test_extensions/test_sklearn_extension/test_sklearn_extension.py
```diff
@@ -1804,7 +1804,6 @@ def test_run_on_model_with_empty_steps(self):
         clf = sklearn.pipeline.Pipeline([
             ('dummystep', 'passthrough'),  # adding 'passthrough' as an estimator
             ('prep', clf),
-            ('variancethreshold', None),  # adding 'None' as an estimator
```
mfeurer: Is there a reason you dropped this and didn't replace it by `drop`?

openml/extensions/sklearn/extension.py
```diff
@@ -883,7 +879,12 @@ def flatten_all(list_):
                             raise ValueError(msg)
                         else:
                             pass
-                    elif not isinstance(sub_component, (OpenMLFlow, type(None))):
+                    elif isinstance(sub_component, type(None)):
+                        msg = 'Cannot serialize objects of None type. Please replace with a '\
```
mfeurer: Could you please add a unit test that serializing `None` indeed fails? Also, could you please phrase it as `Please use a valid placeholder for None`?

mfeurer:
> Should I then try to change the code to see how this can be handled or skip that particular unit test?

You can skip that unit test.

tests/test_extensions/test_sklearn_extension/test_sklearn_extension.py
Neeratyoy: Yeah, since Pipeline supports only _None_ and _passthrough_.

openml/extensions/sklearn/extension.py
Neeratyoy: Done.

openml/extensions/sklearn/extension.py
Neeratyoy: This was fixed.

Issue #468, opened 2018-05-10
Listing calls & URL size limit of OpenML server


janvanrijn:
The OpenML server has a URL size limit, which prevents us from doing listing calls with too many filters. For example, when I want to filter on 1000 task id's it is likely that the call will fail. It would be great if we can somehow automatically detect and catch this in the list all function. However, this is a though problem, as sometimes there are multiple filters (e.g., per task and per setup)

Extending this limit is not a legitimate fix, as the problem will re-occur when using bigger filters.

janvanrijn:
Just realized that there is a straightforward fix for this, i.e., allowing also post variables for the listing functions (although POST means authentication)

@mfeurer what do you think?

mfeurer:
Is the maximum URL length documented somewhere? Also, what's the server response? If the server response is something we can parse/display nicely, is there a reason to act upon this in addition?

Also, how would you handle this except raising an exception? Do you want to "parse" the query on the python side to chunk it? We could go back to post again, but we reverted that a while ago to not have to authenticate, so I don't think that's a great idea. Maybe we can catch a server exception 'URL too long' and then try a post request?

janvanrijn:
I think this issue will mostly occur with setups, as these can not be filtered by other means (tasks, runs, etc)

I added a function to the openml contrib library:
https://github.com/openml/openml-python-contrib/blob/master/openmlcontrib/setups/functions.py

mfeurer:
This can be reproduced with:
```python 
import openml

openml.datasets.list_datasets(
    data_id=list(range(10000))
)
```

@janvanrijn we can add a helper to either call the get or post request depending on the length of the URI. However, to do so we need to know the maximal length the server allows.

PR #689, opened 2019-05-06
Feature/doc

MichaelMMeskhi:
#### Reference Issue
Issue No. 611

#### What does this PR implement/fix? Explain your changes.
Updated contributions documentation page. `Contributing` on OpenML docs now has CONTRIBUTING.md's information and is better structured. 

#### How should this PR be tested?
Contributing.rst was updated with more information. No checks required.

#### Any other comments?
My first pull request regarding my work for issue No. 611.

PR #881, opened 2019-11-12
add better error message for too-long URI

mfeurer:
In response to https://github.com/openml/openml-python/issues/468

### DIFF
--- a/openml/_api_calls.py
+++ b/openml/_api_calls.py
@@ -5,7 +5,7 @@ import hashlib
 import logging
 import requests
 import xmltodict
-from typing import Dict
+from typing import Dict, Optional
 
 from . import config
 from .exceptions import (OpenMLServerError, OpenMLServerException,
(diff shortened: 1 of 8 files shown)


openml/_api_calls.py
```diff
@@ -159,9 +231,12 @@ def _parse_server_exception(
     try:
         server_exception = xmltodict.parse(response.text)
     except Exception:
-        raise OpenMLServerError(
-            'Unexpected server error when calling {}. Please contact the developers!\n'
-            'Status code: {}\n{}'.format(url, response.status_code, response.text))
+        if response.status_code == 414:
```
PGijsbers: Does it not make sense to add this before the try/except block? I don't know the details, but sounds like you would always assume the parsing to fail if the response had code `414`.

PGijsbers:
It looks like there is no unit test verifying these long uris now raise a descriptive error?

openml/_api_calls.py
mfeurer: > Does it not make sense to add this before the try/except block?

Yes, it does.

> but sounds like you would always assume the parsing to fail if the response had code 414.

No, a lot of things can happen with the server output...

mfeurer:
I just added a test.

PGijsbers:
Approved and merged. Should we add a guideline not to move functions around? It's hard to spot what changes, if any, are made in a moved function as it is caught in the diff regardless). Unless it's a specific refactor PR or the dev explicitly says which lines have been changed (if any) for each moved function? Generally speaking, moving functions around won't (shouldn't) happen often, and the overhead introduced for the dev to point out changes (if any) is far smaller than the overhead for the reviewer trying to identify where changes are made (if any).

mfeurer:
That makes sense. Would you like to add this to the contribution guidelines (maybe finishing up the PR #689 on the way)?

PR #889, opened 2019-11-26
Changes proposed in #885. Don't register handlers by default.

### LINKED ISSUE #885: Package should not configure logging without being asked to do so
I think this is a followup to #736. I think the fact that OpenML is configuring logger at all is problematic, especially because it is doing at the `import openml` time without any way for anyone to prevent that. Currently, there is such code segment;

```
logger = logging.getLogger(__name__)


def configure_logging(console_output_level: int, file_output_level: int):
    """ Sets the OpenML logger to DEBUG, with attached Stream- and FileHandler. """
    # Verbosity levels as defined (https://github.com/openml/OpenML/wiki/Client-API-Standards)
    # don't match Python values directly:
    verbosity_map = {0: logging.WARNING, 1: logging.INFO, 2: logging.DEBUG}

    openml_logger = logging.getLogger('openml')
    openml_logger.setLevel(logging.DEBUG)
    message_format = '[%(levelname)s] [%(asctime)s:%(name)s] %(message)s'
    output_formatter = logging.Formatter(message_format, datefmt='%H:%M:%S')

    console_stream = logging.StreamHandler()
    console_stream.setFormatter(output_formatter)
    console_stream.setLevel(verbosity_map[console_output_level])

    one_mb = 2**20
    log_path = os.path.join(cache_directory, 'openml_python.log')
    file_stream = logging.handlers.RotatingFileHandler(log_path, maxBytes=one_mb, backupCount=1)
    file_stream.setLevel(verbosity_map[file_output_level])
    file_stream.setFormatter(output_formatter)

    openml_logger.addHandler(console_stream)
    openml_logger.addHandler(file_stream)
    return console_stream, file_stream
```

And that method is always called inside `__init__`.

Libraries should not do that. Libraries should log things to logger, but consumers of the libraries should then configure how logging happens. For example, inside `__main__` section or something like that.

The problem with the code above is that it writes to the file system just by importing a package. That can make problems on various computing platforms, especially grid computing, where one cannot just write around the shared systems.

`openml_logger.setLevel(logging.DEBUG)` is unnecessary. Every logger in fact starts with `NOTSET` logging level which is even lower (numerical 0) than `DEBUG` (numerical 10). But setting it might prevent somebody else to configure it before importing `openml`. So I suggest this is removed.

Adding handlers to the logger is generally OK, but might be unnecessary. For example, the user of the `openml` package might not be a person sitting in front of the interpreter, but the system. Which would configure that all logging is send to a central logging service over the network. Having to know that you also have to disable/remove some handlers added to the logger seems tricky.

I suggest this is made as a function one can call to configure, but not something which is done by default at import time. Or maybe one can enable it in the config file.

Alternatively, if you really want this behavior, I suggest this is done at the first call to OpenML API functions, and not at the import.

PGijsbers:
Fixes #885.

First implementation of changes suggested in #885:
 - don't register log handlers by default
 - don't set openml logger log level by default

A user can now specify the console and file log levels through `set_console_log_level` and `set_file_log_level` respectively. As input they expect the log level (both OpenML's 0/1/2 and Python log levels are accepted), and a bool `save_as_default`.  Actually saving to the config file is right now not supported (and will raise a `NotImplementedError`). Calling these functions registers the respective handler if it wasn't yet, set/change it to the appropriate value, and ensures the openml logger is set to the right log level.

Consider this code:

```python
import logging

if __name__ == '__main__':
    log = logging.getLogger('mypackage')
    log.setLevel(logging.DEBUG)

    log.info("Message one")
    # 1: just import OpenML
    import openml
    # 2: Specify console log
    # openml.config.set_console_log_level(logging.DEBUG)
    # 3: Specify file log
    # openml.config.set_file_log_level(logging.DEBUG)
    log.info("Message two")

    from openml.flows import get_flow
    f = get_flow(123)
    f.description = ''
    f._to_dict()  # logging.warning will be called due to empty description.

    from openml.datasets import get_dataset
    iris = get_dataset(61)  # since I have it in cache, a logging.debug statement is called.

```

The output we get, uncommenting respective lines for scenario '2' and '3':

1: `Flow 'weka.LinearRegression's empty description` to console only. This is default behavior for warning level messages. The debug message from the last line is ignored.
2: 
```
[WARNING] [11:51:24:openml.flows.flow] Flow 'weka.LinearRegression's empty description
[DEBUG] [11:51:24:openml.datasets.dataset] Data pickle file already exists and is up to date.
``` 
to console only.

3 (the line specifying console output was commented out again): 
```
[WARNING] [11:53:46:openml.flows.flow] Flow 'weka.LinearRegression's empty description
[DEBUG] [11:53:46:openml.datasets.dataset] Data pickle file already exists and is up to date.
```
written to file, no console output. Note that the warning message is now also not printed to console. 

---
This does not include changes to where the config/cache directories are located.

Writing to the config file is not yet supported. I noticed we hack our way around the configparser module. If we want to support writing to the log file from the python module (which I'm for), we need to look at other options or implement our own (our log file definition is very simple). In fact, 
I would further propose refactoring the openml.config module into smaller submodules as a lot of different things are going on within the same module now (loading, logging, test configuration).

I'm willing to do the above refactoring, but it might take a few weeks (I'm very busy until the holidays).

### DIFF
--- a/openml/config.py
+++ b/openml/config.py
@@ -22,6 +22,10 @@ file_handler = None
 def _create_log_handlers():
     """ Creates but does not attach the log handlers. """
     global console_handler, file_handler
+    if console_handler is not None or file_handler is not None:
+        logger.debug("Requested to create log handlers, but they are already created.")
+        return
+
     message_format = '[%(levelname)s] [%(asctime)s:%(name)s] %(message)s'
     output_formatter = logging.Formatter(message_format, datefmt='%H:%M:%S')
 
--- a/openml/config.py
+++ b/openml/config.py
@@ -7,7 +7,7 @@ Store module level information like the API key, cache directory and the server
 import logging
 import logging.handlers
 import os
-from typing import cast, Optional, Tuple, Any
+from typing import Optional, Tuple
 
 from io import StringIO
 import configparser
@@ -48,11 +48,7 @@ def _convert_log_levels(log_level: int) -> Tuple[int, int]:
     return openml_level, python_level
 
 
-def _set_level_register_and_store(
-        handler: logging.Handler,
-        log_level: int,
-        store_under_key: Optional[str] = None
-):
+def _set_level_register_and_store(handler: logging.Handler, log_level: int):
     """ Set handler log level, register it if needed, save setting to config file if specified. """
     oml_level, py_level = _convert_log_levels(log_level)
     handler.setLevel(py_level)
@@ -63,32 +59,17 @@ def _set_level_register_and_store(
     if handler not in openml_logger.handlers:
         openml_logger.addHandler(handler)
 
-    if store_under_key is not None:
-        _save_setting_to_configuration(store_under_key, oml_level)
 
-
-def set_console_log_level(
-        console_output_level: Optional[int] = None,
-        save_as_default: bool = False
-):
+def set_console_log_level(console_output_level: Optional[int] = None):
     """ Set console output to the desired level and register it with openml logger if needed. """
     global console_handler
-    store_under_key = 'verbosity' if save_as_default else None
-    _set_level_register_and_store(console_handler, console_output_level, store_under_key)
+    _set_level_register_and_store(console_handler, console_output_level)
 
 
-def set_file_log_level(
-        file_output_level: Optional[int] = None,
-        save_as_default: bool = False
-):
+def set_file_log_level(file_output_level: Optional[int] = None):
     """ Set file output to the desired level and register it with openml logger if needed. """
     global file_handler
-    store_under_key = 'file_verbosity' if save_as_default else None
-    _set_level_register_and_store(file_handler, file_output_level, store_under_key)
-
-
-def _save_setting_to_configuration(key: str, value: Any):
-    raise NotImplementedError("Storing values not yet supported.")
+    _set_level_register_and_store(file_handler, file_output_level)
 
 
 # Default values (see also https://github.com/openml/OpenML/wiki/Client-API-Standards)
@@ -223,9 +204,7 @@ def _setup():
 
 
 def _parse_config():
-    """Parse the config file, set up defaults.
-    """
-
+    """ Parse the config file, set up defaults. """
     config = configparser.RawConfigParser(defaults=_defaults)
 
     if not os.path.exists(config_file):
@@ -303,14 +282,4 @@ __all__ = [
 ]
 
 _setup()
-
-
 _create_log_handlers()
-_config = _parse_config()
-
-if _config['FAKE_SECTION'].get('verbosity', None) is not None:
-    _console_log_level = int(_config.get('FAKE_SECTION', 'verbosity'))
-    set_console_log_level(_console_log_level)
-if _config['FAKE_SECTION'].get('file_verbosity', None) is not None:
-    _file_log_level = int(_config.get('FAKE_SECTION', 'file_verbosity'))
-    set_file_log_level(_file_log_level)
--- a/openml/config.py
+++ b/openml/config.py
@@ -286,4 +286,4 @@ __all__ = [
 ]
 
 _setup()
-# _create_log_handlers()
+_create_log_handlers()
--- a/openml/config.py
+++ b/openml/config.py
@@ -7,47 +7,92 @@ Store module level information like the API key, cache directory and the server
 import logging
 import logging.handlers
 import os
-from typing import cast
+from typing import cast, Optional, Tuple, Any
 
 from io import StringIO
 import configparser
 from urllib.parse import urlparse
 
 logger = logging.getLogger(__name__)
+openml_logger = logging.getLogger('openml')
+console_handler = None
+file_handler = None
 
 
-def configure_logging(console_output_level: int, file_output_level: int):
-    """ Sets the OpenML logger to DEBUG, with attached Stream- and FileHandler. """
-    # Verbosity levels as defined (https://github.com/openml/OpenML/wiki/Client-API-Standards)
-    # don't match Python values directly:
-    verbosity_map = {0: logging.WARNING, 1: logging.INFO, 2: logging.DEBUG}
-
-    openml_logger = logging.getLogger('openml')
-    openml_logger.setLevel(logging.DEBUG)
+def _create_log_handlers():
+    """ Creates but does not attach the log handlers. """
+    global console_handler, file_handler
     message_format = '[%(levelname)s] [%(asctime)s:%(name)s] %(message)s'
     output_formatter = logging.Formatter(message_format, datefmt='%H:%M:%S')
 
-    console_stream = logging.StreamHandler()
-    console_stream.setFormatter(output_formatter)
-    console_stream.setLevel(verbosity_map[console_output_level])
+    console_handler = logging.StreamHandler()
+    console_handler.setFormatter(output_formatter)
 
-    one_mb = 2**20
+    one_mb = 2 ** 20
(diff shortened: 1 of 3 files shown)


PGijsbers:
> Also, I don't understand what the code for storing something in the config is supposed to do?

Nothing right now (it's not implemented). But the idea is that a user can adjust their log settings (or perhaps all settings, e.g. apikey) programmatically, and persist those settings in the configuration file. This way a user can make permanent configuration changes without having to manually find/edit the configuration file.

mfeurer:
> This way a user can make permanent configuration changes without having to manually find/edit the configuration file.

Hm, but wouldn't a user put the configuration of logging in the startup of an application, and want it to be different for different applications using OpenML? Basically, I don't think this is something we should handle in OpenML.

PGijsbers:
Then what is the point of including verbosity options in the configuration file?

mfeurer:
> Then what is the point of including verbosity options in the configuration file?

Good point. We decided upon that once upon a time, but I'm not convinced that that was a good idea.

mitar:
I think config file could control only when running inside a interactive interpreter, or Jupyter, or something like that. Something [like this could work](https://stackoverflow.com/questions/6108330/checking-for-interactive-shell-in-a-python-script)?

But yes, I am also not convinced that having that in a configuration is reasonable. Because even if you are using interactive interpreter, it might be that openml library is used as a 3rd party to some other library and you are not really expecting to get messages from it.

openml/config.py
```diff
@@ -7,47 +7,94 @@
...
-    """ Sets the OpenML logger to DEBUG, with attached Stream- and FileHandler. """
-    # Verbosity levels as defined (https://github.com/openml/OpenML/wiki/Client-API-Standards)
-    # don't match Python values directly:
-    verbosity_map = {0: logging.WARNING, 1: logging.INFO, 2: logging.DEBUG}
-
-    openml_logger = logging.getLogger('openml')
-    openml_logger.setLevel(logging.DEBUG)
+def _create_log_handlers():
+    """ Creates but does not attach the log handlers. """
+    global console_handler, file_handler
```
mitar: I suggest checking first if they are still `None`, and return early if they are not. So that calling this multiple times does nothing. Otherwise you might override some later changes done by people to original handlers.

openml/config.py
```diff
@@ -7,47 +7,94 @@
...
+    handler.setLevel(py_level)
+
+    if openml_logger.level > py_level or openml_logger.level == logging.NOTSET:
+        openml_logger.setLevel(py_level)
+
+    if handler not in openml_logger.handlers:
+        openml_logger.addHandler(handler)
+
+    if store_under_key is not None:
+        _save_setting_to_configuration(store_under_key, oml_level)
```
mitar: Personally I prefer adding TODO comments instead of implementing a function which throws an error.

openml/config.py
```diff
@@ -7,47 +7,94 @@
...
+    global console_handler
+    store_under_key = 'verbosity' if save_as_default else None
+    _set_level_register_and_store(console_handler, console_output_level, store_under_key)
+
+
+def set_file_log_level(
+        file_output_level: Optional[int] = None,
+        save_as_default: bool = False
+):
+    """ Set file output to the desired level and register it with openml logger if needed. """
```
mitar: I think "register it with openml logger" is implementation detail. Maybe just saying that "configured logging to a file (if not already) and sets the desired level".

openml/config.py
```diff
@@ -7,47 +7,94 @@
...
+def _set_level_register_and_store(
+        handler: logging.Handler,
+        log_level: int,
+        store_under_key: Optional[str] = None
+):
+    """ Set handler log level, register it if needed, save setting to config file if specified. """
+    oml_level, py_level = _convert_log_levels(log_level)
+    handler.setLevel(py_level)
+
+    if openml_logger.level > py_level or openml_logger.level == logging.NOTSET:
```
mitar: Why this check? So one cannot use `set_console_log_level` to reconfigure log levels at runtime once they are set?

openml/config.py
PGijsbers: Good catch

openml/config.py
PGijsbers: Me too (create a TODO which links to an issue). This was just a temporary thing as the decision still had to be made whether or not this functionality was to be implemented in this PR.

openml/config.py
PGijsbers: `openml_logger` needs to log at the least common log level for console logging and file logging. Blindly setting the log level to the requested level might mean some messages might incorrectly not be passed anymore.

PGijsbers:
Based on the feedback I would consider finishing the PR with no longer supporting the log level properties of the openml configuration file. This means that each time the user wants to configure openml logging, it either does so through conventional Python usage or by calling one of the helper functions in config.

openml/config.py
mitar: OK, given that this will be called only if explicitly requested, I am OK with this. And when it is explicitly requested, you are right that such change makes sense. (I am still confused though why one has to change logger level and not just handler level, but that is discussion for somewhere else.)

mfeurer:
> Based on the feedback I would consider finishing the PR with no longer supporting the log level properties of the openml configuration file.

Fine by me, maybe we should drop this from the OpenML cache format. @janvanrijn @giuseppec are you using the fact that there's a verbosity flag in the OpenML config file in the Java and R toolkit.

openml/config.py
PGijsbers: Yep. That's just design of Python's logging framework. It's unintuitive to me too (I would expect default behavior to pass even `logging.DEBUG`).

PGijsbers:
@mfeurer do you know what is going on with these errors? It seems these errors stem from the `develop` branch.

PGijsbers:
I'm thinking it is due to an expired ssl certificate (https://test.openml.org/ also gives off a warning). I'll ask Joaquin to update the certificate.

PGijsbers:
There seems to be an issue with the scikit-learn downloaded in the Appveyor CI (though it reports it tried to install scikit-learn 0.20).

```
 @unittest.skipIf(LooseVersion(sklearn.__version__) < "0.20",
                     reason="columntransformer introduction in 0.20.0")
    def test_run_and_upload_column_transformer_pipeline(self):
        import sklearn.compose
>       import sklearn.impute
tests\test_runs\test_run_functions.py:535: 
c:\miniconda35-x64\lib\site-packages\sklearn\impute\__init__.py:4: in <module>
    from ._knn import KNNImputer
    import numpy as np
    
    from ._base import _BaseImputer
    from ..utils.validation import FLOAT_DTYPES
    from ..metrics import pairwise_distances
>   from ..metrics.pairwise import _NAN_METRICS
E   ImportError: cannot import name '_NAN_METRICS'
c:\miniconda35-x64\lib\site-packages\sklearn\impute\_knn.py:10: ImportError
```

mfeurer:
> There seems to be an issue with the scikit-learn downloaded in the Appveyor CI (though it reports it tried to install scikit-learn 0.20).

Hm, would increasing the version help here?

mitar:
Use [Let's Encrypt](https://letsencrypt.org/) and stop worrying about SSL. :-)

PGijsbers:
For Appveyor, scikit-learn now installed through pip instead of conda, and that seems to avoid the issues. Also bumped the scikit-learn version installed with Appveyor to 0.21 (no particular reason as that is still left over from trying to see if changing the version avoided the issue). If you want me to revert it to 0.20, or bump it to 0.22 instead I can do that too.

PR #892, opened 2019-12-03
store uint8 data in uint8 not float

### LINKED ISSUE #891: pickle for SVHN bigger than arff file
For dataset 41081, I get:
```
1.1G	dataset.arff
2.3G	dataset.pkl.py3
```
Maybe we should be using joblib instead? My hard drive is getting full because I'm trying to run CC-18 stuff, which is a bit annoying.

amueller:
Fixes #891.
Most of the large datasets in cc-18 are actually "image" dataset with entries from 0-255. Storing them as uint8 will drastically remove the storage space required as well as storing and loading times.

This is special-casing but I think it's a common and important special case. The CIFAR-10 dataset pickle went from 1.4G to 176MB with this patch.

### DIFF
--- a/openml/datasets/dataset.py
+++ b/openml/datasets/dataset.py
@@ -388,6 +388,15 @@ class OpenMLDataset(OpenMLBase):
                                                     'boolean'):
                     col.append(self._unpack_categories(
                         X[column_name], categories_names[column_name]))
+                elif attribute_dtype[column_name] in ('floating',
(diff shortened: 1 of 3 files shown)


amueller:
Is there a small dataset to test this or should I mock something? Or any other good ideas for tests?

mfeurer:
Superseded by #983.

PR #894, opened 2020-01-15
Feather investigation

sahithyaravi1493:
Change the caching of datasets from pickle to feather (pyarrow)

#### Any other comments?
-  Feather option as default for caching
- Feather does not support sparse datasets, hence we still use pickle in this case.
- An option has been added to get_dataset, to use pickle for caching data 
- attribute names and categorical variables are pickled when using feather (feather cannot save non tabular data)


mfeurer:
Hey, could you please explain why you want to change the library used for caching? And why you want to use feather and not for example joblib?

sahithyaravi1493:
> Hey, could you please explain why you want to change the library used for caching? And why you want to use feather and not for example joblib?

The main motivation for changing from pickle is that the loading is really slow for large datasets (large no. of instances). Arrow is much faster for large datasets when compared to pickle.  Please find the  comparison results here and let me know what you think :).

https://docs.google.com/document/d/163I1mP_N0Rs8qZgcEH8l18N2F801K_lyOwfS6mv9Cdg/edit

mfeurer:
Thanks for the comparison. Could you please give me your definition of "really slow" and what kind of workflow is hurt by that? I do see that pickle takes up to twice as much time, but the highest number I saw was 5 seconds and so far I didn't have problems where loading the data once was an issue (where loading it multiple times was prohibitive, but even the feather speed wouldn't have been enough).

Also, is there a reason to use a new library instead of panda's built-in feather support (https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_feather.html)?

sahithyaravi1493:
The 2x speed up of feather would make a difference when we are loading datasets with large number of instances for visualizations on the new.openml.org.  This was one of the reasons I looked into this.

Yes, I can change it to use pd.read_feather instead, as it uses the same feather. read_feather internally. I think pandas may throw an installation error if pyarrow is not installed, I will check on that.

openml/datasets/dataset.py
```diff
@@ -36,6 +36,8 @@ class OpenMLDataset(OpenMLBase):
         Description of the dataset.
     format : str
         Format of the dataset which can be either 'arff' or 'sparse_arff'.
+    cache_format : str, optional
```
mfeurer: It's not optional, but has a default. Could you please change that?

openml/datasets/dataset.py
```diff
@@ -99,7 +101,8 @@ class OpenMLDataset(OpenMLBase):
         Serialized arff dataset string.
     """
     def __init__(self, name, description, format=None,
```
mfeurer: I'm somewhat surprised that this one isn't annotated. @Neeratyoy could you please add to your stack to figure out why this is legal given that we have mypy running?

openml/datasets/dataset.py
```diff
@@ -127,6 +130,10 @@ def __init__(self, name, description, format=None,
         self.name = name
         self.version = int(version) if version is not None else None
         self.description = description
+        if cache_format not in ['feather', 'pickle']:
+            raise ValueError("cache_format must be one of 'feather' or 'pickle'")
```
mfeurer: Could you please also print the incorrect cache format? That sometimes makes it easier to debug if something fails.

openml/datasets/dataset.py
```diff
@@ -396,18 +405,20 @@ def _parse_data_from_arff(
 
         return X, categorical, attribute_names
 
-    def _create_pickle_in_cache(self, data_file: str) -> str:
+    def _create_pickle_in_cache(self, data_file: str) -> Tuple[str, str, str]:
         """ Parse the arff and pickle the result. Update any old pickle objects. """
         data_pickle_file = data_file.replace('.arff', '.pkl.py3')
-        if os.path.exists(data_pickle_file):
+        data_feather_file = data_file.replace('.arff', '.feather')
+        feather_attribute_file = data_file.replace('.arff', '_attribute.pkl.py3')
```
mfeurer: hm, how about using `.feather.attributes.pkl.py3` as the suffic? This would make it clear that this belongs to the feather file.

openml/datasets/dataset.py
```diff
@@ -416,32 +427,61 @@ def _create_pickle_in_cache(self, data_file: str) -> str:
...
-            pickle.dump((X, categorical, attribute_names), fh, pickle.HIGHEST_PROTOCOL)
-        logger.debug("Saved dataset {did}: {name} to file {path}"
-                     .format(did=int(self.dataset_id or -1),
-                             name=self.name,
-                             path=data_pickle_file)
-                     )
-
-        return data_pickle_file
+        # Feather format does not work for sparse datasets, so we use pickle for sparse datasets
+        if self.cache_format == "feather" and type(X) != scipy.sparse.csr_matrix:
```
mfeurer: Please use `scipy.sparse.issparse`.

openml/datasets/dataset.py
```diff
@@ -416,32 +427,61 @@ def _create_pickle_in_cache(self, data_file: str) -> str:
...
+            logger.info("feather write")
```
mfeurer: Please add which dataset you write and where you write it to.

openml/datasets/dataset.py
```diff
@@ -416,32 +427,61 @@ def _create_pickle_in_cache(self, data_file: str) -> str:
...
+            X.to_feather(data_feather_file)
+            with open(feather_attribute_file, "wb") as fh:
+                pickle.dump((categorical, attribute_names), fh, pickle.HIGHEST_PROTOCOL)
+        else:
+            logger.info("pickle write")
```
mfeurer: Same here.

openml/datasets/functions.py
```diff
@@ -451,7 +451,8 @@ def get_dataset(
     dataset_id: Union[int, str],
     download_data: bool = True,
     version: int = None,
-    error_if_multiple: bool = False
+    error_if_multiple: bool = False,
+    cache_format: str = 'pickle'
```
mfeurer: Could you please add a comma at the end of the line?

openml/datasets/functions.py
```diff
@@ -479,12 +480,18 @@ def get_dataset(
...
+        Format for caching the dataset - may be feather or pickle
+        Note that the default 'pickle' option may load slower than feather when
+        no.of.rows is very high.
     Returns
     -------
     dataset : :class:`openml.OpenMLDataset`
         The downloaded dataset.
     """
+    if cache_format not in ['feather', 'pickle']:
+        raise ValueError("cache_format must be one of 'feather' or 'pickle'")
```
mfeurer: As above.

tests/test_datasets/test_dataset_functions.py
```diff
@@ -1316,3 +1316,18 @@ def test_list_qualities(self):
         qualities = openml.datasets.list_qualities()
         self.assertEqual(isinstance(qualities, list), True)
         self.assertEqual(all([isinstance(q, str) for q in qualities]), True)
+
+    def test_get_dataset_cache_format_pickle(self):
+        # Feather format cant be tested without installing pyarrow
```
mfeurer: Could you please add unit tests for feather?

sahithyaravi1493: We are not adding pyarrow in our installation. So, I am not sure how to test feather

> Could you please add unit tests for feather?

We are not adding pyarrow in our installation. So, I am not sure how to test feather. Do you have some suggestion for this?

openml/datasets/dataset.py
Neeratyoy: Sure.
Just to confirm the task: I have to check why the missing annotation for _cache_format_ was never caught.

mfeurer: Yes, that's correct.

openml/datasets/dataset.py
```diff
@@ -109,6 +112,7 @@ def __init__(self, name, description, format=None,
                  paper_url=None, update_comment=None,
                  md5_checksum=None, data_file=None, features=None,
                  qualities=None, dataset=None):
+        print(cache_format)
```
mfeurer: Could you please remove this print statement?

openml/datasets/functions.py
```diff
@@ -975,6 +983,7 @@ def _create_dataset_from_description(
         features: Dict,
         qualities: List,
         arff_file: str = None,
+        cache_format: str = 'pickle'
```
mfeurer: Could you please add a comma at the end o the line?

tests/test_datasets/test_dataset_functions.py
```diff
@@ -1316,3 +1316,33 @@ def test_list_qualities(self):
...
+        X, y, categorical, attribute_names = dataset.get_data()
+        self.assertIsInstance(X, pd.DataFrame)
+        self.assertEqual(X.shape, (898, 39))
+        self.assertEqual(len(categorical), X.shape[1])
+        self.assertEqual(len(attribute_names), X.shape[1])
+
+    def test_get_dataset_cache_format_feather(self):
+        # Feather format cant be tested without installing pyarrow
+        # this test case checks if pickle option works
+        dataset = openml.datasets.get_dataset('iris', cache_format='feather')
```
mfeurer: Could you please add a test that the correct files are actually written to disk?

mfeurer: I'm afraid this doesn't test the right thing. We change the cache directory for unit tests, so you'd need to use a slightly different data_folder here (https://github.com/openml/openml-python/blob/develop/openml/testing.py#L92)

Issue #836, opened 2019-10-16
SparseDataFrame is deprecated


mfeurer:
see [here](https://pandas.pydata.org/pandas-docs/stable/user_guide/sparse.html#sparse-migration), so we need to migrate.

mitar:
Yes, having regular dataframes with some columns being sparse is much better anyway.

mfeurer:
SparseDataFrame is now dropped via #897.

PR #897, opened 2020-02-11
Removing support for deprecated pandas SparseDataFrame

Neeratyoy:
#### Reference Issue

Addresses #836. 

All support for pandas SparseDataFrame have been removed. OpenML users will have to upgrade to Pandas 1.0.0 or higher to handle sparse dataframes, or as the pandas migration implies, data frames with sparse entries.

For pandas version >= 1.0.0
`sparse_dataset = openml.datasets.get_dataset(4136, download_data=False)`
`rval, *_  = sparse_dataset.get_data()`
`print(rval.info())`

For pandas version < 1.0.0
`sparse_dataset = openml.datasets.get_dataset(4136, download_data=False)`
`rval, *_  = sparse_dataset.get_data()`
_Exception: Current pandas version found 0.25.3. OpenML supports pandas 1.0.0 or higher._


mfeurer:
We'd need to decide whether we drop Python3.5 support for this one. Pandas 1.X.X only supports Python3.6 or higher, but has some major changes. I don't think we can and should invest the time of supporting both old and new versions of pandas, and should therefore drop Python3.5 support. @janvanrijn @amueller @PGijsbers

PGijsbers:
In favor.

doc/progress.rst
```diff
@@ -10,11 +10,14 @@ Changelog
 ~~~~~~
 
 * FIX #873: Fixes an issue which resulted in incorrect URLs when printing OpenML objects after
-  switching the server
+  switching the server.
 * FIX #885: Logger no longer registered by default. Added utility functions to easily register
   logging to console and file.
 * MAINT #767: Source distribution installation is now unit-tested.
+* MAINT #836: OpenML supports only pandas version 1.0.0 or above to match the removal
```
mfeurer: You can drop the `to match the removal of SparseDataFrame support.`

openml/datasets/dataset.py
```diff
@@ -509,7 +510,12 @@ def _encode_if_category(column):
                 )
         elif array_format == "dataframe":
             if scipy.sparse.issparse(data):
-                return pd.SparseDataFrame(data, columns=attribute_names)
+                # SparseDataFrame removed in pandas 1.0.0
+                if LooseVersion(pd.__version__) >= "1.0.0":
```
mfeurer: I don't think this is necessary here. If there are ever problems, we should place such checks in a central place, for example the main `__init__.py`.

setup.py
```diff
@@ -90,7 +90,5 @@
                               'Operating System :: POSIX',
                               'Operating System :: Unix',
                               'Operating System :: MacOS',
-                              'Programming Language :: Python :: 3',
```
mfeurer: Is there a reason why you dropped the Python :: 3 tag?

tests/test_datasets/test_dataset.py
```diff
@@ -284,9 +287,14 @@ def test_get_sparse_dataset(self):
         self.assertEqual(len(attribute_names), 20001)
         self.assertTrue(all([isinstance(att, str) for att in attribute_names]))
 
+    @unittest.skipIf(LooseVersion(pd.__version__) < "1.0.0",
```
mfeurer: This shouldn't happen, right?

tests/test_datasets/test_dataset.py
```diff
@@ -284,9 +287,14 @@ def test_get_sparse_dataset(self):
...
+                     reason="SparseDataFrame support removed from pandas 1.0.0 and onwards.")
     def test_get_sparse_dataframe(self):
         rval, *_ = self.sparse_dataset.get_data()
-        self.assertTrue(isinstance(rval, pd.SparseDataFrame))
+        dtypes = all([dtype == pd.SparseDtype(np.float32, fill_value=0.0)
+                      for dtype in rval.dtypes])
+        self.assertTrue(isinstance(rval, pd.DataFrame))
```
mfeurer: Could you please use `self.assertIsInstance` here?

tests/test_datasets/test_dataset.py
```diff
@@ -284,9 +287,14 @@ def test_get_sparse_dataset(self):
...
+        self.assertTrue(dtypes)
```
mfeurer: Could you please use `np.testing.assertAllEqual` here?

tests/test_datasets/test_dataset_functions.py
```diff
@@ -963,7 +960,7 @@ def test_create_dataset_pandas(self):
         )
         self.assertEqual(
             _get_online_dataset_format(dataset.id),
-            'sparse_arff',
+            'arff',
```
mfeurer: Shouldn't this still be sparse arff? I guess something changed internally based on the fact that the dataframe is now a regular dataframe, but I think this shouldn't be the case in practice.

setup.py
Neeratyoy: I guess I should have asked to be sure. I presumed that it covers as a <= check.
I'll put it back?

tests/test_datasets/test_dataset.py
Neeratyoy: yeah, this was done before we removed support so forgot to undo it
will do

tests/test_datasets/test_dataset.py
Neeratyoy: I couldn't find that function but I am using `np.testing.assert_array_equal` now

tests/test_datasets/test_dataset_functions.py
Neeratyoy: Based on the Pandas update this PR addresses, the dataframe `df` created on _L931_ is a standard Pandas data frame. However, `df.dtypes` will reveal that each column is instead a sparse array. Maybe since the outer abstraction is no longer a SparseDataFrame but a standard DataFrame, it is being stored as arff and not sparse-arff? Since I had to make that change for the test case to get it to work.

Neeratyoy: @mfeurer I changed the check inside `create_dataset` to detect a sparse data frame based on the new pandas check. However, that for the concerned test case here, it falters at `publish()` with an `OpenMLServerException: Error parsing dataset ARFF file - Arff error in dataset file: trailing characters on data line  (l.9)`

setup.py
mfeurer: These are actually not checks, but tags displayed on pypi.org.

openml/datasets/functions.py
```diff
@@ -708,8 +707,9 @@ def create_dataset(name, description, creator, contributor,
             )
 
     if hasattr(data, "columns"):
-        if isinstance(data, pd.SparseDataFrame):
-            data = data.to_coo()
+        # if isinstance(data, pd.SparseDataFrame):
```
mfeurer: Could you please remove this commented line?

Neeratyoy:
@mfeurer how do I re-trigger an appveyor test exclusively?

mfeurer:
There used to be a button. I must admit I can only see one year old builds. @PGijsbers do you see any recent builds?

PGijsbers:
image

This is what I saw. I pressed rebuild commit.

Neeratyoy:
@PGijsbers I think I wasn't able to see that option despite signing in. Is there something else that needs to be done to enable that?
also, thanks!

@mfeurer all test cases pass now.

examples/30_extended/create_upload_tutorial.py
```diff
@@ -286,12 +286,16 @@
 # Dataset is a pandas sparse dataframe
 # ====================================
 
+# sparse_data = coo_matrix((
```
mfeurer: Please do not keep old code.

examples/30_extended/create_upload_tutorial.py
```diff
@@ -286,12 +286,16 @@
 # Dataset is a pandas sparse dataframe
```
mfeurer: could you please update this header to reflect the fact that pandas doesn't have this class any more.

PGijsbers:
From the Appveyor interface it looks like @mfeurer is an administrator and organization access it granted to Appveyor. It does say "AppVeyor is authorized to act on behalf of PGijsbers GitHub account with admin:repo_hook, read:org, repo:status scope. ".

Let's see if we can figure out what's wrong at the hackathon. I think it would be easier when we can immediately see both sides.

PR #903, opened 2020-02-25
Remove __version__ from __all__ in openml\__init__.py

### LINKED ISSUE #871: __version__ is imported with `from openml import *`
Not that you should ever really want to use `import *`, but '\_\_version\_\_' is listed in `openml\__init__.py:__all__`  which means it gets imported:
```
>>> from openml import *
>>> __version__
'0.9.0'
```
This is generally not done, so we might remove '\_\_version\_\_' from our `__all__` list.

Rong-Inspur:
#### Reference Issue
Fix #871 

Remove '__version__' from our __all__ list in openml\__init__.py

### DIFF
--- a/openml/__init__.py
+++ b/openml/__init__.py
@@ -114,7 +114,6 @@ __all__ = [
     'study',
     'utils',
     '_api_calls',
-    '__version__',
 ]
 
 # Load the scikit-learn extension by default
--- a/openml/__init__.py
+++ b/openml/__init__.py
@@ -46,7 +46,7 @@ from . import setups
 from .setups import OpenMLSetup, OpenMLParameter
 
 
-from .__version__ import __version__
+from .__version__ import __version__  # noqa: F401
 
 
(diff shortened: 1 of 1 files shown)


PGijsbers:
The issue might have been too vague. The desired behavior is to be able to do this:
```python
import openml
print(openml.__version__)  # 0.11.0dev
```
but not this:
```python
from openml import *
__version__
```
The latter currently works fine (`__version__` evaluates to the version `str`), but it should instead raise `NameError: name '__version__' is not defined`.

Using `openml.__version__` is the normal way to check a package version. It should remain functional (and is also used in unit tests).

From my understanding this means the only change required would be to remove [this line](https://github.com/openml/openml-python/blob/develop/openml/__init__.py#L117). But you would need to verify that.

Rong-Inspur:
> Looks good to me! Thank you for your contribution 🎉

Thanks a lot for the review and suggestions!

Issue #804, opened 2019-10-08
Rename arguments of list_evaluations to indicate they expect lists of ids


PGijsbers:
See also #796. It was a step forward, but the argument names still don't convey that we are looking for a list (i.e. argument name should be plural) or ids as opposed to OpenML objects (e.g. by a suffix `_id`).

I would argue that we also should update the type hints (to `Optional[List[Union[str, int]]]`) and add parameter descriptions in the docstring.

mfeurer:
Added via #933

PR #907, opened 2020-04-09
Rename arguments of list_evaluations to indicate they expect lists of ids

Bilgecelik:
#804

### DIFF
--- a/openml/evaluations/functions.py
+++ b/openml/evaluations/functions.py
@@ -17,11 +17,11 @@ def list_evaluations(
     function: str,
     offset: Optional[int] = None,
     size: Optional[int] = None,
-    task: Optional[List] = None,
-    setup: Optional[List] = None,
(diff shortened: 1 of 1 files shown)


openml/evaluations/functions.py
```diff
@@ -337,23 +331,23 @@ def list_evaluations_setups(
         # array_split -length % N sub-arrays of size length//N + 1 and the rest of size length//N.
         setup_chunks = np.array_split(ary=evals['setup_id'].unique(),
                                       indices_or_sections=((length - 1) // N) + 1)
-        setups = pd.DataFrame()
+        setupp = pd.DataFrame()
```
PGijsbers, line 334: Maybe use `setup_data` instead.

Bilgecelik:
Updated after new version - [https://github.com/openml/openml-python/pull/933](url)

Issue #910, opened 2020-05-18
Documentation typo


ArlindKadra:
The code is related to removing tasks from a suite or study, however, in the argument documentation it is written as link instead of de-link.

https://github.com/openml/openml-python/blob/249abc901c34bc87c4b283e42e754308a5a2d629/openml/study/functions.py#L380-L389

https://github.com/openml/openml-python/blob/249abc901c34bc87c4b283e42e754308a5a2d629/openml/study/functions.py#L407

mfeurer:
Fixed via #914.

Issue #909, opened 2020-04-25
Possible typo


ArlindKadra:
https://github.com/openml/openml-python/blob/249abc901c34bc87c4b283e42e754308a5a2d629/examples/30_extended/create_upload_tutorial.py#L221

No column `rnd_str` in the example dataset.

PGijsbers:
It was originally added in f22c39360f82bd118d6b17ebff2e34ee7c174ef4. Not only is there no column named `rnd_str` in the dataset, there is no string data (other than the converted categorical columns). I think it should be corrected to 
> ... dtype while the columns 'temperature' and 'humidity' are kept as numeric columns. ...

mfeurer:
Fixed via #914.

PR #914, opened 2020-06-17
Fixing documentation typo

Neeratyoy:
#### Reference Issue
Addresses #909 and #910.

### DIFF
--- a/openml/study/functions.py
+++ b/openml/study/functions.py
@@ -386,7 +386,7 @@ def detach_from_suite(suite_id: int, task_ids: List[int]) -> int:
         OpenML id of the study
 
     task_ids : list (int)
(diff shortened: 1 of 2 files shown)


openml/study/functions.py
```diff
@@ -386,7 +386,7 @@ def detach_from_suite(suite_id: int, task_ids: List[int]) -> int:
         OpenML id of the study
 
     task_ids : list (int)
-        List of entities to link to the collection
+        List of entities to  de-link (detach) from the collection
```
PGijsbers, line 389: I think just using "detach" is sufficient. If you want to keep both, I believe 'delink' is a single word, but 'unlink' is probably preferred. (also goes for L#407 below )

Issue #898, opened 2020-02-13
get_dataset pickle protocol problem


ieCecchetti:
#### Steps/Code to Reproduce
```python
# load as dataframe
curr_ds = openml.datasets.get_dataset(2)
X, y, categorical_indicator, attribute_names = curr_ds.get_data(
    dataset_format='dataframe',
    target=curr_ds.default_target_attribute
)
```

#### Expected Results
 X and y matrix of anneal DS

#### Actual Results
Traceback (most recent call last):
  File "<input>", line 1, in <module>
  File "C:\Users\name\AppData\Local\JetBrains\PyCharm 2019.3.2\plugins\python\helpers\pydev\_pydev_bundle\pydev_umd.py", line 197, in runfile
    pydev_imports.execfile(filename, global_vars, local_vars)  # execute the script
  File "C:\Users\name\AppData\Local\JetBrains\PyCharm 2019.3.2\plugins\python\helpers\pydev\_pydev_imps\_pydev_execfile.py", line 18, in execfile
    exec(compile(contents+"\n", file, 'exec'), glob, loc)
  File "C:/Users/name/PycharmProjects/metaLearning/openMLFeatureExtractor.py", line 31, in <module>
    curr_ds = openml.datasets.get_dataset(int(curr_ds_starts[0]))
  File "C:\Users\name\AppData\Local\Continuum\anaconda3\envs\metaenv\lib\site-packages\openml\datasets\functions.py", line 530, in get_dataset
    description, features, qualities, arff_file
  File "C:\Users\name\AppData\Local\Continuum\anaconda3\envs\metaenv\lib\site-packages\openml\datasets\functions.py", line 1023, in _create_dataset_from_description
    qualities=qualities,
  File "C:\Users\name\AppData\Local\Continuum\anaconda3\envs\metaenv\lib\site-packages\openml\datasets\dataset.py", line 183, in __init__
    self.data_pickle_file = self._create_pickle_in_cache(data_file)
  File "C:\Users\name\AppData\Local\Continuum\anaconda3\envs\metaenv\lib\site-packages\openml\datasets\dataset.py", line 406, in _create_pickle_in_cache
    data, categorical, attribute_names = pickle.load(fh)
ValueError: unsupported pickle protocol: 5. 

#### Versions
Windows-10-10.0.16299-SP0
Python 3.7.6 (default, Jan  8 2020, 20:23:39) [MSC v.1916 64 bit (AMD64)]
NumPy 1.18.1
[...]

PR #925, opened 2020-07-06
Improve error handling and error message when loading datasets

### LINKED ISSUE #918: Pickle error: No module named 'pandas.core.categorical'
Hi

I think this (below) is a problem with different pickle or different pandas versions. Does the example below work with other versions (possibly without downgrading pandas?). 

Here are my versions:

openml==0.10.2
pandas==1.0.5
Python 3.7.4

And here's the dataset loading example.

```
import openml
dataset = openml.datasets.get_dataset(40536)

---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
<ipython-input-18-ddb30cada139> in <module>
      1 import openml
----> 2 dataset = openml.datasets.get_dataset(40536)
      3 X, y, categorical_indicator, _ = dataset.get_data(
      4   dataset_format='dataframe',
      5   target=dataset.default_target_attribute

~/anaconda3/lib/python3.7/site-packages/openml/datasets/functions.py in get_dataset(dataset_id, download_data, version, error_if_multiple)
    528 
    529     dataset = _create_dataset_from_description(
--> 530         description, features, qualities, arff_file
    531     )
    532     return dataset

~/anaconda3/lib/python3.7/site-packages/openml/datasets/functions.py in _create_dataset_from_description(description, features, qualities, arff_file)
   1021         data_file=arff_file,
   1022         features=features,
-> 1023         qualities=qualities,
   1024     )
   1025 

~/anaconda3/lib/python3.7/site-packages/openml/datasets/dataset.py in __init__(self, name, description, format, data_format, dataset_id, version, creator, contributor, collection_date, upload_date, language, licence, url, default_target_attribute, row_id_attribute, ignore_attribute, version_label, citation, tag, visibility, original_data_url, paper_url, update_comment, md5_checksum, data_file, features, qualities, dataset)
    181 
    182         if data_file is not None:
--> 183             self.data_pickle_file = self._create_pickle_in_cache(data_file)
    184         else:
    185             self.data_pickle_file = None

~/anaconda3/lib/python3.7/site-packages/openml/datasets/dataset.py in _create_pickle_in_cache(self, data_file)
    404             with open(data_pickle_file, "rb") as fh:
    405                 try:
--> 406                     data, categorical, attribute_names = pickle.load(fh)
    407                 except EOFError:
    408                     # The file is likely corrupt, see #780.

ModuleNotFoundError: No module named 'pandas.core.categorical'

```

mfeurer:
Closes #918 

Issue #918 was caused by having a pickle file created with pandas < 1.0 and then being read with pandas >= 1.0. pandas changed the internals of handling categorical values and therefore the dataframe cannot be unpickled any more. This PR changes the behavior of OpenML-Python, and instead of crashing when failing on unpickling the dataset, it reads the data from arff and emits a warning that the pickle is not readable.

### DIFF
--- a/openml/datasets/dataset.py
+++ b/openml/datasets/dataset.py
@@ -456,10 +456,17 @@ class OpenMLDataset(OpenMLBase):
                     # The file is likely corrupt, see #780.
                     # We deal with this when loading the data in `_load_data`.
                     return data_pickle_file, data_feather_file, feather_attribute_file
-                except Exception:
+                except ModuleNotFoundError:
                     # There was some issue loading the file, see #918
                     # We deal with this when loading the data in `_load_data`.
                     return data_pickle_file, data_feather_file, feather_attribute_file
+                except ValueError as e:
+                    if "unsupported pickle protocol" in e.args[0]:
+                        # There was some issue loading the file, see #898
+                        # We deal with this when loading the data in `_load_data`.
+                        return data_pickle_file, data_feather_file, feather_attribute_file
+                    else:
+                        raise
 
             # Between v0.8 and v0.9 the format of pickled data changed from
             # np.ndarray to pd.DataFrame. This breaks some backwards compatibility,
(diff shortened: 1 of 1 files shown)


mfeurer:
Should also fix #898.

openml/datasets/dataset.py
```diff
@@ -456,6 +456,10 @@ def _create_pickle_in_cache(self, data_file: str) -> Tuple[str, str, str]:
                     # The file is likely corrupt, see #780.
                     # We deal with this when loading the data in `_load_data`.
                     return data_pickle_file, data_feather_file, feather_attribute_file
+                except Exception:
```
PGijsbers, line 459: Could you specifically catch `ModuleNotFoundError` and any others that are expected? I would prefer not to use a catch-all as if new issues arise, we could possibly just fix them instead.

PGijsbers: You can add a check for the error of #898 too. And document in comments in one sentence why an error is expected. Including the issue number is good, but since they are easily summarized it is nice to be able to have the documentation within the code, e.g.
```
catch ModuleNotFoundError:
    # 780: Pickled dataframe is likely of pandas<1.0 while attempting to load with pandas>=1.0
    return ...
catch ValueError:
   (maybe check for the specific message)
   # 898: Pickled dataframe pickled with protocol 5 (Py3.8), but loaded with protocol 4 (<Py3.8).
   return ...
```

mfeurer: Good point, I just updated the code.

PR #926, opened 2020-07-06
Increase unit test stability

mfeurer:
Currently, a lot of unit tests are failing because the server cannot process run traces fast enough. This PR increases the time to wait for a trace to reduce the amount of unit tests failing when using continuous integration.

### DIFF
--- a/tests/test_runs/test_run_functions.py
+++ b/tests/test_runs/test_run_functions.py
@@ -84,7 +84,7 @@ class TestRun(TestBase):
             if len(run.evaluations) > 0:
                 return
             else:
-                time.sleep(3)
+                time.sleep(10)
         raise RuntimeError(
             "Could not find any evaluations! Please check whether run {} was "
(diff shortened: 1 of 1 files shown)


PGijsbers:
Looks reasonable to me. Waiting for CI reports.

mfeurer:
> Looks reasonable to me. Waiting for CI reports.

Doesn't work as well as expected. It fails at one instance where `_wait_for_processed_run` was already waiting for 400s. I'll restart that one.

mfeurer:
Okay, with the latest change it didn't fail for a missing trace any more (there only was one issue with an illegal dataset transition which is not within the scope of this PR which I reran). So in my opinion this is good. (in case travis appears to not be done, please click on the travis link to get to travis-ci.org. It's actually done, but this is not reflected on github)

PGijsbers:
Annoying, I've had bad connection between travis/github in my gama repo as well.

PR #931, opened 2020-07-09
adding config file to user guide

### LINKED ISSUE #901: The API docs should document the openml config options
#### Description
The API Docs: http://openml.github.io/openml-python/master/api.html
do not cover the API config options, such as 'server, api_key, verbosity,...'

Could this be added?

#### Steps/Code to Reproduce
Try to look up the OpenML-Python config options on 
http://openml.github.io/openml-python/master/api.html

#### Expected Results
Maybe a section 'Configuration' next to 'Top level Classes', 'Extensions' and 'Modules'?

#### Actual Results
Nothing there.

#### Versions
Irrelevant

marcoslbueno:
#### Reference Issue

Fixes #901 

I added documentation to the user guide (file doc/usage.rst) about the configuration file of openml-python.

#### Any other comments?
The new documentation is based on https://github.com/openml/OpenML/wiki/Client-API-Standards, which has fields including 'verbosity' and 'arff.reader'. However, these fields are apparently not used in the parser script (i.e. doc/config.py), so maybe they should be removed from the new documentation. The fields 'avoid_duplicate_runs' and 'connection_n_retries' were added to the new documentation as they are used in the parser script.

### DIFF
--- a/doc/usage.rst
+++ b/doc/usage.rst
@@ -38,28 +38,26 @@ directory of the user and is called config. It consists of ``key = value`` pairs
 which are separated by newlines. The following keys are defined:
 
 * apikey:
-    * required to access the server
+    * required to access the server. The `OpenML setup <https://openml.github.io/openml-python/master/examples/20_basic/introduction_tutorial.html#authentication>`_ describes how to obtain an API key.
 
 * server:
-    * default: ``http://www.openml.org``
+    * default: ``http://www.openml.org``. Alternatively, use ``test.openml.org`` for the test server.
 
 * cachedir:
-    * if not given, will default to ``file.path(tempdir(), "cache")``.
+    * if not given, will default to ``~/.openml/cache``
 
 * avoid_duplicate_runs:
+    * if set to ``True``, when ``run_flow_on_task`` or similar methods are called a lookup is performed to see if there already exists such a run on the server. If so, download those results instead.
     * if not given, will default to ``True``.
 
 * connection_n_retries:
+    * number of connection retries.
     * default: 2. Maximum number of retries: 20.
 
 * verbosity:
     * 0: normal output
     * 1: info output
     * 2: debug output
-
-* arff.reader:
-    * ``RWeka``: This is the standard Java parser used in Weka.
-    * ``farff``: The `farff package <http://www.github.com/mlr-org/farff>`_ lives below the mlr-org and is a newer, faster parser without Java.
 
 ~~~~~~~~~~~~
 Key concepts
--- a/doc/usage.rst
+++ b/doc/usage.rst
@@ -29,6 +29,37 @@ the OpenML Python connector, followed up by a simple example.
 
 * `Introduction <examples/introduction_tutorial.html>`_
 
+~~~~~~~~~~~~~
+Configuration
+~~~~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


doc/usage.rst
```diff
@@ -29,6 +29,37 @@ the OpenML Python connector, followed up by a simple example.
...
+
+The configuration file resides in a directory ``.openml`` in the home
+directory of the user and is called config. It consists of ``key = value`` pairs
+which are separated by newlines. The following keys are defined:
+
+* apikey:
+    * required to access the server
+
+* server:
+    * default: ``http://www.openml.org``
```
PGijsbers, line 44: List ``test.openml.org`` as an option.

doc/usage.rst
```diff
@@ -29,6 +29,37 @@ the OpenML Python connector, followed up by a simple example.
...
+~~~~~~~~~~~~~
+Configuration
+~~~~~~~~~~~~~
+
+The configuration file resides in a directory ``.openml`` in the home
+directory of the user and is called config. It consists of ``key = value`` pairs
+which are separated by newlines. The following keys are defined:
+
+* apikey:
+    * required to access the server
```
PGijsbers, line 41: I know it should be documented elsewhere too, maybe reference that part of the documentation on where to get the API key.

doc/usage.rst
```diff
@@ -29,6 +29,37 @@ the OpenML Python connector, followed up by a simple example.
...
+
+* server:
+    * default: ``http://www.openml.org``
+
+* cachedir:
+    * if not given, will default to ``file.path(tempdir(), "cache")``.
```
PGijsbers, line 47: Check what this normally resolves to by default and list that instead. I think it should be `~/.openml/cache`.

doc/usage.rst
```diff
@@ -29,6 +29,37 @@ the OpenML Python connector, followed up by a simple example.
...
+
+* avoid_duplicate_runs:
+    * if not given, will default to ``True``.
```
PGijsbers, line 50: Explain what this does, i.e. when `run_flow_on_task` or similar methods are called, when set to `True` a lookup is performed to see if there already exists such a run on the server, and if so download those results instead.

doc/usage.rst
```diff
@@ -29,6 +29,37 @@ the OpenML Python connector, followed up by a simple example.
...
+
+* connection_n_retries:
+    * default: 2. Maximum number of retries: 20.
```
PGijsbers, line 53: Again, also explain what it does (even if the name is fairly self explanatory).

doc/usage.rst
```diff
@@ -29,6 +29,37 @@ the OpenML Python connector, followed up by a simple example.
...
+
+* verbosity:
+    * 0: normal output
+    * 1: info output
+    * 2: debug output
+
+* arff.reader:
```
PGijsbers, line 60: `arff.reader` is not supported for the python package, we always use liac-arff.

marcoslbueno: I removed this bit because I don't know the possible values for `arff.reader` (maybe they are the same as those of  `arff.reader`? Please let me know if so).

PGijsbers:
Thanks! 🎉

Issue #753, opened 2019-07-29
Custom OpenMLFlow


PGijsbers:
I want to register a flow referring to the setup we used to conduct the [automl benchmark](https://github.com/openml/automlbenchmark) presented at the ICML workshop this year. We want to do this to more easily share results through OpenML. I want to run this by you all.

Now, I know that creating your own flow in `openml-python` is discouraged ('Flows should not be generated manually'), but for this scenario I don't really see a better way? Other than using e.g. `openml-r`.

I want to describe that we used the code at our Github repo, and a specific tag, and provide a url to the download for the code. I figured to tag the used AutoML tool (in this case `auto-sklearn`).
There is no direct way to instantiate the script with the right settings currently, but flow below should help/be enough to configure a reproducible run. Again, the flow is created mainly to link and share results, and provide the best possible pointers for reproducing the flow.

Sketching out how I think the flow should look like:
```python
auto_sklearn_flow = openml.flows.get_flow(15275)  # auto-sklearn 0.5.1
amlb_flow = openml.flows.OpenMLFlow(
	name='automlbenchmark_autosklearn',
	description='Auto-sklearn as set up by the AutoML Benchmark',
	external_version='amlb==0.9',
	parameters=OrderedDict(
		time='240',
		memory='32',
		cores='8'
	),
	parameters_meta_info=OrderedDict(
		time=dict(data_type='int', description='time in minutes'),
		memory=dict(data_type='int', description='memory in gigabytes'),
		cores=dict(data_type='int', description='number of available cores')
	),
	language='English',
	components=OrderedDict(automl_tool=auto_sklearn_flow),	
)
```

I would still need to find a place for the following information:
```
source: https://github.com/openml/automlbenchmark/releases/tag/v0.9
subfolder of particular interest: /frameworks/autosklearn
```
I do see `url` fields in `OpenMLFlow.__init__` but they are set by the server.

I think there are a couple of questions here:
 - Do you think I should even try doing this through the Python API?

And if so,
 - Should I create only one flow for the entire benchmark, and take the 'automl tool'  as a parameter? I think it's more generic, but might make the link to the automl tool's flow harder to find.
 - Where to include the code-url.
 - Any other things I need to know?

mfeurer:
> Do you think I should even try doing this through the Python API?

In my opinion yes, it should be flexible enough to allow this to an experienced user. We should update the docs based on this experience.

[...]

PR #932, opened 2020-07-10
Feature #753

PGijsbers:
Looking for feedback. For the example I still need to finalize how the predictions are generated/formatted to make it clear and not distract from the overall example.


examples/30_extended/custom_flow_tutorial.py
```diff
@@ -0,0 +1,183 @@
...
+#
+# Note: flow 15275 is not actually the right flow on the test server,
+# but that does not matter for this demonstration.
+
+autosklearn_flow = openml.flows.get_flow(15275)  # auto-sklearn 0.5.1
+subflow = dict(components=OrderedDict(automl_tool=autosklearn_flow),)
+
+####################################################################################################
+# With all parameters of the flow defined, we can now initialize the OpenMLFlow and publish.
+# Explicitly set the model of the flow to `None`, because we provided all the details already!
```
mfeurer, line 86: I would state this differently. The model is set to `None`, because there is no model to access and parse here, right?

examples/30_extended/custom_flow_tutorial.py
```diff
@@ -0,0 +1,183 @@
...
+    for index in task.get_train_test_split_indices(fold, repeat)[1]
+]
+
+# random class probabilities (Iris has 150 samples and 3 classes):
+r = np.random.rand(150 * n_repeats, 3)
+# scale the random values so that the probabilities of each sample sum to 1:
+y_proba = r / r.sum(axis=1).reshape(-1, 1)
+y_pred = y_proba.argmax(axis=1)
+class_map = dict(zip(range(3), task.class_labels))
+y_true = ["Iris-setosa"] * 50 + ["Iris-versicolor"] * 50 + ["Iris-virginica"] * 50
```
mfeurer, line 146: I think it would be better for the example if you retrieved these from the dataset.

examples/30_extended/custom_flow_tutorial.py
```diff
@@ -0,0 +1,183 @@
...
+        index=index,
+        prediction=class_map[yp],
+        truth=y,
+        proba={c: pb for (c, pb) in zip(task.class_labels, proba)},
+    )
+    ps.append(p)
+
+####################################################################################################
+# Finally we can create the OpenMLRun object and upload.
+# We use the "setup string" because the used flow was a script.
```
mfeurer, line 169: Can you rephrase this, it took me a while to understand what you mean here. Maybe write "We use the argument ``setup_string`` because the used flow was a script."

examples/30_extended/custom_flow_tutorial.py
```diff
@@ -0,0 +1,183 @@
...
+        task=task,
+        repeat=repeat,
+        fold=fold,
+        sample=sample,
+        index=index,
+        prediction=class_map[yp],
+        truth=y,
+        proba={c: pb for (c, pb) in zip(task.class_labels, proba)},
+    )
+    ps.append(p)
```
mfeurer, line 165: This is not used, right? Was this the variable that you mentioned we shouldn't have a look at yesterday?

PGijsbers: I don't recall what I mentioned yesterday, but yes this was left over from development/checking the function runs fine. But yes I will remove it :)

PGijsbers:
Thanks for the feedback :) will process after the meeting.

examples/30_extended/custom_flow_tutorial.py
```diff
@@ -0,0 +1,183 @@
+"""
+================================
+Creating and Using a Custom Flow
+================================
+
+The most convenient way to create a flow for your machine learning workflow is to generate it
+automatically as described in <>. However, there are scenarios where this is not possible, such
```
Bilgecelik, line 7: Maybe one sentence before this to explain why they need flows at the first place?

examples/30_extended/custom_flow_tutorial.py
```diff
@@ -0,0 +1,183 @@
...
+####################################################################################################
+# 1. Defining the flow
+# ====================
+# The first step is to define all the hyperparameters of your flow.
+# Check ... for the descriptions of each variable.
+# Note that `external version` and `name` together should uniquely identify a flow.
+#
+# The AutoML Benchmark runs AutoML systems across a range of tasks.
+# We can not use the flows of the AutoML systems directly, as the benchmark adds performs
+# preprocessing as required.
```
Bilgecelik, line 39: The second part is not clear to me

examples/30_extended/custom_flow_tutorial.py
```diff
@@ -0,0 +1,183 @@
...
+        cores=OrderedDict(description="number of available cores", data_type="int"),
+        memory=OrderedDict(description="memory in gigabytes", data_type="int"),
+        time=OrderedDict(description="time in minutes", data_type="int"),
+    ),
+)
+
+####################################################################################################
+# It is possible for flows to contain subflows. In this example, the auto-sklearn flow is a
+# subflow, this means that the subflow is entirely executed as part of this flow.
+# Using this modularity also allows your runs to specify which hyperparameters of the
```
Bilgecelik, line 75: Is it also possible to define a custom flow without defining it as a subflow? So we need subflows when we want to do multiple runs with different hyperparameter settings for each subflow?

examples/30_extended/custom_flow_tutorial.py
```diff
@@ -0,0 +1,183 @@
...
+    OrderedDict([("oml:name", "time"), ("oml:value", 120), ("oml:component", flow_id)]),
+]
+
+task_id = 1408  # Iris Task
+task = openml.tasks.get_task(task_id)
+dataset_id = task.get_dataset().dataset_id
+
+
+####################################################################################################
+# The last bit of information for the run we need are the predicted values.
```
Bilgecelik, line 116: I think this part may use some more explanation/categorization but not sure how=) Maybe explain what is different for custom flow here, and why.

examples/30_extended/custom_flow_tutorial.py
```diff
@@ -0,0 +1,183 @@
...
+    ),
+    external_version="amlb==0.9",
+    language="English",
+    tags=["amlb", "benchmark", "study_218"],
+    dependencies="amlb==0.9",
+)
+
+####################################################################################################
+# Next we define the flow hyperparameters. We define their name and default value in `parameters`,
+# and provide meta-data for each parameter through `parameters_meta_info`.
```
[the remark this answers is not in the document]
PGijsbers: Yes, though the function arguments make it a little confusing.

examples/30_extended/custom_flow_tutorial.py
PGijsbers: In the benchmark case, there is not. However in the general case there could be a python object model, but with no extension for automatically populating e.g. `parameters` fields.

examples/30_extended/custom_flow_tutorial.py
PGijsbers: Changes as 

```
# OpenML stores Flows for each AutoML system. However, the AutoML benchmark adds
# preprocessing to the flow, so should be described in a new flow.
```

is that clear? Or could you try to identify the part that raises confusion?

examples/30_extended/custom_flow_tutorial.py
PGijsbers: It is possible to have custom flows with no subflows. Custom flows (such as this one) can be defined with their own hyperparameters (such as time or memory for this one).

examples/30_extended/custom_flow_tutorial.py
PGijsbers: There is nothing inherently different for custom flows here. But when you run flows through extensions, (i.e. use `run_flow_on_task`) all of this formatting is done for you automatically.

examples/30_extended/custom_flow_tutorial.py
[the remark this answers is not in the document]
mfeurer: hm, yes :( Maybe you should briefly mention this in the example, too?

examples/30_extended/custom_flow_tutorial.py
mfeurer: Yes, I see. But it sounds like you don't need to provide a model if all information is already parsed. I think it would be better if you could briefly mention the possible cases and which one you chose in this example.

examples/30_extended/custom_flow_tutorial.py
```diff
@@ -142,31 +150,31 @@
...
         truth=y,
         proba={c: pb for (c, pb) in zip(task.class_labels, proba)},
     )
-    ps.append(p)
+    predictions.append(prediction)
 
 ####################################################################################################
 # Finally we can create the OpenMLRun object and upload.
-# We use the "setup string" because the used flow was a script.
+# We use the argument setup_string because the used flow was a script."
```
mfeurer, line 177: I think you accidentally added a trailing quotation mark here?

examples/30_extended/custom_flow_tutorial.py
Bilgecelik: No it is clear now.

examples/30_extended/custom_flow_tutorial.py
Bilgecelik: I understand, i am still a bit unsure of flow / custom flow / subflow differences. Maybe also additional value of using subflows instead of multiple custom flows, but that can be just me=)

examples/30_extended/custom_flow_tutorial.py
Bilgecelik: okay then it is fine

examples/30_extended/custom_flow_tutorial.py
Bilgecelik: I think it is quite clear now.

PGijsbers:
@mfeurer I don't see any reduced coverage, (where) am I reading the codecov report wrong?

mfeurer:
I think it complains that not enough of the diff is tested. 76% of your diff is tested, while codecov expects 87.65% of the diff being tested. I guess the 87.65% is our current test coverage.

PGijsbers:
Ah, I was looking at the wrong tab 😓  looks like I'll have to add some tests that check the error cases / learning curve tasks. Thanks!

examples/30_extended/custom_flow_tutorial.py
```diff
@@ -0,0 +1,202 @@
...
+    external_version="amlb==0.9",
+    language="English",
+    tags=["amlb", "benchmark", "study_218"],
+    dependencies="amlb==0.9",
+)
+
+####################################################################################################
+# Next we define the flow hyperparameters. We define their name and default value in `parameters`,
+# and provide meta-data for each hyperparameter through `parameters_meta_info`.
+# Note that eventhough the argument name is `parameters` they describe the hyperparameters.
```
Neeratyoy, line 61: "eventhough" -> "even though"

examples/30_extended/custom_flow_tutorial.py
```diff
@@ -0,0 +1,202 @@
...
+####################################################################################################
+# It is possible to build a flow which uses other flows.
+# For example, the Random Forest Classifier is a flow, but you could also construct a flow
+# which uses a Random Forest Classifier in a ML pipeline. When constructing the pipeline flow,
+# you can use the Random Forest Classifier flow as a *subflow*. It allows for
+# all hyperparameters of the Random Classifier Flow to also be specified in your pipeline flow.
+#
+# In this example, the auto-sklearn flow is a subflow: the auto-sklearn flow is entirely executed as part of this flow.
+# This allows people to specify auto-sklearn hyperparameters used in this flow.
+# Using a subflow is not required.
```
Neeratyoy, line 82: I think I understood the earlier part about an existing flow included as a *subflow* in the new custom flow. However, I am a little confused with these lines here where it is mentioned that the auto-sklearn flow is executed as a subflow, but then what does "using a subflow is not required" mean here?

examples/30_extended/custom_flow_tutorial.py
```diff
@@ -0,0 +1,202 @@
...
+#
+# Note: flow 15275 is not actually the right flow on the test server,
+# but that does not matter for this demonstration.
+
+autosklearn_flow = openml.flows.get_flow(15275)  # auto-sklearn 0.5.1
+subflow = dict(components=OrderedDict(automl_tool=autosklearn_flow),)
```
Neeratyoy, line 88: The dict key of *automl_tool* is a registered/expected key or the user is free to rename it while creating their flow?
If it is the former, could we then mention or link here how the user could look up other such keys they may require to be specified?

examples/30_extended/custom_flow_tutorial.py
```diff
@@ -0,0 +1,202 @@
...
+# - p_1, ..., p_c: for each class the predicted probability of the sample
+#   belonging to that class. (no elements for regression tasks)
+#   Make sure the order of these elements follows the order of `task.class_labels`.
+# - the predicted class/value for the sample
+# - the true class/value for the sample
+#
+# When using openml-python extensions (such as through `run_model_on_task`),
+# all of this formatting is automatic.
+# Unfortunately we can not automate this procedure for custom flows,
+# which means a little additional effort is required.
```
Neeratyoy, line 145: For me personally, the minor gap in understanding that exists is the behaviour of functions available in the OpenML API. Given that the example so far creates a *custom* flow, pulls a task and a dataset and then jumps to how the predictions should be formatted for a legitimate OpenML run format, leaves me wondering whether I can use any OpenML function to reach till the prediction formatting stage. 
As it appears, one would have to pull the dataset using *task* API such as `get_X_and_y`, and explicitly call fit(), predict() or the equivalent, and then format the output predictions as you mention later in the example.

If I have understood it correctly, then I think that having at least comments talking briefly about the next imminent steps to go from flow, task, data to predictions might be helpful. Though on second thought, what I said feels like quite the obvious next steps when dealing with such a flow, maybe being explicit might help the new OpenML users. 

To summarise, the fact that `run_model_on_task` may not work on a custom flow without a registered extension could be more explicit, if that is indeed what is supposed to happen.

examples/30_extended/custom_flow_tutorial.py
PGijsbers: It is just to say that not every flow has to contain a subflow, but this one does because it's useful. What do you think about

> In general, using a subflow is not required.

instead?

examples/30_extended/custom_flow_tutorial.py
PGijsbers: It's essentially the name I give to the subflow (see https://openml.org/f/15509). You're free to pick the name. For example in the scikit-learn pipeline of [flow 8817](https://www.openml.org/f/8817) you find components like `ColumnTransformer` and `SVC` instead.

examples/30_extended/custom_flow_tutorial.py
PGijsbers: > As it appears, one would have to pull the dataset using task API such as get_X_and_y, and explicitly call fit(), predict() or the equivalent, and then format the output predictions as you mention later in the example.

Correct, as custom flows are used when OpenML can't interface directly with the machine learning flow.

> If I have understood it correctly, then I think that having at least comments talking briefly about the next imminent steps to go from flow, task, data to predictions might be helpful.

I am not sure what you are missing here, all of this is documented step by step in this tutorial? Exactly how to feed data to the model and produce predictions is dependent on the framework you work with. In our case we just generate predictions instead, but in general we don't know the interface which is why it's not available as an extension in the first place.

> To summarise, the fact that run_model_on_task may not work on a custom flow without a registered extension could be more explicit, if that is indeed what is supposed to happen.

In which way do you feel the opening paragraphs do not address this? (I did just notice I forgot to put in the reference link, so perhaps that's why)

examples/30_extended/custom_flow_tutorial.py
Neeratyoy: aah okay, got it
all good then

examples/30_extended/custom_flow_tutorial.py
Neeratyoy: Yeah, sounds much clearer for me personally

examples/30_extended/custom_flow_tutorial.py
Neeratyoy: yes, the link makes it clear :)

openml/runs/functions.py
```diff
@@ -1073,4 +1075,7 @@ def format_prediction(
                 sample = 0
         probabilities = [proba[c] for c in task.class_labels]
         return [repeat, fold, sample, index, *probabilities, truth, prediction]
-    return [repeat, fold, index, truth, prediction]
+    elif isinstance(task, OpenMLRegressionTask):
+        return [repeat, fold, index, truth, prediction]
+    else:
+        raise TypeError(f"Formatting for {type(task)} is not supported.")
```
mfeurer, line 1081: Would a `NotImplementedError` be more logical here?

tests/test_runs/test_run_functions.py
```diff
@@ -1336,3 +1333,44 @@ def test_run_flow_on_task_downloaded_flow(self):
...
+        openml.config.server = self.production_server
+        clustering = openml.tasks.get_task(126033, download_data=False)
+        ignored_input = [0] * 5
+        with self.assertRaises(TypeError):
+            format_prediction(clustering, *ignored_input)
+
+    def test_format_prediction_classification_no_probabilities(self):
+        classification = openml.tasks.get_task(self.TEST_SERVER_TASK_SIMPLE[0], download_data=False)
+        ignored_input = [0] * 5
+        with self.assertRaises(ValueError):
```
mfeurer, line 1348: I'd actually prefer to have `self.assertRaisesRegex` so that one sees what exact error is expected. Could you please change that?

tests/test_runs/test_run_functions.py
```diff
@@ -1336,3 +1333,44 @@ def test_run_flow_on_task_downloaded_flow(self):
...
+            format_prediction(classification, *ignored_input, proba=None)
+
+    def test_format_prediction_classification_incomplete_probabilities(self):
+        classification = openml.tasks.get_task(self.TEST_SERVER_TASK_SIMPLE[0], download_data=False)
+        ignored_input = [0] * 5
+        incomplete_probabilities = {c: 0.2 for c in classification.class_labels[1:]}
+        with self.assertRaises(ValueError):
```
mfeurer, line 1355: Same here.

tests/test_runs/test_run_functions.py
```diff
@@ -1336,3 +1333,44 @@ def test_run_flow_on_task_downloaded_flow(self):
...
+            format_prediction(classification, *ignored_input, proba=incomplete_probabilities)
+
+    def test_format_prediction_task_without_classlabels_set(self):
+        classification = openml.tasks.get_task(self.TEST_SERVER_TASK_SIMPLE[0], download_data=False)
+        classification.class_labels = None
+        ignored_input = [0] * 5
+        with self.assertRaises(ValueError):
```
mfeurer, line 1362: and here

tests/test_runs/test_run_functions.py
```diff
@@ -1336,3 +1333,44 @@ def test_run_flow_on_task_downloaded_flow(self):
...
+            format_prediction(classification, *ignored_input, proba={})
+
+    def test_format_prediction_task_learning_curve_sample_not_set(self):
+        learning_curve = openml.tasks.get_task(801, download_data=False)
+        probabilities = {c: 0.2 for c in learning_curve.class_labels}
+        ignored_input = [0] * 5
+        with self.assertRaises(ValueError):
```
mfeurer, line 1369: and here as well

examples/30_extended/custom_flow_tutorial.py
```diff
@@ -0,0 +1,203 @@
...
+Creating and Using a Custom Flow
+================================
+
+The most convenient way to create a flow for your machine learning workflow is to generate it
+automatically as described in the `Obtain Flow IDs <https://openml.github.io/openml-python/master/examples/30_extended/flow_id_tutorial.html#sphx-glr-examples-30-extended-flow-id-tutorial-py>`_ tutorial.  # noqa E501
+However, there are scenarios where this is not possible, such
+as when the flow uses a framework without an extension or when the flow is described by a script.
+
+In those cases you can still create a custom flow by following the steps of this tutorial.
+As an example we will use the flows generated for the AutoML Benchmark (...),
```
mfeurer, line 12: I believe there is a link missing here (and it shouldn't be `(...)`)

examples/30_extended/custom_flow_tutorial.py
```diff
@@ -0,0 +1,203 @@
...
+from openml import OpenMLClassificationTask
+from openml.runs.functions import format_prediction
+
+openml.config.start_using_configuration_for_example()
+
+####################################################################################################
+# 1. Defining the flow
+# ====================
+# The first step is to define all the hyperparameters of your flow.
+# Check ... for the descriptions of each variable.
```
mfeurer, line 35: Is there a link missing here, too?

examples/30_extended/custom_flow_tutorial.py
```diff
@@ -0,0 +1,203 @@
...
+
+####################################################################################################
+# It is possible to build a flow which uses other flows.
+# For example, the Random Forest Classifier is a flow, but you could also construct a flow
+# which uses a Random Forest Classifier in a ML pipeline. When constructing the pipeline flow,
+# you can use the Random Forest Classifier flow as a *subflow*. It allows for
+# all hyperparameters of the Random Classifier Flow to also be specified in your pipeline flow.
+#
+# In this example, the auto-sklearn flow is a subflow: the auto-sklearn flow is entirely executed as part of this flow.
+# This allows people to specify auto-sklearn hyperparameters used in this flow.
```
mfeurer, line 82: I am wondering if this is necessary for the example. Isn't the Auto-sklearn flow itself sufficient? And wouldn't it be better to create a custom flow from scratch to explain the idea of creating flows as a user?

examples/30_extended/custom_flow_tutorial.py
```diff
@@ -0,0 +1,203 @@
...
+    flow_id=flow_id,
+    dataset_id=dataset_id,
+    parameter_settings=parameters,
+    setup_string=benchmark_command,
+    data_content=predictions,
+    tags=["study_218"],
+    description_text="Run generated by the Custom Flow tutorial.",
+)
+my_run.publish()
+print("run created:", my_run.run_id)
```
mfeurer, line 203: I think you should use the "stop using the test server command" in the end of the example to go back to the default setting.

mfeurer (commented):
Hey Pieter, I'm afraid I have some more questions on this example (and a few minor change requests)

examples/30_extended/custom_flow_tutorial.py
PGijsbers: In the context of the example, yes we could use a flow without subflow. It would make the example easier, though then there is no documentation on how subflows should be referenced. If you think it is better to leave that out, I can simplify the example.

The AutoML benchmark flows are structured this way (with a subflow), because the benchmark actually performs operations on the data: it will encode categorical data and impute missing values if the framework is not able to work directly with missing values or string data.

tests/test_runs/test_run_functions.py
PGijsbers: I've actually been meaning to ask. Why do we use the `unittest` module for some things, and `pytest` for others?
For instance, instead of `self.assertRaisesRegex`  you would use `pytest.raises(ValueError, match=r'regex string')`.

examples/30_extended/custom_flow_tutorial.py
mfeurer: > If you think it is better to leave that out, I can simplify the example.

Ideally we'd have both, not sure how we can easily get this done without having two examples in the end

tests/test_runs/test_run_functions.py
mfeurer: Historical reasons ;)

I think one could actually only use `assert a == b` and not use the `unittest` or `pytest` module any more, see https://docs.pytest.org/en/stable/assert.html

PGijsbers: Yes, but just `assert` does not work for checking exceptions. The page you linked describes the `pytest.raises(ValueError, match=r'regex string')` call I referred too as well.

PGijsbers: I'll use the `unittest` `self.assertRaisesRegex` for consistency.

openml/runs/functions.py
PGijsbers: Sure, updated.

examples/30_extended/custom_flow_tutorial.py
PGijsbers: Would it be acceptable to extend the example with a flow without subflow later? This example at least covers that implicitly anyway. I could add it when I get back from vacation or during the hackathon.

mfeurer: > Would it be acceptable to extend the example with a flow without subflow later?

Yes, I think so.

mfeurer (approved):

PGijsbers:
I did a rebase on develop, locally it looked like it worked fine. I thought the old commits should've simply been replaced by the ones patched onto the develop head? (i.e. same code diff, different commit id)

mfeurer:
> I thought the old commits should've simply been replaced by the ones patched onto the develop head?

Yeah, I don't know how to do this probably either, I'm usually rebasing to avoid such hassle.

> I can try to fix this, but I guess we are squashing anyway so it doesn't matter much?

No, shouldn't matter.

Shall I do a final review and then merge?

PGijsbers:
> Yeah, I don't know how to do this probably either, I'm usually rebasing to avoid such hassle.

I normally merge because it's less of a hassle. After the failed rebase I actually found the general advise not to rebase if the work already lives on remote.

> Shall I do a final review and then merge?

That would be greatly appreciated!

Issue #929, opened 2020-07-07
Data edit api for new website


sahithyaravi1493:
As discussed, this api can edit some meta-features of the dataset:
List of  meta-features can  be edited/ updated via data_edit API call, see PR https://github.com/openml/OpenML/pull/1044

1. description
2. creator
3. collection_date
4. language
5. citation
6. original_data_url
7. paper_url

These meta-features/ data will just need to call create_dataset api and create a new version: **(python api)**
1. attributes
2. data - the data itself
3. default_target_attribute
4. ignore_attribute
5. row_id_attribute

After the data_edit API is ready, I can create a single **python API** to handle both of these cases.
Based on the arguments (None or otherwise). We can either create a new dataset or call the data_edit API.

PGijsbers:
Closed through #935

PR #935, opened 2020-07-22
Edit api

sahithyaravi1493:
#### Reference Issue
Edit api #929 

As discussed, this api can edit some meta-features of the dataset:
We have two cases based on the fields to edit:

**Case 1:**
List of meta-features can be edited/ updated via data_edit REST API call: **Modifies existing version - only uploader or admin can do this**
1. description
2. creator
3. contributor
4. collection_date
5. language
6. citation
7. original_data_url
8. paper_url

**Case 2:**
These meta-features/ data will just need to call create_dataset api and create a new version by cloning the old version and changing only the specified field - **Creates new version**
1. attributes
2. data - the data itself
3. default_target_attribute
4. ignore_attribute
5. row_id_attribute

If fields with both Case1 and Case2 are specifed => Case 2 or creation of new version is chosen

#### Any other comments?

- Has not been added to the docs yet, will do so after feedback or review
- We are planning for another "fork" API which forks a dataset for allowing a user to fork and edit his own copy of a dataset


openml/datasets/functions.py
```diff
@@ -799,6 +799,154 @@ def status_update(data_id, status):
...
+    language=None,
+    attributes=None,
+    data=None,
+    default_target_attribute=None,
+    ignore_attribute=None,
+    citation=None,
+    row_id_attribute=None,
+    original_data_url=None,
+    paper_url=None,
+):
```
PGijsbers, line 817: Add the `return` to the signature.

openml/datasets/functions.py
```diff
@@ -799,6 +799,154 @@ def status_update(data_id, status):
...
+            update_comment=dataset.update_comment,
+            version_label=dataset.version_label,
+        )
+        dataset_new.publish()
+        return dataset_new.dataset_id
+
+    # case 2, changing any of these fields will update existing dataset
+    # compose data edit parameters as xml
+    form_data = {"data_id": data_id}
+    xml = OrderedDict()
```
PGijsbers, line 924: I assume for these the server filters out the None values?

openml/datasets/functions.py
```diff
@@ -799,6 +799,154 @@ def status_update(data_id, status):
...
+            del xml["oml:data_edit_parameters"][k]
+
+    file_elements = {"edit_parameters": ("description.xml", xmltodict.unparse(xml))}
+    result_xml = openml._api_calls._perform_api_call(
+        "data/edit", "post", data=form_data, file_elements=file_elements
+    )
+    result = xmltodict.parse(result_xml)
+    data_id = result["oml:data_edit"]["oml:id"]
+    return int(data_id)
+
```
PGijsbers, line 948: I'd probably favor the two scenarios to be extracted to two submethods which are called from this function. But since we are adding support for edits of other object types, I think we can do this at that point in time.

tests/test_datasets/test_dataset_functions.py
```diff
@@ -1331,3 +1332,44 @@ def test_get_dataset_cache_format_feather(self):
...
+    def test_data_edit(self):
+
+        # admin key for test server (only admins or owners can edit datasets).
+        # all users can edit their own datasets)
+        openml.config.apikey = "d488d8afd93b32331cf6ea9d7003d4c3"
+
+        # case 1, existing version edit
+        did = 564
+        result = edit_dataset(
+            did,
```
PGijsbers, line 1345: For each case, please add a comment which fields trigger whether or not they are edits or new versions.
Maybe iterate over all the attributes in a loop so that you can test each attribute separately for the "new version" one? Though if it increases the unit test runtime too much, I don't think that would be worth it.

openml/datasets/functions.py
sahithyaravi1493: yes, it does. we remove the none values from the xml,  but if dont provide atleast1 field to edit, the server throws an error.
Unknown dataset is also thrown if data_id isnt found,

openml/datasets/functions.py
sahithyaravi1493: I agree,  I can do this is an improvement a bit later. But for now, I am keeping it as is, as we need it for the new website

examples/30_extended/datasets_tutorial.py
```diff
@@ -108,3 +109,39 @@
     alpha=0.8,
     cmap="plasma",
 )
+
+
+############################################################################
+# Edit a created dataset
+# =================================================
+# Use test server
```
PGijsbers, line 117: Please elaborate why (i.e. "Use the test server to avoid editing datasets on the live server.").

examples/30_extended/datasets_tutorial.py
```diff
@@ -108,3 +109,39 @@
...
+openml.config.start_using_configuration_for_example()
+############################################################################
+# Changes to these field edits existing version
```
PGijsbers, line 120: Add something to the effect of "A dataset may only be edited by their original uploader."

examples/30_extended/datasets_tutorial.py
```diff
@@ -108,3 +109,39 @@
...
+    paper_url="",
+    citation="kaggle",
+    language="English",
+)
+edited_dataset = get_dataset(data_id)
+print(f"Edited dataset ID: {data_id}")
+
+
+############################################################################
+# Changes to these fields: attributes, default_target_attribute,
```
PGijsbers, line 136: I assume these changes can be made by anyone, as it generates a new dataset anyway. Then add something to that effect.

openml/datasets/functions.py
PGijsbers: Could you add two small unit tests to make sure both those scenarios raise easy to understand errors?

PR #936, opened 2020-07-27
Adding support for scikit-learn > 0.22

Neeratyoy:
Adds support for scikit-learn 0.22.2 and above by ensuring all unit tests pass.

Checking travis and appveyor unit tests.

#### Any other comments?
The changes here reflect higher scikit-learn versions (>0.22) support for Python 3.6 and 3.7


openml/testing.py
```diff
@@ -223,6 +225,10 @@ def _check_fold_timing_evaluations(
             "wall_clock_time_millis_testing": (0, max_time_allowed),
             "wall_clock_time_millis": (0, max_time_allowed),
         }
+        if LooseVersion(sklearn.__version__) >= "0.23":
```
PGijsbers, line 228: Why is it these can't/don't need to get tested?

Neeratyoy: This was a placeholder check since there was only [one unit test](https://github.com/openml/openml-python/blob/dcf2d22e4b92b6afc5bfb15b69c4f8d37f15d57a/tests/test_extensions/test_sklearn_extension/test_sklearn_extension.py#L1809) which had a slightly different failure point.

Looked into it and had to make [this change](https://github.com/openml/openml-python/blob/dcf2d22e4b92b6afc5bfb15b69c4f8d37f15d57a/openml/extensions/sklearn/extension.py#L1349) to account for [this change](https://github.com/scikit-learn/scikit-learn/blob/0.23.1/sklearn/cluster/_kmeans.py#L273) in K-means in scikit-learn 0.23.1

Not sure if this is the most scalable, long-standing solution.

Neeratyoy:
@mfeurer, @PGijsbers 

For [this unit test](https://github.com/openml/openml-python/blob/dcf2d22e4b92b6afc5bfb15b69c4f8d37f15d57a/tests/test_extensions/test_sklearn_extension/test_sklearn_extension.py#L1223), I have currently edited [this](https://github.com/openml/openml-python/blob/dcf2d22e4b92b6afc5bfb15b69c4f8d37f15d57a/tests/test_extensions/test_sklearn_extension/test_sklearn_extension.py#L1250) to get it to pass with scikit-learn 0.23.1.

However, this is of course not a neat solution. I think that this problem stems from how the following 2 changes interact:
* [Change](https://github.com/scikit-learn/scikit-learn/blob/0.23.1/sklearn/ensemble/_forest.py#L1443) in how keyword arguments and positional arguments will be handled in sklearn`__init__`
* [Updates](https://docs.python.org/3/library/inspect.html#inspect.getfullargspec) to the `inspect` package where they seem to have deprecated `getfullargspec` (at least doesn't work with sklearn 0.23 but works with 0.22) --- `inspect.signature` works across the sklearn versions I tested but yields a different output s.t. [this function](https://github.com/openml/openml-python/blob/dcf2d22e4b92b6afc5bfb15b69c4f8d37f15d57a/openml/extensions/sklearn/extension.py#L978) needs updating

In summary, for sklearn 0.23.1, `inspect.getfullargspec` of `sklearn.ensemble.RandomForestRegressor.__init__` (and others) returns `FullArgSpec(args=[], varargs='args', varkw='kwargs', defaults=None, kwonlyargs=[], kwonlydefaults=None, annotations={})`.

Thoughts, please?

PGijsbers:
Given the linked `inspect.getfullargspec` documentation:
> Note that signature() and Signature Object provide the recommended API for callable introspection

and
 
> This function is retained primarily for use in code that needs to maintain compatibility with the Python 2 inspect module API.

I propose we update our code to use `signature`.

.travis.yml
```diff
@@ -15,6 +15,10 @@ env:
   - TEST_DIR=/tmp/test_dir/
   - MODULE=openml
   matrix:
+  - DISTRIB="conda" PYTHON_VERSION="3.7" SKLEARN_VERSION="0.23.1" TEST_DIST="true"
```
mfeurer, line 18: Would it be possible that you sort these a bit better? I would suggest that we sort them by scikit-learn version, and then by Python version; and have the coverage and flake tests at the top.

openml/extensions/sklearn/extension.py
```diff
@@ -1346,7 +1346,7 @@ def _can_measure_cputime(self, model: Any) -> bool:
         # check the parameters for n_jobs
         n_jobs_vals = SklearnExtension._get_parameter_values_recursive(model.get_params(), "n_jobs")
         for val in n_jobs_vals:
-            if val is not None and val != 1:
+            if val is not None and val != 1 and val != "deprecated":
```
mfeurer, line 1349: Could you briefly explain why this is necessary?

tests/test_extensions/test_sklearn_extension/test_sklearn_extension.py
```diff
@@ -161,11 +168,18 @@ def test_serialize_model_clustering(self):
...
+            fixture_name = "sklearn.cluster.{}.KMeans".format(cluster_name)
             fixture_short_name = "sklearn.KMeans"
             # str obtained from self.extension._get_sklearn_description(model)
-            fixture_description = "K-Means clustering"
+            fixture_description = "K-Means clustering{}".format(
+                "" if LooseVersion(sklearn.__version__) < "0.22" else "."
+            )
             version_fixture = "sklearn==%s\nnumpy>=1.6.1\nscipy>=0.9" % sklearn.__version__
+
+            n_jobs_val = "null" if LooseVersion(sklearn.__version__) < "0.23" else '"deprecated"'
```
mfeurer, line 180: This could be further extended so that we can get rid of the if/else statement below which is also about n_jobs, right?

tests/test_extensions/test_sklearn_extension/test_sklearn_extension.py
```diff
@@ -1195,20 +1234,33 @@ def test__get_fn_arguments_with_defaults(self):
...
 
         for fn, num_params_with_defaults in fns:
             defaults, defaultless = self.extension._get_fn_arguments_with_defaults(fn)
             self.assertIsInstance(defaults, dict)
             self.assertIsInstance(defaultless, set)
             # check whether we have both defaults and defaultless params
             self.assertEqual(len(defaults), num_params_with_defaults)
-            self.assertGreater(len(defaultless), 0)
+            if sklearn_version < "0.23":
+                self.assertGreater(len(defaultless), 0)
```
mfeurer, line 1263: You could add an else here to check that the number of defaultless hyperparameters is actually zero.

Neeratyoy: Done.

.travis.yml
Neeratyoy: Done.

openml/extensions/sklearn/extension.py
Neeratyoy: This ties with the next comment you made.

The common issue arises from the `deprecated` value assigned to the `n_jobs` parameter for `KMeans` for sklearn 0.23 and above.

The difference is that in this [unit test](https://github.com/openml/openml-python/blob/dcf2d22e4b92b6afc5bfb15b69c4f8d37f15d57a/tests/test_extensions/test_sklearn_extension/test_sklearn_extension.py#L167) the `n_jobs_val` is verified explicitly through `assertEqual` as part of the `fixture\_parameters`. Whereas in this [unit test](https://github.com/openml/openml-python/blob/dcf2d22e4b92b6afc5bfb15b69c4f8d37f15d57a/tests/test_extensions/test_sklearn_extension/test_sklearn_extension.py#L1779) the check is happening through a function defined under `sklearn/extension.py` where during `_run_model_on_fold` the cpu time estimates are I believe calculated/included, conditioned on the boolean returned by the function `_can_measure_cputime`, i.e., the function you highlighted in this review comment.

Hence, I included `deprecated` to be handled at this level in a generic way.

tests/test_extensions/test_sklearn_extension/test_sklearn_extension.py
Neeratyoy: Are you referring to the if-else [here](https://github.com/openml/openml-python/blob/dcf2d22e4b92b6afc5bfb15b69c4f8d37f15d57a/tests/test_extensions/test_sklearn_extension/test_sklearn_extension.py#L184)?

If so, then I added L180 to avoid an `elif` or a nested structure since the if-else block differentiates three divisions of sklearn versions --- "< 0.20", "<0.23", and ">=0.23".

Did I address the issue?

tests/test_extensions/test_sklearn_extension/test_sklearn_extension.py
mfeurer: Sorry, I don't see the 'else' here. What I meant was:
```python
            if sklearn_version < "0.23":
                self.assertGreater(len(defaultless), 0)
            else:
                self.assertEqual(len(defaultless), 0)

tests/test_extensions/test_sklearn_extension/test_sklearn_extension.py
mfeurer: Yes, I do - this looks much simpler now.

openml/extensions/sklearn/extension.py
mfeurer: Okay, I think I got how this works. I'm fine with it. However, we need to rethink the design here I guess because if a job uses OpenMP we can no longer measure CPU time :(

tests/test_extensions/test_sklearn_extension/test_sklearn_extension.py
Neeratyoy: Aah, I should have mentioned.

So, since we moved to `inspect.signature` across sklearn versions as done [here](https://github.com/openml/openml-python/blob/023ac406af607da934ce15c0933d14b8665a19c5/openml/extensions/sklearn/extension.py#L997), we don't need [this](https://github.com/openml/openml-python/blob/023ac406af607da934ce15c0933d14b8665a19c5/tests/test_extensions/test_sklearn_extension/test_sklearn_extension.py#L1262) if-else anymore.

mfeurer: I see.

PGijsbers:
Looks good to me! Thanks! 🎉

PR #938, opened 2020-07-29
Fork api

sahithyaravi1493:
#### Reference Issue
#929 

Fork API to clone a dataset. 

There are some changes to the edit-api as well, as it was using a function get_arff which was a protected member and also, it did not work well for all datasets. Now its replaced with get_data.

```
id, forked_dataset = fork_dataset(did)
time.sleep(60)
dataset = get_dataset(id)
print(dataset.features)
```

#### Any other comments?
For the testing, I have just compared the did. Ideally we want to check if the cloned data==old data and cloned features == old features.  But the time taken for the status of the dataset to change to active varies (may be 1 min).
time.sleep(60) should work most of the time, but I am not sure if the tests may fail at some point.
Let me know if it is a good idea to add this to the test. Something like:
```
old_dataset = get_dataset(did)
id, forked_dataset = fork_dataset(did)
time.sleep(60)
forked_dataset = get_dataset(id)
assertEqual(old_dataset.features == old_dataset.features)
```

### DIFF
--- a/doc/progress.rst
+++ b/doc/progress.rst
@@ -8,6 +8,7 @@ Changelog
 
 0.11.0
 ~~~~~~
+* ADD #929: Add fork API
 * ADD #929: Add data edit API
 * FIX #873: Fixes an issue which resulted in incorrect URLs when printing OpenML objects after
   switching the server.
--- a/examples/30_extended/datasets_tutorial.py
+++ b/examples/30_extended/datasets_tutorial.py
@@ -11,7 +11,7 @@ How to list and download datasets.
 
 import openml
 import pandas as pd
-from openml.datasets.functions import edit_dataset, get_dataset
+from openml.datasets.functions import edit_dataset, get_dataset, fork_dataset
 
 ############################################################################
 # Exercise 0
@@ -143,5 +143,15 @@ new_attributes = [
 ]
 data_id = edit_dataset(564, attributes=new_attributes)
 print(f"Edited dataset ID: {data_id}")
+
+############################################################################
+# Fork an existing dataset
+# =================================================
+# This example continues to use the test server, to avoid creating multiple copies on main server.
+############################################################################
+
+forked_did, forked_dataset = fork_dataset(68)
+print(f"Forked dataset ID: {forked_did}")
+print(forked_dataset)
 
 openml.config.stop_using_configuration_for_example()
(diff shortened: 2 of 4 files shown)


PGijsbers:
I am not entirely sure I agree with the design here, in particular the automatic upload. Can you explain to me why you want to fork datasets? I can imagine wanting to edit the data, e.g. impute values or binarize the target feature, but as is, I don't think this accomplishes that task.

sahithyaravi1493:
> I am not entirely sure I agree with the design here, in particular the automatic upload. Can you explain to me why you want to fork datasets? I can imagine wanting to edit the data, e.g. impute values or binarize the target feature, but as is, I don't think this accomplishes that task.

On the new website, we want to have a fork/duplicate button that allows people to create a version of a dataset that they fix. As some non-uploaders cant edit some features of the dataset as is. They can use the fork button to create a clone and then edit it as if it were their own dataset right?

PGijsbers:
Well with the current implementation:
1. User clicks fork, [creates a new openml dataset](https://github.com/openml/openml-python/blob/fork_api/openml/datasets/functions.py#L1005)
2. User edits a sensitive feature (e.g. marks a row id attribute), and creates [another copy](https://github.com/openml/openml-python/blob/fork_api/openml/datasets/functions.py#L883)

Result is three datasets: A, A (Copy) and B (corrected row id).

Your suggestion of returning the dataset copy without a publish on `fork_dataset` would prevent this.

sahithyaravi1493:
> Well with the current implementation:
> 
> 1. User clicks fork, [creates a new openml dataset](https://github.com/openml/openml-python/blob/fork_api/openml/datasets/functions.py#L1005)
> 2. User edits a sensitive feature (e.g. marks a row id attribute), and creates [another copy](https://github.com/openml/openml-python/blob/fork_api/openml/datasets/functions.py#L883)
> 
> Result is three datasets: A, A (Copy) and B (corrected row id).
>  I dint like this as well, we are better off with having only 2-copies, let us wait for Joaquin's comments

> Your suggestion of returning the dataset copy without a publish on `fork_dataset` would prevent this.
Still it has to be published by the user if he wants a fork, so it will result in 3-copies if he edits sensitive data again

PGijsbers:
I'm closing this PR. We have discussed that we will first make changes to the server to allow for data edits and forks without duplicating whole datasets. After those changes are made, we can add/update the support for this functionality in the Python API.

PR #940, opened 2020-08-06
Fix edit api

sahithyaravi1493:
the get_arff function is not reliable and sometimes causes errors for dense vs sparse datasets.
I have fixed this by using get_data instead.
The get_arff did not work for some dense datasets. When we use the data returned by get_arff to construct the new dataset, it resulted in errors during dataset publish.

#### Any other comments?
This api is going to change based on server changes in future. We are going to get rid of the cloning except when data itself changes

### DIFF
--- a/doc/progress.rst
+++ b/doc/progress.rst
@@ -8,6 +8,7 @@ Changelog
 
 0.11.0
 ~~~~~~
+* ADD #929: Fix data edit API
 * ADD #929: Add data edit API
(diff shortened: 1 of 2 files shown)


doc/progress.rst
```diff
@@ -8,6 +8,7 @@ Changelog
 
 0.11.0
 ~~~~~~
+* ADD #929: Fix data edit API
```
PGijsbers, line 11: I don't think this extra line is necessary, as it is a fix for a bug that only ever existed in the `develop` branch and never made it to an official release.

PR #941, opened 2020-08-19
change edit_api to reflect server

sahithyaravi1493:
#### Reference Issue
#929 

The server has made changes to the edit_api. Refer https://github.com/openml/OpenML/pull/1058.
Summary of the changes:
**Edit API** - Edit **non-critical data fields** for everyone. Edit **critical fields** only for the owner, provided there are no tasks. If the owner already created a task, we will just throw an error.

**Fork API** - Clone the row as such with change in dataset ID and uploader ID.

#### Any other comments?

1. Fork API changes will be made in a new PR. This is more urgent as it is breaking the openml-python build.
2. Note that  "data" changes are not supported. If the data itself requires changes, the user has to create a new dataset. We will not be doing this through edit API or fork API, as it changes the arff file. 
3. In future we want to support versioning for descriptions (For now we allow direct changes without any version history.)

### DIFF
--- a/examples/30_extended/datasets_tutorial.py
+++ b/examples/30_extended/datasets_tutorial.py
@@ -21,7 +21,7 @@ from openml.datasets.functions import edit_dataset, get_dataset
 #
 #   * Use the output_format parameter to select output type
 #   * Default gives 'dict' (other option: 'dataframe', see below)
-
+#
 openml_list = openml.datasets.list_datasets()  # returns a dict
 
 # Show a nice table with some key data properties
@@ -117,15 +117,19 @@ _ = pd.plotting.scatter_matrix(
 # This example uses the test server, to avoid editing a dataset on the main server.
 openml.config.start_using_configuration_for_example()
 ############################################################################
-# Changes to these field edits existing version: allowed only for dataset owner
+# Change the non-critical fields
+desc = (
+    "This data sets consists of 3 different types of irises' "
+    "(Setosa, Versicolour, and Virginica) petal and sepal length,"
+    " stored in a 150x4 numpy.ndarray"
+)
+did = 128
 data_id = edit_dataset(
-    564,
-    description="xor dataset represents XOR operation",
-    contributor="",
-    collection_date="2019-10-29 17:06:18",
-    original_data_url="https://www.kaggle.com/ancientaxe/and-or-xor",
-    paper_url="",
-    citation="kaggle",
+    did,
+    description=desc,
+    creator="R.A.Fisher",
+    collection_date="1937",
+    citation="The use of multiple measurements in taxonomic problems",
     language="English",
 )
 edited_dataset = get_dataset(data_id)
@@ -133,15 +137,11 @@ print(f"Edited dataset ID: {data_id}")
 
 
 ############################################################################
-# Changes to these fields: attributes, default_target_attribute,
-# row_id_attribute, ignore_attribute generates a new edited version: allowed for anyone
-
-new_attributes = [
-    ("x0", "REAL"),
-    ("x1", "REAL"),
-    ("y", "REAL"),
-]
-data_id = edit_dataset(564, attributes=new_attributes)
+# Changes to these fields: default_target_attribute, row_id_attribute,
+# ignore_attribute can only be performed by owner
+# To edit critical fields of a dataset owned by you, configure the API key:
+# openml.config.apikey = 'FILL_IN_OPENML_API_KEY'
+data_id = edit_dataset(564, default_target_attribute="y")
 print(f"Edited dataset ID: {data_id}")
 
 openml.config.stop_using_configuration_for_example()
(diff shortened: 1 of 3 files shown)


tests/test_datasets/test_dataset_functions.py
```diff
@@ -1341,47 +1341,34 @@ def test_get_dataset_cache_format_feather(self):
...
 
-        column_names = [
-            ("input1", "REAL"),
-            ("input2", "REAL"),
-            ("y", "REAL"),
-        ]
+        # Case 2
+        # only admins or owners can edit all critical fields of datasets
+        # admin key for test server
+        openml.config.apikey = "d488d8afd93b32331cf6ea9d7003d4c3"
```
PGijsbers, line 1365: I had somehow overlooked this previously. Is it possible to generate a new admin API token? I think we should be able to configure the api key as an environment variable through our travis/appveyor accounts so that we don't need to expose this high access api key.

PGijsbers (commented):
Do we want to allow everyone to edit the data? Is this temporary? I can't imagine this is healthy long term, someone will come along and be destructive.

Of course we do still need a way for other users to at least _suggest_ changes to be approved by the original uploader or someone from openml (hopefully mostly the former). Are there plans for this?

tests/test_datasets/test_dataset_functions.py
sahithyaravi1493: > Do we want to allow everyone to edit the data? Is this temporary? I can't imagine this is healthy long term, someone will come along and be destructive.
> 
> Of course we do still need a way for other users to at least _suggest_ changes to be approved by the original uploader or someone from openml (hopefully mostly the former). Are there plans for this?

Yes this is temporary and we have plans to modify this in future such that the owner has some control. Owners can choose to allow or not allow community edits for non-critical features in future. For reference, here is the whole dicussion  openml/OpenML#1058.

sahithyaravi1493: > I had somehow overlooked this previously. Is it possible to generate a new admin API token? I think we should be able to configure the api key as an environment variable through our travis/appveyor accounts so that we don't need to expose this high access api key.

Instead I could change this test such that this admin API key is not required. The test is for whether an  owner(or admin) can edit a critical feature, so it can be changed to edit a dataset owned by CI.

mfeurer:
> Do we want to allow everyone to edit the data? Is this temporary?

Reading this, I'm also a bit worried about this.

sahithyaravi1493:
> > Do we want to allow everyone to edit the data? Is this temporary?
> 
> Reading this, I'm also a bit worried about this.

It is temporary that everyone can edit non-critical fields. 
This is to avoid unnecessary forks(even with the same file ID) for changes to a non-critical field such as citation, paper_url etc.
For now, description also falls under non-critical category.
However, In the near future, we plan to:

1. We will provide a choice to the owner of the dataset to allow/disallow non-critical data edits from the community. If the owner chooses not to allow this, then only the owner can perform edits.
2. We will support versioning for dataset descriptions. This way we dont need different dataset versions for a change in description.

mfeurer:
> It is temporary that everyone can edit non-critical fields.

Wouldn't it make sense to at least restrict this to registered users?

sahithyaravi1493:
> > It is temporary that everyone can edit non-critical fields.
> 
> Wouldn't it make sense to at least restrict this to registered users?

yes only registered users can do this , otherwise the server will throw an error "Authentication failed"

tests/test_datasets/test_dataset_functions.py
PGijsbers: > The goal of the test is primarily to test whether an owner(or ofcourse admin) can edit a critical feature, so it can be changed to edit a dataset owned by CI (eg: 315) or Test Test (eg: 564)

I would prefer that (and reset the admin API key after - it should be considered public right now).

examples/30_extended/datasets_tutorial.py
```diff
@@ -117,31 +117,31 @@
 # This example uses the test server, to avoid editing a dataset on the main server.
 openml.config.start_using_configuration_for_example()
 ############################################################################
-# Changes to these field edits existing version: allowed only for dataset owner
+# Change the non-critical fields
```
mfeurer, line 120: Could you please give more details here, like what are critical fields, what fields can be edited by whom etc?

mfeurer (commented):
This looks good to me, although I would prefer to have a bit more text in the tutorial what critical and non-critical fields are.

PGijsbers (approved):
Looks good to me. We might want to refer to the documentation on API key configuration rather than leaving the commented out code in (https://openml.github.io/openml-python/master/examples/20_basic/introduction_tutorial.html#sphx-glr-examples-20-basic-introduction-tutorial-py). Either way works for me. I'll leave the merge for @mfeurer since he had left feedback on the PR.

tests/test_datasets/test_dataset_functions.py
mfeurer: > I think we should be able to configure the api key as an environment variable through our travis/appveyor accounts so that we don't need to expose this high access api key.

That does not work for PRs from outside the OpenML organization. Such secrets will be visible to the PR, and if someone does a PR, that person has access to the secret. It would make it less obvious though that we use such a secret. I'm also not sure what to do with local tests then.

mfeurer (approved):

PR #943, opened 2020-08-27
Better support for passthrough and drop in sklearn extension 

mfeurer:

This PR fixes a bug where the pipeline and column transformer elements `'passthrough'` and `'drop'` could be serialized into an `OpenMLFlow`, but not into the xml representation of a flow. 

This PR has some basic checks. If the proposed route is fine with everybody, I will add more tests.

TODOs

* [x] more tests
* [x] specific test for serialization of `("drop", "passthrough")`
* [x] extract the tuple `("passthrough", "drop")` into a module level variable
* [x] there's an unnecessary function right now which only returns the input value and can be dropped again
* [x] make strings like component reference etc. magic variables in the code.

### DIFF
--- a/openml/extensions/sklearn/extension.py
+++ b/openml/extensions/sklearn/extension.py
@@ -56,6 +56,10 @@ SIMPLE_NUMPY_TYPES = [
 ]
 SIMPLE_TYPES = tuple([bool, int, float, str] + SIMPLE_NUMPY_TYPES)
 
+SKLEARN_PIPELINE_STRING_COMPONENTS = ("drop", "passthrough")
+COMPONENT_REFERENCE = "component_reference"
+COMPOSITION_STEP_CONSTANT = "composition_step_constant"
+
 
 class SklearnExtension(Extension):
     """Connect scikit-learn to OpenML-Python."""
@@ -249,8 +253,11 @@ class SklearnExtension(Extension):
     ) -> Any:
         """Recursive function to deserialize a scikit-learn flow.
 
-        This function delegates all work to the respective functions to deserialize special data
-        structures etc.
+        This function inspects an object to deserialize and decides how to do so. This function
+        delegates all work to the respective functions to deserialize special data structures etc.
+        This function works on everything that has been serialized to OpenML: OpenMLFlow,
+        components (which are flows themselves), functions, hyperparameter distributions (for
+        random search) and the actual hyperparameter values themselves.
 
         Parameters
         ----------
@@ -258,8 +265,9 @@ class SklearnExtension(Extension):
             the object to deserialize (can be flow object, or any serialized
             parameter value that is accepted by)
 
-        components : dict
-
+        components : Optional[dict]
+            Components of the current flow being de-serialized. These will not be used when
+            de-serializing the actual flow, but when de-serializing a component reference.
 
         initialize_with_defaults : bool, optional (default=False)
             If this flag is set, the hyperparameter values of flows will be
@@ -307,10 +315,10 @@ class SklearnExtension(Extension):
                     rval = self._deserialize_rv_frozen(value)
                 elif serialized_type == "function":
                     rval = self._deserialize_function(value)
-                elif serialized_type in ("composition_step_constant", "component_reference"):
-                    if serialized_type == "composition_step_constant":
-                        value = self._deserialize_parameter_step_constants(value)
-                    elif serialized_type == "component_reference":
+                elif serialized_type in (COMPOSITION_STEP_CONSTANT, COMPONENT_REFERENCE):
+                    if serialized_type == COMPOSITION_STEP_CONSTANT:
+                        pass
+                    elif serialized_type == COMPONENT_REFERENCE:
                         value = self._deserialize_sklearn(
                             value, recursion_depth=depth_pp, strict_version=strict_version
                         )
@@ -415,7 +423,7 @@ class SklearnExtension(Extension):
         elif (
             isinstance(o, (list, tuple))
             and len(o) == 2
-            and o[1] in ("passthrough", "drop")
+            and o[1] in SKLEARN_PIPELINE_STRING_COMPONENTS
             and isinstance(parent_model, sklearn.pipeline._BaseComposition)
         ):
             rval = o
@@ -723,7 +731,10 @@ class SklearnExtension(Extension):
         for key in subcomponents:
             if isinstance(subcomponents[key], OpenMLFlow):
                 name = subcomponents[key].name
-            elif isinstance(subcomponents[key], str):  # 'drop', 'passthrough' can be passed
+            elif (
+                isinstance(subcomponents[key], str)
+                and subcomponents[key] in SKLEARN_PIPELINE_STRING_COMPONENTS
+            ):
                 name = subcomponents[key]
             else:
                 raise TypeError(type(subcomponents[key]))
@@ -800,6 +811,7 @@ class SklearnExtension(Extension):
 
         external_versions = set()
 
+        # The model is None if the flow is a placeholder flow such as 'passthrough' or 'drop'
         if model is not None:
             model_package_name = model.__module__.split(".")[0]
             module = importlib.import_module(model_package_name)
@@ -814,8 +826,7 @@ class SklearnExtension(Extension):
         external_versions.add(openml_version)
         external_versions.add(sklearn_version)
         for visitee in sub_components.values():
-            # 'drop', 'passthrough', None can be passed as estimators
-            if isinstance(visitee, str):
+            if isinstance(visitee, str) and visitee in SKLEARN_PIPELINE_STRING_COMPONENTS:
                 continue
             for external_version in visitee.external_version.split(","):
                 external_versions.add(external_version)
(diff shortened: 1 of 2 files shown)


openml/extensions/sklearn/extension.py
```diff
@@ -307,11 +307,16 @@ def _deserialize_sklearn(
                     rval = self._deserialize_rv_frozen(value)
                 elif serialized_type == "function":
                     rval = self._deserialize_function(value)
-                elif serialized_type == "component_reference":
+                elif serialized_type in ("composition_step_constant", "component_reference"):
```
PGijsbers, line 310: Are these scikit-learn names?

openml/extensions/sklearn/extension.py
```diff
@@ -307,11 +307,16 @@ def _deserialize_sklearn(
...
+                    if serialized_type == "composition_step_constant":
+                        value = self._deserialize_parameter_step_constants(value)
+                    elif serialized_type == "component_reference":
+                        value = self._deserialize_sklearn(
+                            value, recursion_depth=depth_pp, strict_version=strict_version
+                        )
+                    else:
+                        raise NotImplementedError(serialized_type)
                     assert components is not None  # Necessary for mypy
```
PGijsbers, line 319: Not your code but any idea why the type hint is `Optional[dict]` and the doc-string has type `dict`? There is no documentation of what this function argument actually is.

openml/extensions/sklearn/extension.py
```diff
@@ -777,17 +797,20 @@ def _get_external_version_string(
...
-        external_version = self._format_external_version(
-            model_package_name, model_package_version_number,
-        )
-        openml_version = self._format_external_version("openml", openml.__version__)
-        sklearn_version = self._format_external_version("sklearn", sklearn.__version__)
 
         external_versions = set()
-        external_versions.add(external_version)
+
+        if model is not None:
```
PGijsbers, line 803: When is the model `None`?

PGijsbers (commented):
I am growing more concerned with this code.
First, I am a bit worried that what looks like a minor addition requires this much change in our code base (I realize it's also in part refactoring/method extraction).
Second, generally I find the (de)serialization code hard to follow. It's definitely possible to get into it (I have worked on and extended this code before), but requires quite some effort to understand what each case does (for me, anyway). 

I don't have any specific requested changes (I think a refactor should be a separate PR anyway), but I'd like to hear your thoughts first. Do you share these concerns? (If not, why not?)

openml/extensions/sklearn/extension.py
mfeurer: No, these are magic constants used by us -> I should actually have them as magic constants at the top of the file.

tests/test_extensions/test_sklearn_extension/test_sklearn_extension.py
```diff
@@ -311,6 +333,23 @@ def test_serialize_model_with_subcomponent(self):
         self.assertEqual(new_model_params, model_params)
         new_model.fit(self.X, self.y)
 
+        xml = serialization._to_dict()
+        new_model2 = self.extension.flow_to_model(OpenMLFlow._from_dict(xml))
```
Neeratyoy, line 337: Is this to primarily check the ```_from_dict``` function?
If so, given this unit test is really a long one, and since *xml* only depends on a ```model_to_flow``` call, does it make sense to have a smaller unit test checking this?

openml/extensions/sklearn/extension.py
mfeurer: > Not your code but any idea why the type hint is Optional[dict] and the doc-string has type dict?

That's a mistake. It should be optional in the docstring, too.

mfeurer: > There is no documentation of what this function argument actually is.

Will add this.

openml/extensions/sklearn/extension.py
mfeurer: The model is None if the flow is a placeholder flow such as 'passthrough' or 'drop'. I'll add this to the docs.

openml/extensions/sklearn/extension.py
```diff
@@ -407,6 +412,13 @@ def _serialize_sklearn(self, o: Any, parent_model: Optional[Any] = None) -> Any:
         if self.is_estimator(o):
             # is the main model or a submodel
             rval = self._serialize_model(o)
+        elif (
+            isinstance(o, (list, tuple))
+            and len(o) == 2
+            and o[1] in ("passthrough", "drop")
+            and isinstance(parent_model, sklearn.pipeline._BaseComposition)
```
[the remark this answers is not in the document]
mfeurer: I think drop and passthrough can only occur within pipelines and column transformers.

openml/extensions/sklearn/extension.py
```diff
@@ -713,6 +725,8 @@ def _serialize_model(self, model: Any) -> OpenMLFlow:
                 name = subcomponents[key].name
             elif isinstance(subcomponents[key], str):  # 'drop', 'passthrough' can be passed
```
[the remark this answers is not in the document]
mfeurer: Yes, that's a good idea.

openml/extensions/sklearn/extension.py
```diff
@@ -879,12 +904,14 @@ def flatten_all(list_):
...
                     identifier = sub_component_tuple[0]
                     sub_component = sub_component_tuple[1]
-                    # sub_component_type = type(sub_component_tuple)
+                    sub_component_type = type(sub_component_tuple)
                     if not 2 <= len(sub_component_tuple) <= 3:
                         # length 2 is for {VotingClassifier.estimators,
                         # Pipeline.steps, FeatureUnion.transformer_list}
                         # length 3 is for ColumnTransformer
-                        msg = "Length of tuple does not match assumptions"
+                        msg = "Length of tuple of type {} " "does not match assumptions".format(
```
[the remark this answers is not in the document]
mfeurer: Yes. But that's an artifact from the black facelift. I'll fix that.

openml/extensions/sklearn/extension.py
```diff
@@ -1730,26 +1793,66 @@ def is_subcomponent_specification(values):
...
             # _flow is openml flow object, _param dict maps from flow name to flow
             # id for the main call, the param dict can be overridden (useful for
             # unit tests / sentinels) this way, for flows without subflows we do
             # not have to rely on _flow_dict
             exp_parameters = set(_flow.parameters)
-            exp_components = set(_flow.components)
-            model_parameters = set([mp for mp in component_model.get_params() if "__" not in mp])
-            if len((exp_parameters | exp_components) ^ model_parameters) != 0:
-                flow_params = sorted(exp_parameters | exp_components)
+            if isinstance(component_model, str) and component_model in ("passthrough", "drop"):
```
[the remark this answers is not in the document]
mfeurer: Yes, that makes sense. It's one of the TODOs I mentioned above.

tests/test_extensions/test_sklearn_extension/test_sklearn_extension.py
mfeurer: > Is this to primarily check the _from_dict function?

Yes

> If so, given this unit test is really a long one, and since xml only depends on a model_to_flow call, does it make sense to have a smaller unit test checking this?

Yes and no. I'll think about whether this can easily be done without adding too much boilerplate code.

mfeurer:
> I am growing more concerned with this code.

I am doing so as well.

> First, I am a bit worried that what looks like a minor addition requires this much change in our code base (I realize it's also in part refactoring/method extraction - but I count seven "passthrough" in the code).

Yes, it also took me way too much time to do so. But on the other hand, allowing strings in a place where we had only objects of type OpenMLFlow before is actually breaking with a very strong assumption.

> Second, generally I find the (de)serialization code hard to follow. It's definitely possible to get into it (I have worked on and extended this code before), but requires quite some effort to understand what each case does (for me, anyway).
> I don't have any specific requested changes (I think a refactor should be a separate PR anyway), but I'd like to hear your thoughts first. Do you share these concerns? (If not, why not?)

I fully agree. I think this refactor should happend if and when OpenML moves to a new flow description.

mfeurer:
Okay, I added additional unit tests. A lot of tests can probably be unified. Before starting such an endeavor I would like to know whether everything else is fine to you.

PGijsbers (approved):
I think it looks OK. Would appreciate cleaning up the tests.

openml/extensions/sklearn/extension.py
```diff
@@ -1730,26 +1806,75 @@ def is_subcomponent_specification(values):
...
+            else:
+                _ = set([mp for mp in component_model.get_params(deep=False)])
+                model_components = set(
+                    [
+                        mp
+                        for mp in component_model.get_params(deep=True)
+                        if "__" not in mp and mp not in _
+                    ]
+                )
+            if len(exp_components ^ model_components) != 0:
```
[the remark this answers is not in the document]
mfeurer: I never thought about this, but it makes total sense. Just updated the PR with this suggestion.

Neeratyoy (approved):

PR #944, opened 2020-09-03
fork api

sahithyaravi1493:
Refer https://github.com/openml/OpenML/pull/1058.
**Fork API** - Clone the row as such with change in dataset ID and uploader ID.
Uses the same ARFF file.

data_id = fork_dataset(ID)


examples/30_extended/datasets_tutorial.py
```diff
@@ -146,4 +146,13 @@
 data_id = edit_dataset(564, default_target_attribute="y")
 print(f"Edited dataset ID: {data_id}")
 
+
+############################################################################
+# Fork dataset
+# Used to create a copy of the dataset with a different owner
```
mfeurer, line 152: I think it would be good to give more details what this means - one can read up in the PR implementing this for the server - but maybe a glossary somewhere under `doc.openml.org` would be great.

openml/datasets/functions.py
```diff
@@ -897,6 +897,31 @@ def edit_dataset(
     return int(data_id)
 
 
+def fork_dataset(data_id) -> int:
+    """
+    Create a fork of the dataset with the calling authorized user as owner.
```
mfeurer, line 902: Isn't it the authenticated user which becomes the new owner?

openml/datasets/functions.py
```diff
@@ -897,6 +897,31 @@ def edit_dataset(
     return int(data_id)
 
 
+def fork_dataset(data_id) -> int:
```
mfeurer, line 900: Could you please add a type annotation to the argument?

tests/test_datasets/test_dataset_functions.py
```diff
@@ -27,6 +27,7 @@
 from openml.datasets.functions import (
     create_dataset,
     edit_dataset,
+    fork_dataset,
```
mfeurer, line 30: Could you please also add the function to the documentation and also to https://github.com/openml/openml-python/blob/develop/openml/datasets/__init__.py and maybe not import from functions.py as done here (I know you didn't write the original code, but we should try to get rid of importing from functions.py if not necessary).

mfeurer (commented):
The PR itself looks good to me, but I'm worried that the workflows of edit and fork are only documented in a PR in a different repository. I think there should be some more documentation on openml.org about such concepts.

openml/datasets/functions.py
sahithyaravi1493: Yes

examples/30_extended/datasets_tutorial.py
sahithyaravi1493: Yes, I will add this to the documentation. Especially the API itself is not documented, except in the examples in python API.

PGijsbers: I think I would be more explicit, e.g. prefer "Used to create a copy of the dataset with ~a different~ you as the owner".

PGijsbers: I agree that the general description of what forking a dataset entails and why you would do it should be hosted in the cross-platform documentation. However, I would still dedicate one or two sentences to recap it here as well (best to assume users are lazy).

openml/datasets/functions.py
```diff
@@ -815,12 +815,16 @@ def edit_dataset(
 ) -> int:
     """
       Edits an OpenMLDataset.
-      Specify atleast one field to edit, apart from data_id
-       - For certain fields, a new dataset version is created : attributes, data,
-       default_target_attribute, ignore_attribute, row_id_attribute.
+      Specify atleast one field to edit, apart from data_id.
```
PGijsbers, line 818: `at least` requires a space

openml/datasets/functions.py
```diff
@@ -815,12 +815,16 @@ def edit_dataset(
...
 
-       - For other fields, the uploader can edit the exisiting version.
-        Noone except the uploader can edit the exisitng version.
+      This API has non-critical and critical fields.
```
PGijsbers, line 820: I'd word it as `This function allows editing of both non-critical and critical fields.`.

openml/datasets/functions.py
```diff
@@ -815,12 +815,16 @@ def edit_dataset(
...
+      Critical features are default_target_attribute, ignore_attribute, row_id_attribute.
```
PGijsbers, line 821: "Critical ~features~ fields" to keep naming consistent.

openml/datasets/functions.py
```diff
@@ -815,12 +815,16 @@ def edit_dataset(
...
+
+       - Editing non-critical data fields is allowed for all authenticated users.
+       - Editing critical fields is allowed only for the owner, provided there are no tasks.
```
PGijsbers, line 824: "... no tasks associated with this dataset."

openml/datasets/functions.py
```diff
@@ -862,7 +866,7 @@ def edit_dataset(
 
       Returns
       -------
-      data_id of the existing edited version or the new version created and published"""
+      data_id of the edited dataset """
```
PGijsbers, line 869: "data_id" -> Dataset id

The documentation does not need to expose variable names internal to the function.

openml/datasets/functions.py
PGijsbers: After some consideration, I would word this slightly differently (because this might imply you are editing the dataset id too):

> In addition to providing the dataset id of the dataset to edit (through `dataset_id`), you must specify a value for at least one of the optional function arguments, i.e. one value for a field to edit.

It's a little bit more verbose, but I think it is a bit clearer, especially since this is end-user facing documentation (i.e. this function should not only be called by us internally, but also users that only use the package and not develop it).

openml/datasets/functions.py
```diff
@@ -897,6 +901,46 @@ def edit_dataset(
     return int(data_id)
 
 
+def fork_dataset(data_id: int) -> int:
+    """
+     Creates a new dataset version, with possibly a different owner.
+     The authenticated user becomes the new owner.
```
PGijsbers, line 907: Merge two sentences to "Creates a new dataset version, with the authenticated user as the new owner."

openml/datasets/functions.py
```diff
@@ -897,6 +901,46 @@ def edit_dataset(
...
+     This function does not result in any change in the dataset ARFF file.
```
PGijsbers, line 908: This might be a bit low level for function documentation?
Perhaps instead something like
  "The forked dataset can have distinct dataset meta-data, but the actual data itself is shared with the original version."?

tests/test_datasets/test_dataset_functions.py
```diff
@@ -1407,3 +1407,12 @@ def test_data_edit_errors(self):
             data_id=128,
             default_target_attribute="y",
         )
+
+    def test_data_fork(self):
+        did = 1
+        result = fork_dataset(did,)
```
PGijsbers, line 1413: Remove extra `,`

PGijsbers (changes requested):
Looks good, I left some remarks about the documentation.

openml/datasets/functions.py
```diff
@@ -862,7 +869,7 @@ def edit_dataset(
 
       Returns
       -------
-      data_id of the existing edited version or the new version created and published"""
+      Dataset id """
```
PGijsbers, line 872: Sorry if I wasn't clear, I still like the description of which dataset id is returned - just for it not to be referred to as `dataset_id`  which is the internal variable name.

PGijsbers:
Thanks! It looks good! I was perhaps a bit unclear with one part of the feedback, so I left a comment there.

openml/datasets/functions.py
sahithyaravi1493: The edit only edits the existing dataset now, it does not create a new version.
So that description is outdated.
We could possibly have
"Dataset id of the edited dataset" but its the same as the one we wanted to edit.

PGijsbers: Thanks! Apologies, I missed that this was the `edit`  function. Is there a reason we even should parse for the id at this point? Wouldn't just parsing the response to verify the edit was done successfully (and if not, raise a clear error) be enough?
If an error has occurred, raise some kind of `OpenMLException` and if everything went OK, do nothing (not even return a dataset id - as it is just the one we provided in the first place?).

PGijsbers: My bad. I forgot `_perform_api_call` checks for the server messages already.

PGijsbers (approved):

PGijsbers:
@mfeurer I think it looks good to me. Can you check that your review comments are addressed to your satisfaction?

examples/30_extended/datasets_tutorial.py
mfeurer: Thanks for updating the description. What I'm wondering though is, how is the dataset then finalized (i.e. critical fields should not be editable at all, so there must be a way of finalizing the dataset such that even the fork cannot be changed)? Could you please extend the example by that?

openml/datasets/functions.py
```diff
@@ -815,12 +815,19 @@ def edit_dataset(
 ) -> int:
     """
       Edits an OpenMLDataset.
-      Specify atleast one field to edit, apart from data_id
-       - For certain fields, a new dataset version is created : attributes, data,
-       default_target_attribute, ignore_attribute, row_id_attribute.
+      In addition to providing the dataset id of the dataset to edit (through data_id),
+      you must specify a value for at least one of the optional function arguments,
+       i.e. one value for a field to edit.
```
mfeurer, line 820: The `i.e.` appears to be indented too far

mfeurer:
Thanks for the reminder @PGijsbers. I just had a look and would like to have a further clarification in the example.

examples/30_extended/datasets_tutorial.py
PGijsbers: The dataset is "finalized" when a task is created for it. After a task has been created, critical fields can no longer be edited, not even by the dataset owner. (But yes, it is a good idea to write this down in the documentation)

mfeurer: Thanks @PGijsbers for the explanation. @sahithyaravi1493 could you please add that to the docs?

mfeurer (approved):

PGijsbers (approved):

doc/progress.rst
```diff
@@ -8,7 +8,7 @@ Changelog
 
 0.11.0
 ~~~~~~
-* ADD #929: Add data edit API
+* ADD #929: Add data edit API and Fork API
```
PGijsbers, line 11:
```suggestion
* ADD #929: Add ``edit_dataset`` and ``fork_dataset`` to allow editing and forking of uploaded datasets.
```

tests/test_datasets/test_dataset_functions.py
```diff
@@ -1407,3 +1407,12 @@ def test_data_edit_errors(self):
...
             default_target_attribute="y",
         )
+
+    def test_data_fork(self):
+        did = 1
+        result = fork_dataset(did)
+        self.assertNotEqual(did, result)
+        # Check server exception when unknown dataset is provided
+        self.assertRaisesRegex(
+            OpenMLServerException, "Unknown dataset", fork_dataset, data_id=100000,
```
PGijsbers, line 1417:
```suggestion
            OpenMLServerException, "Unknown dataset", fork_dataset, data_id=999999,
```

PR #958, opened 2020-10-02
Class to enum

### LINKED ISSUE #951: Convert TaskType Class to a TaskType Enum
OpenML features different task types (e.g. classification, regression), which we map to integers through [this class](https://github.com/openml/openml-python/blob/d303cedf5498e9eaf41f084f25d49d032f7630f4/openml/tasks/task.py#L472).
Since Python 3.4 added [Enums](https://docs.python.org/3/library/enum.html) to the language, so we can update our code!

In particular, we should:
 - convert the TaskType to an Enum
 - refactor the name (just `TaskType` is sufficient)
 - update every *use* and *type hint* and *doc string* where `TaskType` is an `int` to use the enum `TaskType` instead, for example [here](https://github.com/openml/openml-python/blob/d303cedf5498e9eaf41f084f25d49d032f7630f4/openml/tasks/functions.py#L461):
```python
def create_task(
    task_type_id: TaskType,
    ...
)
```
 - bonus: make `OpenMLTask.task_type` (a string representation of the task type) based on the enum directly: if `task_type` is your stored `TaskType` enum variable, the string name is accessed with `task_type.name`.

This issue requires little knowledge of the modules, but might be a bit of work to find different spellings of `task_type`.
Make sure to use your IDE for renaming (e.g. `shift+F6` in PyCharm on Windows) to rename all references to a variable at once!
Use find across multiple files (e.g. PyCharm `ctrl+shift+F`) through all files for terms like "task_type_id" to help find variables and documentation to update.

abrahamfrancis:
#### Reference Issue

To Resolve #951 

This PR converts the TaskTypeEnum class to an enum.Enum object

### DIFF
--- a/examples/30_extended/tasks_tutorial.py
+++ b/examples/30_extended/tasks_tutorial.py
@@ -31,7 +31,7 @@ import pandas as pd
 #
 # We will start by simply listing only *supervised classification* tasks:
 
-tasks = openml.tasks.list_tasks(task_type_id=TaskType.SUPERVISED_CLASSIFICATION)
+tasks = openml.tasks.list_tasks(task_type=TaskType.SUPERVISED_CLASSIFICATION)
 
 ############################################################################
 # **openml.tasks.list_tasks()** returns a dictionary of dictionaries by default, which we convert
@@ -47,7 +47,7 @@ print(tasks.head())
 # As conversion to a pandas dataframe is a common task, we have added this functionality to the
 # OpenML-Python library which can be used by passing ``output_format='dataframe'``:
 tasks_df = openml.tasks.list_tasks(
-    task_type_id=TaskType.SUPERVISED_CLASSIFICATION, output_format="dataframe"
+    task_type=TaskType.SUPERVISED_CLASSIFICATION, output_format="dataframe"
 )
 print(tasks_df.head())
 
@@ -158,7 +158,7 @@ print(tasks[0])
 #
 # Creating a task requires the following input:
 #
-# * task_type_id: The task type ID, required (see below). Required.
+# * task_type: The task type ID, required (see below). Required.
 # * dataset_id: The dataset ID. Required.
 # * target_name: The name of the attribute you aim to predict. Optional.
 # * estimation_procedure_id : The ID of the estimation procedure used to create train-test
@@ -190,7 +190,7 @@ openml.config.start_using_configuration_for_example()
 
 try:
     my_task = openml.tasks.create_task(
-        task_type_id=TaskType.SUPERVISED_CLASSIFICATION,
+        task_type=TaskType.SUPERVISED_CLASSIFICATION,
         dataset_id=128,
         target_name="class",
         evaluation_measure="predictive_accuracy",
--- a/examples/30_extended/tasks_tutorial.py
+++ b/examples/30_extended/tasks_tutorial.py
@@ -30,7 +30,7 @@ import pandas as pd
 #
 # We will start by simply listing only *supervised classification* tasks:
 
-tasks = openml.tasks.list_tasks(task_type_id=1)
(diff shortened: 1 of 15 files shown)


examples/30_extended/tasks_tutorial.py
```diff
@@ -30,7 +30,7 @@
 #
 # We will start by simply listing only *supervised classification* tasks:
 
-tasks = openml.tasks.list_tasks(task_type_id=1)
+tasks = openml.tasks.list_tasks(task_type_id=tasktypes.SUPERVISED_CLASSIFICATION)
```
PGijsbers, line 33: `tasktypes` is not defined here. Add an import statement: `from openml.tasks import TaskType`, then just used it directly (`TaskType.SUPERVISED_CLASSIFICATION`).

examples/30_extended/tasks_tutorial.py
```diff
@@ -186,7 +188,7 @@
 openml.config.start_using_configuration_for_example()
 
 try:
-    tasktypes = openml.tasks.TaskTypeEnum
+    tasktypes = openml.tasks.TaskType
     my_task = openml.tasks.create_task(
         task_type_id=tasktypes.SUPERVISED_CLASSIFICATION,
```
PGijsbers, line 193: Just get rid of the assignment to `tasktypes` and use `TaskType` directly.

openml/tasks/functions.py
```diff
@@ -119,7 +119,7 @@ def _get_estimation_procedure_list():
 
 
 def list_tasks(
-    task_type_id: Optional[int] = None,
+    task_type_id: Optional[TaskType] = None,
```
PGijsbers, line 122: Could you rename this to `task_type` (also in the documentation)?

openml/tasks/functions.py
```diff
@@ -195,7 +195,7 @@ def _list_tasks(task_type_id=None, output_format="dict", **kwargs):
     Filter task_type_id is separated from the other filters because
     it is used as task_type_id in the task description, but it is named
     type when used as a filter in list tasks call.
-    task_type_id : int, optional
+    task_type_id : TaskType, optional
```
PGijsbers, line 198: Could you rename the parameter of this function as well (to `task_type`)?

openml/tasks/task.py
```diff
@@ -36,7 +49,7 @@ class OpenMLTask(OpenMLBase):
     def __init__(
         self,
         task_id: Optional[int],
-        task_type_id: int,
+        task_type_id: TaskType,
```
PGijsbers, line 52: And a rename to `task_type` here as well.

PGijsbers (changes requested):
Thanks a ton! I think it's nearly ready for a merge, but I'd like to suggest two changes (see the comments):
 - the `tasktypes` is not defined in the example, which leads to an error
 - could you also please update the parameter names from `task_type_id` to `task_type` to better signal we moved away from using integers in the code?

openml/tasks/task.py
abrahamfrancis: Hey, there is a str `task_type` already being passed to this function. Should I change its name?

abrahamfrancis:
I made the changes you said. Not sure about what I should do with the `OpenMLTask` class and its children.

openml/tasks/task.py
PGijsbers: Ah. I am having some troubles right now with figuring out what the effect of that change will be. It's fine to leave it like it is, I'll have a look at it myself later 👍

PGijsbers (approved):

PGijsbers:
Thanks for your contribution! 🎉
If you're interested, there will a virtual OpenML workshop will be held October 26-30. It is a week long event where we get together to talk about and work on all things OpenML (this includes `openml-python` development). Visit [OpenML](https://new.openml.org/meet#upcoming) to find out more!

PR #960, opened 2020-10-16
Improved documentation of example

joaquinvanschoren:
The dataset creation example did not explain how to add multiple ignore attributes.

I tried comma-separated but then the server doesn't recognize them (but also doesn't return an error).
It needs to be a list.

### DIFF
--- a/examples/30_extended/create_upload_tutorial.py
+++ b/examples/30_extended/create_upload_tutorial.py
@@ -100,8 +100,8 @@ diabetes_dataset = create_dataset(
     # The attribute that represents the row-id column, if present in the
     # dataset.
     row_id_attribute=None,
-    # Array of attributes that should be excluded in modelling, such as 
-    # identifiers and indexes. E.g. ["feat1","feat2"]
+    # Attribute or list of attributes that should be excluded in modelling, such as 
+    # identifiers and indexes. E.g. "feat1" or ["feat1","feat2"]
     ignore_attribute=None,
     # How to cite the paper.
     citation=citation,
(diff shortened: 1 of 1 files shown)


examples/30_extended/create_upload_tutorial.py
```diff
@@ -100,8 +100,8 @@
     # The attribute that represents the row-id column, if present in the
     # dataset.
     row_id_attribute=None,
-    # Attributes that should be excluded in modelling, such as identifiers and
-    # indexes.
+    # Array of attributes that should be excluded in modelling, such as 
+    # identifiers and indexes. E.g. ["feat1","feat2"]
```
PGijsbers, line 104:
```suggestion
    # Attribute or list of attributes that should be excluded in modelling, such as 
    # identifiers and indexes. E.g. "feat1" or ["feat1","feat2"]
```

PGijsbers: My suggestion to reflect that it can be either a single attribute or a list of attributes (and change "array" to "list").

PGijsbers (changes requested):
Please look at my suggested change and commit it if you feel it improves the PR. Either way works for me.

mfeurer:
The docs suggest that this can be a string or a list. Should we maybe take you running into this issue as a reason to check if the attributes actually exist?

PGijsbers:
Missed that detail of the original post. Agreed that we should also add a check to verify the attributes actually exist 👍

PGijsbers (approved):

PR #961, opened 2020-10-20
Updating contribution to aid debugging

Neeratyoy:
Adds a *NOTE* to the contribution guidelines to help debug failing examples.
Currently, if the first failing example switched the server from live to test or vice-versa, and the subsequent example expects the other server, the following examples fail to be built as well.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -239,8 +239,10 @@ You may then run a specific module, test case, or unit test respectively:
   $ pytest tests/test_datasets/test_dataset.py::OpenMLDatasetTest::test_get_data
 ```
 
-*NOTE*: In the case the examples build fails during the Continuous Integration test online, please check if
-the first failing example file changed the server that the example fetches from.
+*NOTE*: In the case the examples build fails during the Continuous Integration test online, please 
+fix the first failing example. If the first failing example switched the server from live to test 
+or vice-versa, and the subsequent examples expect the other server, the ensuing examples will fail 
+to be built as well.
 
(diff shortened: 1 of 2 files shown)


mfeurer (approved):

CONTRIBUTING.md
```diff
@@ -239,6 +239,9 @@ You may then run a specific module, test case, or unit test respectively:
   $ pytest tests/test_datasets/test_dataset.py::OpenMLDatasetTest::test_get_data
 ```
 
+*NOTE*: In the case the examples build fails during the Continuous Integration test online, please check if
+the first failing example file changed the server that the example fetches from.
```
PGijsbers, line 243: You mean if the test server is used (or production, depending on the example)? I think it would help to state it explicitly and why this may lead to failure. Why would it only fail on CI, and not locally?

PGijsbers (commented):
See note

CONTRIBUTING.md
Neeratyoy: Updated it. Not sure if a person contributing to the repo will be calling `make html` to test the examples docs. I thought that it is more likely to such errors during the Travis run, hence I mentioned the online failure. Also, I haven't myself tested if such errors appear on local builds.

PGijsbers: Makes sense. The clarification now makes me understand what kind of problem I am hoping to identify :) looks good to me!

PGijsbers (approved):

mfeurer (approved):

PR #967, opened 2020-10-23
Additional fixes to PR 777

Neeratyoy:
#### Reference Issue
PR #777. 

Makes the changes/improvements suggested by @PGijsbers that were not addressed prior to merge.

### DIFF
--- a/examples/30_extended/run_setup_tutorial.py
+++ b/examples/30_extended/run_setup_tutorial.py
@@ -38,7 +38,6 @@ from sklearn.pipeline import make_pipeline, Pipeline
 from sklearn.compose import ColumnTransformer
 from sklearn.impute import SimpleImputer
 from sklearn.preprocessing import OneHotEncoder, FunctionTransformer
-from sklearn.experimental import enable_hist_gradient_boosting
 
 
 openml.config.start_using_configuration_for_example()
@@ -55,7 +54,7 @@ task = openml.tasks.get_task(6)
 # many potential hyperparameters. Of course, the model can be as complex and as
 # easy as you want it to be
 
-from sklearn.ensemble import HistGradientBoostingClassifier
+from sklearn.ensemble import RandomForestClassifier
 from sklearn.decomposition import TruncatedSVD
 
 
@@ -74,18 +73,16 @@ cat_imp = make_pipeline(
     TruncatedSVD(),
 )
 ct = ColumnTransformer([("cat", cat_imp, cat), ("cont", "passthrough", cont)])
-model_original = Pipeline(
-    steps=[("transform", ct), ("estimator", HistGradientBoostingClassifier()),]
-)
+model_original = Pipeline(steps=[("transform", ct), ("estimator", RandomForestClassifier()),])
 
 # Let's change some hyperparameters. Of course, in any good application we
 # would tune them using, e.g., Random Search or Bayesian Optimization, but for
 # the purpose of this tutorial we set them to some specific values that might
 # or might not be optimal
 hyperparameters_original = {
-    "estimator__loss": "auto",
-    "estimator__learning_rate": 0.15,
-    "estimator__max_iter": 50,
+    "estimator__criterion": "gini",
+    "estimator__n_estimators": 50,
+    "estimator__max_depth": 10,
     "estimator__min_samples_leaf": 1,
 }
 model_original.set_params(**hyperparameters_original)
--- a/examples/30_extended/run_setup_tutorial.py
+++ b/examples/30_extended/run_setup_tutorial.py
@@ -34,9 +34,6 @@ In this tutorial we will
 
 import numpy as np
 import openml
-import sklearn.ensemble
-import sklearn.impute
-import sklearn.preprocessing
 from sklearn.pipeline import make_pipeline, Pipeline
 from sklearn.compose import ColumnTransformer
 from sklearn.impute import SimpleImputer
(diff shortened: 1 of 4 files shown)


PGijsbers:
Checks manually cancelled.

Neeratyoy:
@PGijsbers wrt to why an experimental `HistGradientBoostingClassifier` is used in [this example](https://github.com/openml/openml-python/pull/777#discussion_r510715345):

To handle mixed-type data frames with missing values, we need to create a transformation which imputes these missing values for both categorical and continuous types. Under the current OpenML setup, we cannot create a flow with two instances of the same class, a `SimpleImputer` in this case. So far, for [various](https://github.com/openml/openml-python/blob/bf3cd2ebaac10bd05809a1ce90e346248c4c61b1/tests/test_runs/test_run_functions.py#L570) [unit tests](https://github.com/openml/openml-python/blob/bf3cd2ebaac10bd05809a1ce90e346248c4c61b1/tests/test_extensions/test_sklearn_extension/test_sklearn_extension.py#L1706), we have been handling it via an alias such as [this](https://github.com/openml/openml-python/blob/develop/openml/testing.py#L261). 

To prevent importing from `openml.testing` in a documentation example, it was decided to bypass the need of a second imputer. @mfeurer suggested the `HistGradientBoostingClassifier` since it can handle numeric missing values implicitly. Therefore, for this example, a `passthrough` works.

PGijsbers:
That makes sense (even though it is a bit unfortunate). However, why not consider creating the example on a task which does not have two types of missing values? While all examples implicitly depend on some version(s) of scikit-learn, it feels using an experimental feature has exceptionally narrow support. Old versions (`<=0.20`) won't have this, future versions won't have this.

openml/extensions/sklearn/extension.py
```diff
@@ -1608,23 +1610,22 @@ def _prediction_to_probabilities(y: np.ndarray, model_classes: List[Any]) -> pd.
...
-            if isinstance(task, (OpenMLClassificationTask, OpenMLLearningCurveTask)):
-                if task.class_labels is not None:
-                    if isinstance(y_train, np.ndarray) and isinstance(task.class_labels[0], str):
-                        # mapping (decoding) the predictions to the categories
-                        # creating a separate copy to not change the expected pred_y type
-                        y = [task.class_labels[pred] for pred in y]
-                else:
-                    raise ValueError("The task has no class labels")
-            else:
+            if not isinstance(task, (OpenMLClassificationTask, OpenMLLearningCurveTask)):
```
PGijsbers, line 1613: Still, as far as I can tell this function is only called whenever this condition is already satisfied?

openml/extensions/sklearn/extension.py
```diff
@@ -1608,23 +1610,22 @@ def _prediction_to_probabilities(y: np.ndarray, model_classes: List[Any]) -> pd.
...
                 return None
 
-            # y: list or numpy array of predictions
+            if task.class_labels is None:
```
PGijsbers, line 1616: Still, as far as I can tell this function is only called whenever this condition is already satisfied?

PGijsbers (changes requested):

openml/extensions/sklearn/extension.py
Neeratyoy: I now understand. I think I can do away with the if...else under the exception block and make it cleaner?
And abstract away the checks to the `_prediction_to_probabilities` function.

Neeratyoy:
> However, why not consider creating the example on a task which does not have two types of missing values? 

Yeah I was wondering why that was not done here but then I think we need to change the design of the example. At this point, it samples 3 tasks randomly from the suite. Do we then fix this task list? If so, do we provide a justification? 

Either way, I shall try and find 3 tasks where missing values if present is restricted to one of the types. We can then re-iterate and replace the `HistGradientBoostingClassifier` with a more stable classifier.

examples/30_extended/run_setup_tutorial.py
```diff
@@ -58,7 +54,7 @@
 # many potential hyperparameters. Of course, the model can be as complex and as
 # easy as you want it to be
 
-from sklearn.ensemble import HistGradientBoostingClassifier
+from sklearn.ensemble import RandomForestClassifier
```
mfeurer, line 57: Could you please move these import up to the other imports.

examples/30_extended/study_tutorial.py
```diff
@@ -104,7 +103,7 @@ def cat(X):
     [("cat", cat_imp, cat), ("cont", FunctionTransformer(lambda x: x, validate=False), cont)]
 )
 clf = sklearn.pipeline.Pipeline(
-    steps=[("transform", ct), ("estimator", HistGradientBoostingClassifier()),]
+    steps=[("transform", ct), ("estimator", HistGradientBoostingClassifier())]
```
mfeurer, line 106: Here we actually also have the imports in two places, could you please clean these as well?

mfeurer (commented):
Looks good to me. I'll approve once I see that the tests are green.

examples/30_extended/study_tutorial.py
```diff
@@ -77,38 +72,18 @@
...
-    return X.dtypes != "category"
-
-
-def cat(X):
-    return X.dtypes == "category"
+# Model to be used
+clf = RandomForestClassifier()
 
+# We'll create a study with one run on 5 datasets present in the suite
+tasks = [115, 259, 307]
```
PGijsbers, line 81: I count three?

Neeratyoy: aah missed it during re-iteration
I'll push now or wait for tests to go green first?

PGijsbers: I normally just push immediately.

PGijsbers (approved):

PR #976, opened 2020-10-26
update home page example to numerical dataset (pendigits)

### LINKED ISSUE #973: Home page example fails with 0.11
The example on the [documentation homepage](https://openml.github.io/openml-python/master/) fails.
This is due to:
 - openml-python 0.11 working by dataframes by default, which means that categorical features are no longer automatically converted to a numerical representations, and
 - the example pipeline is not compatible with non-numeric input.

This behavior is intended, so we should update the example to use a task with only numeric data instead.

a-moadel:
#### Reference Issue

Fixes #973 

change the task id in the home page example.

Execute the code in the first example and it should run without an error.

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -32,7 +32,7 @@ Example
     )
     # Download the OpenML task for the german credit card dataset with 10-fold
     # cross-validation.
-    task = openml.tasks.get_task(31)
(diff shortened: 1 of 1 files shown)


PGijsbers (approved):
Thanks for both raising the issue and submitting the fix 🎉

Issue #964, opened 2020-10-22
Verify `ignore_attributes` and `row_id_attribute` on `create_dataset`


PGijsbers:
From #960.

When [`create_dataset`](https://github.com/openml/openml-python/blob/88b7cc0292bb5a7b86a9f45cf29d1733ee3cc300/openml/datasets/functions.py#L580) is called, the user can specify `ignored_attributes` and `row_id_attribute`. These are attributes in the dataset which should be ignored when building models, because they represent e.g. an id field or a feature with label leakage. However, there is currently no check that the attributes listed are actually present in the data.

Using the [dataset upload tutorial](https://openml.github.io/openml-python/master/examples/30_extended/create_upload_tutorial.html#sphx-glr-examples-30-extended-create-upload-tutorial-py), if we overwrite `row_id_attribute=None`  or `ignore_attribute=None` in the `create_dataset` calls to instead say  `row_id_attribute="not exist"` or  `ignore_attribute="not exist"` no error is raised, even though there is no "not exist" attribute in the dataset.

We expect either scenario to raise a `ValueError` clearly stating which id or ignored attribute is not present in the dataset.

joaquinvanschoren:
FYI, when you use `ignored_attributes="attribute1,attribute2"` as a string, it will store "attribute1,attribute2" on the server, and not "attribute1","attribute2". That's no good. Hence, if one uses a comma-separated string instead of a list, it should either return an error or auto-parse and check if all attributes exist.

PGijsbers:
I think this will automatically be caught if this issue is resolved, as the dataset will not have an attribute named `"attribute1,attribute2"`.

a-moadel:
@PGijsbers I would like start working in this task.

PGijsbers:
#978 added client-side checks

PR #978, opened 2020-10-28
add validation for ignore_attributes and default_target_attribute at …

a-moadel:
…craete_dataset

#### Reference Issue

#964

##### What does this PR implement/fix? Explain your changes.

1. First for ignore_attributes and default_target_attribute expand to list if they are comma sperated strings.
2. check for both attributes that all values are ie columns.

on call of create_dataset , we can try values for ignore_attributes and default_target_attribute that are part of the columns and some values that are not.

### DIFF
--- a/openml/datasets/functions.py
+++ b/openml/datasets/functions.py
@@ -333,6 +333,28 @@ def _load_features_from_file(features_file: str) -> Dict:
         return xml_dict["oml:data_features"]
 
 
+def _expand_parameter(parameter):
+    expanded_parameter = []
+    if isinstance(parameter, str):
+        expanded_parameter = [x.strip() for x in parameter.split(",")]
+    elif isinstance(parameter, list):
+        expanded_parameter = parameter
+    return expanded_parameter
+
+
+def _validated_data_attributes(attributes, data_attributes, parameter_name):
+    if attributes is not None:
+        for attribute_ in attributes:
+            is_row_id_an_attribute = any([attr[0] == attribute_ for attr in data_attributes])
+            if not is_row_id_an_attribute:
+                raise ValueError(
+                    "all attribute of '{}' should be one of the data attribute. "
+                    " Got '{}' while candidates are {}.".format(
+                        parameter_name, attribute_, [attr[0] for attr in data_attributes]
+                    )
+                )
+
+
 def check_datasets_active(dataset_ids: List[int]) -> Dict[int, bool]:
     """
     Check if the dataset ids provided are active.
@@ -636,6 +658,7 @@ def create_dataset(
     ignore_attribute : str | list
         Attributes that should be excluded in modelling,
         such as identifiers and indexes.
+        Can have multiple values, comma separated.
     citation : str
         Reference(s) that should be cited when building on this data.
     version_label : str, optional
@@ -687,6 +710,11 @@ def create_dataset(
                     attributes_[attr_idx] = (attr_name, attributes[attr_name])
     else:
         attributes_ = attributes
+    ignore_attributes = _expand_parameter(ignore_attribute)
+    _validated_data_attributes(ignore_attributes, attributes_, "ignore_attribute")
+
+    default_target_attributes = _expand_parameter(default_target_attribute)
+    _validated_data_attributes(default_target_attributes, attributes_, "default_target_attribute")
 
     if row_id_attribute is not None:
         is_row_id_an_attribute = any([attr[0] == row_id_attribute for attr in attributes_])
--- a/openml/datasets/functions.py
+++ b/openml/datasets/functions.py
@@ -333,7 +333,7 @@ def _load_features_from_file(features_file: str) -> Dict:
         return xml_dict["oml:data_features"]
 
 
-def _expand_parameter(parameter):
+def _expand_parameter(parameter: Union[str, list]):
     expanded_parameter = []
     if isinstance(parameter, str):
         expanded_parameter = [x.strip() for x in parameter.split(",")]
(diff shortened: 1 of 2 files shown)


openml/datasets/functions.py
```diff
@@ -333,6 +333,28 @@ def _load_features_from_file(features_file: str) -> Dict:
         return xml_dict["oml:data_features"]
 
 
+def _expand_parameter(parameter):
```
PGijsbers, line 336: please add type hints

openml/datasets/functions.py
```diff
@@ -333,6 +333,28 @@ def _load_features_from_file(features_file: str) -> Dict:
...
+    expanded_parameter = []
+    if isinstance(parameter, str):
+        expanded_parameter = [x.strip() for x in parameter.split(",")]
+    elif isinstance(parameter, list):
+        expanded_parameter = parameter
+    return expanded_parameter
+
+
+def _validated_data_attributes(attributes, data_attributes, parameter_name):
```
PGijsbers, line 345: please add type hints

openml/datasets/functions.py
```diff
@@ -333,6 +333,28 @@ def _load_features_from_file(features_file: str) -> Dict:
...
+    if attributes is not None:
```
PGijsbers, line 346: looking at the flow, it looks like `attributes` is never `None` since that is not a possible return value for `_expand_parameter`.

openml/datasets/functions.py
```diff
@@ -333,6 +333,28 @@ def _load_features_from_file(features_file: str) -> Dict:
...
+        for attribute_ in attributes:
+            is_row_id_an_attribute = any([attr[0] == attribute_ for attr in data_attributes])
```
PGijsbers, line 348: The variable name does not make a lot of sense. It just indicates of the attribute is actually an attribute of the dataset, right?

tests/test_datasets/test_dataset_functions.py
```diff
@@ -876,6 +876,94 @@ def test_get_online_dataset_format(self):
...
+                ignore_attribute=None,
+                citation=citation,
+                attributes="auto",
+                data=df,
+                version_label="test",
+                original_data_url=original_data_url,
+                paper_url=paper_url,
+            )
+
+    def test_create_dataset_ignore_attribute_validation(self):
```
PGijsbers, line 923: also a test case for row_id_attribute?

tests/test_datasets/test_dataset_functions.py
```diff
@@ -876,6 +876,94 @@ def test_get_online_dataset_format(self):
             "The format of the ARFF files is different",
         )
 
+    def test_create_dataset_default_target_attribute_validation(self):
```
PGijsbers, line 879: I know we haven't used this functionality before in this project, but would you be able to refactor these similar tests using `pytest.mark.parametrize`? It allows you to easily change only a few variables with otherwise same test setup.

tests/test_datasets/test_dataset_functions.py
```diff
@@ -909,7 +996,7 @@ def test_create_dataset_pandas(self):
             collection_date=collection_date,
             language=language,
             licence=licence,
-            default_target_attribute=default_target_attribute,
+            default_target_attribute="play",
```
PGijsbers, line 999: Why did you change these tests?

PGijsbers (changes requested):
I left some questions and feedback that I'd like to see addressed 👍

tests/test_datasets/test_dataset_functions.py
a-moadel: This one will fail because of the new logic , as "play" is not in the dataset columns.

openml/datasets/functions.py
```diff
@@ -333,6 +333,27 @@ def _load_features_from_file(features_file: str) -> Dict:
         return xml_dict["oml:data_features"]
 
 
+def _expand_parameter(parameter: Union[str, list]):
```
PGijsbers, line 336:
```suggestion
def _expand_parameter(parameter: Union[str, List]) ->List[str]:
```

- `List` should be used as a type hint instead of `list` (`list` is only supported for type hints in Py3.9+)
- type hints include a return value (`-> None` of no return value)

openml/datasets/functions.py
```diff
@@ -333,6 +333,27 @@ def _load_features_from_file(features_file: str) -> Dict:
...
+    expanded_parameter = []
+    if isinstance(parameter, str):
+        expanded_parameter = [x.strip() for x in parameter.split(",")]
+    elif isinstance(parameter, list):
+        expanded_parameter = parameter
+    return expanded_parameter
+
+
+def _validated_data_attributes(attributes: list, data_attributes: list, parameter_name: str):
```
PGijsbers, line 345: So here it should also say `-> None`. I am not adding it as a commit suggestion here, because I think it would violate the line length constraint we have. So best to do this manually and then have the `pre-commit` hook re-format if necessary.

tests/test_datasets/test_dataset_functions.py
PGijsbers: But previously it was also set to `"play"`? And it says `column_names = ["rnd_str", "outlook", "temperature", "humidity", "windy", "play"]` just a few lines up. I am a bit confused about this. Maybe sitting together for this works better.

PGijsbers:
I forgot about this earlier, could you also add a test which tests the new functionality that allows a number of attributes to be specified by a comma-separated string? (e.g. `ignore_attribute="windy,outlook"`

PGijsbers (approved):

PGijsbers:
Thanks 🎉

PR #980, opened 2020-10-28
Improving the performance of check_datasets_active, modifying unit test

### LINKED ISSUE #671: Checking active datasets
https://github.com/openml/openml-python/blob/3dc6dee62c1835e9b4f2bc9b6f9208d902263606/openml/datasets/functions.py#L295-L296

Maybe we should `list_datasets(status='active')` and if a `dataset_id` from the iterable of `dataset_ids` that we want to check is in the results, we can mark it as True, otherwise False.

ArlindKadra:
#### Reference Issue
Fixes #671 

Improves the performance of the function `check_datasets_active` by querying only for the given datasets, instead of all datasets.  Furthemore, the error throwing behavior can be controller by the user.

Existing unit test which was modified.

### DIFF
--- a/doc/progress.rst
+++ b/doc/progress.rst
@@ -11,6 +11,7 @@ Changelog
 
 0.11.0
 ~~~~~~
+* MAINT #671: Improved the performance of ``check_datasets_active`` by only querying the given list of datasets in contrast to querying all datasets. Modified the corresponding unit test.
 * ADD #753: Allows uploading custom flows to OpenML via OpenML-Python.
 * ADD #777: Allows running a flow on pandas dataframes (in addition to numpy arrays).
 * ADD #888: Allow passing a `task_id` to `run_model_on_task`.
--- a/doc/progress.rst
+++ b/doc/progress.rst
@@ -8,10 +8,10 @@ Changelog
 
 0.11.1
 ~~~~~~
+* MAINT #671: Improved the performance of ``check_datasets_active`` by only querying the given list of datasets in contrast to querying all datasets. Modified the corresponding unit test.
 
 0.11.0
 ~~~~~~
-* MAINT #671: Improved the performance of ``check_datasets_active`` by only querying the given list of datasets in contrast to querying all datasets. Modified the corresponding unit test.
 * ADD #753: Allows uploading custom flows to OpenML via OpenML-Python.
 * ADD #777: Allows running a flow on pandas dataframes (in addition to numpy arrays).
 * ADD #888: Allow passing a `task_id` to `run_model_on_task`.
--- a/openml/datasets/functions.py
+++ b/openml/datasets/functions.py
@@ -333,27 +333,40 @@ def _load_features_from_file(features_file: str) -> Dict:
         return xml_dict["oml:data_features"]
 
 
-def check_datasets_active(dataset_ids: List[int]) -> Dict[int, bool]:
+def check_datasets_active(
+    dataset_ids: List[int],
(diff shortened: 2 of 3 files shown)


doc/progress.rst
```diff
@@ -11,6 +11,7 @@ Changelog
 
 0.11.0
 ~~~~~~
+* MAINT #671: Improved the performance of ``check_datasets_active`` by only querying the given list of datasets in contrast to querying all datasets. Modified the corresponding unit test.
```
PGijsbers, line 14: Should be under `0.11.1`

openml/datasets/functions.py
```diff
@@ -333,27 +333,40 @@ def _load_features_from_file(features_file: str) -> Dict:
...
     Check if the dataset ids provided are active.
 
+    Raises an error if a dataset_id in the given list
+    of dataset_ids does not exist on the server.
+
     Parameters
     ----------
     dataset_ids : List[int]
         A list of integers representing dataset ids.
+    raise_error_if_not_exist : bool, optional (default=True)
```
PGijsbers, line 350:
```suggestion
    raise_error_if_not_exist : bool (default=True)
```

It's not `optional`, `optional` means it can take a `None` value (as opposed to having a default value).

PGijsbers (changes requested):
Looks good 👍  only two minor remarks

openml/datasets/functions.py
```diff
@@ -333,27 +333,40 @@ def _load_features_from_file(features_file: str) -> Dict:
...
+        given dataset ids do not exist on the server.
 
     Returns
     -------
     dict
         A dictionary with items {did: bool}
     """
-    dataset_list = list_datasets(status="all")
+    dataset_list = list_datasets(
+        dataset_ids=dataset_ids,
```
PGijsbers, line 360:
```suggestion
        data_id=dataset_ids,
```

PGijsbers: Actually this is a rather important one that hadn't caught my eye first time around 😅

openml/datasets/functions.py
ArlindKadra: @PGijsbers I also thought the same. However, going up I noticed this:
https://github.com/openml/openml-python/blob/e84cdf928b8de713ce0cc2a528f9f675025b783e/openml/datasets/functions.py#L207
So I tried to be consistent, I will change it.

doc/progress.rst
ArlindKadra: oki.

openml/datasets/functions.py
ArlindKadra: Strange, I got a suggestion for that and the function had no argument like that at all :S. Nice catch.

openml/datasets/functions.py
PGijsbers: You'd be correct to point out that annotation is wrong too :)

openml/datasets/functions.py
PGijsbers: like because of the `**kwargs` (used to allow all filters)

PGijsbers (changes requested):
LGTM (pending CI) 👍

PGijsbers (approved):

openml/datasets/functions.py
ArlindKadra: I will post an issue regarding that, since I also observed a function that had no documentation at the module.

mfeurer (approved):

PR #981, opened 2020-10-29
Added unit test to verify how the dataset object handles comparisons

### LINKED ISSUE #577: Unittest Dataset Equality
#### Description
Add a unit test for dataset equality. Something which was changed in #540.

ArlindKadra:
#### Reference Issue
Fixes #577 
Adds unit test to verify the implemented comparison for the dataset object.
With the newly added unit test.

### DIFF
--- a/tests/test_datasets/test_dataset.py
+++ b/tests/test_datasets/test_dataset.py
@@ -219,6 +219,12 @@ class OpenMLDatasetTest(TestBase):
         self.assertIsInstance(xy, pd.DataFrame)
         self.assertEqual(xy.shape, (150, 5))
 
+    def test_equality_comparison(self):
+        # verify the implemented comparison of the dataset object
+        self.assertTrue(self.iris == self.iris)
+        self.assertFalse(self.iris == self.titanic)
+        self.assertFalse(self.titanic == 'Wrong_object')
+
 
 class OpenMLDatasetTestOnTestServer(TestBase):
     def setUp(self):


tests/test_datasets/test_dataset.py
```diff
@@ -219,6 +219,12 @@ def test_get_data_corrupt_pickle(self):
         self.assertIsInstance(xy, pd.DataFrame)
         self.assertEqual(xy.shape, (150, 5))
 
+    def test_equality_comparison(self):
+        # verify the implemented comparison of the dataset object
```
PGijsbers, line 223: Typically comments should explain the *why* not *what*. Just from reading the code it's easy to tell what this test is testing. (also the *why* is obvious, so I think it would have been better to leave the comment out altogether)

PGijsbers (dismissed):
👍

PGijsbers:
I'll wait for Appveyor to complete. Github Actions seems to fail due to too high workload on the server. I will throttle the number of parallel jobs further in the next PR :/

ArlindKadra:
> I'll wait for Appveyor to complete. Github Actions seems to fail due to too high workload on the server. I will throttle the number of parallel jobs further in the next PR :/

Yes, I also noticed that the Github Actions were failing, however, the tests that failed were not affected by the unit test added with what I saw.  Let us wait for Appveyor.

mfeurer:
I just had a look at the attributes of the dataset, and while the test obviously works, I'm wondering whether a few other attributes should also be excluded from checking, and how we would best test that?

The attributes would be `_dataset`, `data_pickle_file`, `data_feather_file`, `feather_attribute_file`. Maybe one should automatically change one attribute at a time and check that the equals comparison still works as expected?

tests/test_datasets/test_dataset.py
```diff
@@ -219,6 +219,11 @@ def test_get_data_corrupt_pickle(self):
         self.assertIsInstance(xy, pd.DataFrame)
         self.assertEqual(xy.shape, (150, 5))
 
+    def test_equality_comparison(self):
+        self.assertTrue(self.iris == self.iris)
```
mfeurer, line 223: Is there a reason to not go for `self.assertEqual(self.iris, self.iris)`?

mfeurer (changes requested):

tests/test_datasets/test_dataset.py
ArlindKadra: No, it was just the first thing that came to my mind. I changed it as you suggested.

openml/datasets/dataset.py
```diff
@@ -297,6 +297,10 @@ def __eq__(self, other):
             "url",
             "dataset",
```
mfeurer, line 298: Could you please remove this line? There's no attribute `dataset` on the `OpenMLDataset`.

ArlindKadra: Yep, removed what was previously there. I guess someone mistook it for `_dataset`.

openml/datasets/dataset.py
```diff
@@ -295,8 +295,11 @@ def __eq__(self, other):
             "version",
             "upload_date",
             "url",
-            "dataset",
             "data_file",
+            "_dataset",
+            "data_pickle_file",
+            "data_feather_file",
+            "feather_attribute_file",
```
PGijsbers, line 302: I'd like to see these defined as a separate set (e.g. `cache_fields`), because these are not server fields.

PGijsbers (changes requested):

PR #982, opened 2020-10-29
Adding helper functions to support ColumnTransformer

Neeratyoy:
#### Reference Issue

Addresses #773. 

PR #777 made the important fixes, while this PR adds implicit support for the usage of sklearn ColumnTransformer, which is essential to handle mixed data types under the new DataFrame default of OpenML.

```python
import openml
from openml.extensions.sklearn import cat, cont

task = openml.tasks.get_task(23)
X, y = task.get_X_and_y('dataframe')

print(cat(X))
print(cont(X))
```


openml/extensions/sklearn/__init__.py
```diff
@@ -7,3 +7,19 @@
 __all__ = ["SklearnExtension"]
 
 register_extension(SklearnExtension)
+
+
+def cont(X):
+    """Returns True for all non-categorical columns, False for the rest.
```
mfeurer, line 13: Could you please extend the docstring to say why this function exists?

openml/extensions/sklearn/__init__.py
```diff
@@ -7,3 +7,19 @@
...
+    """
+    if not hasattr(X, "dtypes"):
+        raise AttributeError("Not a Pandas DataFrame with 'dtypes' as attribute!")
+    return X.dtypes != "category"
+
+
+def cat(X):
+    """Returns True for all categorical columns, False for the rest.
```
mfeurer, line 21: Same here.

mfeurer (commented):
This looks good to me. As this PR would close #773, maybe we should add a unit test that actually checks that the same sklearn model on two different datasets will result in the same flow?

openml/extensions/sklearn/__init__.py
Neeratyoy: Is it adequate now?

PGijsbers:
So to be clear, this PR adds no new functionality, but just makes it easier to use the column transformer with a callable hyperparameter (which ensures the flow is serialized in the same way regardless of dataset)?

Neeratyoy:
> So to be clear, this PR adds no new functionality, but just makes it easier to use the column transformer with a callable hyperparameter (which ensures the flow is serialized in the same way regardless of dataset)?

Yes, and also now we can have complicated pipelines even under examples, by importing these helper functions directly from `openml.extensions` instead of `openml.testing` or declaring it every time. Hence, this PR has a few cosmetic changes for the documentation examples too.

PGijsbers:
Considering the primary focus of this code is for examples, and `cat` and `cont` are not scikit-learn specific, should we consider having these helper functions elsewhere? e.g. `openml.utils`?

mfeurer:
> Considering the primary focus of this code is for examples, and cat and cont are not scikit-learn specific, should we consider having these helper functions elsewhere

I think they are pretty scikit-learn specific in the sense that they are for `sklearn.compose.ColumnTransformer`.

openml/extensions/sklearn/__init__.py
```diff
@@ -7,3 +7,31 @@
 __all__ = ["SklearnExtension"]
 
 register_extension(SklearnExtension)
+
+
+def cont(X):
+    """Returns True for all non-categorical columns, False for the rest.
+
+    This function is required to work with default OpenML datasets as DataFrames allowing
```
mfeurer, line 15:
```suggestion
    This is a helper function for OpenML datasets encoded as DataFrames simplifying the handling of
```

mfeurer: And the same in the other function, too.

openml/extensions/sklearn/__init__.py
mfeurer: Yes, it is!

tests/test_extensions/test_sklearn_extension/test_sklearn_extension.py
```diff
@@ -2215,3 +2206,37 @@ def cat(X):
...
+            # make pipeline
+            clf = SVC(gamma="scale", random_state=1)
+            pipe = make_pipeline(preprocessor, clf)
+            # run task
+            run = openml.runs.run_model_on_task(pipe, task, avoid_duplicate_runs=True)
+            run.publish()
+            new_run = openml.runs.get_run(run.run_id)
+            return new_run.setup_id
+
+        setup1 = column_transformer_pipe(23)
```
mfeurer, line 2239: You chose two numeric-only datasets here, how about using [kr-vs-kp](https://test.openml.org/d/2) as an alternative?

mfeurer (changes requested):
This looks like everything works now :)

However, it would be great if you could extend the following example https://openml.github.io/openml-python/master/examples/30_extended/flows_and_runs_tutorial.html#train-machine-learning-models and demonstrate that manually passing the indicators is not necessary.

mfeurer (changes requested):
Looks good, but did you already update the example?

Neeratyoy:
> Looks good, but did you already update the example?

yeah I pushed the example now

mfeurer (approved):

tests/test_extensions/test_sklearn_extension/test_sklearn_extension.py
```diff
@@ -2215,3 +2206,37 @@ def cat(X):
...
+                transformers=[
+                    ("num", StandardScaler(), cont),
+                    ("cat", OneHotEncoder(handle_unknown="ignore"), cat),
+                ]
+            )
+            # make pipeline
+            clf = SVC(gamma="scale", random_state=1)
+            pipe = make_pipeline(preprocessor, clf)
+            # run task
+            run = openml.runs.run_model_on_task(pipe, task, avoid_duplicate_runs=True)
```
PGijsbers, line 2234: This seems to fail on several jobs (but strangely not all-but-one, perhaps due to race conditions?). And shouldn't that be correct? 
There is no part of this pipeline which has dynamic setup, which means after the first run upload [isn't it correct to raise an error](https://github.com/openml/openml-python/pull/982/checks?check_run_id=1347584596#step:8:883)?

Neeratyoy: That's a good point that I missed. 
I think for now the correct approach might be to mark the run/setup to me removed using the `_mark_entity_for_removal` feature
I'll push with that change and see the tests and then ask for a review if it passes!

Neeratyoy: Also, @mfeurer is cleaning the test servers, and he recommended waiting this week out before judging if this error is a result of those changes. Nevertheless, I made the push with the change that should be there in any case.

PGijsbers: Okay we'll wait.

>  if this error is a result of those changes. 

I imagine the times it _did_ pass are those where a previous upload had been deleted by the clean up script in between.

PGijsbers:
I got pinged as review requested. Do you think we should merge while there are still this many test server issues? If so, go ahead, but if you think those can be resolved first, I would prefer to wait.

tests/test_extensions/test_sklearn_extension/test_sklearn_extension.py
mfeurer: I just had a look at this PR and it appears that this error and an error building the docs are the only things that hold back merging this PR? @Neeratyoy could you please have a look into this?

PR #984, opened 2020-10-29
Retry on database error to reduce number of test failures

mfeurer:

This PR adds a retry mechanism to api calls in case it detects a database error.

New unit test. Also, an overloaded test server.

### DIFF
--- a/openml/_api_calls.py
+++ b/openml/_api_calls.py
@@ -188,28 +188,26 @@ def _send_request(
                     response = session.post(url, data=data, files=files)
                 else:
                     raise NotImplementedError()
-                pass
-            except (requests.exceptions.ConnectionError, requests.exceptions.SSLError,) as e:
+                __check_response(response=response, url=url, file_elements=files)
+                break
+            except (
+                requests.exceptions.ConnectionError,
+                requests.exceptions.SSLError,
+                OpenMLServerException,
+            ) as e:
+                if isinstance(e, OpenMLServerException):
+                    if e.code != 107:
+                        # 107 is a database connection error - only then do retries
+                        raise
+                    else:
+                        wait_time = 0.3
+                else:
+                    wait_time = 0.1
                 if i == n_retries:
                     raise e
                 else:
-                    time.sleep(0.1 * i)
+                    time.sleep(wait_time * i)
                     continue
-
-            try:
-                __check_response(response=response, url=url, file_elements=files)
-            except OpenMLServerException as e:
-                # Do retries in case of a database connection error
-                if e.code == 107:
-                    if i == n_retries:
-                        raise e
-                    else:
-                        time.sleep(i * 0.3)
-                        continue
-                else:
-                    raise e
-
-            break
     if response is None:
         raise ValueError("This should never happen!")
     return response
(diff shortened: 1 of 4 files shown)


openml/_api_calls.py
```diff
@@ -198,17 +188,42 @@ def __send_request(
                     response = session.post(url, data=data, files=files)
                 else:
                     raise NotImplementedError()
-                break
+                pass
```
PGijsbers, line 191: `pass` has no function here

openml/_api_calls.py
```diff
@@ -198,17 +188,42 @@ def __send_request(
...
+            except OpenMLServerException as e:
+                # Do retries in case of a database connection error
+                if e.code == 107:
+                    if i == n_retries:
+                        raise e
+                    else:
+                        time.sleep(i * 0.3)
+                        continue
+                else:
+                    raise e
```
PGijsbers, line 210: Maybe simplify the flow by calling `check_response` where you currently have the `pass` statement. You can then catch the `OpenMLServerException` in the same `except` clause and use shared logic on whether or not an error should be raised or a new connection attempt should be made.

PGijsbers (changes requested):
I think it makes sense to treat a connection error the same as a `server is busy` is error. I have a minor suggestion for making sure the code reflects that. Other than that I agree with the PR.

openml/_api_calls.py
mfeurer: You mean like this?
```
            try:
                if request_method == "get":
                    response = session.get(url, params=data)
                elif request_method == "delete":
                    response = session.delete(url, params=data)
                elif request_method == "post":
                    response = session.post(url, data=data, files=files)
                else:
                    raise NotImplementedError()
                break
            except (
                requests.exceptions.ConnectionError, 
                requests.exceptions.SSLError,
                OpenMLServerException,
            ) as e:
                if isinstance(e, OpenMLServerException) and e.code != 107:
                    # 107 is a database connection error - only then do retries
                    raise e
                if i == n_retries:
                    raise e
                else:
                    time.sleep(0.1 * i)
                    continue
```
Or a bit more verbose like that:
```
            try:
                if request_method == "get":
                    response = session.get(url, params=data)
                elif request_method == "delete":
                    response = session.delete(url, params=data)
                elif request_method == "post":
                    response = session.post(url, data=data, files=files)
                else:
                    raise NotImplementedError()
                break
            except (
                requests.exceptions.ConnectionError,
                requests.exceptions.SSLError,
            ) as e:
                if i == n_retries:
                    raise e
                else:
                    time.sleep(0.1 * i)
                    continue
            except OpenMLServerException as e:
                # 107 is a database connection error - only then do retries
                if e.code != 107:
                    raise
                if i == n_retries:
                    raise e
                else:
                    time.sleep(0.3 * i)
                    continue
```
In any case, merging the two try/excepts is an improvement to the code.

PGijsbers: Fine with either, but I think I prefer the first one (with a `wait_time` variable if you want to preserve the difference in sleep time).

PGijsbers (approved):
(let's see some green from CI first, but otherwise 👍 )

joaquinvanschoren:
@all-contributors please add @mfeurer for code

PR #985, opened 2020-10-30
Improve unit tests

mfeurer:
* increase wait time when uploading a run to 600s. It's better to wait a very long time instead of failing that test (on CI)
* reduce the warnings a bit to have less output online
* remove deprecated `format` argument to `OpenMLDataset`
* add load balancing for parallel unit tests

### DIFF
--- a/tests/test_datasets/test_dataset_functions.py
+++ b/tests/test_datasets/test_dataset_functions.py
@@ -373,9 +373,9 @@ class TestOpenMLDataset(TestBase):
     def test_get_dataset_uint8_dtype(self):
         dataset = openml.datasets.get_dataset(1)
         self.assertEqual(type(dataset), OpenMLDataset)
-        self.assertEqual(dataset.name, 'anneal')
+        self.assertEqual(dataset.name, "anneal")
         df, _, _, _ = dataset.get_data()
-        self.assertEqual(df['carbon'].dtype, 'uint8')
+        self.assertEqual(df["carbon"].dtype, "uint8")
 
     def test_get_dataset(self):
         # This is the only non-lazy load to ensure default behaviour works.
@@ -1157,7 +1157,7 @@ class TestOpenMLDataset(TestBase):
         downloaded_dataset = None
         # fetching from server
         # loop till timeout or fetch not successful
-        max_waiting_time_seconds = 400
+        max_waiting_time_seconds = 600
         # time.time() works in seconds
         start_time = time.time()
         while time.time() - start_time < max_waiting_time_seconds:
--- a/tests/test_datasets/test_dataset_functions.py
+++ b/tests/test_datasets/test_dataset_functions.py
@@ -860,7 +860,7 @@ class TestOpenMLDataset(TestBase):
         decoder = arff.ArffDecoder()
         # check if the arff from the dataset is
         # the same as the arff from _get_arff function
-        d_format = (dataset.format).lower()
+        d_format = (dataset.data_format).lower()
 
         self.assertEqual(
             dataset._get_arff(d_format),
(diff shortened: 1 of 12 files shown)


tests/test_runs/test_run_functions.py
```diff
@@ -518,12 +518,14 @@ def _run_and_upload_regression(
...
 
-    def test_run_and_upload_logistic_regression(self):
+    @unittest.mock.patch("warnings.warn")
+    def test_run_and_upload_logistic_regression(self, warn_mock):
         lr = LogisticRegression(solver="lbfgs")
         task_id = self.TEST_SERVER_TASK_SIMPLE[0]
         n_missing_vals = self.TEST_SERVER_TASK_SIMPLE[1]
         n_test_obs = self.TEST_SERVER_TASK_SIMPLE[2]
         self._run_and_upload_classification(lr, task_id, n_missing_vals, n_test_obs, "62501")
+        self.assertLessEqual(warn_mock.call_count, 3)
```
PGijsbers, line 528: Please explain why we expect three warnings.

tests/test_runs/test_run_functions.py
```diff
@@ -635,11 +638,13 @@ def test_run_and_upload_knn_pipeline(self):
         n_missing_vals = self.TEST_SERVER_TASK_MISSING_VALS[1]
         n_test_obs = self.TEST_SERVER_TASK_MISSING_VALS[2]
         self._run_and_upload_classification(pipeline2, task_id, n_missing_vals, n_test_obs, "62501")
+        self.assertEqual(warnings_mock.call_count, 3)
```
PGijsbers, line 641: Here too.

tests/test_runs/test_run_functions.py
```diff
@@ -635,11 +638,13 @@ def test_run_and_upload_knn_pipeline(self):
...
 
     def test_run_and_upload_gridsearch(self):
         gridsearch = GridSearchCV(
             BaggingClassifier(base_estimator=SVC()),
             {"base_estimator__C": [0.01, 0.1, 10], "base_estimator__gamma": [0.01, 0.1, 10]},
+            cv=3,
```
PGijsbers, line 647: Is this to reduce load? Is there a reason why you don't also reduce the dimensions (e.g. to 2x2 grid)?

PGijsbers (changes requested):
Overall looks good, just would like some minor clarifications.

PGijsbers (changes requested):
Looks good overall, just expecting some minor explanations.

tests/test_datasets/test_dataset_functions.py
```diff
@@ -1370,12 +1375,28 @@ def test_data_edit(self):
 
         # Case 2
         # only owners (or admin) can edit all critical fields of datasets
-        # this is a dataset created by CI, so it is editable by this test
-        did = 315
-        result = edit_dataset(did, default_target_attribute="col_1", ignore_attribute="col_2")
+        # for this, we need to first clone a dataset to do changes
+        did = fork_dataset(1)
```
PGijsbers, line 1379: It would probably also make sense to test each case in a separate unit test? I don't really see a reason to test both scenarios in the same test, and it would help both with separating concerns and load balancing.

tests/test_runs/test_run_functions.py
mfeurer: Honestly, I don't understand the 3 myself and would have expected 1. I will simply increase the max_iter so this test passes.

tests/test_runs/test_run_functions.py
mfeurer: No, this is to make sklearn stop printing a warning that the default value will change in the future.

tests/test_runs/test_run_functions.py
mfeurer: Ah, I now understand. We call run_model_on_task once, and then twice a helper function called `_rerun_model_and_compare_predictions`. Anyway, I think increasing the number iterations is a bit cleaner.

tests/test_datasets/test_dataset_functions.py
mfeurer: Good catch, will change this.

PGijsbers (approved):
LGTM 👍

PR #987, opened 2020-10-30
Rework local openml directory

### LINKED ISSUE #883: OpenML package fails to import if home directory is not writable
#### Description

It looks like OpenML package is trying to write into an user home directory when it is imported. I think this is really not nice. Just importing a package should not have such a side effect. You should write only if there is a request for that.

This broke [D3M](https://gitlab.com/datadrivendiscovery/primitives/-/jobs/350149345) build process now that we added the OpenML package as a dependency to [d3m](https://gitlab.com/datadrivendiscovery/d3m) core package. :-(

#### Steps/Code to Reproduce

Create an user on your system which has `/nonexitent` as its home directory. That directory should not exist. Try `import openml` when running under such user.

#### Expected Results

Package loads.

#### Actual Results

It fails with:

```
File "/usr/local/lib/python3.6/dist-packages/openml/__init__.py", line 18, in <module>
  from . import _api_calls
File "/usr/local/lib/python3.6/dist-packages/openml/_api_calls.py", line 8, in <module>
  from . import config
File "/usr/local/lib/python3.6/dist-packages/openml/config.py", line 241, in <module>
  _setup()
File "/usr/local/lib/python3.6/dist-packages/openml/config.py", line 133, in _setup
  os.mkdir(os.path.expanduser(os.path.join('~', '.openml')))
FileNotFoundError: [Errno 2] No such file or directory: '/nonexistent/.openml'
```

#### Versions

```
>>> import platform; print(platform.platform())
Linux-5.0.0-32-generic-x86_64-with-Ubuntu-18.04-bionic
>>> import sys; print("Python", sys.version)
Python 3.6.8 (default, Oct  7 2019, 12:59:55) 
[GCC 8.3.0]
>>> import numpy; print("NumPy", numpy.__version__)
NumPy 1.17.3
>>> import scipy; print("SciPy", scipy.__version__)
SciPy 1.3.1
>>> import sklearn; print("Scikit-Learn", sklearn.__version__)
Scikit-Learn 0.21.3
>>> import openml; print("OpenML", openml.__version__)
OpenML 0.10.1
```

### LINKED ISSUE #884: Use standard Linux locations for config and cache
Currently OpenML package creates `.openml` in home directory. These days this is not a nice behavior, because [there is a standard where things should go](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html).

* Use `XDG_CONFIG_HOME` environment variable to determine base config directory.
* Use `XDG_CACHE_HOME` environment variable to determine base cache directory.

See [more practical information and details here](https://wiki.archlinux.org/index.php/XDG_Base_Directory).

So, by default, config should go into `~/.config/openml` and cache into `~/.cache/openml`.

### LINKED ISSUE #906: Package crashes if there is no permission to create the cache directory.
When trying to set the cache directory to a directory which exists, but where the directory it resides in does not allow the creation of a new directory, the package erroneously crashes:
```
  File "01_create_commands.py", line 6, in <module>
    from update_metadata_util import classification_tasks, regression_tasks
  File "/tmp/auto-sklearn/scripts/update_metadata_util.py", line 2, in <module>
    import openml
  File "/home/feurerm/miniconda/3-4.5.4/envs/autosklearnmetadata/lib/python3.7/site-packages/openml/__init__.py", line 20, in <module>
    from . import _api_calls
  File "/home/feurerm/miniconda/3-4.5.4/envs/autosklearnmetadata/lib/python3.7/site-packages/openml/_api_calls.py", line 10, in <module>
    from . import config
  File "/home/feurerm/miniconda/3-4.5.4/envs/autosklearnmetadata/lib/python3.7/site-packages/openml/config.py", line 244, in <module>
    _setup()
  File "/home/feurerm/miniconda/3-4.5.4/envs/autosklearnmetadata/lib/python3.7/site-packages/openml/config.py", line 150, in _setup
    os.mkdir(cache_directory)
PermissionError: [Errno 13] Permission denied: '/data/aad/openml/'
```
The directory is:
```
ls -al /data/aad
total 50127952
drwxr-sr-x  27 other     other        4096 Mär 10 10:58 .
drwxr-xr-x   3 root    root               0 Apr  9 09:42 ..
drwxr-xr-x   8 feurerm other        4096 Apr  8 13:24 openml
```

### LINKED ISSUE #972: Give a more informative error message if cache directory can not be created
If your configuration specifies a cache directory which can not be created, the following error is raised:

```
(jupyter3) me-2:datanb me$ python
Python 3.6.5 (default, Apr 25 2018, 14:23:58)
[GCC 4.2.1 Compatible Apple LLVM 9.1.0 (clang-902.0.39.1)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import openml
  File "/Users/me/venv/jupyter3/lib/python3.6/site-packages/openml/__init__.py", line 20, in <module>
    from . import _api_calls
  File "/Users/me/venv/jupyter3/lib/python3.6/site-packages/openml/_api_calls.py", line 10, in <module>
    from . import config
  File "/Users/me/venv/jupyter3/lib/python3.6/site-packages/openml/config.py", line 295, in <module>
    _setup()
  File "/Users/me/venv/jupyter3/lib/python3.6/site-packages/openml/config.py", line 203, in _setup
    os.mkdir(cache_directory)
FileNotFoundError: [Errno 2] No such file or directory: '/var/folders/gj/cm0k4b_s42j30zs376cq_5hh0000gn/T//RtmpFJlUoh/cache'
```

It would be better to have a more informative error message that indicates the cache directory can not be created and where to find the config file to change the cache directory.

mfeurer:
Closes #883 #884 #906 #972

* Fixes #883: package can be imported if the home directory is not writable.
* Fixes #884: use standard linux locations for config and cache
* Fixes #906: package can be imported if there is no permission to create a cache directory.
* Fixes #972: give a more useful error message if a cache directory cannot be created

Unit tests.

#### Any other comments?

@mitar does this resolve the issues you had with the package?

@phmueller does this help you using OpenML inside HPOlib?

### DIFF
--- a/openml/config.py
+++ b/openml/config.py
@@ -9,7 +9,6 @@ import logging.handlers
 import os
 from pathlib import Path
 from typing import Tuple, cast
-import warnings
 
 from io import StringIO
 import configparser
@@ -21,7 +20,7 @@ console_handler = None
 file_handler = None
 
 
-def _create_log_handlers():
+def _create_log_handlers(create_file_handler=True):
     """ Creates but does not attach the log handlers. """
     global console_handler, file_handler
     if console_handler is not None or file_handler is not None:
@@ -34,12 +33,13 @@ def _create_log_handlers():
     console_handler = logging.StreamHandler()
     console_handler.setFormatter(output_formatter)
 
-    one_mb = 2 ** 20
-    log_path = os.path.join(cache_directory, "openml_python.log")
-    file_handler = logging.handlers.RotatingFileHandler(
-        log_path, maxBytes=one_mb, backupCount=1, delay=True
-    )
-    file_handler.setFormatter(output_formatter)
+    if create_file_handler:
+        one_mb = 2 ** 20
+        log_path = os.path.join(cache_directory, "openml_python.log")
+        file_handler = logging.handlers.RotatingFileHandler(
+            log_path, maxBytes=one_mb, backupCount=1, delay=True
+        )
+        file_handler.setFormatter(output_formatter)
 
 
 def _convert_log_levels(log_level: int) -> Tuple[int, int]:
@@ -197,12 +197,20 @@ def _setup():
     if not os.path.exists(expanded_openml_dir):
         try:
             os.mkdir(expanded_openml_dir)
+            cache_exists = True
         except PermissionError:
-            warnings.warn(
-                "No permission to create openml directory at %s! This can result in OpenML-Python "
-                "not working properly." % expanded_openml_dir
-            )
-            pass
+            cache_exists = False
+    else:
+        cache_exists = True
+
+    if cache_exists:
+        _create_log_handlers()
+    else:
+        _create_log_handlers(create_file_handler=False)
+        openml_logger.warning(
+            "No permission to create openml directory at %s! This can result in OpenML-Python "
+            "not working properly." % expanded_openml_dir
+        )
 
     config = _parse_config()
     apikey = config.get("FAKE_SECTION", "apikey")
@@ -216,14 +224,13 @@ def _setup():
         try:
             os.mkdir(cache_directory)
         except PermissionError:
-            warnings.warn(
+            openml_logger.warning(
                 "No permission to create openml cache directory at %s! This can result in "
                 "OpenML-Python not working properly." % cache_directory
             )
-            pass
 
     avoid_duplicate_runs = config.getboolean("FAKE_SECTION", "avoid_duplicate_runs")
-    connection_n_retries = config.get("FAKE_SECTION", "connection_n_retries")
+    connection_n_retries = int(config.get("FAKE_SECTION", "connection_n_retries"))
     max_retries = config.get("FAKE_SECTION", "max_retries")
     if connection_n_retries > max_retries:
         raise ValueError(
@@ -300,4 +307,3 @@ __all__ = [
 ]
 
 _setup()
-_create_log_handlers()
--- a/openml/config.py
+++ b/openml/config.py
@@ -122,8 +122,8 @@ cache_directory = str(_defaults["cachedir"])  # so mypy knows it is a string
 avoid_duplicate_runs = True if _defaults["avoid_duplicate_runs"] == "True" else False
 
 # Number of retries if the connection breaks
-connection_n_retries = _defaults["connection_n_retries"]
-max_retries = _defaults["max_retries"]
+connection_n_retries = int(_defaults["connection_n_retries"])
+max_retries = int(_defaults["max_retries"])
 
 
 class ConfigurationForExamples:
--- a/openml/config.py
+++ b/openml/config.py
@@ -7,7 +7,9 @@ Store module level information like the API key, cache directory and the server
 import logging
 import logging.handlers
 import os
+from pathlib import Path
 from typing import Tuple, cast
+import warnings
 
 from io import StringIO
 import configparser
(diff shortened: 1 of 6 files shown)


openml/config.py
```diff
@@ -184,26 +189,35 @@ def _setup():
...
-    except FileExistsError:
-        # For other errors, we want to propagate the error as openml does not work without cache
-        pass
+    # read config file, create openml directory
+    expanded_openml_dir = os.path.expanduser(os.path.join("~", ".openml"))
+    if not os.path.exists(expanded_openml_dir):
+        try:
+            os.mkdir(expanded_openml_dir)
+        except PermissionError:
+            warnings.warn(
```
mitar, line 198: I would prefer using `logging` module here.

openml/utils.py
```diff
@@ -276,9 +276,11 @@ def _create_cache_directory(key):
     cache = config.get_cache_directory()
     cache_dir = os.path.join(cache, key)
     try:
-        os.makedirs(cache_dir)
-    except OSError:
-        pass
+        os.makedirs(cache_dir, exist_ok=True)
+    except Exception as e:
+        raise openml.exceptions.OpenMLCacheException(
+            f"Cannot create cache directory {cache_dir} due to exception {e}"
```
mitar, line 282: You do not need to have `due to exception {e}` because you have `from e` later on.

mitar:
Not fully. My main issue is that OpenML initializes stuff at package import and not on first use. So people who would not even use OpenML codepath in our package (which depends on `openml`) get an exception (or with this PR a warning).

I do not think you should be calling in `config.py`:

```
_setup()
_create_log_handlers()
```

You should call those at first call which requires any of those. You should also not be configuring logging without requesting. See:

```
log_path = os.path.join(cache_directory, "openml_python.log")
```

If cache directory is not writable, this will fail, no? And it is called from `_create_log_handlers()`, always, for everyone.

mitar:
For reference, current patch we are using: https://gitlab.com/datadrivendiscovery/images/-/blob/master/libs/patches/openml.patch

PhMueller:
@mfeurer: Thanks for the notification. Looks fine by me.

joaquinvanschoren:
@mfeurer Would you have time to look at this or should we ask someone else?

openml/config.py
mfeurer: I updated the code to use logging instead of warnings.

openml/utils.py
mfeurer: Yes, makes sense.

mfeurer:
Alright, thanks for your suggestions @mitar. I tried to incorporate them as much as possible, but I'm afraid I didn't do all you asked for.

> Not fully. My main issue is that OpenML initializes stuff at package import and not on first use. So people who would not even use OpenML codepath in our package (which depends on openml) get an exception (or with this PR a warning).

I agree that this might be nicer. On the other hand, I think the current solution is more in the spirit of 'fail fast' and gives a warning as quickly as possible. In any case, I currently don't have the time to work on this, wouldn't know how to achieves this and think the changes in this PR cover most use cases and issues and will for now leave it is at is (unless there's an issue with my recent changes).

> If cache directory is not writable, this will fail, no? And it is called from _create_log_handlers(), always, for everyone.

I made this conditional on the cache directory being writable, so there shouldn't be any failure from this.

Tests might fail because the server is very flaky at the moment, so we probably have to rebase this on #1000 and rerun.

mfeurer:
@PGijsbers do you think you could have a look at the Windows failures? I believe it'll be much easier to debug for you on a Windows machine than for me by pushing to github

mitar (approved):
Thanks. This looks great now.

PR #988, opened 2020-10-30
Transition other Travis jobs to Github Actions

PGijsbers:
I think this covers the previous work that was done by Travis jobs:
 - build docs, deploy docs (when on a push to `develop`  and `master`)
 - check the distribution
 - run the unit tests, check no file left behind
 - report coverage (accidentally disabled before the hackathon)
 - pre-commit check
 - PEP 561 compliance check

I opted to split workflows where Travis branched merely on conditionals. I think this makes each individual workflow easier to understand. There is probably some improvement to be made by using predefined Github Actions, but I think this is good enough for me right now. I would rather focus on setting up Github Actions for our Windows testing.


.github/workflows/docs.yaml
```diff
@@ -0,0 +1,44 @@
...
+    steps:
+    - uses: actions/checkout@v2
+    - name: Setup Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.8
+    - name: Install dependencies
+      run: |
+        pip install -e .[examples,examples_unix]
+        pip install sphinx-gallery sphinx_bootstrap_theme numpydoc
```
mfeurer, line 16: Could we make these extra requirements for documentation building?

.github/workflows/dist.yaml
```diff
@@ -0,0 +1,30 @@
...
+    - name: Twine check
+      run: |
+        pip install twine
+        last_dist=$(ls -t dist/openml-*.tar.gz | head -n 1)
+        twine check $last_dist
+    - name: Install dist
+      run: |
+        last_dist=$(ls -t dist/openml-*.tar.gz | head -n 1)
+        pip install $last_dist
+    - name: PEP 561 Compliance
```
mfeurer, line 26: I'm not 100% sure if this test is sufficient. We only test if it's installable, but not whether the installed package actually works. Having said that, as we don't compile anything, it's maybe sufficient.

mfeurer (commented):
This looks great, just one minor change request. 

I think it's good to have every test as a separate file, it's much cleaner now like this.

.github/workflows/dist.yaml
PGijsbers: As far as I can tell, unless we're changing something about the structure of the package (i.e. include more than just `openml`), I don't think results should differ. To keep concerns separate and prevent additional load on the server, I decided not to run an additional test suite here. I'd say we wait until we encounter a problem from this before we add it.

mfeurer: Makes sense.

mfeurer (approved):

joaquinvanschoren:
@all-contributors please add @pgijsbers for test

Issue #968, opened 2020-10-23
Raise warning if sklearn model already fitted prior to _run_model_on_task/fold()


Neeratyoy:
#### Description

Give a warning if model is already fitted.
This is acceptable in case of custom experimentation, but not desirable if we want to upload to OpenML. For transparent interpretation of OpenML results, the user should be told of what's happening.

Can be checked for in the `_run_model_on_fold()` function in `openml.extensions.sklearn.extension` prior to calling of `model_copy.fit()`.

Possible solutions:
* https://stackoverflow.com/a/39900933/8363967
* https://scikit-learn.org/stable/modules/generated/sklearn.utils.validation.check_is_fitted.html

Neeratyoy:
Based on this line [here](https://github.com/openml/openml-python/blob/develop/openml/extensions/sklearn/extension.py#L1646) and the documentation of [clone](https://scikit-learn.org/stable/modules/generated/sklearn.base.clone.html) which suggests a new estimator _that has not been fitted_ is being created, this issue is rendered moot right?

I have got a branch locally that fixes this issue, but it appears it is just redundant code, so hence will push only if you feel we should still have an explicit check.

@mfeurer @PGijsbers

PGijsbers:
To me it looks like the intention of this message is that if `_run_model_on_fold` is called with a model that is already fit, a warning should be raised *because* the model that is already fit is ignored and a new one is trained instead. This might not be what the user intended when they provided an already fit model to this function. But @mfeurer wrote this message and the code, so I'm interested to hear if I interpreted this correctly 😉

mfeurer:
Sorry, could you please fill me in which message you are talking about? We should add a message which warns if the estimator is already fitted. There are scikit-learn helpers for checking that.

PGijsbers:
Closed by #989

PR #989, opened 2020-10-30
Warning if fitted sklearn model being used

Neeratyoy:
#### Reference Issue

Addresses #968.

Prints a *warning* if the user passes an already fitted sklearn model to `run_model_on_task`.

```python
import openml


task = openml.tasks.get_task(23)
clf = RandomForestClassifier()

run = openml.runs.run_model_on_task(clf, task)
print("Unfit model run")

X, y = task.get_X_and_y()
clf.fit(X, y)
run = openml.runs.run_model_on_task(clf, task)
print("Fit model run")
```


openml/extensions/sklearn/extension.py
```diff
@@ -1641,6 +1641,22 @@ def _prediction_to_probabilities(
             if X_test is None:
                 raise TypeError("argument X_test must not be of type None")
 
+        try:
+            # check if model is fitted
+            # 'predict' internally calls sklearn.utils.validation.check_is_fitted for every
+            # model-specific attribute it excepts, thus offering a more robust check than
```
PGijsbers, line 1647:
```suggestion
            # model-specific attribute it expects, thus offering a more robust check than
```

openml/extensions/sklearn/extension.py
```diff
@@ -1641,6 +1641,22 @@ def _prediction_to_probabilities(
...
+            from sklearn.exceptions import NotFittedError
+
+            model.predict(X_train)
+            warnings.warn(
+                "The model is already fitted!"
+                " This might cause inconsistency in comparison of results."
+            )
+        except NotFittedError:
+            # model is not fitted, as is required
+            pass
```
PGijsbers, line 1658: I thought in the Python call we discussed that perhaps we would check this at the first call to ``run_model_on_task``?
In either case I would extract this to a separate method `_raise_warning_if_fitted` to make sure the functions don't get too big (they already are).

Neeratyoy: Alright, will make it into a function and push.

As for its placement, I reconsidered it given that irrespective of what is called, `run_model_on_task` or `run_flow_on_task`, this function is what the call is reduced to. Hence went ahead with this placement for this snippet of code.

PGijsbers: `run_model_on_task` actually calls `run_flow_on_task`. That said, then we would need to add a function to the extension interface that will indicate if a model is already fit, otherwise we can't check it in a general way.

PGijsbers: @mfeurer do you think this is something we should want? or do we just leave it to the extension devs to implement a warning if they see it fit?

mfeurer: I'd leave it to the extensions devs. For scikit-learn I'd use [this function](https://scikit-learn.org/stable/modules/generated/sklearn.utils.validation.check_is_fitted.html).

mfeurer (changes requested):
This PR looks good now, but you accidentally committed a .orig file. Could you please remove that again?

Neeratyoy:
> This PR looks good now, but you accidentally committed a .orig file. Could you please remove that again?

Done

mfeurer (approved):

mfeurer:
Hey, it's a bit hard to see this but among all the failures it says:
```
self = <openml.extensions.sklearn.extension.SklearnExtension object at 0x0000009C4334DA58>
model = RandomForestClassifier(bootstrap=True, class_weight=None, criterion='gini',
                       max_depth=None, max...                     n_jobs=None, oob_score=False, random_state=42, verbose=0,
                       warm_start=False)
    def check_if_model_fitted(self, model: Any) -> bool:
        """Returns True/False denoting if the model has already been fitted/trained
    
        Parameters
        ----------
        model : Any
    
        Returns
        -------
        bool
        """
        try:
            # check if model is fitted
            from sklearn.exceptions import NotFittedError
            from sklearn.utils.validation import check_is_fitted
    
>           check_is_fitted(model)  # raises a NotFittedError if the model has not been trained
E           TypeError: check_is_fitted() missing 1 required positional argument: 'attributes'
openml\extensions\sklearn\extension.py:1556: TypeError
```
Could you please have a look?

joaquinvanschoren:
@all-contributors please add @Neeratyoy for code

PGijsbers (commented):
Looks good to me, if the failing unit test that @mfeurer pointed out is fixed.

PGijsbers:
@all-contributors please add @Neeratyoy for code

(sorry for the spam)

Neeratyoy:
> Could you please have a look?

Given the error message, it might be that the older versions of sklearn had `check_is_fitted` behaving differently.

I updated the design of the check to be agnostic to the sklearn versions and also to the kind of model passed.

PGijsbers:
As far as I can tell these failures are a combination of timeouts and server issues, am I overlooking something?

mfeurer:
Yes, except for:
* pytest workers crashing. This is due to the timeout of pytest being lower than the unit test running (timeout is set to 600s, if the unit test now waits for 600s, it runs over the timeout)
* a known failure that previously only existed on Windows. I already asked @Neeratyoy to look into this.
In my opinion, this PR can be merged.

PGijsbers:
> pytest workers crashing. 

Yes this is what I referred to with timeouts (not server timeouts), my bad for leaving it ambiguous.

> In my opinion, this PR can be merged.

Will do.

PR #1000, opened 2020-11-10
Making some unit tests work

Neeratyoy:
* Fixes some of the failing unit tests post the test server clean-up.

* Update docs to ensure that any `publish` to the test server during unit testing is also scheduled for deletion.

### DIFF
--- a/openml/_api_calls.py
+++ b/openml/_api_calls.py
@@ -228,7 +228,6 @@ def _send_request(
                     raise
                 else:
                     time.sleep(retry_counter)
-                    continue
     if response is None:
         raise ValueError("This should never happen!")
     return response
--- a/openml/_api_calls.py
+++ b/openml/_api_calls.py
@@ -5,7 +5,7 @@ import hashlib
 import logging
 import requests
 import xmltodict
-from typing import Dict, Optional
+from typing import Dict, Optional, cast
 
 from . import config
 from .exceptions import (
@@ -103,20 +103,32 @@ def _download_text_file(
         except FileNotFoundError:
             pass
 
+    n_retries = cast(int, config.connection_n_retries)
+    wait_time = 0.2
+    raise_error = None
     logging.info("Starting [%s] request for the URL %s", "get", source)
     start = time.time()
-    response = __read_url(source, request_method="get")
-    downloaded_file = response.text
+    for retry in range(n_retries):
+        response = __read_url(source, request_method="get")
+        downloaded_file = response.text
 
-    if md5_checksum is not None:
-        md5 = hashlib.md5()
-        md5.update(downloaded_file.encode("utf-8"))
-        md5_checksum_download = md5.hexdigest()
-        if md5_checksum != md5_checksum_download:
-            raise OpenMLHashException(
-                "Checksum {} of downloaded file is unequal to the expected checksum {} "
-                "when downloading {}.".format(md5_checksum_download, md5_checksum, source)
-            )
+        if md5_checksum is not None:
+            md5 = hashlib.md5()
+            md5.update(downloaded_file.encode("utf-8"))
+            md5_checksum_download = md5.hexdigest()
+            if md5_checksum == md5_checksum_download:
+                raise_error = False
+                break
+            else:
+                raise_error = True
+                time.sleep(wait_time)
+    # raise_error can be set to True only if the variables md5_checksum_download and md5_checksum
+    # were initialized and compared during retries
+    if raise_error:
+        raise OpenMLHashException(
+            "Checksum {} of downloaded file is unequal to the expected checksum {} "
+            "when downloading {}.".format(md5_checksum_download, md5_checksum, source)
+        )
 
     if output_path is None:
         logging.info(
--- a/openml/_api_calls.py
+++ b/openml/_api_calls.py
@@ -175,10 +175,13 @@ def _send_request(
     request_method, url, data, files=None,
 ):
     n_retries = config.connection_n_retries
+    max_retries = 10
+    retry_counter = 0
     response = None
     with requests.Session() as session:
         # Start at one to have a non-zero multiplier for the sleep
-        for i in range(1, n_retries + 1):
+        while retry_counter < n_retries:
+            retry_counter += 1
             try:
                 if request_method == "get":
                     response = session.get(url, params=data)
@@ -198,15 +201,17 @@ def _send_request(
                 if isinstance(e, OpenMLServerException):
                     if e.code != 107:
                         # 107 is a database connection error - only then do retries
-                        raise
+                        raise e
                     else:
                         wait_time = 0.3
+                        # increase retries if database connection error
+                        n_retries = min(n_retries + 1, max_retries)
                 else:
                     wait_time = 0.1
-                if i == n_retries:
+                if retry_counter == n_retries:
                     raise e
                 else:
-                    time.sleep(wait_time * i)
+                    time.sleep(wait_time * retry_counter)
                     continue
     if response is None:
         raise ValueError("This should never happen!")
--- a/openml/_api_calls.py
+++ b/openml/_api_calls.py
@@ -175,7 +175,7 @@ def _send_request(
     request_method, url, data, files=None,
 ):
     n_retries = config.connection_n_retries
-    max_retries = 10
+    max_retries = config.max_retries
     retry_counter = 0
     response = None
     with requests.Session() as session:
@@ -199,13 +199,13 @@ def _send_request(
                 OpenMLServerException,
             ) as e:
                 if isinstance(e, OpenMLServerException):
-                    if e.code != 107:
-                        # 107 is a database connection error - only then do retries
-                        raise e
-                    else:
+                    if e.code in [107, 500]:
+                        # 107: database connection error
+                        # 500: internal server error
                         wait_time = 0.3
-                        # increase retries if database connection error
-                        n_retries = min(n_retries + 1, max_retries)
+                        n_retries = min(n_retries + 1, max_retries)  # increase retries
+                    else:
+                        raise
                 else:
                     wait_time = 0.1
                 if retry_counter == n_retries:
--- a/openml/_api_calls.py
+++ b/openml/_api_calls.py
@@ -6,7 +6,7 @@ import logging
 import requests
 import xml
 import xmltodict
-from typing import Dict, Optional, cast
+from typing import Dict, Optional
 
 from . import config
 from .exceptions import (
@@ -104,32 +104,10 @@ def _download_text_file(
         except FileNotFoundError:
             pass
 
-    n_retries = cast(int, config.connection_n_retries)
-    wait_time = 0.2
-    raise_error = None
     logging.info("Starting [%s] request for the URL %s", "get", source)
     start = time.time()
-    for retry in range(n_retries):
-        response = __read_url(source, request_method="get")
-        downloaded_file = response.text
-
-        if md5_checksum is not None:
-            md5 = hashlib.md5()
-            md5.update(downloaded_file.encode("utf-8"))
-            md5_checksum_download = md5.hexdigest()
-            if md5_checksum == md5_checksum_download:
-                raise_error = False
-                break
-            else:
-                raise_error = True
-                time.sleep(wait_time)
-    # raise_error can be set to True only if the variables md5_checksum_download and md5_checksum
-    # were initialized and compared during retries
-    if raise_error:
-        raise OpenMLHashException(
-            "Checksum {} of downloaded file is unequal to the expected checksum {} "
-            "when downloading {}.".format(md5_checksum_download, md5_checksum, source)
-        )
+    response = __read_url(source, request_method="get", md5_checksum=md5_checksum)
+    downloaded_file = response.text
 
     if output_path is None:
         logging.info(
@@ -175,25 +153,33 @@ def _read_url_files(url, data=None, file_elements=None):
     return response
 
 
-def __read_url(url, request_method, data=None):
+def __read_url(url, request_method, data=None, md5_checksum=None):
     data = {} if data is None else data
     if config.apikey is not None:
         data["api_key"] = config.apikey
+    return _send_request(
+        request_method=request_method, url=url, data=data, md5_checksum=md5_checksum
+    )
 
-    return _send_request(request_method=request_method, url=url, data=data)
+
+def __is_checksum_equal(downloaded_file, md5_checksum=None):
+    if md5_checksum is None:
+        return True
+    md5 = hashlib.md5()
+    md5.update(downloaded_file.encode("utf-8"))
+    md5_checksum_download = md5.hexdigest()
+    if md5_checksum == md5_checksum_download:
+        return True
+    return False
 
 
-def _send_request(
-    request_method, url, data, files=None,
-):
+def _send_request(request_method, url, data, files=None, md5_checksum=None):
     n_retries = max(1, min(config.connection_n_retries, config.max_retries))
 
-    retry_counter = 0
     response = None
     with requests.Session() as session:
         # Start at one to have a non-zero multiplier for the sleep
-        while retry_counter < n_retries:
-            retry_counter += 1
+        for retry_counter in range(1, n_retries + 1):
             try:
                 if request_method == "get":
                     response = session.get(url, params=data)
@@ -204,12 +190,18 @@ def _send_request(
                 else:
                     raise NotImplementedError()
                 __check_response(response=response, url=url, file_elements=files)
+                if request_method == "get" and not __is_checksum_equal(response.text, md5_checksum):
+                    raise OpenMLHashException(
+                        "Checksum of downloaded file is unequal to the expected checksum {} "
+                        "when downloading {}.".format(md5_checksum, url)
+                    )
                 break
             except (
                 requests.exceptions.ConnectionError,
                 requests.exceptions.SSLError,
                 OpenMLServerException,
                 xml.parsers.expat.ExpatError,
+                OpenMLHashException,
             ) as e:
                 if isinstance(e, OpenMLServerException):
                     if e.code not in [107, 500]:
--- a/openml/_api_calls.py
+++ b/openml/_api_calls.py
@@ -114,9 +114,8 @@ def _download_text_file(
         md5_checksum_download = md5.hexdigest()
         if md5_checksum != md5_checksum_download:
             raise OpenMLHashException(
-                "Checksum {} of downloaded file is unequal to the expected checksum {}.".format(
-                    md5_checksum_download, md5_checksum
-                )
+                "Checksum {} of downloaded file is unequal to the expected checksum {} "
+                "when downloading {}.".format(md5_checksum_download, md5_checksum, source)
             )
 
     if output_path is None:
--- a/openml/_api_calls.py
+++ b/openml/_api_calls.py
@@ -203,7 +203,7 @@ def _send_request(
                         # 107: database connection error
                         # 500: internal server error
                         wait_time = 0.3
-                        n_retries = min(n_retries + 1, max_retries)  # increase retries
+                        n_retries = min(n_retries + 1, max_retries)
                     else:
                         raise
                 else:
--- a/openml/config.py
+++ b/openml/config.py
@@ -87,7 +87,7 @@ _defaults = {
     "server": "https://www.openml.org/api/v1/xml",
     "cachedir": os.path.expanduser(os.path.join("~", ".openml", "cache")),
     "avoid_duplicate_runs": "True",
-    "connection_n_retries": 2,
+    "connection_n_retries": 5,
     "max_retries": 20,
 }
 
--- a/openml/config.py
+++ b/openml/config.py
@@ -88,6 +88,7 @@ _defaults = {
     "cachedir": os.path.expanduser(os.path.join("~", ".openml", "cache")),
     "avoid_duplicate_runs": "True",
     "connection_n_retries": 2,
+    "max_retries": 20,
 }
 
 config_file = os.path.expanduser(os.path.join("~", ".openml", "config"))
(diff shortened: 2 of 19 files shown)


Neeratyoy:
@mfeurer [this step](https://github.com/openml/openml-python/blob/e0af15e751b4c74d68940da2d41c7714694b2e94/tests/test_datasets/test_dataset_functions.py#L1377) appears to be required in order to test the subsequent clause under which the failure message is raised. That is, without uploading a task, the test passes without exceptions.

However, on local tests with the current change, it gave the following error:
`openml.exceptions.OpenMLServerException: https://test.openml.org/api/v1/xml/task/ returned code 623: Tasks can only be created upon active datasets. - problematic input: source_data, dataset not active.`

Neeratyoy:
@mfeurer many more tests fail now
This is a common error:
```
openml.exceptions.OpenMLServerException: https://test.openml.org/api/v1/xml/flow/exists returned code 104: This is a read-only account, it does not have permission for write operations. - API calls of the read-only user can only be of type GET.
```
While other fetches from the server return empty, which seems to make many of the assertions fail too!

mfeurer:
@Neeratyoy I just discussed with @joaquinvanschoren and he gave the API key write permissions

Neeratyoy:
@mfeurer 

1) If this function is in line with what we discussed, I can then extend it to other tests: https://github.com/openml/openml-python/pull/1000/files#diff-cce4323af466db2f54edb9161a510818e26857ea520e94923c3b987aff28dfc8R35

2) By the design of `create_task`, I wasn't able to create a *meta_data* that can be uniformly used to create a task, publish, retrieve and check the meta-data, so I needed this line: https://github.com/openml/openml-python/blob/431447c6e607cd8c05bb52b9ad320b51b735887b/tests/test_runs/test_run_functions.py#L935

openml/utils.py
```diff
@@ -31,6 +33,52 @@
     pass
 
 
+def check_task_existence(task_meta_data: dict) -> Union[int, None]:
```
mfeurer, line 36: Could you please put this into `testing.py` as it is only used for unit tests?

openml/utils.py
```diff
@@ -31,6 +33,52 @@
...
+                    # even if one of the meta-data key mismatches, then task_id is not a match
+                    task_match.pop(-1)
+                    break
+            # if task_id is retained in the task_match list, it passed all meta key-value matches
+            if len(task_match) == 1:
+                return_val = task_id
+                break
+        if len(task_match) == 0:
+            return_val = None
+    except openml.exceptions.OpenMLServerException:
```
mfeurer, line 77: Please don't catch and not act on exceptions. I think this try/except should just not be there at all.

openml/utils.py
```diff
@@ -31,6 +33,52 @@
...
+
+    Return
+    ------
+    int, None
+    """
+    return_val = None
+    try:
+        tasks = openml.tasks.list_tasks(output_format="dataframe")
+        tasks = typing.cast(pd.DataFrame, tasks).loc[
+            tasks["task_type"] == task_meta_data["task_type"]
```
mfeurer, line 52: Could you please add the task type to `list_tasks` to reduce overhead? Also, could you please make the argument the individual elements of `task_meta_data` to make this function easier to read?

openml/utils.py
```diff
@@ -31,6 +33,52 @@
...
+        if len(tasks) == 0:
+            return None
+        tasks = tasks.loc[tasks.did == task_meta_data["dataset_id"]]
+        if len(tasks) == 0:
+            return None
+        tasks = tasks.loc[tasks.target_feature == task_meta_data["target_name"]]
+        if len(tasks) == 0:
+            return None
+        task_match = []
+        for task_id in tasks.tid.values:
```
mfeurer, line 63: Please don't use this syntax from pandas, it's deprecated. It should be `tasks['tid'].to_list()`.

tests/test_datasets/test_dataset_functions.py
```diff
@@ -1350,7 +1351,7 @@ def test_data_edit_errors(self):
             "original_data_url, default_target_attribute, row_id_attribute, "
             "ignore_attribute or paper_url to edit.",
             edit_dataset,
-            data_id=564,
+            data_id=64,
```
mfeurer, line 1354: Could you please add the dataset name as a comment?

tests/test_runs/test_run_functions.py
```diff
@@ -44,7 +45,7 @@ class TestRun(TestBase):
     # diabetis dataset, 768 observations, 0 missing vals, 33% holdout set
     # (253 test obs), no nominal attributes, all numeric attributes
     TEST_SERVER_TASK_SIMPLE: Tuple[Union[int, List], ...] = (119, 0, 253, [], [*range(8)])
-    TEST_SERVER_TASK_REGRESSION: Tuple[Union[int, List], ...] = (738, 0, 718, [], [*range(8)])
+    TEST_SERVER_TASK_REGRESSION: Tuple[Union[int, List], ...] = (1605, 0, 2178, [], [*range(8)])
```
mfeurer, line 48: Could you please update the comment to reflect what is in these lists?

tests/test_runs/test_run_functions.py
```diff
@@ -55,6 +56,24 @@ class TestRun(TestBase):
         [1, 2, 7, 10, 13, 14],
     )
 
+    # if task IDs are deleted during test server maintenance, these meta data should still allow
+    # unit tests to pass by uploading a similar task at runtime
+    TASK_META_DATA = {
+        1605: {
+            "task_type": "Supervised Regression",
+            "dataset_id": 123,
```
mfeurer, line 64: Could you please add that this is the dataset quake? Also, is this the same information as in the change above? If yes, can they be merged?

tests/test_runs/test_run_functions.py
```diff
@@ -498,7 +517,7 @@ def _run_and_upload_classification(
     def _run_and_upload_regression(
         self, clf, task_id, n_missing_vals, n_test_obs, flow_expected_rsv, sentinel=None
     ):
-        num_folds = 1  # because of holdout
+        num_folds = 10  # because of holdout
```
mfeurer, line 520: Comment and number don't match any more

tests/test_runs/test_run_functions.py
```diff
@@ -1453,7 +1511,24 @@ def test_format_prediction_task_learning_curve_sample_not_set(self):
...
+            task_id = new_task.task_id
+            # mark to remove the uploaded task
+            TestBase._mark_entity_for_removal("task", task_id)
+            TestBase.logger.info("collected from test_run_functions: {}".format(task_id))
+
+        regression = openml.tasks.get_task(task_id, download_data=False)
         ignored_input = [0] * 5
         res = format_prediction(regression, *ignored_input)
         self.assertListEqual(res, [0] * 5)
+        self.assertListEqual(res, [0] * 5)
```
mfeurer, line 1534: This line is now duplicated.

tests/test_runs/test_run_functions.py
```diff
@@ -1453,7 +1511,24 @@ def test_format_prediction_task_learning_curve_sample_not_set(self):
             format_prediction(learning_curve, *ignored_input, sample=None, proba=probabilities)
 
     def test_format_prediction_task_regression(self):
-        regression = openml.tasks.get_task(self.TEST_SERVER_TASK_REGRESSION[0], download_data=False)
+        task_id = self.TEST_SERVER_TASK_REGRESSION[0]
```
mfeurer, line 1514: Is there a reason why you only needed to update three tests? Are the other tests relying on the 'more complicated' datasets from the OpenML100?

tests/test_tasks/test_regression_task.py
```diff
@@ -11,9 +16,36 @@ class OpenMLRegressionTaskTest(OpenMLSupervisedTaskTest):
     __test__ = True
 
     def setUp(self, n_levels: int = 1):
-
         super(OpenMLRegressionTaskTest, self).setUp()
-        self.task_id = 625
+
+        task_meta_data = {
+            "task_type": "Supervised Regression",
+            "dataset_id": 105,
```
mfeurer, line 23: Could you please add the dataset name?

openml/_api_calls.py
```diff
@@ -175,10 +175,13 @@ def _send_request(
     request_method, url, data, files=None,
 ):
     n_retries = config.connection_n_retries
+    max_retries = 10
```
mfeurer, line 178: Is there a reason you changed this here and not in the config file? Also, what's the intuition of moving from a for loop to the while loop?

openml/_api_calls.py
```diff
@@ -198,15 +201,17 @@ def _send_request(
                 if isinstance(e, OpenMLServerException):
                     if e.code != 107:
                         # 107 is a database connection error - only then do retries
-                        raise
+                        raise e
```
mfeurer, line 204: Could you please also check for the `500 internal server error` and try again on that?

openml/testing.py
```diff
@@ -252,6 +253,58 @@ def _check_fold_timing_evaluations(
...
+def check_task_existence(
+    task_type: TaskType, dataset_id: int, target_name: str, **kwargs
+) -> Union[int, None]:
+    """Checks if any task with exists on test server that matches the meta data.
+
+    Parameter
+    ---------
+    task_type : openml.tasks.TaskType
+        ID of the task type as detailed `here <https://www.openml.org/search?type=task_type>`_.
+        - Supervised classification: 1
```
mfeurer, line 265: You can drop this enumeration as `task_type` is an ENUM now and all legal values are described in there.

tests/test_runs/test_run_functions.py
```diff
@@ -921,7 +986,34 @@ def test_initialize_model_from_run(self):
...
+            "task_type": TaskType.SUPERVISED_CLASSIFICATION,
+            "dataset_id": 128,  # iris
+            "estimation_procedure_id": 1,
+            "target_name": "class",
+        }
+        _task_id = check_task_existence(**task_meta_data)
+        if _task_id is not None:
+            task_id = _task_id
+        else:
+            task_meta_data["task_type"] = TaskType.SUPERVISED_CLASSIFICATION
```
mfeurer, line 999: This code now exists twice, can this be put into its own function?

openml/_api_calls.py
PGijsbers: Side node, I believe `raise` and `raise e` are equivalent here since `e` is the currently handled exception.

openml/_api_calls.py
Neeratyoy: Yes, I should add it to config instead, that's true. I'll do so.

Wrt replacing the for-loop with a while-loop, I am now incrementing the number of retries by 1 every time a database connection error happens. Since `n_retries` determine the iterations, I think the for-loop iterator is not mutable from within the loop. A while-loop allowed me to do that easily and therefore I felt it was more readable.

openml/_api_calls.py
Neeratyoy: @mfeurer I edited the structure a bit to allow further scaling of such retries if needed. Hope this is okay.

@PGijsbers didn't know that, thanks!

tests/test_study/test_study_functions.py
```diff
@@ -213,9 +213,8 @@ def test_study_attach_illegal(self):
     def test_study_list(self):
         study_list = openml.study.list_studies(status="in_preparation")
         # might fail if server is recently resetted
-        self.assertGreater(len(study_list), 2)
+        self.assertGreaterEqual(len(study_list), 2)
 
     def test_study_list_output_format(self):
         study_list = openml.study.list_studies(status="in_preparation", output_format="dataframe")
         self.assertIsInstance(study_list, pd.DataFrame)
-        self.assertGreater(len(study_list), 2)
```
Neeratyoy, line 221: @mfeurer I removed this since the previous unit test takes care of it and we shouldn't get two tests failing. 
Or I shouldn't have done it since we no longer have a tighter check on both the output format and the list returned?

openml/config.py
```diff
@@ -88,6 +88,7 @@ def set_file_log_level(file_output_level: int):
     "cachedir": os.path.expanduser(os.path.join("~", ".openml", "cache")),
     "avoid_duplicate_runs": "True",
     "connection_n_retries": 2,
+    "max_retries": 20,
```
mfeurer, line 91: Could you please increase the default number of retries to let's say 5 or so? There are still a few stochastic failures that can potentially be remedied by trying again.

mfeurer (commented):
This looks good, I just realized that `get_dataset_arff` doesn't go through `openml_api_calls._perform_api_call` and is therefore not repeated. Could you please add a repeat option there, too, so that we retry in case the hast doesn't match?

tests/test_runs/test_run_functions.py
mfeurer: This one is still open, right?

Neeratyoy: I think I updated it, now the metadata is defined all at [one place](https://github.com/openml/openml-python/blob/d95b5e6b836a52870a76ca79c76d9954a39d00fe/tests/test_runs/test_run_functions.py#L989), and no key is being updated separately.

mfeurer (approved):

openml/_api_calls.py
```diff
@@ -174,11 +186,14 @@ def __read_url(url, request_method, data=None):
...
 ):
-    n_retries = config.connection_n_retries
+    n_retries = max(1, min(config.connection_n_retries, config.max_retries))
+
+    retry_counter = 0
     response = None
     with requests.Session() as session:
         # Start at one to have a non-zero multiplier for the sleep
-        for i in range(1, n_retries + 1):
+        while retry_counter < n_retries:
```
PGijsbers, line 195: This looks like a `for retry_counter in range(n_retries)` loop with extra steps to me.

openml/_api_calls.py
```diff
@@ -103,21 +104,32 @@ def _download_text_file(
...
-        md5.update(downloaded_file.encode("utf-8"))
-        md5_checksum_download = md5.hexdigest()
-        if md5_checksum != md5_checksum_download:
-            raise OpenMLHashException(
-                "Checksum {} of downloaded file is unequal to the expected checksum {}.".format(
-                    md5_checksum_download, md5_checksum
-                )
-            )
+    for retry in range(n_retries):
+        response = __read_url(source, request_method="get")
```
PGijsbers, line 113: I believe `__read_url` internally already does retries too, so we allow up to `n_retries**2` retries? Should we consider propagating the checksum down a few levels?

openml/_api_calls.py
```diff
@@ -103,21 +104,32 @@ def _download_text_file(
...
+            else:
+                raise_error = True
+                time.sleep(wait_time)
+    # raise_error can be set to True only if the variables md5_checksum_download and md5_checksum
+    # were initialized and compared during retries
+    if raise_error:
+        raise OpenMLHashException(
+            "Checksum {} of downloaded file is unequal to the expected checksum {} "
+            "when downloading {}.".format(md5_checksum_download, md5_checksum, source)
+        )
```
PGijsbers, line 132: I think there are a couple of things that can make this flow better:
 - you should only retry if you have a checksum, so we can represent this in the code flow. (note that now actually you download every file 10 times if no checksum is provided?)
 - you can use a `for/else` to avoid keeping `raise_error` flags.
 
Provided we keep the outer loop of retries:
```python
for retry in range(n_retries):
    ... download ...
    if checksum is None:
        break
     
    ... generate checksum ...
    if checksum is correct:
        break
    elif retry < (n_entries - 1):  # optional optimization, would hardly ever be beneficial
        time.sleep(wait_time)
else:
    # this is only reached if the loop never broke.
    raise OpenMLHashException(...)
```
Then you don't need to keep track of help variable `raise_error`, and don't nest as deep either.

PGijsbers (changes requested):
I only looked at `_api_calls`. I think the main point of discussion is whether or not we want two loops of retries for server connects. Currently the outer loop is ran as long as the response content is incorrect (not same hash), and the inner loop also connects up to `n_retries` in `_send_request` until a `status: 200`. If we don't want to nest this loop (and have `n_retries`^2 retries, we should probably allow the `_send_request` to check the response against a checksum for `get` queries.

openml/_api_calls.py
Neeratyoy: It was changed to a `while` since an intermediate change made it easier to work with rather than a `for`. However, with the latest changes in this loop, it can be reverted back to a `for`.

openml/_api_calls.py
Neeratyoy: Okay this looks neat, will try it out, thanks!

Neeratyoy:
@PGijsbers the tests that fail are oddly due to either a *worker crashing* or the [issue](https://github.com/openml/OpenML/issues/1080) pertaining to attaching entities to a study.

tests/test_runs/test_run_functions.py
```diff
@@ -367,7 +367,7 @@ def _check_sample_evaluations(
                                 # and/or measurements are not as accurate.
                                 # Either way, windows seems to get an eval-time
                                 # of 0 sometimes.
-                                self.assertGreater(evaluation, 0)
+                                self.assertGreaterEqual(evaluation, 0)
```
PGijsbers, line 370: What is the scenario where this is needed now?
If we weaken the test like this, the conditional on checking the platform is also no longer necessary.
But this also makes the test mostly pointless as it could just not record any time in all scenarios and pass.

PGijsbers (commented):
I think `_api_calls` looks fine now. I also left a comment on a change that was shown to me when I clicked the notification, would like some clarification for that first.

PGijsbers:
Looks mostly like server failure? Though something like this is suspicious:
```
=================================== FAILURES ===================================
____________________ TestStudyFunctions.test_publish_study _____________________
[gw3] linux -- Python 3.6.12 /opt/hostedtoolcache/Python/3.6.12/x64/bin/python

self = <test_study_functions.TestStudyFunctions testMethod=test_publish_study>

    @pytest.mark.flaky()
    def test_publish_study(self):
        # get some random runs to attach
        run_list = openml.evaluations.list_evaluations("predictive_accuracy", size=10)
        self.assertEqual(len(run_list), 10)
    
        fixt_alias = None
        fixt_name = "unit tested study"
        fixt_descr = "bla"
        fixt_flow_ids = set([evaluation.flow_id for evaluation in run_list.values()])
        fixt_task_ids = set([evaluation.task_id for evaluation in run_list.values()])
        fixt_setup_ids = set([evaluation.setup_id for evaluation in run_list.values()])
    
        study = openml.study.create_study(
            alias=fixt_alias,
            benchmark_suite=None,
            name=fixt_name,
            description=fixt_descr,
            run_ids=list(run_list.keys()),
        )
        study.publish()
        # not tracking upload for delete since _delete_entity called end of function
        # asserting return status from openml.study.delete_study()
        self.assertGreater(study.id, 0)
        study_downloaded = openml.study.get_study(study.id)
        self.assertEqual(study_downloaded.alias, fixt_alias)
        self.assertEqual(study_downloaded.name, fixt_name)
        self.assertEqual(study_downloaded.description, fixt_descr)
        self.assertEqual(study_downloaded.main_entity_type, "run")
    
        self.assertSetEqual(set(study_downloaded.runs), set(run_list.keys()))
        self.assertSetEqual(set(study_downloaded.setups), set(fixt_setup_ids))
        self.assertSetEqual(set(study_downloaded.flows), set(fixt_flow_ids))
        self.assertSetEqual(set(study_downloaded.tasks), set(fixt_task_ids))
    
        # test whether the list run function also handles study data fine
        run_ids = openml.runs.list_runs(study=study.id)
        self.assertSetEqual(set(run_ids), set(study_downloaded.runs))
    
        # test whether the list evaluation function also handles study data fine
        run_ids = openml.evaluations.list_evaluations(
            "predictive_accuracy", size=None, study=study.id
        )
        self.assertSetEqual(set(run_ids), set(study_downloaded.runs))
    
        # attach more runs
        run_list_additional = openml.runs.list_runs(size=10, offset=10)
>       openml.study.attach_to_study(study.id, list(run_list_additional.keys()))

tests/test_study/test_study_functions.py:164: 
openml/study/functions.py:371: in attach_to_study
    result_xml = openml._api_calls._perform_api_call(uri, "post", post_variables)
openml/_api_calls.py:61: in _perform_api_call
    response = __read_url(url, request_method, data)
openml/_api_calls.py:161: in __read_url
    request_method=request_method, url=url, data=data, md5_checksum=md5_checksum
openml/_api_calls.py:192: in _send_request

response = <Response [412]>
url = 'https://test.openml.org/api/v1/xml/study/1685/attach'
file_elements = None

    def __check_response(response, url, file_elements):
        if response.status_code != 200:
>           raise __parse_server_exception(response, url, file_elements=file_elements)
E           openml.exceptions.OpenMLServerException: https://test.openml.org/api/v1/xml/study/1685/attach returned code 1045: Problem attaching entities. Please ensure to only attach entities that exist - None

openml/_api_calls.py:230: OpenMLServerException
```

would mean there were no runs on the server (or at most10)?

tests/test_runs/test_run_functions.py
Neeratyoy: I admit I wasn't entirely sure about this and can revert it

Neeratyoy:
> I also left a comment on a change that was shown to me when I clicked the notification, would like some clarification for that first.

I'm afraid I can't see exactly which comment :/

tests/test_runs/test_run_functions.py
PGijsbers: But I imagine you must have ran into this?

PGijsbers:
> I'm afraid I can't see exactly which comment :/

It's the one about the time measurements that you commented on 👍

tests/test_runs/test_run_functions.py
Neeratyoy: well, there were some unit tests stochastically failing with the assertion error that `0.0 is not greater than 0`
while passing for other runs
I was checking if loosening the strictness of this check fixes it entirely, but as you point out, that makes the point of the check moot 
so I reverted it

PGijsbers:
I'm merging this as it's a big improvement to what was before, and so we don't need to carry this over a break.
Hopefully we can sort out the last failures soon. Ultimately we will need a better test structure one day, but that will be a major undertaking.

PGijsbers (approved):

mfeurer:
Thanks for merging @PGijsbers 

The flaky error you found was reported by @Neeratyoy in https://github.com/openml/OpenML/issues/1080

Issue #1005, opened 2020-11-18
Initializing models from runs, dependency issues


arpasan:
#### Description
Hi, I'm encountering the following error when trying to initialize a model from Run ID:
`
ValueError: Trying to deserialize a model with dependency sklearn==0.21.dev0 not satisfied.
`

Specifically:
<img width="952" alt="Screenshot 2020-11-18 at 06 26 31" src="https://user-images.githubusercontent.com/54488689/99490308-e37f2e80-2969-11eb-9f7b-57535345c7d9.png">

I tried installing `openml` from the dev branch...
```python
! pip3 install git+https://github.com/openml/openml-python@develop
```
...but that doesn't do the trick.

On the other hand, if I hone in on a model/pipeline containing the exact same `sklearn` that I have, I get another error:
<img width="948" alt="Screenshot 2020-11-18 at 06 26 57" src="https://user-images.githubusercontent.com/54488689/99490258-c8acba00-2969-11eb-8c35-afb41df49f3c.png">

I'm not sure what to do about this. Any hints?

#### Versions
<img width="560" alt="Screenshot 2020-11-18 at 06 44 59" src="https://user-images.githubusercontent.com/54488689/99490455-2a6d2400-296a-11eb-96d9-924946f7316b.png">

Thanks!

PGijsbers:
Not being able to reinitialize a `scikit-learn 0.21.dev0` model when that package is not installed is intended default behavior.
However I think we should allow loose reinitialization as we do with the `flow_to_model`. I created a branch that allows this (`loose_initialization`). The following code will hopefully allow you to reinitialize the model (provided all dependencies are installed):

```python
import openml
tree = openml.runs.initialize_model_from_run(10417577)
```
I used a different run as the one you provided had a dependency (`hpo4svm`) that I did not have installed. Please see if this work on the `loose_initialization` branch helps you.

@mfeurer I also noticed there is very little documentation on the scikit-learn extension. We should probably include the Extension interface methods (e.g. `flow_to_model`).

arpasan:
Thanks @PGijsbers for answering. That's the exact same code that broke for me earlier though. (And I'm afraid it gives the same kind of an error message.)
<img width="936" alt="Screenshot 2020-11-18 at 18 39 46" src="https://user-images.githubusercontent.com/54488689/99566930-cf6d1880-29cd-11eb-9400-20ec58565f34.png">

[...]

PR #1006, opened 2020-11-18
Allow loose reinitialization from runs also

PGijsbers:
See #1005.

If someone is able to complete this PR (e.g. add unit tests), then that would be greatly appreciated. Otherwise this will go stale for at least two weeks. Maybe @Neeratyoy ?

### DIFF
--- a/openml/runs/functions.py
+++ b/openml/runs/functions.py
@@ -321,7 +321,7 @@ def get_run_trace(run_id: int) -> OpenMLRunTrace:
     return run_trace
 
 
-def initialize_model_from_run(run_id: int) -> Any:
+def initialize_model_from_run(run_id: int, strict_version: bool = True) -> Any:
     """
     Initialized a model based on a run_id (i.e., using the exact
     same parameter settings)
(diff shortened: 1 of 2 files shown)


Neeratyoy:
Sure, I can take this up. If not this week, next week for sure. 
To clarify, the main _to-do_ would be to have some unit tests that work across sklearn versions with the strict version parameter set to `False` ?

PGijsbers:
For this PR, that would be the extent of it. In general we need more documentation for the extension but that can also be a separate PR later (possibly even by someone else).

Neeratyoy:
Alright, I think I shall address this next week from Monday!

PGijsbers:
Sorry this went off my radar. It looks like the added test fails?

Neeratyoy:
My bad. I made this push as a potential design for how the unit test can be written, wasn't expecting it to pass. Moreover, I think this PR needs to be rebased first. I am working on some other things, for now, targeting a release this week. Will pick this up hopefully next week!

PR #1015, opened 2020-12-21
Fix 1013: Store run `setup_string`

### LINKED ISSUE #1013: `setup_string` of Run is not saved
MWE:
```python
import openml
from sklearn.tree import DecisionTreeClassifier

run = openml.runs.run_model_on_task(DecisionTreeClassifier(), 59)
print('Run is not yet uploaded:', run.id == None)
run.setup_string = "test.py -o --setup-string"
run.publish()
print('local run:', run.id, run.setup_string)
server_run = openml.runs.get_run(run.id)
print('server run:', server_run.id, server_run.setup_string)
```
current output:
```
Run is not yet uploaded: True
local run: 10559751 test.py -o --setup-string
server run: 10559751 None
```
expected output:
```
Run is not yet uploaded: True
local run: 10559751 test.py -o --setup-string
server run: 10559751 test.py -o --setup-string
```

I imagine adding the field to `OpenMLRun._to_dict` is enough. 

@mfeurer this is just an oversight, right? I couldn't find/think of a reason as to why this would not be supported.

PGijsbers:
Closes #1013.


tests/test_runs/test_run_functions.py
```diff
@@ -518,6 +518,18 @@ def _run_and_upload_regression(
...
 
+    def test_upload_run_with_setup_string(self):
+        SETUP_STRING = "setup-string"
+        run = openml.runs.run_model_on_task(
+            DummyClassifier(strategy="prior"),
+            self.TEST_SERVER_TASK_SIMPLE[0],
+            avoid_duplicate_runs=False,
+        )
+        run.setup_string = SETUP_STRING
+        run.publish()
```
mfeurer, line 529: Do we need the server round trip here to assert that this is working or would an xml round trip suffice?

PGijsbers: What are you proposing here? That we check the run xml and don't actually publish the run? I'd considered it and prefer that actually, but this decoupling is not generally done in our unit tests which is why I did it this way.
We can also forgo doing an actual run, have something like:

```
run = OpenMLRun(task_id=0, flow_id=0, dataset_id=0, setup_string=SETUP_STRING)
xml = run._to_xml()
run = xmltodict.parse(xml)["oml:run"]
assert "oml:setup_string" in run
assert run["oml:setup_string"] == SETUP_STRING
```

should make the test take virtually no time, and also avoid disk operations.

mfeurer: > That we check the run xml and don't actually publish the run?

Yes, I was thinking about that and was wondering if it's possible.

> but this decoupling is not generally done in our unit tests

Yeah, it would be great to have that, but I guess we can't easily get that :+1: 

>  We can also forgo doing an actual run, have something like:

Hm, maybe we should initialize the Run() ?

mfeurer: Not sure if it would work, though: `openml.runs.function._create_run_from_xml(xml, from_server=False)`

PGijsbers: > Hm, maybe we should initialize the Run() ?

Doesn't the first line take care of that?

PGijsbers: Ah, didn't see the follow up message.

PGijsbers: Updated (and test moved to other file).

openml/runs/functions.py
```diff
@@ -762,6 +762,7 @@ def obtain_field(xml_obj, fieldname, from_server, cast=None):
     if "oml:output_data" not in run:
         if from_server:
             raise ValueError("Run does not contain output_data " "(OpenML server error?)")
+        predictions_url = None
```
PGijsbers, line 765: In the unit test this path was reached, and originally that would mean you got an `UnboundError` because `predictions_url` was never defined when creating the run.

mfeurer:
Hey, I think the changes are good, but it's really hard to tell given the massive amount of failing unit tests. We have fixed most of them in PR #1000, but it would be great if you could check the changes in `_api_calls.py`. We can then merge it and you could rebase this one so the tests are more indicative of whether the changes actually work?

mfeurer (approved):

PGijsbers:
I noticed `description_text` is also not stored, will add.

PGijsbers:
I don't see description in the xsd, but there is `run_details`: https://github.com/openml/OpenML/blob/develop/openml_OS/views/pages/api_new/v1/xsd/openml.run.upload.xsd 
Do we use that?

PGijsbers:
Oops. Looks like `description_text` relates to the arff description. Adding run_details separately.

openml/runs/run.py
```diff
@@ -57,7 +57,9 @@ class OpenMLRun(OpenMLBase):
     run_id: int
     description_text: str, optional
         Description text to add to the predictions file.
-        If left None,
+        If left None, is set to the time the arff file is generated.
+    run_details: str, optional (default=None)
```
mfeurer, line 61: This needs to also be used in the constructor in `openml.functions.runs._create_run_from_xml`.

PGijsbers: I forgot about this, but it's because [it's not in the xml](https://github.com/openml/OpenML/issues/1087).
I'll just retrieve it safely (with a default value) so we can do a release if there's no response tomorrow (and then should work when/if the server gets an update). Ok?

PGijsbers:
@mfeurer I think it's done now. Updated the progress log so it doesn't need a separate PR. The `extension.py` file apparently hadn't been made `mypy` compliant yet, so I had to add some `ignores` (I wasn't sure why the attributes weren't recognized, but I don't have time to sort that out right now- I hope that's ok).

PR #1016, opened 2021-01-08
Adding sklearn 0.24 support

Neeratyoy:

Adds support for [sklearn 0.24](https://scikit-learn.org/stable/whats_new/v0.24.html), especially making example documentation support only this latest sklearn release.

Passing the new sklearn unit tests that will be spawned for sklearn 0.24 to run on Python 3.6 and higher.

#### Any other comments?

Main changes are wrt this feature added: https://github.com/scikit-learn/scikit-learn/pull/17317


Neeratyoy:
@mfeurer 
[This unit test](https://github.com/openml/openml-python/blob/develop/tests/test_flows/test_flow_functions.py#L328) uses a [flow](https://www.openml.org/f/8175) which was built on `sklearn=0.19` using `from sklearn.ensemble.forest import RandomForestClassifier`. 
`0.23` allows the import of this with a warning that `0.24` will only support `from sklearn.ensemble import RandomForestClassifier` and rightly so, this unit test fails for `sklearn=0.24`.

We can skip the unit test for `sklearn=0.24`, or allow the unit test by using a different flow? For the latter, can I pick any flow?

mfeurer (approved):
Approving this, but as I made a lot of changes myself, too, I'm also asking @PGijsbers to have a look at this.

PGijsbers (approved):

Issue #1009, opened 2020-12-07
Give possibility to not download the dataset qualities


mfeurer:
#### Description

When downloading a lot of datasets from OpenML in order to get detailed information about the individual features it is not necessary to also download the `qualities.xml` file, and one could therefore save one call to the server per dataset (and also avoid errors in case a dataset wasn't processed).

a-moadel:
Hi @mfeurer,
Can I pick this feature up?

mfeurer:
Sure, please go ahead. Do you need any further pointers?

mfeurer:
Added via #1017. Thanks @a-moadel!

PR #1017, opened 2021-01-12
Feature/give possibility to not download the dataset qualities

a-moadel:
We can Ignore downloading the dataset qualities if its not exist while getting datasets.

Clear the cache , and calling get_dataset with download_qualities set to false : this result in qulaities.xml to not be downloaded.
Clear the cache , and calling get_dataset with download_qualities set to default : this result in qulaities.xml to be downloaded.

### DIFF
--- a/openml/datasets/functions.py
+++ b/openml/datasets/functions.py
@@ -311,6 +311,7 @@ def _name_to_id(
 
 def get_datasets(
     dataset_ids: List[Union[str, int]], download_data: bool = True,
+    download_qualities: bool = True
 ) -> List[OpenMLDataset]:
     """Download datasets.
 
@@ -334,7 +335,7 @@ def get_datasets(
     """
     datasets = []
     for dataset_id in dataset_ids:
-        datasets.append(get_dataset(dataset_id, download_data))
+        datasets.append(get_dataset(dataset_id, download_data, download_qualities))
     return datasets
 
 
@@ -345,7 +346,7 @@ def get_dataset(
     version: int = None,
     error_if_multiple: bool = False,
     cache_format: str = "pickle",
-) -> OpenMLDataset:
+    download_qualities: bool = True) -> OpenMLDataset:
     """ Download the OpenML dataset representation, optionally also download actual data file.
 
     This function is thread/multiprocessing safe.
@@ -405,7 +406,7 @@ def get_dataset(
         features_file = _get_dataset_features_file(did_cache_dir, dataset_id)
 
         try:
-            qualities_file = _get_dataset_qualities_file(did_cache_dir, dataset_id)
+            qualities_file = _get_dataset_qualities_file(did_cache_dir, dataset_id,download_qualities)
         except OpenMLServerException as e:
             if e.code == 362 and str(e) == "No qualities found - None":
                 logger.warning("No qualities found for dataset {}".format(dataset_id))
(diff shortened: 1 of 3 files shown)


openml/datasets/functions.py
```diff
@@ -290,6 +290,8 @@ def _name_to_id(
     error_if_multiple : bool (default=False)
         If `False`, if multiple datasets match, return the least recent active dataset.
         If `True`, if multiple datasets match, raise an error.
+    download_qualities : bool, optional
```
PGijsbers, line 293:
```suggestion
    download_qualities : bool, optional (default=True)
```

openml/datasets/functions.py
```diff
@@ -326,6 +328,8 @@ def get_datasets(
         make the operation noticeably slower. Metadata is also still retrieved.
         If False, create the OpenMLDataset and only populate it with the metadata.
         The data may later be retrieved through the `OpenMLDataset.get_data` method.
+    download_qualities : bool, optional
```
PGijsbers, line 331:
```suggestion
    download_qualities : bool, optional (default=True)
```

PGijsbers (changes requested):
@mfeurer Do you think the qualities should be loaded if they are present in cache? That would mean different results of the function depending on the cache state. I would think we can skip it altogether?
@a-moadel Please also add unit tests that test the feature works as intended (after we agree on the desired behavior).

I noticed the `pre-commit` workflow is not triggered. Does anyone have an idea why this would not trigger?

Closes #1009.

mfeurer:
This looks good to me now. @PGijsbers you had a comment about persistent behavior, that's not yet addressed, right?

tests/test_datasets/test_dataset_functions.py
```diff
@@ -433,6 +433,11 @@ def test__get_dataset_qualities(self):
         qualities_xml_path = os.path.join(self.workdir, "qualities.xml")
         self.assertTrue(os.path.exists(qualities_xml_path))
 
+    def test__get_dataset_qualities_skip_download(self):
+        qualities = _get_dataset_qualities_file(self.workdir, 2, False)
```
PGijsbers, line 437: Could you use keywords here (e.g. `download_qualities=False`)?

openml/datasets/functions.py
```diff
@@ -405,7 +412,9 @@ def get_dataset(
         features_file = _get_dataset_features_file(did_cache_dir, dataset_id)
 
         try:
-            qualities_file = _get_dataset_qualities_file(did_cache_dir, dataset_id)
+            qualities_file = _get_dataset_qualities_file(
```
PGijsbers, line 415: I think it makes more sense to use `download_qualities` here and not call the `_get_dataset_qualities_file` if we don't want to process the qualities.

openml/datasets/functions.py
```diff
@@ -290,6 +290,8 @@ def _name_to_id(
     error_if_multiple : bool (default=False)
         If `False`, if multiple datasets match, return the least recent active dataset.
         If `True`, if multiple datasets match, raise an error.
+    download_qualities : bool, optional (default=True)
+        If `True`, also download qualities.xml file. If false use the file if it was cached.
```
PGijsbers, line 294: Please update this and the other references where it says the cached version is used if `false` is passed, as this is not the behavior anymore.

PGijsbers (changes requested):
First off, sorry for the delay. I think it works as intended now 🥳 however I do have a few minor comments that I would prefer to see addressed before merging. I'd like to see references to the old behavior updated, and I suggested a minor code change which puts the responsibility of not loading `qualities` in the `get_dataset` function instead. I don't think it makes sense to pass down a `bool` that signifies you just want `""` back - then we don't need to call down in the first place.

PGijsbers (approved):

PR #1018, opened 2021-01-15
Refactor data loading/storing

PGijsbers:
I stumbled on the dataset loading/caching code and found the flow very difficult to parse. Additionally there was a lot of duplicate code around. The main goal of this PR is to make the flow of loading and storing compressed data easier to read, and reduce the amount of duplicate code. In the process I streamlined the flow a little, which should lead to some performance gains (less excessive data loading).

Differences in behavior:
  - When the OpenMLDataset is constructed, outdated pickle data is no longer pro-actively updated. Instead the numpy pickle files are updated in `_load_data`.
  - When `_load_data` is called on a dataset which is not yet compressed (or needs to be updated), it is loaded only once instead of twice.
  - When the OpenMLDataset is constructed, the arff file is not immediately converted to a compressed format. Instead this happens the first time the data is required in `_load_data`.
  - OpenMLDataset's `data_pickle_file`, `data_feather_file` and `feather_attribute_file` members more accurately reflect the presence of the file. Previously the cache format files for the format that was not used would also be set, while they are never generated.

All unit tests passed without modification, except for the `test_get_dataset_cache_format_feather` which relied on the assumption that the data was compressed and stored on disk before the first load. However I also updated the pickle test which "tests" if the file can be loaded from the compressed format, otherwise it would now test if they are loaded from arff (the first load is from arff but stores to pickle, subsequent loads are from pickle).

### DIFF
--- a/openml/datasets/dataset.py
+++ b/openml/datasets/dataset.py
@@ -217,10 +217,11 @@ class OpenMLDataset(OpenMLBase):
             self.qualities = None
 
         if data_file is not None:
-            rval = self._create_pickle_in_cache(data_file)
+            rval = self._compressed_cache_file_paths(data_file)
             self.data_pickle_file = rval[0]  # type: Optional[str]
             self.data_feather_file = rval[1]  # type: Optional[str]
             self.feather_attribute_file = rval[2]  # type: Optional[str]
+            self._cache_compressed_file_from_arff(self.data_file)
         else:
             self.data_pickle_file, self.data_feather_file, self.feather_attribute_file = (
                 None,
@@ -278,7 +279,7 @@ class OpenMLDataset(OpenMLBase):
             "upload_date",
             "url",
             "dataset",
-            "data_file",
+            "arff_file",
         }
 
         # check that the keys are identical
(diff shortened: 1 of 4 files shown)


openml/datasets/dataset.py
```diff
@@ -217,16 +217,14 @@ def find_invalid_characters(string, pattern):
             self.qualities = None
 
         if data_file is not None:
-            rval = self._create_pickle_in_cache(data_file)
-            self.data_pickle_file = rval[0]  # type: Optional[str]
-            self.data_feather_file = rval[1]  # type: Optional[str]
-            self.feather_attribute_file = rval[2]  # type: Optional[str]
+            rval = self._compressed_cache_file_paths(data_file)
+            self.data_pickle_file = rval[0] if os.path.exists(rval[0]) else None
```
mfeurer, line 221: I have a weird request: could you please list the directory first and then check if the files are there? Apparently, it is really expensive to check the content of a directory on some filesystems, and this way it would reduce the load by a factor of three.

mfeurer (commented):
Looks good to me, I only have one minor change request.

mfeurer (approved):

PR #1021, opened 2021-02-04
improve path detection

mfeurer:
Fixes an issue that could make unit tests crashed if the OpenML-Python repository was not in the path. Reported in https://github.com/conda-forge/staged-recipes/pull/13860

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -35,7 +35,6 @@ logger = logging.getLogger("unit_tests")
 logger.setLevel(logging.DEBUG)
 
 file_list = []
-directory = None
 
 
 def worker_id() -> str:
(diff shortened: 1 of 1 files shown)


PGijsbers (approved):

PGijsbers:
It looks like the tests are now leaving files behind: e.g. https://github.com/openml/openml-python/pull/1021/checks?check_run_id=1830686974

PGijsbers (changes requested):
setting to request changes because the issue needs to be looked into.

mfeurer:
Hm, I don't understand why my change might fix this, but it appears to do so?

PGijsbers (approved):

PR #1029, opened 2021-02-22
Parquet Support

PGijsbers:
Download the Parquet file from the MinIO server, if a link to it is provided by the server in the dataset description xml (currently only on the test server).
Unlike the ARFF equivalent, checksums are not available yet, so there's no internal file integrity check.

On Friday tests were green, but there are currently authentication issues with the test server, @sahithyaravi1493 is looking into it.


openml/datasets/functions.py
```diff
@@ -424,6 +424,10 @@ def get_dataset(
                 raise
 
         arff_file = _get_dataset_arff(description) if download_data else None
+        if "oml:minio_url" in description and download_data:
```
mfeurer, line 427: Could you please explain why we potentially download both files? I guess this is easier to handle at the moment?

openml/_api_calls.py
```diff
@@ -68,6 +72,45 @@ def _perform_api_call(call, request_method, data=None, file_elements=None):
     return response.text
 
 
+def _download_minio_file(
+    source: str, destination: Union[str, pathlib.Path], exists_ok: bool = True,
```
mfeurer, line 76: Not related to the content, but how does `black` accept this? Shouldn't it split this up into three lines?

openml/_api_calls.py
```diff
@@ -68,6 +72,45 @@ def _perform_api_call(call, request_method, data=None, file_elements=None):
...
+    """ Download file ``source`` from a MinIO Bucket and store it at ``destination``.
+
+    Parameters
+    ----------
+    source : Union[str, pathlib.Path]
+        URL to a file in a MinIO bucket.
+    destination : str
+        Path to store the file to, if a directory is provided the original filename is used.
+    exists_ok : bool, optional (default=True)
+        If False, raise FileExists if a file already exists in ``destination``.
```
mfeurer, line 87: The return values is not specified here.

openml/_api_calls.py
```diff
@@ -68,6 +72,45 @@ def _perform_api_call(call, request_method, data=None, file_elements=None):
...
+    parsed_url = urllib.parse.urlparse(source)
+
+    # expect path format: /BUCKET/path/to/file.ext
+    bucket, object_name = parsed_url.path[1:].split("/", maxsplit=1)
+    if destination.is_dir():
+        destination = pathlib.Path(destination, object_name)
+    if destination.is_file() and not exists_ok:
+        raise FileExistsError(f"File already exists in {destination}.")
+
+    client = minio.Minio(endpoint=parsed_url.netloc, secure=False,)
```
mfeurer, line 99: What means `secure=False`?

openml/_api_calls.py
```diff
@@ -68,6 +72,45 @@ def _perform_api_call(call, request_method, data=None, file_elements=None):
...
+        file_meta = client.fget_object(
+            bucket_name=bucket, object_name=object_name, file_path=str(destination),
+        )
+    except minio.error.S3Error as e:
+        if e.message.startswith("Object does not exist"):
+            raise FileNotFoundError(f"Object at '{source}' does not exist.") from e
+        # e.g. permission error, or a bucket does not exist (which is also interpreted as a
+        # permission error on minio level).
+        raise FileNotFoundError("Bucket does not exist or is private.") from e
+    return file_meta
```
mfeurer, line 111: Is this return value used and or needed?

tests/test_datasets/test_dataset_functions.py
```diff
@@ -407,6 +410,85 @@ def test__getarff_path_dataset_arff(self):
         self.assertIsInstance(arff_path, str)
         self.assertTrue(os.path.exists(arff_path))
 
+    def test__download_minio_file_object_does_not_exist(self):
+        self.assertRaisesRegex(
+            FileNotFoundError,
+            r"Object at .* does not exist",
+            _download_minio_file,
+            source="http://openml1.win.tue.nl/dataset20/i_do_not_exist.pq",
```
mfeurer, line 418: Is this a new test server?

mfeurer (commented):
I just had a brief look and I think this looks good and I'm really looking forward to having this live instead of arff files :)

openml/datasets/functions.py
PGijsbers: The main reason is that the test server currently always returns a `minio_url` entry, regardless of whether or not the parquet file actually exists. I suppose you could turn it around and only download the arff file of the parquet file needed to download. 
Some people also might still be interested in having the arff file (for now), and we would not have a public API for downloading that file, so in the interest of merging this before a PyPI release, I figured keeping that behavior for now is better.

openml/_api_calls.py
PGijsbers: If I remember correctly this was actually formatted by `black` because it fits on one line :)

openml/_api_calls.py
PGijsbers: It means we don't have to supply credentials to the minio server, because the files are publicly available.

openml/_api_calls.py
PGijsbers: No, I thought earlier on it might be useful. I ended up not using it so I'll remove it from the function.

tests/test_datasets/test_dataset_functions.py
PGijsbers: There is currently no distinction between a test and production `minio` server. Will need updating once that's sorted out 😔

openml/datasets/functions.py
mfeurer: That makes sense. Do you think we should open an issue to track the move from arff to parquet?

mfeurer (commented):
This looks good and the code looks good enough to be merged. However, it currently breaks documentation building and dataset upload, could you please ahve a look at that?

PGijsbers:
I was working on the last errors but was having a lunch break 😓 wasn't ready for re-review yet.

mfeurer:
> wasn't ready for re-review yet.

Sorry, I was too eager :)

PGijsbers:
@mfeurer I can't reproduce the doc errors locally and the given errors are not informative. Are you able to reproduce the errors locally?

mfeurer:
> Are you able to reproduce the errors locally?

No, I'm not :( And it seems that the errors are parsed wrongly in that they don't show the `from e`. Maybe you can add a print statement and raise the original exception?

>  Appveyor seems to fail due to two tests that I think should be unix only.

Good question. Would it be possible that you have a look whether the tests could make sense for Windows, too?

PGijsbers:
> Maybe you can add a print statement and raise the original exception?

I suppose so, for debugging on CI at least. Hopefully it explains why `from e` doesn't work in this scenario (I've seen it fail locally earlier in the development, but then `from e` correctly showed both exceptions/tracebacks as it should). 

> Would it be possible that you have a look whether the tests could make sense for Windows, too?

I tried some simple stuff to make it cross-platform, but it's not working so far. Are you ok with it being skipped on Windows for now? Adding unit tests for a completely unrelated issue seems like a good candidate for a separate PR anyway.

openml/datasets/functions.py
PGijsbers: Seems like a good idea: #1032.

mfeurer (approved):

Issue #248, opened 2017-05-05
Store runtime of instances of BaseSearchCV


mfeurer:
Currently not possible because of https://github.com/scikit-learn/scikit-learn/issues/8833.

PGijsbers:
It seems @mfeurer added `refit_time_ attribute` [here](https://github.com/scikit-learn/scikit-learn/pull/11310), so this can be added.

PR #1031, opened 2021-02-23
Measuring runtimes

Neeratyoy:
#### Reference Issue

Addresses #248.

### DIFF
--- a/openml/runs/functions.py
+++ b/openml/runs/functions.py
@@ -662,7 +662,7 @@ def _run_task_get_arffcontent_parallel_helper(
         raise NotImplementedError(task.task_type)
     config.logger.info(
         "Going to execute flow '%s' on task %d for repeat %d fold %d sample %d.",
-        flow.name,
+        str(model),
         task.task_id,
         rep_no,
         fold_no,
--- a/openml/runs/functions.py
+++ b/openml/runs/functions.py
@@ -271,7 +271,6 @@ def run_flow_on_task(
 
     # execute the run
     res = _run_task_get_arffcontent(
-        flow=flow,
         model=flow.model,
         task=task,
         extension=flow.extension,
@@ -432,7 +431,6 @@ def run_exists(task_id: int, setup_id: int) -> Set[int]:
 
 
 def _run_task_get_arffcontent(
-    flow: OpenMLFlow,
     model: Any,
     task: OpenMLTask,
     extension: "Extension",
@@ -476,7 +474,6 @@ def _run_task_get_arffcontent(
     job_rvals = Parallel(verbose=0, n_jobs=n_jobs)(
         delayed(_run_task_get_arffcontent_parallel_helper)(
             extension=extension,
-            flow=flow,
             fold_no=fold_no,
             model=model,
             rep_no=rep_no,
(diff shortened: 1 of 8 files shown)


tests/test_extensions/test_sklearn_extension/test_sklearn_extension.py
```diff
@@ -2253,3 +2253,15 @@ def column_transformer_pipe(task_id):
...
         TestBase._mark_entity_for_removal("run", run2.run_id)
         self.assertEqual(run1.setup_id, run2.setup_id)
+
+    def test_for_refit_time_in_basesearchCV(self):
+        X, y = sklearn.datasets.load_iris(return_X_y=True)
+        rs = sklearn.model_selection.GridSearchCV(
+            estimator=sklearn.ensemble.RandomForestClassifier(),
+            param_grid={"n_estimators": [2, 3, 4, 5]},
+        )
+        rs.fit(X, y)
```
mfeurer, line 2263: Could you please try to integrate this test into an existing unit test? Right now this only tests scikit-learn, which I don't think we have to do.

mfeurer (commented):
Looks good, except for the failure that you mentioned offline

Neeratyoy:
> for the failure that you mentioned offline

Should pass with the changes made.
For the long term, we can have a new PR that removes the `flow` parameter from `_run_task_get_arffcontent`'s argument list?

openml/runs/functions.py
```diff
@@ -662,7 +662,7 @@ def _run_task_get_arffcontent_parallel_helper(
         raise NotImplementedError(task.task_type)
     config.logger.info(
         "Going to execute flow '%s' on task %d for repeat %d fold %d sample %d.",
-        flow.name,
+        str(model),
```
mfeurer, line 665: Hey, what's the reason to change this? And if you change this, you should also change the rest of the logging message.

mfeurer (commented):
I think you can also remove the `flow` in this PR as it is not needed and removing it is helpful in reaching this PRs goal.

Neeratyoy:
Added a `flaky` decorator, and also followed some of the suggestions from [here](https://joblib.readthedocs.io/en/latest/auto_examples/serialization_and_wrappers.html) to test if that prevents, what clearly is a stochastic failure. Moreover, the old `failure to attach entities` are reappearing now randomly again 😕

tests/test_study/test_study_functions.py
```diff
@@ -133,8 +135,8 @@ def test_publish_study(self):
             run_ids=list(run_list.keys()),
         )
         study.publish()
-        # not tracking upload for delete since _delete_entity called end of function
-        # asserting return status from openml.study.delete_study()
+        TestBase._mark_entity_for_removal("study", study.id)
```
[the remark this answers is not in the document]
Neeratyoy: I'd say we can let it be. Since this statement simply marks the study ID that will be deleted at the end of all unit tests. 

The `delete_study` should delete the study if this unit test passes, which I think was the original design and should remain. In this scenario, when at the end of unit testing this study is being tried to be deleted again, it fails to do so without hampering the rest of the flow. 

When this particular unit test fails, the `delete_study` is not reached and the study remains on the test server. Collecting the ID in this manner should safeguard even against such a scenario and ensure the study uploaded during unit testing gets deleted be it a success/failure.

mfeurer (approved):

Issue #895, opened 2020-01-17
Why is computation time not reported if n_jobs != 1 or != None?


NicolasHug:
I'm running a big benchmark suite with `RandomizedSearchCV(n_jobs=-1)`.

Unfortunately, computation time is reported only if `n_jobs` is `None` or `1`.

I don't understand the reason https://github.com/openml/openml-python/issues/229. Why isn't the interpretation left out to the user?

----

As a side note: `n_jobs=None` can be overridden with a context manager:

```py
from joblib import parallel_backend
with parallel_backend('loky', n_jobs=-1):
	RandomizedSearchCV(n_jobs=None)
```

This is equivalent to just calling `RandomizedSearchCV(n_jobs=-1)`.

With the latter, openml won't report computation time, but as far as I understand, the former will run just fine and report the computation time. So it seems that the check isn't properly enforced anyway.

CC @amueller

amueller:
ping @janvanrijn @mfeurer ;)

amueller:
Should we add a ``wallclock_time_millis_training`` instead maybe?

amueller:
The reason it is wrong for ``n_jobs != 1`` is that internally it uses ``process_time`` which will not count any of the subprocess time, and it's not using wall-clock time.

mfeurer:
> ping @janvanrijn @mfeurer ;)

I'll come back to you after the ICML deadline.

mfeurer:
Thanks for raising this issue, it seems that there are indeed one or two problems here.

I believe the reason why the wallclock time is not reported if the number of cores is -1 is because we can't figure out on how many cores it was executed and the number then only makes limited sense. Currently, this is a very restrictive assumption that can be circumvented in plenty of ways (as you showed). Do you have any suggestions on how to improve on this?

> Should we add a wallclock_time_millis_training additionally maybe which can always be computed?

That exists and is computed if `n_jobs != -1`.

In order to get the times of each base run you can check optimization trace which should have the time for each model fit. However, we currently don't seem to store the refit time correctly (or at all?), which to me currently seems like the biggest bug here.

NicolasHug:
> Do you have any suggestions on how to improve on this?

I think you can use `effective_n_jobs` from joblib: https://github.com/joblib/joblib/blob/master/joblib/parallel.py#L366

mfeurer:
Yet another issue we have to think about is the recent use of OpenMP in scikit-learn which might make it harder for us to get a useful estimate of the used time.

[...]

PR #1038, opened 2021-03-12
Standardize use of n_jobs and reporting of computation time

Neeratyoy:
#### Reference Issue
Addresses #895.

### DIFF
--- a/doc/usage.rst
+++ b/doc/usage.rst
@@ -145,10 +145,14 @@ obtained on. Learn how to share your datasets in the following tutorial:
 
 * `Upload a dataset <examples/30_extended/create_upload_tutorial.html>`_
 
-~~~~~~~~~~~~~~~~~~~~~~~
+***********************
 Extending OpenML-Python
-~~~~~~~~~~~~~~~~~~~~~~~
+***********************
 
 OpenML-Python provides an extension interface to connect other machine learning libraries than
 scikit-learn to OpenML. Please check the :ref:`api_extensions` and use the
 scikit-learn extension in :class:`openml.extensions.sklearn.SklearnExtension` as a starting point.
+
+Runtime measurement is incorporated in the OpenML sklearn-extension. Example usage and potential
+usage for Hyperparameter Optimisation can be found in the example tutorial:
+`HPO using OpenML <examples/30_extended/fetch_runtimes.html>`
--- a/doc/usage.rst
+++ b/doc/usage.rst
@@ -155,4 +155,4 @@ scikit-learn extension in :class:`openml.extensions.sklearn.SklearnExtension` as
 
 Runtime measurement is incorporated in the OpenML sklearn-extension. Example usage and potential
 usage for Hyperparameter Optimisation can be found in the example tutorial:
-`HPO using OpenML <examples/30_extended/fetch_runtimes.html>`
+`HPO using OpenML <examples/30_extended/fetch_runtimes.html>`_
--- a/examples/30_extended/fetch_runtimes.py
+++ b/examples/30_extended/fetch_runtimes.py
@@ -32,6 +32,8 @@ from sklearn.ensemble import RandomForestClassifier
 from sklearn.model_selection import GridSearchCV, RandomizedSearchCV
 
 
+openml.config.apikey = "7ff6d43e00cc810a00c01bd770996dfc"
+
 ############################################################################
 # Preparing tasks and scikit-learn models
 # ***************************************
@@ -90,7 +92,7 @@ for repeat, val1 in measures["predictive_accuracy"].items():
 # related as:
 # usercpu_time_millis = usercpu_time_millis_training + usercpu_time_millis_testing
 # wall_clock_time_millis = wall_clock_time_millis_training + wall_clock_time_millis_testing
-#
+
 # The timing measures recorded as `*_millis_training` contain the per
 # repeat-fold timing incurred for the executing of the `.fit()` procedure
 # of the model. For `usercpu_time_*` the time recorded using `time.process_time()`
@@ -135,11 +137,14 @@ print_compare_runtimes(measures)
 # includes this in the `wall_clock_time_millis_training` measure recorded.
 
 clf = RandomForestClassifier(n_estimators=10, n_jobs=2)
+
 # GridSearchCV model
+n_iter = 5
 grid_pipe = GridSearchCV(
     estimator=clf,
-    param_grid={"n_estimators": np.linspace(start=1, stop=50, num=6).astype(int).tolist()},
+    param_grid={"n_estimators": np.linspace(start=1, stop=50, num=n_iter).astype(int).tolist()},
     cv=2,
+    n_jobs=2,
 )
 
 run4 = openml.runs.run_model_on_task(model=grid_pipe, task=task, n_jobs=2)
@@ -147,21 +152,43 @@ measures = run4.fold_evaluations
 print_compare_runtimes(measures)
 print()
 
-# Scikit-learn HPO estimators often have a `refit=True` parameter set by default,
-# and thus records the time to refit the model with the best found parameters.
-# This can be extracted in the following manner:
-for repeat, val1 in measures["refit_time"].items():
-    for fold, val2 in val1.items():
-        print("Repeat #{}-Fold #{}: {:.4f}".format(repeat, fold, val2))
-    print()
-
 # Like any optimisation problem, scikit-learn's HPO estimators also generate
 # a sequence of configurations which are evaluated as the best found
-# configuration is tracked throughout the trace. Along with the GridSearchCV
-# already used above, we demonstrate how such optimisation traces can be
-# retrieved by showing an application of these traces - comparing the
-# speed of finding the best configuration using RandomizedSearchCV and
-# GridSearchCV available with scikit-learn.
+# configuration is tracked throughout the trace.
+# The OpenML run object stores these traces as OpenMLRunTrace objects accessible
+# using keys of the pattern (repeat, fold, iterations). Since `iterations` is part
+# of the scikit-learn model here, the runtime recorded per repeat-per fold is for
+# this entire `fit()` procedure.
+
+# We earlier extracted the number of repeats and folds for this task:
+print("# repeats: {}\n# fold: {}".format(n_repeats, n_folds))
+
+# To extract the training runtime of the first repeat, first fold:
+print(run4.fold_evaluations["wall_clock_time_millis_training"][0][0])
+
+# To extract the training runtime of the 1-st repeat, 4-th fold and also
+# to fetch the parameters and performance of the evaluations made during
+# the 1-st repeat, 4-th fold evaluation by the Grid Search model
+_repeat = 0
+_fold = 3
+print(
+    "Total runtime for repeat {}'s fold {}: {:4f} ms".format(
+        _repeat, _fold, run4.fold_evaluations["wall_clock_time_millis_training"][_repeat][_fold]
+    )
+)
+for i in range(n_iter):
+    key = (_repeat, _fold, i)
+    r = run4.trace.trace_iterations[key]
+    print(
+        "n_estimators: {:>2} - score: {:.3f}".format(
+            r.parameters["parameter_n_estimators"], r.evaluation
+        )
+    )
+
+# Along with the GridSearchCV already used above, we demonstrate how such
+# optimisation traces can be retrieved by showing an application of these
+# traces - comparing the speed of finding the best configuration using
+# RandomizedSearchCV and GridSearchCV available with scikit-learn.
 
 # RandomizedSearchCV model
 rs_pipe = RandomizedSearchCV(
@@ -170,29 +197,62 @@ rs_pipe = RandomizedSearchCV(
         "n_estimators": np.linspace(start=1, stop=50, num=15).astype(int).tolist()
     },
     cv=2,
-    n_iter=6,
+    n_iter=n_iter,
+    n_jobs=2,
 )
-run5 = openml.runs.run_model_on_task(model=rs_pipe, task=task, n_jobs=2)
+run5 = openml.runs.run_model_on_task(model=rs_pipe, task=task)
 
-# This function tracks the best so-far seen accuracy as the regret from 100% accuracy
+
+# Since for the call to `openml.runs.run_model_on_task` the parameter
+# `n_jobs` is set to its default None, the evaluations across the OpenML folds
+# are not parallelized. Hence, the time recorded is agnostic to the `n_jobs`
+# being set at both the HPO estimator `GridSearchCV` as well as the base
+# estimator `RandomForestClassifier` in this case. OpenML only records the
+# time taken for the completion of the complete `fit()` call, per-repeat per-fold.
+
+# This notion can be used to extract and plot the best found performance per
+# fold by the HPO model and the corresponding time taken for search across
+# that fold. Moreover, since `n_jobs=None` for `openml.runs.run_model_on_task`
+# the runtimes per fold can be cumulatively added to plot the trace against time.
+
+
+def extract_trace_data(run, n_repeats, n_folds, n_iter, key=None):
+    key = "wall_clock_time_millis_training" if key is None else key
+    data = {"score": [], "runtime": []}
+    for i_r in range(n_repeats):
+        for i_f in range(n_folds):
+            data["runtime"].append(run.fold_evaluations[key][i_r][i_f])
+            for i_i in range(n_iter):
+                r = run.trace.trace_iterations[(i_r, i_f, i_i)]
+                if r.selected:
+                    data["score"].append(r.evaluation)
+                    break
+    return data
+
+
 def get_incumbent_trace(trace):
-    # `trace` is an OpenMLRunTrace object
     best_score = 1
     inc_trace = []
-    for i, r in enumerate(trace.trace_iterations.values()):
-        # `r` is an OpenMLTraceIteration object
-        if i == 0 or (r.selected and (1 - r.evaluation) < best_score):
-            best_score = 1 - r.evaluation
+    for i, r in enumerate(trace):
+        if i == 0 or (1 - r) < best_score:
+            best_score = 1 - r
         inc_trace.append(best_score)
     return inc_trace
 
 
+grid_data = extract_trace_data(run4, n_repeats, n_folds, n_iter)
+rs_data = extract_trace_data(run5, n_repeats, n_folds, n_iter)
+
 plt.clf()
-plt.plot(get_incumbent_trace(run4.trace), label="Grid Search")
-plt.plot(get_incumbent_trace(run5.trace), label="Random Search")
-plt.yscale("log")
+plt.plot(
+    np.cumsum(grid_data["runtime"]), get_incumbent_trace(grid_data["score"]), label="Grid Search"
+)
+plt.plot(
+    np.cumsum(rs_data["runtime"]), get_incumbent_trace(rs_data["score"]), label="Random Search"
+)
 plt.xscale("log")
-plt.xlabel("Number of evaluations")
+plt.yscale("log")
+plt.xlabel("Wallclock time (in milliseconds)")
 plt.ylabel("1 - Accuracy")
 plt.title("Optimisation Trace Comparison")
 plt.legend()
--- a/examples/30_extended/fetch_runtimes.py
+++ b/examples/30_extended/fetch_runtimes.py
@@ -26,22 +26,24 @@ runtimes retrieved
 
 import openml
 import numpy as np
+from matplotlib import pyplot as plt
 
 from sklearn.ensemble import RandomForestClassifier
-from sklearn.model_selection import GridSearchCV
+from sklearn.model_selection import GridSearchCV, RandomizedSearchCV
 
 
 ############################################################################
 # Preparing tasks and scikit-learn models
 # ***************************************
 
-task_id = 1792  # 1778  # supervised classification on mfeat-fourier
+task_id = 168908
+
 task = openml.tasks.get_task(task_id)
 print(task)
 
 # Viewing associated data
-X, y = task.get_X_and_y(dataset_format="dataframe")
-print(X.head)
+X, y = task.get_X_and_y(dataset_format="array")
+# print(X.head)
 n_repeats, n_folds, n_samples = task.get_split_dimensions()
 print(
     "Task {}: number of repeats: {}, number of folds: {}, number of samples {}.".format(
@@ -56,7 +58,7 @@ def print_compare_runtimes(measures):
     for repeat, val1 in measures["usercpu_time_millis_training"].items():
         for fold, val2 in val1.items():
             print(
-                "Repeat #{}-Fold#{}: CPU-{:.3f} vs Wall-{:.3f}".format(
+                "Repeat #{}-Fold #{}: CPU-{:.3f} vs Wall-{:.3f}".format(
                     repeat, fold, val2, measures["wall_clock_time_millis_training"][repeat][fold]
                 )
             )
@@ -66,11 +68,11 @@ def print_compare_runtimes(measures):
 ############################################################################
 # Case 1: Running a Random Forest model on an OpenML task
 # *******************************************************
-# We'll run a Random Forest model and obtain an OpenML run-object. We can
+# We'll run a Random Forest model and obtain an OpenML run object. We can
 # see the evaluations recorded per fold for the dataset and the information
 # available for this run.
 
-run1 = openml.runs.run_model_on_task(model=clf, task=task)
+run1 = openml.runs.run_model_on_task(model=clf, task=task, dataset_format="array")
 measures = run1.fold_evaluations
 
 # The timing and performance metrics available are
@@ -81,7 +83,7 @@ for key in measures.keys():
 # and can be retrieved as:
 for repeat, val1 in measures["predictive_accuracy"].items():
     for fold, val2 in val1.items():
-        print("Repeat #{}-Fold#{}: {}".format(repeat, fold, val2))
+        print("Repeat #{}-Fold #{}: {:.4f}".format(repeat, fold, val2))
     print()
 
 # The remaining entries recorded in `measures` are the runtime records
@@ -102,13 +104,25 @@ print_compare_runtimes(measures)
 ############################################################################
 # Case 2: Running a Random Forest model on an OpenML task in parallel
 # *******************************************************************
-# Running the same model on the same task but with parallelism (n_jobs > 1)
+# Redefining the model to allow parallelism with `n_jobs=2`
+clf = RandomForestClassifier(n_estimators=10, n_jobs=2)
 
-run2 = openml.runs.run_model_on_task(model=clf, task=task, n_jobs=4)
+run2 = openml.runs.run_model_on_task(model=clf, task=task, dataset_format="array")
 measures = run2.fold_evaluations
-
-# Comparing the CPU and wall-clock training times of the Random Forest model
+# The wall-clock time recorded per fold should be lesser than Case 1 above
 print_compare_runtimes(measures)
+
+# Redefining the model to use all available cores with `n_jobs=-1`
+clf = RandomForestClassifier(n_estimators=10, n_jobs=-1)
+
+run3 = openml.runs.run_model_on_task(model=clf, task=task, dataset_format="array")
+measures = run3.fold_evaluations
+# The wall-clock time recorded per fold should be lesser than the case above,
+# if more than 2 CPU cores are available
+print_compare_runtimes(measures)
+
+# It should be noted that there are multiple levels at which parallelism can
+# occur here.
 
 ############################################################################
 # Case 3: Running and benchmarking HPO algorithms with their runtimes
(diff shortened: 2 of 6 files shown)


doc/usage.rst
```diff
@@ -145,10 +145,14 @@ obtained on. Learn how to share your datasets in the following tutorial:
...
-~~~~~~~~~~~~~~~~~~~~~~~
+***********************
 
 OpenML-Python provides an extension interface to connect other machine learning libraries than
 scikit-learn to OpenML. Please check the :ref:`api_extensions` and use the
 scikit-learn extension in :class:`openml.extensions.sklearn.SklearnExtension` as a starting point.
+
+Runtime measurement is incorporated in the OpenML sklearn-extension. Example usage and potential
+usage for Hyperparameter Optimisation can be found in the example tutorial:
+`HPO using OpenML <examples/30_extended/fetch_runtimes.html>`
```
mfeurer, line 158:
```suggestion
`HPO using OpenML <examples/30_extended/fetch_runtimes.html>`_
```

examples/30_extended/fetch_runtimes.py
```diff
@@ -0,0 +1,132 @@
...
+
+from sklearn.ensemble import RandomForestClassifier
+from sklearn.model_selection import GridSearchCV
+
+
+############################################################################
+# Preparing tasks and scikit-learn models
+# ***************************************
+
+task_id = 1792  # 1778  # supervised classification on mfeat-fourier
```
mfeurer, line 38:
```suggestion
task_id = 1792  # supervised classification on optdigits
```

examples/30_extended/fetch_runtimes.py
```diff
@@ -0,0 +1,132 @@
...
+                    repeat, fold, val2, measures["wall_clock_time_millis_training"][repeat][fold]
+                )
+            )
+        print()
+
+
+############################################################################
+# Case 1: Running a Random Forest model on an OpenML task
+# *******************************************************
+# We'll run a Random Forest model and obtain an OpenML run-object. We can
```
mfeurer, line 69:
```suggestion
# We'll run a Random Forest model and obtain an OpenML run object. We can
```

examples/30_extended/fetch_runtimes.py
```diff
@@ -0,0 +1,132 @@
...
+# to record the time entry for `wall_clock_time_*`. *_millis_testing`
+# follows the same procedure but for time taken for the `.predict()` procedure.
+
+# Comparing the CPU and wall-clock training times of the Random Forest model
+print_compare_runtimes(measures)
+
+############################################################################
+# Case 2: Running a Random Forest model on an OpenML task in parallel
+# *******************************************************************
+# Running the same model on the same task but with parallelism (n_jobs > 1)
```
mfeurer, line 105: Could you please also give an example of setting `n_jobs` for the random forest? That should reduce both the wallclock and the cpu time, right? And could you please also highlight that fact?

examples/30_extended/fetch_runtimes.py
```diff
@@ -0,0 +1,132 @@
...
+
+############################################################################
+# Case 3: Running and benchmarking HPO algorithms with their runtimes
+# *******************************************************************
+# We shall now optimize the same RandomForest model for the same task using
+# scikit-learn HPO support by using GridSearchCV to optimize our earlier
+# RandomForest model's parameter `n_estimators`. sklearn also provides a
+# `refit_time_` for such HPO models, i.e., the time incurred by training
+# and evaluating the model on the best found parameter setting. OpenML
+# includes this in the `wall_clock_time_millis_training` measure recorded.
```
mfeurer, line 121: Could you please also show how to see the fit times stored in the trajectory and the refit time?

tests/test_extensions/test_sklearn_extension/test_sklearn_extension.py
```diff
@@ -2260,3 +2254,15 @@ def column_transformer_pipe(task_id):
         run2 = column_transformer_pipe(23)  # only numeric
         TestBase._mark_entity_for_removal("run", run2.run_id)
         self.assertEqual(run1.setup_id, run2.setup_id)
+
+    def test_for_refit_time_in_basesearchCV(self):
```
mfeurer, line 2258: Could you please explain the difference to the test above?

Neeratyoy: This branch was initially created after checking out from the earlier PR we did which added the refit times. This unit test was removed from there as you pointed out this unit test is testing for scikit-learn. We should instead incorporate this check into an existing unit test, which I resulted in [this unit test](https://github.com/openml/openml-python/blob/fix_895/tests/test_extensions/test_sklearn_extension/test_sklearn_extension.py#L1248).
I did rebase this branch but this function remained.

I'll remove it in the next push.

examples/30_extended/fetch_runtimes.py
Neeratyoy: I had to commit again to remove the inclusion of `refit_time` in the `user_defined_measures` recorded in a run object under `fold_evaluations` since the OpenML server recognises it as an illegal extra field when trying to publish such run objects. 

In order to bypass this and show an example of how one could extract refit time, we could make the following change,
```
# currently
wall_clock_time_millis = wall_clock_time_millis_training + wall_clock_time_millis_testing
# where
wall_clock_time_millis_training = wall_clock_time_millis_training + refit_time
```
we can change to:
```
wall_clock_time_millis = wall_clock_time_millis_training + refit_time + wall_clock_time_millis_testing
```
Then in the example, we can show (although quite verbose) way to extract refit time as,
```python
refit_time = run.fold_evaluations["wall_clock_time_millis"][repeat][fold] - \
     run.fold_evaluations["wall_clock_time_millis_training"][repeat][fold] - \
     run.fold_evaluations["wall_clock_time_millis_testing"][repeat][fold]
```

Wrt fit times for the trace/trajectory, I did add some things based on my interpretation. Would gladly re-iterate or change based on feedback.

examples/30_extended/fetch_runtimes.py
Neeratyoy: Done.

mfeurer (commented):
Hey, I renamed the file so I could render it on my local machine. Also, I left yet another few comments :)

examples/30_extended/fetch_runtimes.py
```diff
@@ -66,11 +70,11 @@ def print_compare_runtimes(measures):
 ############################################################################
 # Case 1: Running a Random Forest model on an OpenML task
 # *******************************************************
-# We'll run a Random Forest model and obtain an OpenML run-object. We can
+# We'll run a Random Forest model and obtain an OpenML run object. We can
 # see the evaluations recorded per fold for the dataset and the information
 # available for this run.
 
-run1 = openml.runs.run_model_on_task(model=clf, task=task)
+run1 = openml.runs.run_model_on_task(model=clf, task=task, dataset_format="array")
```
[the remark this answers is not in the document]
Neeratyoy: I left it like this for a couple of reasons.

Firstly, based on local runs, for certain datasets, the difference in runtimes noticeable `visually` was more pronounced with arrays instead of data frames. Though of course, this is an entirely empirical observation conditioned on my local machine.

Secondly, while testing for various datasets/tasks, the `array` formatting allowed for quick experimentation without complicating the sklearn model/pipeline.

examples/30_extended/fetch_runtimes.py
```diff
@@ -26,22 +26,26 @@
...
 
 
+openml.config.apikey = "7ff6d43e00cc810a00c01bd770996dfc"
+
 ############################################################################
 # Preparing tasks and scikit-learn models
 # ***************************************
 
-task_id = 1792  # 1778  # supervised classification on mfeat-fourier
+task_id = 168908
```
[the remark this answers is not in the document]
Neeratyoy: True...would you have some `medium` sized dataset suggestions?

With small datasets, the difference in runtimes is not that pronounced, which I think makes sense too.
I found some large datasets which make the difference clear but naturally the example takes really long to run.
This dataset appeared to be small in rows but has a lot of features.

I'll try to look for other datasets but after addressing other issues with this PR.

examples/30_extended/fetch_runtimes.py
Neeratyoy: Currently, the example doesn't show how to retrieve `refit_time`. Should I make the change as I detailed above or let it be as it is now?

examples/30_extended/fetch_runtimes.py
```diff
@@ -120,13 +136,124 @@ def print_compare_runtimes(measures):
...
+run4 = openml.runs.run_model_on_task(model=grid_pipe, task=task, n_jobs=2)
+measures = run4.fold_evaluations
 print_compare_runtimes(measures)
+print()
+
+# Like any optimisation problem, scikit-learn's HPO estimators also generate
+# a sequence of configurations which are evaluated as the best found
+# configuration is tracked throughout the trace.
+# The OpenML run object stores these traces as OpenMLRunTrace objects accessible
+# using keys of the pattern (repeat, fold, iterations). Since `iterations` is part
```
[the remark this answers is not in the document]
Neeratyoy: I tried rewording but might need one more reiteration.

examples/30_extended/fetch_runtimes.py
```diff
@@ -120,13 +136,124 @@ def print_compare_runtimes(measures):
...
+# of the scikit-learn model here, the runtime recorded per repeat-per fold is for
+# this entire `fit()` procedure.
+
+# We earlier extracted the number of repeats and folds for this task:
+print("# repeats: {}\n# fold: {}".format(n_repeats, n_folds))
+
+# To extract the training runtime of the first repeat, first fold:
+print(run4.fold_evaluations["wall_clock_time_millis_training"][0][0])
+
+# To extract the training runtime of the 1-st repeat, 4-th fold and also
```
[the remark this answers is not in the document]
Neeratyoy: I tried rewording this too but might need one more reiteration.

examples/30_extended/fetch_runtimes.py
```diff
@@ -0,0 +1,259 @@
...
+# Redefining the model to use all available cores with `n_jobs=-1`
+clf = RandomForestClassifier(n_estimators=10, n_jobs=-1)
+
+run3 = openml.runs.run_model_on_task(model=clf, task=task, dataset_format="array")
+measures = run3.fold_evaluations
+# The wall-clock time recorded per fold should be lesser than the case above,
+# if more than 2 CPU cores are available
+print_compare_runtimes(measures)
+
+# It should be noted that there are multiple levels at which parallelism can
```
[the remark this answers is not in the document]
Neeratyoy: Made a 3 point elaboration. Is this better?

examples/30_extended/fetch_runtimes_tutorial.py
```diff
@@ -88,32 +86,38 @@ def print_compare_runtimes(measures):
...
+run2 = openml.runs.run_model_on_task(
+    model=clf, task=task, n_jobs=4, upload_flow=False, avoid_duplicate_runs=False,
+)
 measures = run2.fold_evaluations
 # The wall-clock time recorded per fold should be lesser than Case 1 above
 print_compare_runtimes(measures)
 
+####################################################################################
+# Running a Random Forest model on an OpenML task in parallel (all cores available):
+
```
mfeurer, line 120: This cell tries to upload the flow implicitly - could you please set `upload_flow=False`? I'm wondering whether we should change the overall behavior here? CC @PGijsbers

examples/30_extended/fetch_runtimes_tutorial.py
```diff
@@ -202,7 +224,7 @@ def print_compare_runtimes(measures):
 )
 run5 = openml.runs.run_model_on_task(model=rs_pipe, task=task)
 
-
+################################################################################
```
mfeurer, line 227: Could you please add a `#` to render the 2nd paragraph into the text block, too?

examples/30_extended/fetch_runtimes_tutorial.py
```diff
@@ -0,0 +1,281 @@
...
+# Running a Random Forest model on an OpenML task in parallel (all cores available):
+
+# Redefining the model to use all available cores with `n_jobs=-1`
+clf = RandomForestClassifier(n_estimators=10, n_jobs=-1)
+
+run3 = openml.runs.run_model_on_task(model=clf, task=task, dataset_format="array")
+measures = run3.fold_evaluations
+# The wall-clock time recorded per fold should be lesser than the case above,
+# if more than 2 CPU cores are available
+print_compare_runtimes(measures)
```
mfeurer, line 128: The results on my local machine are really interesting and I assume similar things will happen on CI, too:
```
Repeat #0-Fold #0: CPU-205.384 vs Wall-2071.654
Repeat #0-Fold #1: CPU-78.619 vs Wall-661.656
Repeat #0-Fold #2: CPU-63.022 vs Wall-520.443
Repeat #0-Fold #3: CPU-66.916 vs Wall-543.427
Repeat #0-Fold #4: CPU-66.177 vs Wall-488.355
Repeat #0-Fold #5: CPU-67.053 vs Wall-495.918
Repeat #0-Fold #6: CPU-65.344 vs Wall-474.387
Repeat #0-Fold #7: CPU-68.329 vs Wall-689.225
Repeat #0-Fold #8: CPU-80.341 vs Wall-637.022
Repeat #0-Fold #9: CPU-83.629 vs Wall-649.633
```
Could you please add a comment on the fact that the CPU time no longer contains the actual computation and can therefore not be used in a meaningful way?

examples/30_extended/fetch_runtimes_tutorial.py
```diff
@@ -0,0 +1,281 @@
...
+#############################################################################
+# It should be noted that there are multiple levels at which parallelism can
+# occur here.
+#
+# * At the outermost level, OpenML tasks contain fixed data splits, on which the
+#   defined model/flow is executed. Thus, a model can be fit on each OpenML dataset fold
+#   in parallel using the `n_jobs` parameter to `run_model_on_task` or `run_flow_on_task`.
+#
+# * The model/flow specified can also include scikit-learn models that perform their own
+#   parallelisation. For instance, by specifying `n_jobs` in the Random Forest model definition.
```
mfeurer, line 139:
```suggestion
#   parallelization. For instance, by specifying `n_jobs` in the Random Forest model definition
#   as done in this section of the example.
```

examples/30_extended/fetch_runtimes_tutorial.py
```diff
@@ -0,0 +1,281 @@
+"""
+=================
+Fetching Runtimes
+=================
+
+The runtime of machine learning models on specific datasets can be a deciding
+factor on the choice of algorithms. Especially for benchmarking and comparison
```
mfeurer, line 7:
```suggestion
factor on the choice of algorithms, especially for benchmarking and comparison
```

examples/30_extended/fetch_runtimes_tutorial.py
```diff
@@ -0,0 +1,281 @@
...
+run5 = openml.runs.run_model_on_task(model=rs_pipe, task=task)
+
+################################################################################
+# Since for the call to `openml.runs.run_model_on_task` the parameter
+# `n_jobs` is set to its default None, the evaluations across the OpenML folds
+# are not parallelized. Hence, the time recorded is agnostic to the `n_jobs`
+# being set at both the HPO estimator `GridSearchCV` as well as the base
+# estimator `RandomForestClassifier` in this case. OpenML only records the
+# time taken for the completion of the complete `fit()` call, per-repeat per-fold.
+
```
mfeurer, line 234:
```suggestion
#
```

examples/30_extended/fetch_runtimes_tutorial.py
```diff
@@ -0,0 +1,281 @@
...
+#
+# * The sklearn model can further be an HPO estimator and contain it's own parallelisation.
+#   If the base estimator used supports `parallelisation`, then there's a 2-level nested definition
+#   for parallelisation possible. Such an example is explored next.
+
+#####################################################################
+# Case 3: Running and benchmarking HPO algorithms with their runtimes
+# *******************************************************************
+# We shall now optimize a similar RandomForest model for the same task using
+# scikit-learn HPO support by using GridSearchCV to optimize our earlier
```
mfeurer, line 149:
```suggestion
# scikit-learn's HPO support by using GridSearchCV to optimize our earlier
```

examples/30_extended/fetch_runtimes_tutorial.py
```diff
@@ -0,0 +1,281 @@
...
+# RandomForest model's parameter `n_estimators`. scikit-learn also provides a
```
mfeurer, line 150:
```suggestion
# RandomForest model's hyperparameter `n_estimators`. scikit-learn also provides a
```

examples/30_extended/fetch_runtimes_tutorial.py
```diff
@@ -0,0 +1,281 @@
...
+    estimator=clf,
+    param_grid={"n_estimators": np.linspace(start=1, stop=50, num=n_iter).astype(int).tolist()},
+    cv=2,
+    n_jobs=2,
+)
+
+run4 = openml.runs.run_model_on_task(model=grid_pipe, task=task, n_jobs=2)
+measures = run4.fold_evaluations
+print_compare_runtimes(measures)
+print()
```
mfeurer, line 169:
```suggestion
```

This is no longer necessary as you're starting a new cell afterwards

examples/30_extended/fetch_runtimes_tutorial.py
```diff
@@ -0,0 +1,281 @@
...
+    "Total runtime for repeat {}'s fold {}: {:4f} ms".format(
+        _repeat, _fold, run4.fold_evaluations["wall_clock_time_millis_training"][_repeat][_fold]
+    )
+)
+for i in range(n_iter):
+    key = (_repeat, _fold, i)
+    r = run4.trace.trace_iterations[key]
+    print(
+        "n_estimators: {:>2} - score: {:.3f}".format(
+            r.parameters["parameter_n_estimators"], r.evaluation
```
mfeurer, line 205: Do we also have the fit time for the individual model or only the score?

examples/30_extended/fetch_runtimes.py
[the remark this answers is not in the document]
mfeurer: Maybe `adult` is better as it has less features? Or `credit-g`, but it might be too small?

examples/30_extended/fetch_runtimes.py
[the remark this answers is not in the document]
mfeurer: Looks better, but I still miss the word `inner cross-validation`

examples/30_extended/fetch_runtimes_tutorial.py
```diff
@@ -0,0 +1,281 @@
...
+# configuration is tracked throughout the trace.
+# The OpenML run object stores these traces as OpenMLRunTrace objects accessible
+# using keys of the pattern (repeat, fold, iterations). Here `fold` implies the
+# outer-cross validation fold as obtained from the task data splits available in OpenML.
+# Since we are running `GridSearchCV` here which itself performs a `2-fold` cross validation,
+# the runtime recorded per repeat-per fold in the run object is for the entire `fit()`
+# procedure of GridSearchCV thus subsuming the runtimes of the 2-fold CV search performed.
+
+# We earlier extracted the number of repeats and folds for this task:
+print("# repeats: {}\n# fold: {}".format(n_repeats, n_folds))
```
mfeurer, line 183:
```suggestion
print("# repeats: {}\n# folds: {}".format(n_repeats, n_folds))
```

examples/30_extended/fetch_runtimes.py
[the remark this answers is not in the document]
mfeurer: Also, I think the differences don't need to be super-pronounced, it suffices if they're there.

examples/30_extended/fetch_runtimes.py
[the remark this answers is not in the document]
mfeurer: Yes, I'm currently just wondering whether this is the best spot to put this information or whether there's a nicer place.

examples/30_extended/fetch_runtimes.py
mfeurer: I think it would be great if we can add this

mfeurer:
Hey, I think the example is really getting along well. Based on the current status I'm wondering three things:

1. Should we have a summary section which summarizes the behavior of OpenML-Python, scikit-learn and joblib?
2. Should we have a section which shows that the measured numbers will differ based on the backend?
3. We currently don't have case 2 from https://github.com/openml/openml-python/issues/895#issuecomment-770889839

examples/30_extended/fetch_runtimes.py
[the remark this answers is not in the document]
Neeratyoy: What if we move these 3 points to the top before we list the 3 cases we show in the subsequent example?

examples/30_extended/fetch_runtimes.py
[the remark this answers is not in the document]
Neeratyoy: Will add with the next commit.

examples/30_extended/fetch_runtimes_tutorial.py
Neeratyoy: Aah I realize that I am not being consistent with the call signature to `run_model_on_task` in this cell compares to the previous ones. Here, I am not passing `upload_flow` and `avoid_duplicate_runs`. 
The main reason I believe I did that in the previous two cells is to make it explicit and acquaint the user with these parameters, given this is a tutorial.
I'll make the call consistent here.

However, `upload_flow=False` is the default `run_model_on_task`.

examples/30_extended/fetch_runtimes_tutorial.py
Neeratyoy: Done.

examples/30_extended/fetch_runtimes.py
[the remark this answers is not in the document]
Neeratyoy: Tried this: https://www.openml.org/d/41027

Brought the total file runtime down by half to under 2 minutes...

examples/30_extended/fetch_runtimes_tutorial.py
Neeratyoy: Added a bit of snippet.

examples/30_extended/fetch_runtimes_tutorial.py
Neeratyoy: No, I wanted to ask you the same 😅 

However, to answer you, nope! I couldn't find any aspect of the runtime being captured by the `OpenMLRunTrace` class definition.

Neeratyoy:
> 1. Should we have a summary section which summarizes the behavior of OpenML-Python, scikit-learn and joblib?

Do you mean their interaction and behaviour? It would be useful I guess, but the only concern or question is what (exactly) to summarize I guess.

> 2. Should we have a section which shows that the measured numbers will differ based on the backend?

Though we could, from the OpenML API standpoint, we can also ignore it. If the duty of parallelization is delegated to scikit-learn or OpenML (as we show in the example), I feel it is likely that the user may not set the backend with a `parallel_backend` context. 
Having said that, it also might be fair to show a 4th case in the example, giving an example of changing the backend.

On second thought, this might necessitate the mention of how the backends change in the entire stack of function calls going through OpenML to scikit-learn pipelines. This is obviously quite complicated.

> 3. We currently don't have case 2 from [#895 (comment)](https://github.com/openml/openml-python/issues/895#issuecomment-770889839)

In the latest [SGDClassifier documentation](https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.SGDClassifier.html), the parallelization happens through `joblib` which I thought we are already covering. 
While the [HistGradientBoosting](https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.HistGradientBoostingClassifier.html), still appears to be *experimental*, so didn't quite think about including it..

examples/30_extended/fetch_runtimes.py
Neeratyoy: Done.

examples/30_extended/fetch_runtimes_tutorial.py
mfeurer: Alright, just added this request to https://github.com/openml/OpenML/issues/413

examples/30_extended/fetch_runtimes.py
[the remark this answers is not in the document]
mfeurer: Sounds good to me.

examples/30_extended/fetch_runtimes_tutorial.py
mfeurer: But we do know the number of cores here, don't we? We have the case that subprocesses are started and they are not recorded in the CPU time, right?

mfeurer:
> Do you mean their interaction and behaviour? It would be useful I guess, but the only concern or question is what (exactly) to summarize I guess.

I think the main caveats to pay attention to, like the gist of https://github.com/openml/openml-python/issues/895

> Though we could, from the OpenML API standpoint, we can also ignore it.

Probably I mixed up something. But an example on how this can be changed might be a good idea!

> In the latest SGDClassifier documentation, the parallelization happens through joblib which I thought we are already covering.
While the HistGradientBoosting, still appears to be experimental, so didn't quite think about including it..

Yes, but that's for fitting multiple classifiers for a "one vs all" scheme. I was more thinking about something like the neural network where the number of used cores can't be set via the API.

examples/30_extended/fetch_runtimes_tutorial.py
```diff
@@ -11,13 +12,33 @@
...
+
+* The model/flow specified can also include scikit-learn models that perform their own
+  parallelization. For instance, by specifying `n_jobs` in a Random Forest model definition
+  (covered under Case 2 below).
+
+* The sklearn model can further be an HPO estimator and contain it's own parallelization.
+  If the base estimator used also supports `parallelization`, then there's at least a 2-level nested
+  definition for parallelization possible (covered under Case 3 below).
+
+We shall cover these 3 representative scenarios for:
```
mfeurer, line 30:
```suggestion
We shall cover these 4 representative scenarios for:
```

examples/30_extended/fetch_runtimes_tutorial.py
```diff
@@ -89,23 +116,26 @@ def print_compare_runtimes(measures):
...
 # to record the time entry for `wall_clock_time_*`. The `*_millis_testing` entry
 # follows the same procedure but for time taken for the `.predict()` procedure.
 
 # Comparing the CPU and wall-clock training times of the Random Forest model
 print_compare_runtimes(measures)
 
 ######################################################################
-# Case 2: Running a Random Forest model on an OpenML task in parallel
+# Case 2: Running Scikit-learn models on an OpenML task in parallel
 # ********************************************************************
```
mfeurer, line 136:
```suggestion
# *****************************************************************
```

examples/30_extended/fetch_runtimes_tutorial.py
```diff
@@ -325,3 +357,73 @@ def get_incumbent_trace(trace):
...
+
+dt = DecisionTreeClassifier()
+run6 = openml.runs.run_model_on_task(
+    model=dt, task=task, upload_flow=False, avoid_duplicate_runs=False, n_jobs=2
+)
+measures = run6.fold_evaluations
+print_compare_runtimes(measures)
+
+################################################################################
+# Running a Neural Network from scikit-learn that uses scikit-learn independent
```
mfeurer, line 379: Can we maybe add a link to the documentation in scikit-learn?

examples/30_extended/fetch_runtimes_tutorial.py
```diff
@@ -325,3 +357,73 @@ def get_incumbent_trace(trace):
...
+run7 = openml.runs.run_model_on_task(
+    model=mlp, task=task, upload_flow=False, avoid_duplicate_runs=False
+)
+measures = run7.fold_evaluations
+print_compare_runtimes(measures)
+
+################################################################################
+# Summmary
+# *********
+# OpenML records model runtimes for the CPU-clock and the wall-clock times. The above
```
mfeurer, line 392: Hm, should we explicitly say "the scikit-learn extension"? Everything we're writing here is exclusive about the scikit-learn extension, so it could be confusing otherwise.

examples/30_extended/fetch_runtimes_tutorial.py
```diff
@@ -325,3 +357,73 @@ def get_incumbent_trace(trace):
...
+#   * Reports the total time for performing search on each of the OpenML data split, subsuming
+#     any sort of parallelism that happened as part of the HPO estimator or the underlying
+#     base estimator
+#   * Also allows extraction of the `refit_time` that scikit-learn measures using `time.time()`
+#     for retraining the model per outer fold, for the best found configuration
+#
+# * `CPU-time` & `wallclock-time` for models that scikit-learn doesn't parallelize
+#
+#   * Models like Decision Trees don't parallelize and thus both the wallclock and CPU times are
+#     accurate runtime representations as long as `n_job=1` or `n_jobs=None` at the OpenML call
```
mfeurer, line 426: Are you sure? Shouldn't it be accurate even when `n_job>1` as the time is measured inside each fold independently?

examples/30_extended/fetch_runtimes_tutorial.py
```diff
@@ -325,3 +357,73 @@ def get_incumbent_trace(trace):
...
+#   * Scikit-learn Neural Networks can undergo parallelization implicitly owing to thread-level
+#     parallelism involved in the linear algebraic operations and thus only the wallclock-time
```
mfeurer, line 428: I'm not sure if I agree here. The wallclock-time is what the user has to wait, but the CPU time can still be higher (depending on the number of parallel threads). How about saying that the wallclock time and cpu time can differ?

examples/30_extended/fetch_runtimes_tutorial.py
```diff
@@ -130,26 +160,25 @@ def print_compare_runtimes(measures):
 # larger datasets.
 print_compare_runtimes(measures)
 
+####################################################################################
+# Running the same task with a different parallel backend. Joblib provides multiple
+# backends: {`loky` (default), `multiprocessing`, `dask`, `threading`, `sequential`}.
+# The backend can be explicitly set using a joblib context manager. The behaviour of
+# the job distribution can change and therefore the scale of runtimes recorded too.
+
+with parallel_backend(backend="multiprocessing", n_jobs=-1):
```
mfeurer, line 169: Is the `-1` intentional? I think it parallelizes the outer fold and therefore runs sequentially on the inner fold, making the CPU and wallclock time identical.

examples/30_extended/fetch_runtimes_tutorial.py
Neeratyoy: Yes, that's true. I'll remove the mention of `n_jobs`.

Neeratyoy: I agree, shouldn't make any difference for Decision Trees and have similar CPU and wallclock times.

examples/30_extended/fetch_runtimes_tutorial.py
Neeratyoy: Alright, made the change.

examples/30_extended/fetch_runtimes_tutorial.py
Neeratyoy: The `-1` was intentional.

What would `inner folds` imply for this case? My understanding was that OpenML's parallel run would be running multiple RFs on the different OpenML data splits, while each RF fit would be distributing its trees too?

On my machine, the runtimes don't look `identical`, in fact quite erratic.
Screenshot from 2021-04-08 15-45-06

Based on the analysis we did earlier, the first table shows that the inner context does change into a `threading` backend, but somehow the Case 2 table suggests that speed-up does happen? 
multiprocessing-full

What do you suggest I do here?
To be honest, the use of a context manager and the resulting behaviour does confuse me.

examples/30_extended/fetch_runtimes_tutorial.py
```diff
@@ -133,7 +133,7 @@ def print_compare_runtimes(measures):
 
 ######################################################################
 # Case 2: Running Scikit-learn models on an OpenML task in parallel
```
mfeurer, line 135:
```suggestion
# Case 2: Running Scikit-learn model on an OpenML task in parallel
```

examples/30_extended/fetch_runtimes_tutorial.py
```diff
@@ -389,9 +396,10 @@ def get_incumbent_trace(trace):
 ################################################################################
 # Summmary
 # *********
-# OpenML records model runtimes for the CPU-clock and the wall-clock times. The above
-# examples illustrated how these recorded runtimes can be extracted, especially when
-# using a scikit-learn model and under parallel setups too. To summarize, OpenML measures the:
+# OpenML records model runtimes for the CPU-clock and the wall-clock times for
```
mfeurer, line 399: There was a comment which got resolved which read like

> Hm, should we explicitly say "the scikit-learn extension"? Everything we're writing here is exclusive about the scikit-learn extension, so it could be confusing otherwise.

and I cannot add text there any more, but I don't think it's resolved yet. It still says OpenML everywhere, though this is only about the scikit-learn extension.

examples/30_extended/fetch_runtimes_tutorial.py
mfeurer: So if you do threading (because of the fallback), jobs which leave the GIL (such as the decision tree) can still run in parallel...super confusing. I guess we need to add another case? Do you know of an algorithm that does neither use blas nor the joblib? A linear SVM?

mfeurer:
I still can't annotate all my comments... anyway, linear SVM also releases the GIL: https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/svm/_liblinear.pyx#L61

Maybe [naive bayes](https://scikit-learn.org/stable/modules/generated/sklearn.naive_bayes.GaussianNB.html#sklearn.naive_bayes.GaussianNB) doesn't release the GIL?

mfeurer (approved):

PR #1049, opened 2021-04-08
Configuration ci

### LINKED ISSUE #1045: Add a command line interface for (API key) configuration

PGijsbers:
Closes #1045 

Adding a CLI which can be used to configure the fields in the configuration file.
This is a draft PR because it's feature incomplete (some fields are missing), but I wanted some input to see if this is an acceptable way to do it.

Installing openml with `pip install openml` will also add a cli script `openml`. It'll expect something like `openml ROUTINE ROUTINE_ARGS`, the current routine is only `configure`, but I figured some day we might use it for e.g. dataset download.
The signature of `openml configure` looks like this:

```
usage: openml configure [-h] [{apikey,server,cachedir,avoid_duplicate_runs,connection_n_retries,max_retries,all,none}]

Set or read variables in your configuration file. For more help also see 'https://openml.github.io/openml-
python/master/usage.html#configuration'.

positional arguments:
  {apikey,server,cachedir,avoid_duplicate_runs,connection_n_retries,max_retries,all,none}
                        The field you wish to edit.Choosing 'all' lets you configure all fields one by one.Choosing
                        'none' will print out the current configuration.

optional arguments:
  -h, --help            show this help message and exit
```

[Video](https://www.loom.com/share/8e106734f0fc4947b59d74619782b460) of current functionality.
~I'm also planning for a direct option that avoids the explanation/interactive routine (e.g. `openml config apikey 1234567890abcdef1234567890abcdef`).~
Update: It's also possible to use a non-interactive direct command, e.g.: `openml configure apikey 1234567890abcdef1234567890abcdef`.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -5,6 +5,7 @@ select = C,E,F,W,B,T
 ignore = E203, E402, W503
 per-file-ignores =
     *__init__.py:F401
+    *cli.py:T001
 exclude =
     venv
     examples
--- a/doc/progress.rst
+++ b/doc/progress.rst
@@ -10,6 +10,7 @@ Changelog
 ~~~~~~
 
 * FIX #1035: Render class attributes and methods again.
+* ADD #1049: Add a command line tool for configuration openml-python.
 
 0.12.0
 ~~~~~~
--- a/doc/usage.rst
+++ b/doc/usage.rst
@@ -59,6 +59,10 @@ which are separated by newlines. The following keys are defined:
     * 1: info output
     * 2: debug output
 
+This file is easily configurable by the ``openml`` command line interface.
+To see where the file is stored, and what its values are, use `openml configure none`.
+Set any field with ``openml configure FIELD`` or even all fields with just ``openml configure``.
+
 ~~~~~~~~~~~~
 Key concepts
 ~~~~~~~~~~~~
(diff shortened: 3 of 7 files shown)


PGijsbers:
I do notice some downsides to using the `configparser`, e.g. you can't see which values are explicitly set or which are default (if the explicitly set value matches the default value). It's also seems to complicate the code. Considering our configurations haven't gotten any more complicated for years, I think we should be safe to assume the simple `field = value` format will last us a while longer, and it might be more easily readable to just write a custom parser for it. It should not take more than a few lines of codes, e.g.

```python
set_fields ={}
with open(file, 'r') as fh:
    for line in fh.readlines():
        field, value = line.split('=', 1)
        set_fields[field.strip()] = value.strip()
config = {**_defaults, **set_fields}
```

Not entirely sure if I want to do that right now (I'd like this CLI done before my leave next week), but do you think we should consider that?

PGijsbers:
Remarks while working on this:
 - `max_retries` is contained in `_defaults` but seems like it should be set by developers only. After discussing we probably want to avoid having such a number at all as long as we employ a decent retry policy (that waits increasingly long times).
 - the `configparser` note from above.
 - `verbosity` is currently not used in `openml-python`, I think this is a bug introduced by #987.

mfeurer (approved):

PR #1053, opened 2021-04-14
Fix temp directory creation on container

mfeurer:
#### Reference Issue

Fixes a bug which could occur when using OpenML-Python in a docker container. Initially reported as https://github.com/conda-forge/openml-feedstock/pull/1#issuecomment-817332813 by @bastianzim

When OpenML-Python was trying to create a cache directory as `/home/${USER}/.config/openml` it would only try to create the `openml` directory. If the `.config` directory was also missing, it would fail.

This PR creates both directories.

New unit test, or in a docker container built with:
```
apt update
apt upgrade
apt install python3-dev ipython3 python3-pip git
pip3 install openml
```

### DIFF
--- a/openml/config.py
+++ b/openml/config.py
@@ -204,7 +204,7 @@ def _setup(config=None):
     # read config file, create directory for config file
     if not os.path.exists(config_dir):
         try:
-            os.mkdir(config_dir)
+            os.makedirs(config_dir, exist_ok=True)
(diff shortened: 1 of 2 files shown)


BastianZim:
Thank you!

ArlindKadra (approved):

joaquinvanschoren (approved):

PR #1056, opened 2021-04-14
Speed up dataset unit tests

mfeurer:
by only loading necessary datasets

### DIFF
--- a/tests/test_datasets/test_dataset.py
+++ b/tests/test_datasets/test_dataset.py
@@ -22,16 +22,6 @@ class OpenMLDatasetTest(TestBase):
         super(OpenMLDatasetTest, self).setUp()
         openml.config.server = self.production_server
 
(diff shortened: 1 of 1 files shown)


PGijsbers (commented):
Wouldn't it be better to just work with a lazy-load property?
e.g.
```python
self._titanic = None

@property
def titantic(self):
     if self._titanic is None:
         self._titanic = download..
     return self._titanic
```
I guess the fact that it requires less changes is a bit moot at this point, but it does make more use of code re-use (which is more maintainable and less error prone).

mfeurer:
> I guess the fact that it requires less changes is a bit moot at this point, but it does make more use of code re-use (which is more maintainable and less error prone).

Yes, I agree on code reuse. It'll also be easier if we at one point move to pytest.

PGijsbers (approved):
Exactly :) Conversion to fixtures should be no issue at all.

Issue #1057, opened 2021-04-15
Task fetching labels as numpy when expected dataframe


Neeratyoy:
#### Steps/Code to Reproduce

```python
import openml
task = openml.tasks.get_task(12731)
X, y = task.get_X_and_y(dataset_format="dataframe")
print(type(X))  # prints pandas DataFrame
print(type(y))  # prints numpy array
```

#### Expected Results
```python
print(type(y))
```
`y` is a pandas Series.

#### Actual Results

```python
print(type(y))
```
`y` is a numpy array.

#### Versions

OpenML `'0.12.1'

PR #1059, opened 2021-04-15
Convert sparse labels to pandas series

Neeratyoy:
#### Reference Issue

Addresses #1057.

Adds a check when fetching and converting datasets that converts labels as sparse arrays to dataframes.

```python
import openml
task = openml.tasks.get_task(12731)
_, y = task.get_X_and_y(dataset_format="dataframe")
print(type(y))  # print pandas Series
```


openml/datasets/dataset.py
```diff
@@ -629,6 +629,8 @@ def _encode_if_category(column):
         elif array_format == "dataframe":
             if scipy.sparse.issparse(data):
                 return pd.DataFrame.sparse.from_spmatrix(data, columns=attribute_names)
+            elif isinstance(data, np.ndarray):
+                return pd.Series(data)
```
mfeurer, line 633: Hey, what's the reason not to apply `columns=attribute_names` here?

mfeurer: Also, how can this become a numpy array in the first place? If I understand this correctly, this is called by `get_data`, which splits off `y` by doing `y = data.iloc[:, targets]` from which point on it should be and stay a series, right?

tests/test_tasks/test_task_functions.py
```diff
@@ -234,3 +234,9 @@ def test_deletion_of_cache_dir(self):
         self.assertTrue(os.path.exists(tid_cache_dir))
         openml.utils._remove_cache_dir_for_id("tasks", tid_cache_dir)
         self.assertFalse(os.path.exists(tid_cache_dir))
+
+    def test_supervised_task_target_format(self):
```
mfeurer, line 238: Why is this a test for a task and not for the respective dataset?

mfeurer: Also, do we have sufficient tests that the data is loaded correctly for sparse datasets?

Neeratyoy: I added this unit test to datasets: https://github.com/openml/openml-python/blob/fix_1057/tests/test_datasets/test_dataset.py#L306

and removed this unit test from tasks as this new unit test makes the one under tasks redundant
I also checked this unit test with the dataset used in the #1057 and this unit test passes for it too as it is a more generalized test given the changes made in this PR

tests/test_datasets/test_dataset.py
```diff
@@ -303,7 +303,22 @@ def test_get_sparse_dataset_with_target(self):
...
+        self.assertEqual(X.shape, (600, 20000))
+
+        self.assertIsInstance(y, pd.Series)
+        self.assertIsInstance(y.dtypes, pd.SparseDtype)
+        self.assertEqual(y.shape, (600,))
+
+        self.assertEqual(len(attribute_names), 20000)
+        self.assertNotIn("class", attribute_names)
+
+    def test_get_sparse_dataset_array(self):
```
mfeurer, line 321: As we check for the format of `y` in the new test above, maybe we should also check for the format of `y` here?

tests/test_datasets/test_dataset.py
```diff
@@ -303,7 +303,22 @@ def test_get_sparse_dataset_with_target(self):
...
+    def test_get_sparse_dataset_dataframe_with_target(self):
+        X, y, _, attribute_names = self.sparse_dataset.get_data(
+            dataset_format="dataframe", target="class"
+        )
+        self.assertIsInstance(X, pd.DataFrame)
+        self.assertIsInstance(X.dtypes[0], pd.SparseDtype)
+        self.assertEqual(X.shape, (600, 20000))
+
+        self.assertIsInstance(y, pd.Series)
+        self.assertIsInstance(y.dtypes, pd.SparseDtype)
```
mfeurer, line 315: Do you know if scikit-learn can handle `pd.SparseDtype`?

mfeurer (approved):

Issue #1028, opened 2021-02-22
Public `delete` methods


PGijsbers:
Currently `openml-python` only allows for the deletion of `studies` but the rest API (and our internal `openml.utils._delete_entity`) support the deletion of other entities as well (dataset, flow, task, run, user). ~I don't know if this was deliberate, but if it wasn't, they should be added.~

mfeurer:
This was a decision made by me in the early days of OpenML-Python. I think that decision should be reverted and these functions should be implemented.

Mirkazemi:
I'm working on the issue but I could not assign it to myself.

PR #1060, opened 2021-04-20
Issue 1028: public delete functions for run, task, flow and database

Mirkazemi:
#### Reference Issue
Public delete methods #1028

The new functions allow the user to delete run, task, flow and database using API:
openml.runs.delete_run()
openml.tasks.delete_task()
openml.flows.delete_flow()
openml.datasets.delete_dataset()

The unit tests are added to the 
tests/test_runs/test_run_functions.py
tests/test_tasks/test_task_functions.py
tests/test_flows/test_flow_functions.py
tests/test_datasets/test_dataset_functions.py


PGijsbers (commented):
@mfeurer Should we want additional error wrapping?

E.g. deleting a task with runs says:

```
>>> openml.tasks.delete_task(4263)
  File "E:\repositories\openml-python\openml\tasks\functions.py", line 537, in delete_task
    return openml.utils._delete_entity("task", task_id)
  File "E:\repositories\openml-python\openml\utils.py", line 175, in _delete_entity
    result_xml = openml._api_calls._perform_api_call(url_suffix, "delete")
  File "E:\repositories\openml-python\openml\_api_calls.py", line 61, in _perform_api_call
    response = __read_url(url, request_method, data)
  File "E:\repositories\openml-python\openml\_api_calls.py", line 160, in __read_url
    return _send_request(
  File "E:\repositories\openml-python\openml\_api_calls.py", line 192, in _send_request
  File "E:\repositories\openml-python\openml\_api_calls.py", line 230, in __check_response
    raise __parse_server_exception(response, url, file_elements=file_elements)
openml.exceptions.OpenMLServerException: https://www.openml.org/api/v1/xml/task/4263 returned code 454: Task is executed in some runs. Delete these first - None
```

Deleting something not owned by you:
```
  File "E:\repositories\openml-python\openml\runs\functions.py", line 1208, in delete_run
    return openml.utils._delete_entity("run", run_id)
  File "E:\repositories\openml-python\openml\utils.py", line 175, in _delete_entity
    result_xml = openml._api_calls._perform_api_call(url_suffix, "delete")
  File "E:\repositories\openml-python\openml\_api_calls.py", line 61, in _perform_api_call
    response = __read_url(url, request_method, data)
  File "E:\repositories\openml-python\openml\_api_calls.py", line 160, in __read_url
    return _send_request(
  File "E:\repositories\openml-python\openml\_api_calls.py", line 192, in _send_request
  File "E:\repositories\openml-python\openml\_api_calls.py", line 230, in __check_response
    raise __parse_server_exception(response, url, file_elements=file_elements)
openml.exceptions.OpenMLServerException: https://www.openml.org/api/v1/xml/run/10559886 returned code 393: Run is not owned by you - None
```

The messages are pretty clear though the stack trace perhaps a bigger than expected by an end-user. We could also use e.g. [`PermissionError`](https://docs.python.org/3/library/exceptions.html#PermissionError) for trying to delete others' entities.

@Mirkazemi Please include tests for common errors (deleting something that is not yours, deleting an entity with attached other entities (e.g. task with runs)). I suggest you wait with this until Matthias has chimed in on whether or not the exceptions should be wrapped.

mfeurer:
I agree, we could add two new exceptions for the cases you mentioned above to make it easier for the user to understand what's going on and automatically parse the exceptions.

tests/test_runs/test_run_functions.py
```diff
@@ -1686,3 +1691,22 @@ def test_joblib_backends(self, parallel_mock):
...
+    def test_delete_run(self):
+        rs = 1
+        clf = sklearn.pipeline.Pipeline(
+            steps=[("imputer", SimpleImputer()), ("estimator", DecisionTreeClassifier())]
+        )
+        task = openml.tasks.get_task(32)  # diabetes; crossvalidation
+
+        run = openml.runs.run_model_on_task(model=clf, task=task, seed=rs)
+        run.publish()
+        TestBase._mark_entity_for_removal("run", run.run_id)
```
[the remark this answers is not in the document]
Mirkazemi: I agree so I'll remove the _mark_entity_for_removal().

PGijsbers:
@mfeurer Can you have a quick look at the progress? In particular the testing against cached `xml` responses from the server. See for example the [flow tests](https://github.com/Mirkazemi/openml-python/blob/230035b00e93cfccf7f86f89355f6a126682456d/tests/test_flows/test_flow_functions.py#L411-L522), which feature both an integration test that Mirkazemi had added, but below that also multiple new tests against cached responses. 
The downside of the integrations tests are plenty:
 - they can be flaky as they depend on server stability
 - they are slower (by nature of having to communicate with the server)
 - and require either an extensive setup to keep independent from other tests (e.g., creating both a new task and a run to test for `task can not be deleted because of associated runs`), or would create dependencies (e.g., trying to delete already existing tasks/tasks generated by other tests, which is both hard to orchestrate and in case of server bugs might even result in unintended loss of data).
 
With the new server implementation, the server tests themselves will be responsible for ensuring the right behaviour for provided input. However, this will take a while to roll out (and for openml-python to use), so there is also a pro to actually doing integration tests (the `openml-python` tests are also used as server tests for now). My questions are thus:
 1. To what extend do you think we should keep/add integration tests (if any)?
 2. What do you think about the mocked xml response tests, and how they are currently structured?

tests/test_datasets/test_dataset_functions.py
```diff
@@ -1672,3 +1674,130 @@ def test_valid_attribute_validations(default_target_attribute, row_id_attribute,
...
+        self.assertTrue(openml.datasets.delete_dataset(_dataset_id))
+
+
+@mock.patch.object(requests.Session, "delete")
+def test_delete_dataset_not_owned(mock_get, test_files_directory):
+    openml.config.start_using_configuration_for_example()
+    content_file = (
+        test_files_directory / "mock_responses" / "datasets" / "data_delete_not_owned.xml"
+    )
+    mock_get.return_value = create_request_response(status_code=353, content_filepath=content_file)
```
mfeurer, line 1731: From my understanding of the docs, this should be HTTP error code `412`, and the OpenML error code `353` should only be used internally. If this correct, this needs to be changed for every function.

PGijsbers: You are right. I updated the tests. The reason the tests pass because `openml-python` does not look at the status code except that it is `!= 200`.

PGijsbers: But otherwise you're good with this testing approach?

tests/conftest.py
```diff
@@ -186,3 +187,8 @@ def pytest_addoption(parser):
 @pytest.fixture(scope="class")
 def long_version(request):
     request.cls.long_version = request.config.getoption("--long")
+
+
+@pytest.fixture
+def test_files_directory() -> pathlib.Path:
+    return pathlib.Path(__file__).parent / "files"
```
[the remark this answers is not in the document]
mfeurer: You're right, my bad.

tests/test_datasets/test_dataset_functions.py
mfeurer: In general yes. I just don't know how we can check for such deviations from the API :( Maybe at some point we need to add a test suite that checks that the answers from the server are still the same XML that we expect :D

tests/test_datasets/test_dataset_functions.py
```diff
@@ -1672,3 +1674,130 @@ def test_valid_attribute_validations(default_target_attribute, row_id_attribute,
...
+            original_data_url=original_data_url,
+            paper_url=paper_url,
+        )
+        dataset.publish()
+        _dataset_id = dataset.id
+        self.assertTrue(openml.datasets.delete_dataset(_dataset_id))
+
+
+@mock.patch.object(requests.Session, "delete")
+def test_delete_dataset_not_owned(mock_get, test_files_directory):
```
mfeurer, line 1726: How about naming the first argument `mock_request`?

tests/test_datasets/test_dataset_functions.py
```diff
@@ -1672,3 +1674,130 @@ def test_valid_attribute_validations(default_target_attribute, row_id_attribute,
...
+
+    with pytest.raises(
+        OpenMLNotAuthorizedError,
+        match="The data can not be deleted because it was not uploaded by you.",
+    ):
+        openml.datasets.delete_dataset(40_000)
+
+    expected_call_args = [
+        ("https://test.openml.org/api/v1/xml/data/40000",),
+        {"params": {"api_key": "c0c42819af31e706efe1f4b88c23c6c1"}},
```
mfeurer, line 1741: Can we read the API key from `testing.py` or somehow else insert it, maybe via a fixture?

tests/test_flows/test_flow_functions.py
```diff
@@ -403,3 +407,116 @@ def test_get_flow_id(self):
...
+        flow.publish()
+        _flow_id = flow.flow_id
+        self.assertTrue(openml.flows.delete_flow(_flow_id))
+
+
+@mock.patch.object(requests.Session, "delete")
+def test_delete_flow_not_owned(mock_get, test_files_directory):
+    openml.config.start_using_configuration_for_example()
+    content_file = test_files_directory / "mock_responses" / "flows" / "flow_delete_not_owned.xml"
+    mock_get.return_value = create_request_response(status_code=323, content_filepath=content_file)
```
mfeurer, line 437: What is status code 323?

mfeurer (commented):
A few minor things, but besides that, this PR looks good to me.

tests/test_flows/test_flow_functions.py
PGijsbers: A sign that I forgot to push my last commit ;) It will be 412. 
`323` is the corresponding OpenML code that's embedded in the XML.

tests/test_datasets/test_dataset_functions.py
```diff
@@ -1672,3 +1674,130 @@ def test_valid_attribute_validations(default_target_attribute, row_id_attribute,
...
+    ]
+    assert expected_call_args == list(mock_get.call_args)
+
+
+@mock.patch.object(requests.Session, "delete")
+def test_delete_dataset_with_run(mock_get, test_files_directory):
```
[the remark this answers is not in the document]
PGijsbers: Yes, my bad ^^"

PGijsbers:
Thanks! I'll process the feedback later and add similar tests for the other entities. I'll ping you when I'm done :)

tests/test_datasets/test_dataset_functions.py
PGijsbers: Renamed it to `mock_delete` since it mocks the `delete` method.

mfeurer (approved):

PGijsbers:
Thanks @Mirkazemi for starting this PR 🎉

Issue #1066, opened 2021-04-28
Update documentation


ArlindKadra:
https://github.com/openml/openml-python/blob/62014cdb80fe7a19d105abd70d999edc8e84c817/openml/datasets/functions.py#L370-L371

It would be good to update the argument description to mention that it also takes a dataset name like in `get_datasets`. I did not notice it the first time I looked around.

Issue #1063, opened 2021-04-26
Use name instead of interger ids


ArlindKadra:
#### Description
Use names instead of integer ids when possible for better readability (e.g., for datasets at least it is possible to use names of the dataset).

PGijsbers:
Should we consider stating the id in the tutorial text/comments? That way if the dataset is not correctly found by name (e.g. because a newer version is uploaded which breaks the example), users can refer to the id?

ArlindKadra:
Hey @PGijsbers, I had the same thought and I fully agree with it. I will do that too.

ArlindKadra:
```
    If dataset is retrieved by name, a version may be specified.
    If no version is specified and multiple versions of the dataset exist,
    the earliest version of the dataset that is still active will be returned.
    If no version is specified, multiple versions of the dataset exist and
    ``exception_if_multiple`` is set to ``True``, this function will raise an exception.
```

On another note, reading the documentation, I think we can put the dataset version in the function call, that should also work.  We can even give the `error_if_multiple` its default value for users to understand that the function has quite a bit to offer.

mfeurer:
> On another note, reading the documentation, I think we can put the dataset version in the function call, that should also work.

Yes, that sounds like a good idea.

> We can even give the error_if_multiple its default value for users to understand that the function has quite a bit to offer.

You could add that to the datasets example.

PR #1070, opened 2021-05-03
Updating examples to use names instead of integer ids whenever possible

ArlindKadra:
#### Reference Issue

#1066, #1063

- [x] Uses names instead of integer ids in the examples whenever possible. 
- [x] Update documentation for `get_dataset`
- [ ] Check implementation of examples for datasets and splits, update them if necessary in case they do not show that they can be accessed with names too.

Existing unit tests.

### DIFF
--- a/examples/30_extended/datasets_tutorial.py
+++ b/examples/30_extended/datasets_tutorial.py
@@ -52,7 +52,7 @@ datalist.query("NumberOfClasses > 50")
 # =================
 
 # This is done based on the dataset ID.
-dataset = openml.datasets.get_dataset(1471)
+dataset = openml.datasets.get_dataset(dataset_id="eeg-eye-state", version=1)
 
 # Print a summary
 print(
@@ -92,7 +92,7 @@ print(X.info())
 # data file. The dataset object can be used as normal.
 # Whenever you use any functionality that requires the data,
 # such as `get_data`, the data will be downloaded.
-dataset = openml.datasets.get_dataset(1471, download_data=False)
+dataset = openml.datasets.get_dataset(dataset_id="eeg-eye-state", version=1, download_data=False)
 
 ############################################################################
 # Exercise 2
--- a/openml/datasets/functions.py
+++ b/openml/datasets/functions.py
@@ -368,7 +368,8 @@ def get_dataset(
     Parameters
     ----------
     dataset_id : int or str
-        Dataset ID of the dataset to download
+        Dataset ID of the dataset to download. It can be an integer or it can be a string
+        of the dataset name.
     download_data : bool, optional (default=True)
         If True, also download the data file. Beware that some datasets are large and it might
         make the operation noticeably slower. Metadata is also still retrieved.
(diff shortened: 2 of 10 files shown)


examples/30_extended/datasets_tutorial.py
```diff
@@ -52,7 +52,7 @@
 # =================
 
 # This is done based on the dataset ID.
```
PGijsbers, line 54:
```suggestion
```

openml/datasets/functions.py
```diff
@@ -368,7 +368,8 @@ def get_dataset(
     Parameters
     ----------
     dataset_id : int or str
-        Dataset ID of the dataset to download
+        Dataset ID of the dataset to download. It can be an integer or it can be a string
+        of the dataset name.
```
PGijsbers, line 372:
```suggestion
        The ID or name of the dataset to download.
```
Since ID's are parsed to integer anyway if they're given as strings.

ArlindKadra: @PGijsbers I can also do that, however, I tried to keep it consistent with https://github.com/openml/openml-python/blob/566943174bb7952721921c8d852112d127ae6901/openml/datasets/functions.py#L323-L325 at `get_datasets`

examples/30_extended/datasets_tutorial.py
ArlindKadra: For this particular example, which actually is a tutorial for datasets, it probably makes sense to comment more on the fact that they can be downloaded with an id which can be an integer or a name and then explaining a bit more that a name would be good to be followed up with a version since the datasets can change between versions.

ArlindKadra:
@PGijsbers  @mfeurer  While making the modifications I noticed a study that actually had no alias, can that be a case?
And also can we add an alias to it? It was at the `30_extended/study_tutorial`, study `https://www.openml.org/s/123`

examples/30_extended/datasets_tutorial.py
PGijsbers: That's fair but the comment did not reflect that. So in that case I think it should be modified to something like `This can also be done based on the dataset ID.` probably accompanied by a commented out example `# dataset = openml.datasets.get_dataset(1471)`

ArlindKadra: Yes, you are right on that. This is also what I was thinking about.

openml/datasets/functions.py
PGijsbers: I think the difference is fine, or should favor the simpler wording.

PGijsbers:
An alias is not required, but I'll ask people with access if they're prepared to add one (I don't believe there's an API call for it).

ArlindKadra:
> An alias is not required, but I'll ask people with access if they're prepared to add one (I don't believe there's an API call for it).

Ah thanks. Yeah, then maybe it is not worth the effort. I can just keep it like that and write a comment at the tutorial like we discussed for the datasets.

PR #1071, opened 2021-05-04
Remove E500 from list of exception to raise

PGijsbers:
OpenML code 500 indicates no results for a flow query, and was likely
confused with the HTTP code 500 for internal server error.

Note that when parsing the error, this is actually annotated correctly [here](https://github.com/openml/openml-python/blob/6e8a9db03fd1af9d3eb0623970101413b531cc69/openml/_api_calls.py#L302-L310).


mfeurer (approved):

mfeurer:
As an additional piece of information Pieter sent me this one offline which explains all the OpenML error codes: https://github.com/openml/OpenML/blob/develop/openml_OS/views/pages/api_new/v1/xml/pre.php

PR #1080, opened 2021-05-12
Extend extensions page

janvanrijn:
adds more documentation how to add extensions to OpenML Python


doc/extensions.rst
```diff
@@ -27,9 +27,14 @@ Connecting new machine learning libraries
...
 ~~~~~~~~~~~~~~~~~~~~~~
 
-To leverage support from the community and to tap in the potential of OpenML, interfacing
-with popular machine learning libraries is essential. However, the OpenML-Python team does
-not have the capacity to develop and maintain such interfaces on its own. For this, we
+To leverage support from the community and to tap in the potential of OpenML,
+interfacing with popular machine learning libraries is essential.
+The OpenML-Python package is capable of downloading meta-data and results (data,
+flows, runs), regardless of the library that was used to upload it.
+However, uploading flows and runs from a specific library, requires an
```
mfeurer, line 34: Hm, I would say that the extensions simplify this process as one can go without them (https://openml.github.io/openml-python/main/examples/30_extended/custom_flow_.html#sphx-glr-examples-30-extended-custom-flow-py)

doc/extensions.rst
```diff
@@ -57,6 +62,71 @@ Once the new extension class has been defined, the openml-python module to
...
+    predictions are exactly the same.)
+    The example in the scikit-learn interface might seem daunting, but note that
+    here some complicated design choices were made, that allow for all sorts of
+    interesting research questions. It is probably good practice to start easy. 
+  * :meth:`model_to_flow`: The inverse of :meth:`flow_to_model`. Serializes a 
+    model into an OpenML Flow. The flow should preserve the class, the library
+    version, and the tunable hyperparameters. 
+  * :meth:`get_version_information`: Return a tuple with the version information
+    of the important libraries. 
+  * :meth:`create_setup_string`: No longer uses, and will be deprecated soon.
```
mfeurer, line 99:
```suggestion
  * :meth:`create_setup_string`: No longer used, and will be deprecated soon. 
```

doc/extensions.rst
```diff
@@ -57,6 +62,71 @@ Once the new extension class has been defined, the openml-python module to
...
+  * :meth:`is_estimator`: Gets as input a class, and checks whether it has the 
+    status of estimator in the library (typically, whether it has a train method
+    and a predict method).
+  * :meth:`seed_model`: Sets a random seed to the model. 
+  * :meth:`_run_model_on_fold`: One of the main requirements for a library to
+    generate run objects for the OpenML server. Obtains a train split (with
+    labels) and a test split (without labels) and the goal is to train a model
+    on the train split and return the predictions on the test split. On top of
+    the actual predictions, also the class probabilities should be determined. 
+    For hard-classifiers, this can just be the hot-encoded predicted label. The
```
mfeurer, line 110: I would say "classifiers that do not return class probabilities"

doc/extensions.rst
```diff
@@ -57,6 +62,71 @@ Once the new extension class has been defined, the openml-python module to
...
+    information about a hyperparameter optimization trace can be provided. 
+  * :meth:`obtain_parameter_values`: Obtains the hyperparameters of a given
+    model and the current values. Please note that in the case of a hyperparameter
+    optimization procedure (e.g., random search), you only should return the
+    hyperparameters of this procedure (e.g., the hyperparameter grid, budget,
+    etc) and that the chosen model will be inferred from the optimization trace. 
+  * :meth:`check_if_model_fitted`: Check whether the train method of the model
+    has been called (and as such, whether the predict method can be used).
+* Hyperparameter optimization (optional)
+  * :meth:`instantiate_model_from_hpo_class`: In a given run has recorded the
```
mfeurer, line 123:
```suggestion
  * :meth:`instantiate_model_from_hpo_class`: If a given run has recorded the
```

mfeurer (commented):
Looks good to me. I have a few minor suggestions on how to improve it.

doc/extensions.rst
```diff
@@ -57,6 +62,75 @@ Once the new extension class has been defined, the openml-python module to
...
+    Additionally, information about a hyperparameter optimization trace can be
+    provided. 
+  * :meth:`obtain_parameter_values`: Obtains the hyperparameters of a given
+    model and the current values. Please note that in the case of a hyperparameter
+    optimization procedure (e.g., random search), you only should return the
+    hyperparameters of this procedure (e.g., the hyperparameter grid, budget,
+    etc) and that the chosen model will be inferred from the optimization trace. 
+  * :meth:`check_if_model_fitted`: Check whether the train method of the model
+    has been called (and as such, whether the predict method can be used).
+* Hyperparameter optimization (optional)
```
prabhant, line 126: A small addition that can be added here is
The optional methods have to be initialized as an empty function example:
def instantiate_model_from_hpo_class():
    "This is just an empty function to avoid errors"

It can be confusing for some users to get errors that this optional method is not implemented. So here we can add that though these methods are optional they should still be present in the extension as an empty function.

prabhant (commented):
LGTM, A small change can be added too :)

doc/extensions.rst
janvanrijn: What do you mean exactly? I would just suggest users to raise a `NotImplementedError` (since that method will only be invoked externally)

prabhant: In the end, you still need these optional methods to exist in the extension, While using the extension openml-python does require the optional methods to exist too(at least that was the issue I faced while developing openml-tensorflow).  So I just defined the functions without definitions like here: https://github.com/openml/openml-tensorflow/blob/master/openml_tensorflow/extension.py

But it was confusing for me at first when I was getting those errors, so I think it will be helpful to add a message regarding that here(that we still need to define these methods as empty methods if we don't want to implement them).

doc/extensions.rst
```diff
@@ -57,6 +62,75 @@ Once the new extension class has been defined, the openml-python module to
...
+The following methods should get implemented. Although the documentation in
+the `Extension` interface should always be leading, here we list some additional
+information and best practices. 
+The `Sklearn Extension API :class:`openml.extensions.sklearn.SklearnExtension.html`
+is a good example to follow. Note that most of the methods to be implemented 
+are relatively simple, and can be done in several lines of code. 
+
+* General setup (required)
+  * :meth:`can_handle_flow`: Takes as argument an OpenML flow, and checks
+    whether this can be handled by the current extension. The OpenML database
```
PGijsbers, line 74: The docbuild fails because this indentation will lead to a warning (which we don't allow). In case you have issues building docs locally there's also a [docker image](https://github.com/openml/openml-python/tree/develop/docker) available.

doc/extensions.rst
```diff
@@ -27,9 +27,14 @@ Connecting new machine learning libraries
...
 
-To leverage support from the community and to tap in the potential of OpenML, interfacing
-with popular machine learning libraries is essential. However, the OpenML-Python team does
-not have the capacity to develop and maintain such interfaces on its own. For this, we
+To leverage support from the community and to tap in the potential of OpenML,
+interfacing with popular machine learning libraries is essential.
+The OpenML-Python package is capable of downloading meta-data and results (data,
+flows, runs), regardless of the library that was used to upload it.
+However, in order to simplify the process of uploading flows and runs from a
+specific library, an additional interface can be build.
```
PGijsbers, line 35:
```suggestion
specific library, an additional interface can be built.
```

doc/extensions.rst
```diff
@@ -27,9 +27,14 @@ Connecting new machine learning libraries
...
+The OpenML-Python team does not have the capacity to develop and maintain such
+interfaces on its own. For this, we
```
PGijsbers, line 37:
```suggestion
interfaces on its own. For this reason, we
```

PGijsbers: Just a preference.

doc/extensions.rst
```diff
@@ -57,6 +62,75 @@ Once the new extension class has been defined, the openml-python module to
 :meth:`openml.extensions.register_extension` must be called to allow OpenML-Python to
 interface the new extension.
 
+The following methods should get implemented. Although the documentation in
+the `Extension` interface should always be leading, here we list some additional
+information and best practices. 
+The `Sklearn Extension API :class:`openml.extensions.sklearn.SklearnExtension.html`
+is a good example to follow. Note that most of the methods to be implemented 
+are relatively simple, and can be done in several lines of code.
```
PGijsbers, line 70:
```suggestion
is a good example to follow. Note that most methods are relatively simple and can be implemented in several lines of code. 
```

PGijsbers: Just a preference.

PGijsbers (changes requested):
Looks good! There is one grammatical error (and two suggestions). But the docs currently don't build without warning because of the indentation.

doc/extensions.rst
janvanrijn: should I remove all line breaks?

doc/extensions.rst
janvanrijn: > In the end, you still need these optional methods to exist in the extension

Are you sure? Since the extensin inherits from the `Extension` interface, we technically do not need to implement it in the specific right? 

Just to be sure, I added: 
> Note that although this method is optional, for the extension to work it
> needs to exist. If this function is not required, we recommend to have an
> empty function that raises a `NotImplementedError` with a useful error
> message.

doc/extensions.rst
PGijsbers: Line breaks are OK, but it's indented too far for this parser. Content for the lists should only be indented as far as the asterisks (not sure why the parser wants this, to be honest).

mfeurer: I think this is a misleading error message. I added a line break and it now works. I created a commit which fixes doc building and will now create a commit fixing the merge conflicts arising from this.

doc/extensions.rst
```diff
@@ -64,73 +64,78 @@ interface the new extension.
...
+
   * :meth:`instantiate_model_from_hpo_class`: If a given run has recorded the
     hyperparameter optimization trace, then this method can be used to
     reinstantiate the model with hyperparameters of a given hyperparameter
     optimization iteration. Has some similarities with :meth:`flow_to_model` (as
-    this method also sets the hyperparameters of a model). 
-    
-
+    this method also sets the hyperparameters of a model).
+    Note that although this method is optional, for the extension to work it
```
mfeurer, line 135: Shouldn't it rather be that the implementing the logic is optional but that the function needs to exists? So

```
Note that although this method is required, it is not necessary to implement any logic if hyperparameter optimization is not implemented. Simply raise a `NotImplementedError` then.
```

PGijsbers (approved):

PR #1085, opened 2021-05-19
Don't fail when Parquet server can't be reached

### LINKED ISSUE #1084: Cached datasets unusable without internet access due to missing Parquet files
#### Description
When trying to use a pre-built cache in a remote computing environment that did not have access to the internet, OpenML fails due to an inability to access the internet when checking for (potentially) non-existent Parquet files. This error cascades through `openml.datasets.get_dataset()` and causes the entire cache to be marked as corrupt and deletes it. As a result, the existing cache is lost, so later code that would otherwise have worked perfectly fine by relying on the ARFF cache file fails as well.

#### Steps/Code to Reproduce
To avoid confusion, I will refer to "local" as a node with internet access and "remote" as the node without internet access. If you observe the code of `get_dataset()` [here](https://github.com/openml/openml-python/blob/bb17e72d1866e1d23dcf2eace2ca4bdd73af9d39/openml/datasets/functions.py#L410-L445), it tries to download the ARFF file, followed by the Parquet file and then sets a flag to indicate that the cache should be retained. The workflow on local is, therefore, this: 

1. The ARFF file is downloaded, 
2. The Parquet file is NOT downloaded as there is either a silent failure or no Parquet file exists, 
3. The cache deletion flag is flipped and the cache is retained. 

However, when I copy the cache over to remote and try accessing the dataset again, the workflow changes since remote does not have internet access. In this case, the following workflow occurs: 

1. The ARFF file cache is successfully read, 
2. The Parquet file cache is checked and not found, 
3. OpenML tries to access the internet to download the Parquet file and fails, 
4. This particular error is not handled appropriately and therefore cascades down to the `finally` block without resetting the cache clear flag, and the entire cache gets deleted, including the perfectly valid ARFF file. 

Thus, the later sections of the program that would otherwise have run using the cached ARFF file also fail. For a minimum working example, the following code was run on both local and remote:

```python
import openml
d = openml.datasets.get_dataset(4135)
```

The cache from local was copied to remote before executing the above code on remote and verified to be visible to `openml.config`.

#### Expected Results
The dataset should be successfully read on remote from the ARFF file even if the Parquet file is missing.

#### Actual Results
[This](https://gist.github.com/NeoChaos12/5eb32ee0ba2378b6ad73731ac24ec95c) error on remote (which was expected) is not handled properly and causes the cache itself to be deleted.

#### Versions
Linux-4.4.180-94.100-default-x86_64-with-SuSE-12-x86_64
Python 3.7.10 (default, Feb 26 2021, 18:47:35) 
[GCC 7.3.0]
NumPy 1.19.4
SciPy 1.5.3
Scikit-Learn 0.23.2
OpenML 0.12.1

P.S.: The OpenML [docs](https://openml.github.io/openml-python/develop/usage.html) have a broken link for how to set up the `apikey` parameter for the config file. This threw me off for a spin since I initially thought that the source of the error has to do with an authentication failure while downloading a Parquet file, but I'm not sure if it warrants a whole issue to itself.

PGijsbers:
Fixes #1084.
The Parquet file is optional, and failing to reach it (and download it)
should not prevent the usage of the other cached/downloaded files.

To test run:
```python
import openml

d = openml.datasets.get_dataset(4135)
print(d)
```
once with an internet connection, and once without. Both should work.

### DIFF
--- a/openml/datasets/functions.py
+++ b/openml/datasets/functions.py
@@ -8,6 +8,7 @@ from typing import List, Dict, Union, Optional, cast
 import numpy as np
 import arff
 import pandas as pd
+import urllib3
 
 import xmltodict
 from scipy.sparse import coo_matrix
@@ -425,7 +426,10 @@ def get_dataset(
 
         arff_file = _get_dataset_arff(description) if download_data else None
         if "oml:minio_url" in description and download_data:
-            parquet_file = _get_dataset_parquet(description)
+            try:
+                parquet_file = _get_dataset_parquet(description)
+            except urllib3.exceptions.MaxRetryError:
+                parquet_file = None
         else:
             parquet_file = None
         remove_dataset_cache = False


openml/datasets/functions.py
```diff
@@ -425,7 +426,10 @@ def get_dataset(
 
         arff_file = _get_dataset_arff(description) if download_data else None
         if "oml:minio_url" in description and download_data:
-            parquet_file = _get_dataset_parquet(description)
+            try:
+                parquet_file = _get_dataset_parquet(description)
+            except urllib3.exceptions.MaxRetryError:
```
PGijsbers, line 431: I wasn't sure if I should be conservative here, or just catch all errors. Considering we haven't had any issues otherwise I went with the general best practice of only minimally catching exceptions.

mfeurer (approved):

openml/datasets/functions.py
mfeurer: Looks reasonable to me.

PR #1086, opened 2021-05-19
Allow tasks to be downloaded without dataqualities

PGijsbers:
Previously ``download_qualities`` would be left at the default of `True` with no way to overwrite it, this PR adds the parameter to `get_task` which would allow you to download a task with its dataset without ever downloading qualities.

### DIFF
--- a/openml/tasks/functions.py
+++ b/openml/tasks/functions.py
@@ -4,7 +4,7 @@ from collections import OrderedDict
 import io
 import re
 import os
-from typing import Union, Dict, Optional
+from typing import Union, Dict, Optional, List
 
 import pandas as pd
 import xmltodict
@@ -297,7 +297,9 @@ def __list_tasks(api_call, output_format="dict"):
     return tasks
 
 
-def get_tasks(task_ids, download_data=True):
+def get_tasks(
+    task_ids: List[Union[str, int]], download_data: bool = True, download_qualities: bool = True
+) -> List[OpenMLTask]:
     """Download tasks.
 
     This function iterates :meth:`openml.tasks.get_task`.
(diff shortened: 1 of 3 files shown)


openml/tasks/functions.py
```diff
@@ -306,21 +308,25 @@ def get_tasks(task_ids, download_data=True):
...
     for task_id in task_ids:
-        tasks.append(get_task(task_id, download_data))
+        tasks.append(get_task(task_id, download_data, download_qualities))
     return tasks
 
 
 @openml.utils.thread_safe_if_oslo_installed
-def get_task(task_id: int, download_data: bool = True) -> OpenMLTask:
+def get_task(
+    task_id: int, download_data: bool = True, download_qualities: bool = True
```
mfeurer, line 328: Above we allow `task_id` to be of type `str`, but here it's only int. How about we deprecate the use of anything besides `str` as I really don't see a reason to support it?

openml/tasks/functions.py
```diff
@@ -329,25 +335,30 @@ def get_task(task_id: int, download_data: bool = True) -> OpenMLTask:
...
         Option to trigger download of data along with the meta data.
+    download_qualities : bool (default=True)
+        Option to download 'qualities' meta-data in addition to the minimal dataset description.
 
     Returns
     -------
     task
     """
+    if not isinstance(task_id, int):
+        warnings.warn("Task id must be specified as `int` from 0.14.0 onwards.", DeprecationWarning)
```
PGijsbers, line 350: Went with (what I remember to be) scikit-learn's default behavior of having at least one full minor release with the deprecation warning.

mfeurer (approved):

PR #1108, opened 2021-08-20
Fix #1106

### LINKED ISSUE #1106: Categorical features encoded as booleans
#### Description
I realized that for datasets that have two classes `"True"` and `"False"` according to the arff file the respective attribute is converted to a boolean `True` and `False` and not a categorical with the values `"True"` and `"False"`.

This is not consistent with scikit-learn, which returns a category with two string entries.

The issue is now that the return value of OpenML cannot be parsed by scikit-learn's `sklearn.utils.multiclass.type_of_target`, which prohibits us to run scikit-learn models on such datasets.

Proposed solution: we return the string values and do not cast to bool as the arff specification does not mention booleans.

CC @glemaitre as author of the conversion @jnothman is this intended behavior in scikit-learn? A target with data type categorical should always be parsable by `type_of_target` I guess?

#### Steps/Code to Reproduce

```python
import openml
import sklearn.utils.multiclass

task = openml.tasks.get_task(3902)  # download the OpenML task
X, y = task.get_X_and_y(dataset_format='pandas')

print(y)
print(sklearn.utils.multiclass.type_of_target(y))

import sklearn.datasets

X, y = sklearn.datasets.fetch_openml(data_id=1049, return_X_y=True)

print(y)
print(sklearn.utils.multiclass.type_of_target(y))
```

#### Expected Results

Note that `type_of_target` returns binary both times.

```
0       FALSE
1       FALSE
2       FALSE
3       FALSE
4       FALSE
        ...  
1453    FALSE
1454    FALSE
1455    FALSE
1456    FALSE
1457    FALSE
Name: c, Length: 1458, dtype: category
Categories (2, object): ['FALSE', 'TRUE']
binary
0       FALSE
1       FALSE
2       FALSE
3       FALSE
4       FALSE
        ...  
1453    FALSE
1454    FALSE
1455    FALSE
1456    FALSE
1457    FALSE
Name: c, Length: 1458, dtype: category
Categories (2, object): ['FALSE', 'TRUE']
binary
```

#### Actual Results

Note that `type_of_target` returns unknown for the first call.
```
0       False
1       False
2       False
3       False
4       False
        ...  
1453    False
1454    False
1455    False
1456    False
1457    False
Name: c, Length: 1458, dtype: category
Categories (2, object): [False < True]
unknown
0       FALSE
1       FALSE
2       FALSE
3       FALSE
4       FALSE
        ...  
1453    FALSE
1454    FALSE
1455    FALSE
1456    FALSE
1457    FALSE
Name: c, Length: 1458, dtype: category
Categories (2, object): ['FALSE', 'TRUE']
binary
```

mfeurer:
Fixes #1106

For pandas Dataframe do not encode categories with values "True" and "False" as type boolean but keep them as type category.

### DIFF
--- a/openml/datasets/dataset.py
+++ b/openml/datasets/dataset.py
@@ -414,17 +414,7 @@ class OpenMLDataset(OpenMLBase):
             if isinstance(type_, list):
                 categorical.append(True)
                 categories_names[name] = type_
-                if len(type_) == 2:
-                    type_norm = [cat.lower().capitalize() for cat in type_]
-                    if set(["True", "False"]) == set(type_norm):
-                        categories_names[name] = [
-                            True if cat == "True" else False for cat in type_norm
-                        ]
-                        attribute_dtype[name] = "boolean"
-                    else:
-                        attribute_dtype[name] = "categorical"
-                else:
-                    attribute_dtype[name] = "categorical"
+                attribute_dtype[name] = "categorical"
             else:
                 categorical.append(False)
                 attribute_dtype[name] = ARFF_DTYPES_TO_PD_DTYPE[type_]
--- a/tests/test_datasets/test_dataset.py
+++ b/tests/test_datasets/test_dataset.py
@@ -147,12 +147,12 @@ class OpenMLDatasetTest(TestBase):
         # test to check that we are converting properly True and False even
         # with some inconsistency when dumping the data on openml
         data, _, _, _ = self.jm1.get_data()
-        self.assertTrue(data["defects"].dtype.name == "category")
-        self.assertTrue(set(data["defects"].cat.categories) == {True, False})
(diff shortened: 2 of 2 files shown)


PGijsbers (commented):
With our current ARFF support this makes sense, since ARFF does not officially support BOOLEAN type. However I don't think we should keep the behavior consistent when loading from parquet (i.e. converting to string if it was stored boolean)? I guess that's not for this PR either way.

Above concern aside, I would be happy to have this (and simplify [this](https://github.com/openml/automlbenchmark/pull/324/files) again), but want to wait for a comment from the scikit-learn contributors on #1106.

PGijsbers:
We decided that having boolean values for categoricals of two types (being exactly true/false) is better after all. Consistency with scikit-learn will hopefully come when we fully migrated to parquet.

PR #1117, opened 2021-10-13
Fix/1110

### LINKED ISSUE #1110: Make optional arguments to `create_study` actually optional
From the docs [`create_study`](https://openml.github.io/openml-python/main/generated/openml.study.create_study.html#openml.study.create_study) and `create_benchmark_suite` contain some optional parameters, e.g. `alias`, which don't have defaults which actually makes them required.

1. For the `alias` parameter a `None` default seems sensible.
2. For a `study` the `suite`'s id is also optional, is it by design that we want studies without suites?
3. I would argue that for a `study`, the `run_ids` should also be optional as they could be created later (e.g. in a setup where you define a study, and upload results to it as they come in).

PGijsbers:
Updates `create_study` (fixes #1110):
 - `alias`  parameter is now optional
 - `run_ids` is allowed to be an empty list or `None` (it is possible to create an empty study and upload runs to attach to it later). Internally `[]` will be converted to `None`, otherwise the server gets upset (see #1110 for more info).
 
Note that previously empty studies were not allowed in `openml-python`, but they now are. Previously empty studies were only allowed if they were created before `2019-01-01`, and they were expected to have their `runs`  set to `None`  so I reintroduced this convention for new studies without runs (though I think perhaps an empty list might be better).

### DIFF
--- a/tests/test_study/test_study_functions.py
+++ b/tests/test_study/test_study_functions.py
@@ -133,7 +133,6 @@ class TestStudyFunctions(TestBase):
         study_downloaded = openml.study.get_study(study.id)
         self.assertEqual(study_downloaded.main_entity_type, "run")
         self.assertIsNone(study_downloaded.runs)
-        openml.study.delete_study(study.id)
 
     def test_publish_empty_study_explicit(self):
         self._test_publish_empty_study_is_allowed(explicit=True)


PGijsbers:
I ran unit tests locally, and the `study` ones worked fine (except `test_study_list` which explicitly says failure can be caused by server state). The general suite seemed to have issues which as far as I can tell are due to the test server state?

PR #1128, opened 2022-03-11
Predictions

### LINKED ISSUE #1103: Make predictions from a run easily accessible
If I'm not overlooking anything, there is no convenient way to access predictions that are stored in a run. The `run` object only exposes functions which use the data internally (e.g. [`get_metric_fn`](https://github.com/openml/openml-python/blob/6717e66a1e967a131a6ca7feb96f6d166017bed7/openml/runs/run.py#L370)). Internally predictions are downloaded and processed (here).

I propose we add a `predictions` property which can serve the predictions in dataframe format through lazy loading. The remainder of the code should then also be refactored to use this property.

PGijsbers:
#### Reference Issue

Closes #1103 

Adds a `predictions` property to the OpenMLRun object which will download and format the predictions on-demand.
Only output for a pandas dataframe is supported, since we will slowly move to that standard anyway.
Local run results can also be formatted in this way.
Data is only accessed and transformed when the property is called and does not impact performance otherwise.

I added a line to test where we test other run consistency (download vs local run).

#### Any other comments?
Test server is currently not live so I couldn't test on the whole suite, but since it's an added feature it shouldn't interfere with old tests. I did a test on the live server which worked, but it may not be as comprehensive.

### DIFF
--- a/doc/progress.rst
+++ b/doc/progress.rst
@@ -10,7 +10,7 @@ Changelog
 ~~~~~~
  * FIX#1110: Make arguments to ``create_study`` and ``create_suite`` that are defined as optional by the OpenML XSD actually optional.
  * MAIN#1088: Do CI for Windows on Github Actions instead of Appveyor.
-
+ * ADD#1103: Add a ``predictions`` property to OpenMLRun for easy accessibility of prediction data.
 
 0.12.2
 ~~~~~~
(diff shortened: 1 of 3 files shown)


mfeurer (approved):
This looks good to me. Do you know whether the server is already back up?

PR #1134, opened 2022-04-15
Fixed issue #1125 by setting strip_whitespace=False in xmltodict.parse calls

amastruserio:
#### What does this PR implement/fix? Explain your changes.
Fixed issue #1125 that incorrectly formatted the whitespace on some feature categories. Added flag strip_whitespace=False on all xmltodict.parse() functions. The changes didn't raise any other visible issues in any of the unit tests, and appeared to work as expected.

PR #1136, opened 2022-04-18
Fix/strip whitespace

### LINKED ISSUE #1125: Whitespace stripped from XML files can lead to congruence
We use `xmltodict.parse` with the default `strip_whitespace=True` which can lead to a scenario where the features have categories that don't match the ARFF file categories (e.g., '50000.+' in features and ' 50000.+' in the data).
In principle it's an easy fix, but we should take care to test we don't break anything, and I'd propose to check if this "bug" can also lead to issues in reading other XML files.

For more info, see https://github.com/openml/automlbenchmark/issues/350#issuecomment-1004811045

amastruserio:
#### Reference Issue

Fixes #1125 

Update to pull request #1134. Deletion of test files from tasks folder reverted. Change appears to function as expected with proper formatting and passes unit tests.


PGijsbers:
Thanks for the contributions, and sorry for the delay in our response. This change is quite invasive (as mentioned in the issue) so I want to ensure we can run our unit test suite before merging. However, unit tests currently do not work due to some issues with the test server. We hope to fix this over the coming month and I will revisit this PR.

PGijsbers:
@eddiebergman @LennartPurucker I think we should look at this. It may be nicer to just monkey patch the parse function to a partial with the default set to `strip_whitespace=False`.

LennartPurucker:
Closing this as a result of recent developments: https://github.com/openml/openml-python/issues/1125

Issue #1139, opened 2022-06-01
get_metric_fn can't handle classification metrics using probabilities


pseudotensor:
import openml
import sklearn
import numpy as np
from sklearn.impute import SimpleImputer
from sklearn.tree import DecisionTreeClassifier

openml.config.apikey = "KEY"

# build a scikit-learn classifier
clf = sklearn.pipeline.make_pipeline(SimpleImputer(),
                                     DecisionTreeClassifier())

task_id = 189871

task = openml.tasks.get_task(task_id)  # download the OpenML task
run = openml.runs.run_model_on_task(clf, task)  # run the classifier on the task

# ok:
#score = run.get_metric_fn(sklearn.metrics.accuracy_score)  # print accuracy score
# fails, even if pass labels
score = run.get_metric_fn(sklearn.metrics.log_loss, kwargs=dict(labels=task.class_labels))

print('Data set: %s; Accuracy: %0.2f' % (task.get_dataset().name, float(np.mean(score))))
```

fails with:
```
ValueError: The number of classes in labels is different from that in y_pred. Classes found in labels: ['0' '1' '2' '3' '4']

```

The problem is is old (2017) when function first made: https://github.com/openml/openml-python/commit/1c285a803b58dca963e4c51930251ac334d94d19

This block:

https://github.com/openml/openml-python/blob/develop/openml/runs/run.py#L489-L492

gets an index of the prediction, but this is not the correct thing to pass to log_loss.

There are multiple problems:
1) A classification metric like log_loss should take in probabilities, never labels, but this is impossible to do
2) The original labels should be what is predicted and passed to any metric, as defined by the task class labels, not the index.


I think the only solution is to avoid the helpers in openml-python and directly manage the splits etc.

pseudotensor:
FYI: https://github.com/openml/openml-python/pull/1140  Just to show how it can be made to work.  Not necessarily elegent.

pseudotensor:
In the end I couldn't make this helper work properly.  The scores are just wrong, e.g. AUC of 1.  Had to use "plain" way of just asking task for splits and doing them myself, like many other papers have done.

E.g. even random forest was giving AUC=1 for KDDCup09_appetency with task_id 75105

mfeurer:
Does the code you provide above trigger this issue together with #1140?

pseudotensor:
> Does the code you provide above trigger this issue together with #1140?

The PR is an attempted work-around, but it's incomplete and doesn't really work in general.

[...]

PR #1140, opened 2022-06-01
Allow prob-based metrics to work

pseudotensor:

Handling probabilities with some metrics

code here: https://github.com/openml/openml-python/issues/1139

### DIFF
--- a/openml/runs/run.py
+++ b/openml/runs/run.py
@@ -385,7 +385,7 @@ class OpenMLRun(OpenMLBase):
 
         return arff_dict
 
-    def get_metric_fn(self, sklearn_fn, kwargs=None):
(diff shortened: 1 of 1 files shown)


PGijsbers:
Thanks for the contribution 🎉 I'll try to have a closer look next week.
In the meantime, please have a look at the [pull request checklist](https://github.com/openml/openml-python/blob/main/CONTRIBUTING.md#pull-request-checklist) and fill in the missing gaps, such as:
 - add the new parameter to the docstring
 - create unit tests, in particular also ones that would fail before this PR but succeed after
 - add a note to the progress log

PGijsbers:
Closing this because it is stale. In another rework we will probably base it on scikit-learn scorer objects instead of the metric function so that we have access to whether it needs proba directly instead of as an additional argument.

PR #1145, opened 2022-06-26
Set default value of avoid_duplicate_runs to false for run_model_on_task

### LINKED ISSUE #1143: run_model_on_task: make avoid_duplicate_runs=False the default
#### Description
run_model_on_task has an option to avoid running experiments that already exist on OpenML, called avoid_duplicate_runs. This, however, requires an API key. It is currently the default, meaning that people can't try out this function without setting their API key.
This creates an unnecessary obstacle, especially for beginners who don't know that the avoid_duplicate_runs option can be switched off. 

#### Steps/Code to Reproduce
```
from sklearn import ensemble
from openml import tasks, runs

clf = ensemble.RandomForestClassifier()
task = tasks.get_task(3954)
run = runs.run_model_on_task(clf, task)
```

#### Expected Results
The model should just run. The user may have no intention to upload the run to OpenML later.

#### Actual Results
An API key error is thrown

#### Versions
Linux-5.4.188+-x86_64-with-Ubuntu-18.04-bionic
Python 3.7.13 (default, Apr 24 2022, 01:04:09) 
[GCC 7.5.0]
NumPy 1.21.6
SciPy 1.4.1
Scikit-Learn 1.0.2
OpenML 0.12.2

chadmarchand:
#### Reference Issue

Closes #1143

Sets the default value of `avoid_duplicate_runs` in the `run_model_on_task` function to False. When true, this option avoids running an experiment that already exists on OpenML, but this requires an API key. This change means that an API key is not required unless explicitly setting `avoid_duplicate_runs` to true.

Tested that without an API key, the following code block does not return a 401 error any more after this change has been made.

```python
from sklearn import ensemble
from openml import tasks, runs

clf = ensemble.RandomForestClassifier()
task = tasks.get_task(3954)
run = runs.run_model_on_task(clf, task)
```

I don't believe that an automated test should be required as we are just changing a default value and not changing any implementation, but please let me know if otherwise and I will add a test.

### DIFF
--- a/doc/progress.rst
+++ b/doc/progress.rst
@@ -8,6 +8,7 @@ Changelog
 
 0.13.0
 ~~~~~~
+ * MAINT#1143: Change default argument for ``avoid_duplicate_runs`` of ``run_model_on_task`` to false to avoid requiring an API key by default.
  * FIX#1030: ``pre-commit`` hooks now no longer should issue a warning.
  * FIX#1110: Make arguments to ``create_study`` and ``create_suite`` that are defined as optional by the OpenML XSD actually optional.
  * MAIN#1088: Do CI for Windows on Github Actions instead of Appveyor.
--- a/openml/runs/functions.py
+++ b/openml/runs/functions.py
@@ -48,7 +48,7 @@ RUNS_CACHE_DIR_NAME = "runs"
 def run_model_on_task(
     model: Any,
     task: Union[int, str, OpenMLTask],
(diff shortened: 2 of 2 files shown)


PGijsbers:
Thanks for putting in the effort. After some more thought we have decided that instead we simply want to issue a warning if the user has `avoid_duplicate` set to true when no apikey is set. In the future we expect `flow/exists` to be callable without authentication and then we can remove the warning.

Issue #1058, opened 2021-04-15
__repr__ for OpenML objects are not protected against missing class attributed


Neeratyoy:
#### Description

The `__repr__` print for OpenML objects is called from [here](https://github.com/openml/openml-python/blob/develop/openml/base.py#L19). The `_get_repr_body_fields` function contains formatters for printing each object. However, if for certain reasons, some of the attributes are missing the print fails by throwing an error, while other functions of the API may work as-is. This can be a confusing behaviour for users. 

Having checks to catch and handle such errors in `_get_repr_body_fields` will be useful to have.

#### Steps/Code to Reproduce

Representative example:
```python
import openml
task = openml.tasks.get_task(146806)   
print(task)  # fails
```

#### Expected Results

```
OpenML Classification Task
==========================
Task Type Description: https://www.openml.org/tt/TaskType.SUPERVISED_CLASSIFICATION
Task ID..............: 146806
Task URL.............: https://www.openml.org/t/146806
Estimation Procedure.: holdout
Target Feature.......: label
# of Classes.........: 2
Cost Matrix..........: Available
```

#### Actual Results

`TypeError: object of type 'NoneType' has no len()`

#### Versions

OpenML `0.12.1`.

PGijsbers:
Fixes in #1148

PR #1148, opened 2022-06-27
fix nonetype error during print for tasks without class labels

### LINKED ISSUE #1100: `TypeError` raised by trying to get the `__repr__` of a locally created task
When creating a task locally the `__repr__` function raises a `TypeError` due to the `class_labels` attribute being `None`. This persists even after publishing it, which makes exposure of this bug more prevalent (because the `task.publish()` returns itself which invokes `__repr__` if ran in an interactive shell without assigning the result to a variable):

```python
import openml
openml.config.start_configuration_for_example()
t = openml.tasks.create_task(openml.tasks.task.TaskType.SUPERVISED_CLASSIFICATION, 128, 2, "class", "weighted_recall")
t.publish()
>>> Traceback (most recent call last):
  File "D:\repositories\openml-python\openml\base.py", line 19, in __repr__
    body_fields = self._get_repr_body_fields()
  File "D:\repositories\openml-python\openml\tasks\task.py", line 101, in _get_repr_body_fields
    fields["# of Classes"] = len(getattr(self, "class_labels"))
TypeError: object of type 'NoneType' has no len()
```

Suggested expected behavior: leave it out of the `__repr__` if `class_labels` is `None`, i.e. extend the check of [task#100](https://github.com/openml/openml-python/blob/6717e66a1e967a131a6ca7feb96f6d166017bed7/openml/tasks/task.py#L100) to skip assignment if `class_labels is None`. I don't think it's worth the added complexity to support an accurate `class_labels` in this niche scenario.

willcmartin:
#### Reference Issue
Fixes #1100 and #1058

Previously, if `class_labels` attribute of task is `None` when `__repr__` was called, a `NoneType` error was thrown. Now "# of Classes" is not printed if there are no `class_labels`.

#### Any other comments?
I followed the implementation suggestion from #1100 where "# of Classes" is ignored. Let me know if this is no longer the best solution to this issue.

### DIFF
--- a/doc/progress.rst
+++ b/doc/progress.rst
@@ -8,6 +8,7 @@ Changelog
 
 0.13.0
 ~~~~~~
+ * FIX#1100/#1058: Avoid ``NoneType`` error when printing task without ``class_labels`` attribute.
  * FIX#1030: ``pre-commit`` hooks now no longer should issue a warning.
  * FIX#1110: Make arguments to ``create_study`` and ``create_suite`` that are defined as optional by the OpenML XSD actually optional.
  * MAIN#1088: Do CI for Windows on Github Actions instead of Appveyor.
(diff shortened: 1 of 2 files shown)


PGijsbers:
Yep! Sorry for the late response @willcmartin, but the contribution is much appreciated 🎉

PR #1150, opened 2022-07-01
pre-commit update

### LINKED ISSUE #1146: Pre-commit hook fails due to import error
#### Description

When following the [contribution guidelines](https://github.com/openml/openml-python/blob/develop/CONTRIBUTING.md), I installed all dependencies and ran the pre-commit hook  when working on my PR (#1145). The pre-commit hook failed when committing due to an import error. I manually moved an older version of the `click` package in order to get the hook to work.

#### Steps/Code to Reproduce

Follow the [contribution guidelines](https://github.com/openml/openml-python/blob/develop/CONTRIBUTING.md) (`pip install -e .[test]`,  `pre-commit install`). Then commit a change.

#### Expected Results

 No import error is thrown when running the pre-commit hook.

#### Actual Results

An import error is thrown when running the pre-commit hook.

```
[WARNING] Unstaged files detected.
[INFO] Stashing unstaged files to /home/chad/.cache/pre-commit/patch1656209125-53395.
black....................................................................Failed
- hook id: black
- exit code: 1

  File "/home/chad/.cache/pre-commit/repo9hzu_n53/py_env-python3/bin/black", line 8, in <module>
    sys.exit(patched_main())
  File "/home/chad/.cache/pre-commit/repo9hzu_n53/py_env-python3/lib/python3.8/site-packages/black.py", line 4134, in patched_main
    patch_click()
  File "/home/chad/.cache/pre-commit/repo9hzu_n53/py_env-python3/lib/python3.8/site-packages/black.py", line 4123, in patch_click
    from click import _unicodefun  # type: ignore
ImportError: cannot import name '_unicodefun' from 'click' (/home/chad/.cache/pre-commit/repo9hzu_n53/py_env-python3/lib/python3.8/site-packages/click/__init__.py)

mypy openml..............................................................Passed
mypy tests...........................................(no files to check)Skipped
flake8 openml............................................................Passed
flake8 tests.........................................(no files to check)Skipped
[INFO] Restored changes from /home/chad/.cache/pre-commit/patch1656209125-53395.
```

#### Versions

```
Python 3.8.10 (default, Mar 15 2022, 12:22:08)
NumPy 1.20.1
SciPy 1.6.1
Scikit-Learn 1.1.1
OpenML 0.12.2
```

### Comments

This appears to be the issue: https://stackoverflow.com/questions/71673404/importerror-cannot-import-name-unicodefun-from-click So it seems updating `black` in the pre-commit hook would fix the issue.

PGijsbers:
#### Reference Issue

closes #1146 
The issue identified problems with the outdated fixed versions of the pre-commit file. For me personally, it lead to different issues, so I decided it's probably best to bump all dependencies to their newest versions.

Update all dependencies used in pre-commit hooks to their most recent version.
Other files have been updated accordingly to pass the new pre-commit tests, this mainly involves:
 - black formatting of docstrings, which now no longer have a leading and trailing space.
 - black formatting of many comma-separates lists (e.g., when calling a function), which is now by default reformatted to one-item-per-line, as opposed to only using this if it would otherwise exceed line length. (Personally I am a fan)
 - Resolving a few new mypy complaints.

Not really sure, installing the latest pre-commit hooks and verifying that it works as expected?

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -5,7 +5,7 @@ select = C,E,F,W,B,T
 ignore = E203, E402, W503
 per-file-ignores =
     *__init__.py:F401
-    *cli.py:T001
+    *cli.py:T201
 exclude =
     venv
     examples
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,11 +1,11 @@
 repos:
   - repo: https://github.com/psf/black
-    rev: 19.10b0
+    rev: 22.6.0
     hooks:
       - id: black
         args: [--line-length=100]
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v0.761
+    rev: v0.961
     hooks:
       - id: mypy
         name: mypy openml
@@ -14,15 +14,15 @@ repos:
         name: mypy tests
         files: tests/.*
   - repo: https://gitlab.com/pycqa/flake8
-    rev: 3.8.3
+    rev: 4.0.1
     hooks:
       - id: flake8
         name: flake8 openml
         files: openml/.*
         additional_dependencies:
-          - flake8-print==3.1.4
+          - flake8-print==5.0.0
       - id: flake8
         name: flake8 tests
         files: tests/.*
         additional_dependencies:
-          - flake8-print==3.1.4
+          - flake8-print==5.0.0
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,11 +12,13 @@ repos:
         files: openml/.*
         additional_dependencies:
           - types-requests
+          - types-python-dateutil
       - id: mypy
         name: mypy tests
         files: tests/.*
         additional_dependencies:
           - types-requests
+          - types-python-dateutil
   - repo: https://gitlab.com/pycqa/flake8
     rev: 4.0.1
     hooks:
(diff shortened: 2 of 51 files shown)


mfeurer (commented):
Great PR. Do you know if there's a way we could automate these updates of black etc.?

mfeurer (approved):
Sorry, this was supposed to be an approve.

PGijsbers:
>  Do you know if there's a way we could automate these updates of black etc.?

The reason we use fixed versions is to make sure that everyone has the same environment because, as you see from this PR, rules can change between different versions. Second, since rules can change between versions, as I see it automated version bumps would mean either:
 - updating only the pre-commit file, leading to a state where the repository might not pass its pre-commit checks
 - updating the pre-commit file and auto-formatting (works for black, but mypy issues can't be automatically resolved). Force merging these changes means automatically accepting whichever changes black makes (not a fan, should probably be evaluated by us before accepting the new style guide). Doing it through an automated PR means that effectively very little work is actually automated (a version bump and formatting for review by us), doing this for every release of black/mypy might actually lead to more work?
 
So I am not really sure of any *useful* way to automate this procedure.

mfeurer:
Good points. Maybe something like the dependabot which opens PRs to bump the versions, and we can then add to the PRs to fix these issues? But maybe it's not worth pursuing this?

PGijsbers:
(of course, feel free to start looking into this yourself now if you want to)

Issue #1153, opened 2022-07-11
Look into "depenabot" for pre-commit


PGijsbers:
See #1150, it might be useful to have some automation which can bump the versions for the pre-commit dependencies and automatically notify us of this. We should look at how easy this is to set up (`dependabot`might be helpful, since it serves a similar purpose), and if it's worth pursuing (caveats mentioned in the PR).

eddiebergman:
Hi @PGijsbers CC @mfeurer,

I looked into it and there's 2 solutions, one automatic and one manual:
1. Use a [github workflow](https://github.com/marketplace/actions/pre-commit-autoupdate) that will run everyday and create a PR if it needs updating
2. Run [`pre-commit autoupdate`](https://pre-commit.com/#using-the-latest-version-for-a-repository) on your local code and it will check for updates and apply accordingly, at which point you can manually make a PR.

I [checked dependabot](https://github.com/dependabot/dependabot-core/issues/1524#issuecomment-750307910) and it doesn't seem supported or on their roadmap.

There's a PR for 1. at #1154 if you want, feel free to close if it's not :)

Best,
Eddie

mfeurer:
This is now enabled, see 1250 as an example.

PR #1154, opened 2022-07-11
Add: workflow to update precommit

eddiebergman:
Checks everyday to create PR that is to update pre-commit hooks 

See #1153

### DIFF
--- a/tests/test_runs/test_run_functions.py
+++ b/tests/test_runs/test_run_functions.py
@@ -183,7 +183,11 @@ class TestRun(TestBase):
         predictions_prime = run_prime._generate_arff_dict()
 
         self._compare_predictions(predictions, predictions_prime)
-        pd.testing.assert_frame_equal(run.predictions, run_prime.predictions)
+        pd.testing.assert_frame_equal(
+            run.predictions,
+            run_prime.predictions,
+            check_dtype=False,  # Loaded ARFF reads NUMERIC as float, even if integer.
+        )
 
     def _perform_run(
         self,


PGijsbers:
>  you may want to have the updates as part of the release

Which they will be... the next release. These dependencies are only for developers and govern the code formatting/guidelines and generally have no effect on the functionality of the tool.

> So this should only make changes to the pre-commit config file which shouldn't cause merge conflicts

Only updating the pre-commit file without updating the affected files may lead to a state where CI will do a pre-commit check to make sure everything is formatted/type-hinted correctly and fail. This can happen when e.g., black's formatting rules update and would reformat some files. That said, it shouldn't be hard to update the workflow to add that step to the PR.

eddiebergman:
Yup, your points makes sense, I'll change it later today.

Regarding updates causing format conflicts, most common precommit hooks that would format code are fairly stable, i.e. their updates are primarily new functionality, covering new features, optimizations and weird edge cases. I've had to update mypy once for it to recognize valid typing I wrote but that's the only time an updated pre-commit made a meaningful difference for me.

Best,
Eddie

PGijsbers:
Our last bump changed 50 files (#1150). Granted, we had not updated the dependencies since we added the pre-commit hooks, but that was "just" under 2 years ago. I do concede that the times the dependencies actually change formatting will not be common (most edits will be caused by one or two specific version bumps). Auto-updating the pre-commit file only without showing the affected files (if any) does not seem all that useful to me since I wouldn't do a blind merge for the above reasons and thus still need to figure out which (if any) files are affected manually. But adding a step in the job with `pre-commit run --all-files` before pushing the changes for a PR would likely solve that issue (provided CI reports the `mypy` problems).

eddiebergman:
That's also a great point, I'll add those as extra steps :)

PR #1160, opened 2022-09-10
check number of datasets for run (#1102)

michaelbaluja:
#### Reference Issue

Fixes openml/OpenML#1182 

Fixes issue with `get_run` and `get_runs` where run 1 would throw an error due to multiple datasets being present for the run. The issue is fixed by checking the number of datasets before attempting to access the supposed dictionary (in this case actually a list of dictionaries) containing the run. In the case of multiple datasets, dataset discovery falls back to the `else` clause and retrieves dataset info from the attached task.

Run any current tests for `get_run` and `get_runs` to ensure proper functionality.

### DIFF
--- a/openml/runs/functions.py
+++ b/openml/runs/functions.py
@@ -819,7 +819,7 @@ def _create_run_from_xml(xml, from_server=True):
     # whenever that's resolved, we can enforce it being present (OpenML#1087)
     run_details = obtain_field(run, "oml:run_details", from_server=False)
 
-    if "oml:input_data" in run:
+    if "oml:input_data" in run and len(run["oml:input_data"].get("oml:dataset")) == 1:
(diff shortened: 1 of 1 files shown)


PGijsbers:
Thanks for taking the time to open a pull request! I think this is something that should be fixed on the server side rather than the client side, but I am waiting for a reply on the matter in #1102. 

@mfeurer Do you think we should fix these type of errors regardless? To me it would not be obvious what the correct fallback behaviour should be - in which case I think an error is more appropriate. What does seem reasonable is to instead raise a more user-friendly error similar to "Contact the developers, raise an issue here:`<link>`.".

PGijsbers:
This will be fixed serverside, transfered the related issue: https://github.com/openml/OpenML/issues/1182

PR #1173, opened 2022-10-06
Flow exists GET is deprecated, use POST

PGijsbers:


mfeurer:
Hey, just checking, why is "flow_exists" a POST request? It is stateless and doesn't post anything on the server, so it should be a get request?

mfeurer (approved):

PR #1177, opened 2022-10-17
Remove dtype checking for prediction comparison

PGijsbers:
It looks like the predictions loaded from an arff file are read as floats by the arff reader, which results in a different type (float v int). Because "equality" of values is already checked, I figured dtype is not as imported. That said, I am not sure why there are so many redundant comparisons in the first place? Anyway, the difference should be due to pandas inference behavior, and if that is what we want to test, then we should make a small isolated test case instead of integrating it into every prediction unit test.

### DIFF
--- a/tests/test_runs/test_run_functions.py
+++ b/tests/test_runs/test_run_functions.py
@@ -183,7 +183,11 @@ class TestRun(TestBase):
         predictions_prime = run_prime._generate_arff_dict()
 
         self._compare_predictions(predictions, predictions_prime)
-        pd.testing.assert_frame_equal(run.predictions, run_prime.predictions)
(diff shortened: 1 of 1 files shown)


mfeurer (approved):
Fine by me, but why did this fail in the first place :confused:

mfeurer:
That description makes sense, but why did this work at some point at all?

The behavior of the ARFF reader is correct as ARFF does not specify integers. Maybe OpenML should also move away from ARFF for internal data structures.

PGijsbers:
> That description makes sense, but why did this work at some point at all?

My best guess is different behaviour of Pandas, but I haven't spent time nailing it down exactly (it's interesting, but didn't seem important).

PR #1184, opened 2022-11-24
feat(minio): Allow for proxies

### LINKED ISSUE #1159: NewConnectionError /datasets2/dataset_2.pq
Hi, I'm having difficulties downloading datasets and I need some help debugging it if possible.

Essentially I'm trying to do this simple task on a server but encountering errors, I'm not sure why it doesn't happen locally. This effects the ability for all users of our cluster to use automl_benchmark efficiently as we need to manually download data locally and scp it over to the cluster, causing possible issues with pandas versions when unpickling as has been a noted issue, one such example being #918. All other web traffic is handled without issue so I thought perhaps there was some misspecification in openml-python

```python
import openml
openml.tasks.get_task(2).get_data()
```

The error I eventually run into, which is more of a logged warning is ([source code of urllib3](https://github.com/urllib3/urllib3/blob/dd188963de6dcf51bd81b78f3fac10a79a7215bf/src/urllib3/connectionpool.py#L821-L823)):
```
WARNING:urllib3.connectionpool:Retrying (Retry(total=2, connect=None, read=None, redirect=None, status=None))
after connection broken by 'NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7ff83f3d2e20>:
Failed to establish a new connection: [Errno 110] Connection timed out')': /dataset2/dataset_2.pq
```

Seeing as it works locally, I tried to debug where openml does its called and came across `def _send_request(...)` so I put in some print debug statements which essentially give me the below. What seems to happen is that all normal url requests are handled fine but there is some web request happening outside of `def _send_request(...)` where things are going wrong and it seems to get `url="/dataset2/dataset_2.pq"` which is not a valid url.

```python
HIHIHIHIHIHIHIHIHIHI  # In function
5  # n_retries
{}  # params
https://www.openml.org/api/v1/xml/data/2  # url
1  # response pre checksum
<Response [200]>
2 # response after checksum
<Response [200]>

<Response [200]> # response at returning


HIHIHIHIHIHIHIHIHIHI
5
{}
https://www.openml.org/api/v1/xml/data/features/2
1
<Response [200]>
2
<Response [200]>
<Response [200]>

HIHIHIHIHIHIHIHIHIHI
5
{}
https://www.openml.org/api/v1/xml/data/qualities/2
1
<Response [200]>
2
<Response [200]>
<Response [200]>

HIHIHIHIHIHIHIHIHIHI
5
{}
https://api.openml.org/data/v1/download/1666876/anneal.arff
1
<Response [200]>
2
<Response [200]>
<Response [200]>

# Now it hangs
WARNING:urllib3.connectionpool:Retrying (Retry(total=4, connect=None, read=None, redirect=None, status=None)) after connection broken by 'NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7ff83ff2f6d0>: Failed to establish a new connection: [Errno 110] Connection timed out')': /dataset2/dataset_2.pq
WARNING:urllib3.connectionpool:Retrying (Retry(total=3, connect=None, read=None, redirect=None, status=None)) after connection broken by 'NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7ff83ff24100>: Failed to establish a new connection: [Errno 110] Connection timed out')': /dataset2/dataset_2.pq
WARNING:urllib3.connectionpool:Retrying (Retry(total=2, connect=None, read=None, redirect=None, status=None)) after connection broken by 'NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7ff83f3d2e20>: Failed to establish a new connection: [Errno 110] Connection timed out')': /dataset2/dataset_2.pq
```

I tried looking up `import requests`, `"pq"` and `"dataset` but this didn't help me find where this extra call could have been made. Any advice would be greatly appreciated.

```
openml            0.12.2
```

Best,
Eddie

eddiebergman:
#### Reference Issue
Fixes #1159

The [`Minio` client](https://min.io/docs/minio/linux/developers/python/API.html#constructor) does not automatically resolve proxies as is currently being done for `requests`. This PR uses `requests` strategy to automatically resolve the correct proxy and use this with the `Minio` client. With the shift to `.pq` files which are communcated with by `Minio`, this should help transition away from arff with less complaints.

This cause major annoyances on our cluster due where we have been manually downloading on local machines and copying datasets over. Previously, `Minio` would do repeated calls to the download server but have long timeouts and ignore `openml.config` settings.

I don't know, if you have a proxy, please use it along with this following script:
```bash
# Runs fine now
python test-proxy.py

# Will hang as we disable proxies which is apparently common in curl,wget,requests
no_proxy="*" python test-proxy.py
```

```python
# test-proxy.py
import openml
import shutil
from pathlib import Path

did = 2
path = Path(openml.config.get_cache_directory()) / "datasets" / str(did)
if path.exists():
    shutil.rmtree(str(path))
    
assert not path.exists()

d = openml.datasets.get_dataset(did, download_data=True)
print(d)

assert path.exists()
```

I'm unfamiliar with python's networking libraries and how to test proxies in general so I excluded tests.

#### Any other comments?
* _Proxies are handled without user knowledge with requests and so Minio will do so too_. I think this makes sense but if you wish to make it more explicit somehow, please let me know 
* I left a parameter `proxy: str | bool` in the `_download_minio_file` specifically to document the behavior and the user has no way to interact with this beyond environment variables (as is requests)
* I didn't think this required an example as there is non for requests and proxies, they are now alligned.
* Uni-freiburg and its cluster users would really appreciate a release with these changes! :clinking_glasses:

### DIFF
--- a/openml/_api_calls.py
+++ b/openml/_api_calls.py
@@ -10,6 +10,7 @@ import requests
 import urllib.parse
 import xml
 import xmltodict
+from urllib3 import ProxyManager
 from typing import Dict, Optional, Union
 
 import minio
@@ -22,6 +23,26 @@ from .exceptions import (
     OpenMLHashException,
 )
 
+
+def resolve_env_proxies(url: str) -> Optional[str]:
+    """Attempt to find a suitable proxy for this url.
+
+    Relies on ``requests`` internals to remain consistent. To disable this from the
+    environment, please set the enviornment varialbe ``no_proxy="*"``.
+
+    Parameters
+    ----------
+    url : str
+        The url endpoint
+
+    Returns
+    -------
+    Optional[str]
+        The proxy url if found, else None
+    """
+    resolved_proxies = requests.utils.get_environ_proxies(url)
+    selected_proxy = requests.utils.select_proxy(url, resolved_proxies)
+    return selected_proxy
 
 def _create_url_from_endpoint(endpoint: str) -> str:
     url = config.server
@@ -84,6 +105,7 @@ def _download_minio_file(
     source: str,
     destination: Union[str, pathlib.Path],
     exists_ok: bool = True,
+    proxy: Union[bool, str] = True
 ) -> None:
     """Download file ``source`` from a MinIO Bucket and store it at ``destination``.
 
@@ -95,7 +117,10 @@ def _download_minio_file(
         Path to store the file to, if a directory is provided the original filename is used.
     exists_ok : bool, optional (default=True)
         If False, raise FileExists if a file already exists in ``destination``.
-
+    proxy: Union[bool, str]
+        If True (default), environemnt variables will be parsed using ``requests`` to find a
+        suitable proxy. If False, no proxy will be used. Finally, pass a str to use a
+        custom proxy. To disable this from the environment varialbes, use ``no_proxy="*"``.
     """
     destination = pathlib.Path(destination)
     parsed_url = urllib.parse.urlparse(source)
@@ -107,7 +132,18 @@ def _download_minio_file(
     if destination.is_file() and not exists_ok:
         raise FileExistsError(f"File already exists in {destination}.")
 
-    client = minio.Minio(endpoint=parsed_url.netloc, secure=False)
+    if isinstance(proxy, str):
+        proxy_client = ProxyManager(proxy)
+    else:
+        proxy_url = resolve_env_proxies(parsed_url.geturl()) if proxy else None
+        if proxy_url:
+            proxy_client = ProxyManager(proxy_url)
+
+    client = minio.Minio(
+        endpoint=parsed_url.netloc,
+        secure=False,
+        http_client=proxy_client
+    )
 
     try:
         client.fget_object(


openml/_api_calls.py
```diff
@@ -23,6 +24,26 @@
 )
 
 
+def resolve_env_proxies(url: str) -> Optional[str]:
+    """Attempt to find a suitable proxy for this url.
+
+    Relies on ``requests`` internals to remain consistent. To disable this from the
+    environment, please set the enviornment varialbe ``no_proxy="*"``.
```
PGijsbers, line 31:
```suggestion
    environment, please set the environment variable ``no_proxy="*"``.
```

openml/_api_calls.py
```diff
@@ -95,7 +117,10 @@ def _download_minio_file(
         Path to store the file to, if a directory is provided the original filename is used.
     exists_ok : bool, optional (default=True)
         If False, raise FileExists if a file already exists in ``destination``.
-
+    proxy: Union[bool, str]
```
PGijsbers, line 120:
```suggestion
    proxy: Union[bool, str]  (default=True)
```

openml/_api_calls.py
```diff
@@ -95,7 +117,10 @@ def _download_minio_file(
...
+        If True (default), environemnt variables will be parsed using ``requests`` to find a
+        suitable proxy. If False, no proxy will be used. Finally, pass a str to use a
+        custom proxy. To disable this from the environment varialbes, use ``no_proxy="*"``.
```
PGijsbers, line 123:
```suggestion
        custom proxy. To disable this from the environment variables, use ``no_proxy="*"``.
```

openml/_api_calls.py
```diff
@@ -107,7 +132,19 @@ def _download_minio_file(
...
         raise FileExistsError(f"File already exists in {destination}.")
 
-    client = minio.Minio(endpoint=parsed_url.netloc, secure=False)
+    proxy_client = None
+    if isinstance(proxy, str):
+        proxy_client = ProxyManager(proxy)
+    else:
+        proxy_url = resolve_env_proxies(parsed_url.geturl()) if proxy else None
+        if proxy_url:
+            proxy_client = ProxyManager(proxy_url)
```
PGijsbers, line 141: If we were to change ``proxy`` to be of type `str` and use a default `"auto"`, instead of relying on mixed types, the resulting code would look much nicer (unfortunately I can't make the suggestion in the same batch because of the github web interface). As I understand it, the behavior should still be equivalent.
```suggestion
    if proxy == "auto":
      proxy = resolve_env_proxies(parsed_url.geturl())
    proxy_client = ProxyManager(proxy) if proxy else None
```

PGijsbers (changes requested):
> I'm unfamiliar with python's networking libraries and how to test proxies in general so I excluded tests.

Same here, unfortunately. A manual test at least confirms the defaults don't break when not behind a proxy 😅 

> Uni-freiburg and its cluster users would really appreciate a release with these changes!

I was planning a release today, let's get it in! (btw, is it correct you are not on OpenML Slack?)

eddiebergman:
I put in your `"auto"` suggestion and it's looking a little cleaner, thanks for that!

> Same here, unfortunately. A manual test at least confirms the defaults don't break when not behind a proxy sweat_smile

That's good at least! I would be lying if I said I actually tried it without a proxy, I was hoping the tests would all catch that :) Brief look at the failing tests makes me think it's unrelated?

> (btw, is it correct you are not on OpenML Slack?)
I'm not actually, if it's relatively active I'd be thrilled to join to know what's going on!

PGijsbers (approved):

PGijsbers:
> I'm not actually, if it's relatively active I'd be thrilled to join to know what's going on!

Not always, but it can be very convenient to quickly discuss things or ask a question :) (for example, the server issue last week, or me notifying you of the new openml-python release so you can do a quick test later)

PR #1187, opened 2022-11-25
Download bucket

PGijsbers:
Adds an experimental feature which allows you to download all the files in a dataset bucket. For now, the intended use case is limited to meta-album datasets, where the bucket contains additional folder(s)/archive(s) with images. It's switched off by default.


openml/datasets/functions.py
```diff
@@ -432,7 +443,8 @@ def get_dataset(
             else:
                 raise
 
-        arff_file = _get_dataset_arff(description) if download_data else None
+        # arff_file = _get_dataset_arff(description) if download_data else None
```
PGijsbers, line 446:
```suggestion
```

Issue #1069, opened 2021-05-03
Documentation Update


ArlindKadra:
https://github.com/openml/openml-python/blob/6e8a9db03fd1af9d3eb0623970101413b531cc69/openml/runs/run.py#L28-L63

This probably we should update, considering there are some fields that do not have a description and might be clearer with one.

v-parmar:
Hey, What should you want to update?

PGijsbers:
For each attribute we would want a description of what it represents, e.g.:
```diff
     trace: OpenMLRunTrace 
+     The trace containing information on internal model evaluations of this run.
     model: object 
+     The untrained model that was evaluated in this run.
```

v-parmar:
Hey, I have done the given task. How can I submit it?

PGijsbers:
We documented the process in our [contribution guidelines](https://github.com/openml/openml-python/blob/develop/CONTRIBUTING.md#general-git-workflow), in short:
 - push to your fork
 - check that it meets all the criteria to open a pull request (optionally add things if it doesn't)
 - set up a pull request and wait for review

v-parmar:
Hey @PGijsbers, I have opened the PR on issue #1194 and added the description please check it.

mfeurer:
This was solved by @v-parmar via #1194.

PR #1193, opened 2023-02-05
Add files Documentation.txt for Documentation Update #1069

v-parmar:
#### Reference Issue
Documentation Update #1069

Added files Documentation.txt, which contains descriptions for each attribute.

### DIFF
--- a/Documentation.txt
+++ b/Documentation.txt
@@ -1,0 +1,25 @@
+task_id: 				an integer representing the task id
+flow_id: 				an integer representing the flow id
+dataset_id: 			an integer representing the dataset id
+setup_string: 			a string representing the setup
+output_files: 			a dictionary representing the output files, where the key is the name of the file and the value is the file path
+setup_id: 				an integer representing the setup id
+tags: 				a list of strings representing tags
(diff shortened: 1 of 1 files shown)


PGijsbers:
First, thank you for taking the time. However, I'm sorry to say but there seems to be some miscommunication here. We do not want these descriptions in a separate file, but we want descriptions about the parameters as part of the docstring of the function. The documentation should also, if applicable, tell a little more about what exactly the parameter is used for or what it should contain. I gave some [examples](https://github.com/openml/openml-python/issues/1069#issuecomment-1368940372) in the open issue thread. For example, for `model` I suggested something like `"The untrained model that was evaluated in this run."`. This gives more information to the user: in this case specifically, that the model is not expected to be a trained model. Adding this information does require some knowledge about how to use `openml-python`, so feel free to try out working with `OpenMLRun`s first to understand it.

I'll close this PR now, I don't think the documentation is useful to be transferred elsewhere. If you can provide more complete descriptions as part of the function docstring, feel free to open another PR. If you have any further questions, please leave a message on the issue (#1069).

PR #1194, opened 2023-02-07
Update run.py

v-parmar:
#### Reference Issue
Documentation Update #1069

Here, I added a description for each attribute of what it represents.

### DIFF
--- a/openml/runs/run.py
+++ b/openml/runs/run.py
@@ -31,54 +31,53 @@ class OpenMLRun(OpenMLBase):
     Parameters
     ----------
     task_id: int
-        An integer representing the ID of the OpenML task associated with the run.
+        The ID of the OpenML task associated with the run.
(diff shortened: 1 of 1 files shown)


LennartPurucker:
Hey @v-parmar,

Thank you very much for the update. But I think you only updated some of it. @mfeurer specifically wanted to change the duplicate information. These duplicates still exist in the current version. 

If you compare my version from above and the current version, it is mainly about the start of the parameters' descriptions, which usually start with explaining the type. As far as I can tell, this was the problem.

v-parmar:
Hey @mfeurer, as you said I have updated the description and removed duplicate information. @LennartPurucker Thanks for the hint.

LennartPurucker (approved):

mfeurer (approved):

PR #1200, opened 2023-02-11
Update tests for sklearn 1.2, server issue

PGijsbers:
A few tests got outdated with `scikit-learn >= 1.2`. One tests seems to fail due to the test server, for which I opened an issue [here](https://github.com/openml/OpenML/issues/1180). After these fixes I get all green again (together with #1199).

### DIFF
--- a/tests/test_extensions/test_sklearn_extension/test_sklearn_extension.py
+++ b/tests/test_extensions/test_sklearn_extension/test_sklearn_extension.py
@@ -338,6 +338,7 @@ class TestSklearnExtensionFlowFunctions(TestBase):
                 )
             )
         else:
+            n_init = '"warn"' if LooseVersion(sklearn.__version__) >= "1.2" else "10"
             fixture_parameters = OrderedDict(
                 (
                     ("algorithm", '"lloyd"'),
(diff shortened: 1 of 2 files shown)


tests/test_extensions/test_sklearn_extension/test_sklearn_extension.py
```diff
@@ -1449,22 +1450,19 @@ def test_deserialize_complex_with_defaults(self):
...
-            }
+        impute_strategy = "median" if LooseVersion(sklearn.__version__) < "0.23" else "mean"
+        sparse = LooseVersion(sklearn.__version__) >= "0.23"
+        estimator_name = (
+            "base_estimator" if LooseVersion(sklearn.__version__) < "1.2" else "estimator"
+        )
+        params = {
+            "Imputer__strategy": impute_strategy,
+            "OneHotEncoder__sparse": sparse,
+            "Estimator__n_estimators": 10,
```
mfeurer, line 1461: This was higher in the previous version, is there a reason you had to change this?

tests/test_runs/test_run_functions.py
```diff
@@ -410,10 +410,13 @@ def test_check_erronous_sklearn_flow_fails(self):
 
         # Invalid parameter values
         clf = LogisticRegression(C="abc", solver="lbfgs")
-        with self.assertRaisesRegex(
-            ValueError,
-            r"Penalty term must be positive; got \(C=u?'abc'\)",  # u? for 2.7/3.4-6 compability
-        ):
+        # The exact error message depends on scikit-learn version.
+        # Because the sklearn-extension module is to be separated,
+        # I will simply relax specifics of the raised Error.
```
mfeurer, line 415: Would it be possible to test on a more relaxed regex so we can make sure that we test for the right thing?

mfeurer (approved):
This looks good to me, although the tests fail because they don't clean up local parquet files. But maybe that issue can be fixed in a separate PR.

PR #1202, opened 2023-02-12
Mark sklearn tests

### LINKED ISSUE #1186: Mark all tests that actually use scikit-learn
While eventually we want to remove the extension logic from the core package into a separate extension package, for the time being scikit-learn will be part of openml-python. That said, most of the package does not interact with scikit-learn. By marking tests which specifically use scikit-learn, we can drastically reduce CI/test server load (and false negatives). Currently, the test matrix runs the entire suite every time, while only changing scikit-learn version (which, again, is irrelevant to most tests). We could run the core suite once, and only scikit-learn sensitive tests multiple times (once per supported version). CI would have to be updated to only run on marked tests in those settings.

PGijsbers:
Closes #1186 

Marks 103 tests that use scikit-learn. I hope it is exhaustive, though it is possible I missed a few.
Updated the Github Workflow to _only_ run scikit-learn dependent tests by default, as most matrix jobs are specifically to test different scikit-learn versions (and non sklearn tests should not be affected by scikit-learn version). The code coverage and windows job still run the full suite.

Built on top of #1199, #1200 (hence the draft state).

### DIFF
--- a/tests/test_runs/test_run.py
+++ b/tests/test_runs/test_run.py
@@ -102,6 +102,7 @@ class TestRun(TestBase):
         else:
             self.assertIsNone(run_prime_trace_content)
 
+    @pytest.mark.sklearn
     def test_to_from_filesystem_vanilla(self):
 
         model = Pipeline(
@@ -137,6 +138,7 @@ class TestRun(TestBase):
             "collected from {}: {}".format(__file__.split("/")[-1], run_prime.run_id)
         )
 
+    @pytest.mark.sklearn
     @pytest.mark.flaky()
     def test_to_from_filesystem_search(self):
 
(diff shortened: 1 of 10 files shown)


tests/test_runs/test_run.py
```diff
@@ -189,6 +191,7 @@ def test_to_from_filesystem_no_model(self):
         with self.assertRaises(ValueError, msg="Could not find model.pkl"):
             openml.runs.OpenMLRun.from_filesystem(cache_path)
 
+    @pytest.mark.sklearn
```
mfeurer, line 194: There's a mark missing in the function above.

mfeurer (approved):

Issue #1197, opened 2023-02-10
Can't upload rmse,mae and other results to openml server


SharathKumarReddyAlijarla:
#### Description

Can't upload rmse,mae and other results to openml server
#### Steps/Code to Reproduce

`    root = logging.getLogger()
    root.setLevel(logging.INFO)

    args = parse_args()
    if args.openml_apikey:
        openml.config.apikey = args.openml_apikey
    if args.openml_server:
        openml.config.server = args.openml_server
    else:
        openml.config.server = 'https://test.openml.org/api/v1/'
    clf_pipeline=pipeline()
    run = openml.runs.run_model_on_task(clf_pipeline, args.task_id)
    score = []
    evaluations = run.fold_evaluations['mean_absolute_error'][0]
    for key in evaluations:
        score.append(evaluations[key])
    mae=np.mean(score)
    print(np.mean(score))
    print(np.mean(run.get_metric_fn(sklearn.metrics.mean_squared_error)))
    print(np.mean(run.get_metric_fn(sklearn.metrics.r2_score)))
    logging.info('Task %d - %s; Accuracy: %0.2f' % (args.task_id, task.get_dataset().name, mae))
    
    del_outputdir()
    run.to_filesystem(directory= 'myrun')
    print(args.upload_result)
    
    if args.upload_result:
        run.publish()
        print(run)
        print('Results Uploaded to Openml and stored in local folder myrun')
    else:
        print('Results stored in local folder myrun')`
<!--
Example:
```python
import openml
openml.flows.get_flow(1)
```

```
If the code is too long, feel free to put it in a public gist and link
it in the issue: https://gist.github.com
-->

#### Expected Results
<!-- Example: No error is thrown. Please paste or describe the expected results.-->
want to upload mean absolute error value to openml runs. run is uploaded but this values are not uploaded when run.publish() is executed.I am only able to get results locally,
#### Actual Results
<!-- Please paste or specifically describe the actual output or traceback. -->

#### Versions
<!--
Please run the following snippet and paste the output below.
import platform; print(platform.platform())
import sys; print("Python", sys.version)
import numpy; print("NumPy", numpy.__version__)
import scipy; print("SciPy", scipy.__version__)
import sklearn; print("Scikit-Learn", sklearn.__version__)
import openml; print("OpenML", openml.__version__)
-->


<!-- Thanks for contributing! -->

LennartPurucker:
Solved in #1209

Issue #559, opened 2018-09-27
float precision


joaquinvanschoren:
When a dataset is loaded (from an arff file), the values are converted to floats:

> datasets.py:348:  target_dtype = int if target_categorical[0] else float

That's fine in itself, but when doing regression experiments, we need to upload both the predicted and actual (truth) value. However, when exporting the actual value back to strings (in the arff file), precision is lost.

For instance, here are the first values of the training set of task 738 (test server):
> 6.1, 6.3, 5.8, 6.1, 6.0, 5.9

This is what is exported to the predictions arff file:
> 6.0999999   6.30000019  5.80000019  6.0999999   6.19999981  6.          5.9000001

What would be the best way to fix this? Load the data as Decimals instead of floats? Round the data before writing to arff?

joaquinvanschoren:
Another solution could be to store the exact values (string or decimal) as a component of the Dataset object when it first reads the data. That would ensure that you could pickle and cache it and still get the exact values back?

mfeurer:
Thanks for figuring this out. Indeed, the issue is the conversion of the target column into floating point representation.

> Another solution could be to store the exact values (string or decimal) as a component of the Dataset object when it first reads the data.

That's a great idea, but doing it exactly like this will probably not work because the dataset won't know that it is a regression dataset. However, having a function called `Dataset.get_column()` which returns the values of a column as raw python data types could help here (and also caches it). This would then be called by the task class with the necessary argument (the column). What do you think of this solution?

PGijsbers:
An actual run on the live server is e.g. [10154904](https://www.openml.org/r/10154904).

LennartPurucker:
So I looked into this again, and I am somewhat sure that the problem described above was a result of the problem fixed in #1209 

I am not able to reproduce the problem for task 738. The ground truth in the arff file seems to have the correct precision. 

I used the following code (with the openml's version from the current dev branch): 

```python
from sklearn.dummy import DummyRegressor
from openml import tasks, runs
import openml

openml.config.server = "https://test.openml.org/api/v1/xml"

reg = DummyRegressor()
task = tasks.get_task(738)
run = runs.run_model_on_task(reg, task, avoid_duplicate_runs=False)
run.to_filesystem("./tmp_dir")
```

The exported ground truth values and their precision match the values of the downloaded dataset. 

[...]

PR #1209, opened 2023-02-20
Fix: correctly order the ground truth and prediction for ARFF files in run.data_content

LennartPurucker:
### Bug - What does this PR fix?
The order of the ground truth and predictions is mixed up in the current implementation of how a run stores prediction data and how it uses the prediction data to build an ARFF file. 

As a result, the ground truth is treated as the predictions, and the predictions are treated as the ground truth. 
Consequently, publishing a run to uploads the wrong values for these columns to the OpenML server. 

### Impact 
This is not validated on the server side (to my understanding). Hence, all ARFF files of predictions uploaded using the OpenML Python API are most likely wrong. Moreover, evaluations of such runs also report the wrong scores. This might have impacted the results of papers that used scores of runs uploaded by the Python Client for meta-analysis. 

#### Reference Issues
Multiple issues exist as a result of this. The following issues are likely related to this problem: #1197, ~~#559~~, #1131

### Fix
I changed the order in the appropiate places and added a test for the (IMO) expected/correct behavior.  Additionally, I changed the tests that checked for the old order to the new order. 

#### New Order 
The order I used follows the order of ARFF files uploaded by the R Client API. I used the following code snippet to find the order. 
```python
from openml.runs import get_run

# 1875133 is a run from mlr 
print(get_run(1875133).predictions_url)

# After downloading and opening the file stored under the predictions_url, one can see that the file's order is:
# @attribute 'repeat' numeric
# @attribute 'fold' numeric
# @attribute 'row_id' numeric
# @attribute 'prediction' {'1','2'}
# @attribute 'truth' {'1','2'}
# @attribute 'confidence.1' numeric
# @attribute 'confidence.2' numeric

``` 

### Open Questions
~~The contribution guidelines mention that I should change the progress.rst. I am unsure if this would be below 13.1 and what I should mention there. What do you think?~~ I updated the progress.rst for the newest version. 

### Required (Server-Side) Follow-up Actions
This bug affects a lot of already published runs on OpenML. 
We might need to change/adjust the uploaded ARFF files and re-evaluate all these runs.


tests/test_runs/test_run.py
```diff
@@ -189,47 +197,105 @@ def test_to_from_filesystem_no_model(self):
...
+
+        for fold_id in range(s_d[1]):
+            _, test_indices = task.get_train_test_split_indices(repeat=0, fold=fold_id, sample=0)
+            fold_map[test_indices] = fold_id
+
+        X, y = task.get_X_and_y()
+
+        # Check correctness of y_ture and y_pred in run
+        for fold_id in range(s_d[1]):
+            # Get data for fold
```
mfeurer, line 216: Why can't we simply read the masks from the splits?

tests/test_runs/test_run.py
```diff
@@ -50,10 +52,16 @@ def _test_run_obj_equals(self, run, run_prime):
                     self.assertDictEqual(other, dict())
         self.assertEqual(run._to_xml(), run_prime._to_xml())
 
-        numeric_part = np.array(np.array(run.data_content)[:, 0:-2], dtype=float)
-        numeric_part_prime = np.array(np.array(run_prime.data_content)[:, 0:-2], dtype=float)
-        string_part = np.array(run.data_content)[:, -2:]
-        string_part_prime = np.array(run_prime.data_content)[:, -2:]
+        # Determine which attributes are numeric and which not
```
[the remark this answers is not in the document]
LennartPurucker: The previous implementation hard-coded for the old order which parts are numeric and which parts are strings. 

This way, we can use the ARFF description to determine the types. If the order changes at some point in the future, this should not brake.

tests/test_runs/test_run.py
LennartPurucker: Which function would you use for that / are you referring to?

tests/test_runs/test_run.py
```diff
@@ -189,47 +197,105 @@ def test_to_from_filesystem_no_model(self):
...
-        # obtain run from filesystem
-        loaded_run = openml.runs.OpenMLRun.from_filesystem(cache_path)
-        loaded_run.publish()
-        TestBase._mark_entity_for_removal("run", loaded_run.run_id)
-        TestBase.logger.info(
-            "collected from {}: {}".format(__file__.split("/")[-1], loaded_run.run_id)
-        )
+            # Make sure that the flow has not been uploaded as requested.
+            self.assertFalse(openml.flows.flow_exists(flow.name, flow.external_version))
+            self.assert_run_prediction_data(task, run, clone(model))
```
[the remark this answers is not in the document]
LennartPurucker: Done

tests/test_runs/test_run.py
LennartPurucker: Refactored loops to make this better

mfeurer (approved):
Looks good to me, let's wait for the unit tests to work again and then merge this.

doc/progress.rst
```diff
@@ -9,8 +9,7 @@ Changelog
 0.13.1
 ~~~~~~
 
- * Add new contributions here.
-
+ * FIX #1197 #559 #1131: Fix the order of ground truth and predictions in the ``OpenMLRun`` object and in ``format_prediction``.
```
PGijsbers, line 12: Can you add that it is specifically about regression tasks? thanks

openml/runs/functions.py
```diff
@@ -906,6 +905,14 @@ def obtain_field(xml_obj, fieldname, from_server, cast=None):
 
     tags = openml.utils.extract_xml_tags("oml:tag", run)
 
+    # Make sure default values are used where needed to keep run objects identical
+    if not evaluations:
+        evaluations = None
+    if not fold_evaluations:
+        fold_evaluations = None
+    if not sample_evaluations:
+        sample_evaluations = None
```
PGijsbers, line 914: Could instead pass the parameter value `evaluations or None` which means that it's none with an empty dict, and `evaluations` otherwise.

doc/progress.rst
LennartPurucker: The switched order is a problem for regression and classification tasks (and maybe learning curve).

openml/runs/functions.py
LennartPurucker: Indeed, good suggestion, I updated this. Looks much cleaner.

tests/test_runs/test_run.py
```diff
@@ -189,47 +211,155 @@ def test_to_from_filesystem_no_model(self):
...
+            # Assert identical for offline - offline
+            self._test_run_obj_equals(run, loaded_run)
+
+            # Publish and test for offline - online
+            run.publish()
+            self.assertTrue(openml.flows.flow_exists(flow.name, flow.external_version))
+
+            try:
+                online_run = openml.runs.get_run(run.run_id, ignore_cache=True)
+                self._test_run_obj_equals(run, online_run, only_check_prediction_data=True)
```
PGijsbers, line 356: Why not instead just call `check_prediction_data` directly?

LennartPurucker: Originally I wanted to check that the object is equal, but this does not work as the runs created from the sever have many differences from locally created runs. Hence, it grew like this. 

But at this point, we could just call `check_prediction_data`. Or do you have any other opinion on this?

PGijsbers: I'm not sure what the differences are.. but if they are not important then just calling `check_prediction_data` should be good enough I think.

tests/test_runs/test_run.py
```diff
@@ -189,47 +211,155 @@ def test_to_from_filesystem_no_model(self):
...
+    def assert_run_prediction_data(task, run, model):
+        # -- Get y_pred and y_true as it should be stored in the run
+        s_d = task.get_split_dimensions()
+        if (s_d[0] > 1) or (s_d[2] > 1):
+            raise ValueError("Test does not support this task type's split dimensions.")
+
+        X, y = task.get_X_and_y()
+
+        # Check correctness of y_ture and y_pred in run
+        for fold_id in range(s_d[1]):
```
PGijsbers, line 245:
```suggestion
        n_repeats, n_folds, n_samples = task.get_split_dimensions()
        if (n_repeats > 1) or (n_samples > 1):
            raise ValueError("Test does not support this task type's split dimensions.")

        X, y = task.get_X_and_y()

        # Check correctness of y_true and y_pred in run
        for fold_id in range(n_folds):
```

tests/test_runs/test_run.py
LennartPurucker: Done.

PGijsbers (approved):

mfeurer (approved):

Issue #1144, opened 2022-06-24
Report minimal evaluation results in run printout


joaquinvanschoren:
#### Description
When doing print(run), the result is only some basic info like the task type ID. It shows no aggregated evaluation information. It would be nice if that could be added to the printout.

Alternatively, add a function that returns a dictionary with the local evaluation results.
If you currently run a run locally (not downloaded from OpenML), `run.evaluations` is None.
`run.fold_evaluations`, however, contains accuracy and runtime results, but in a very inconvenient format.

#### Steps/Code to Reproduce
```python
from sklearn import ensemble
from openml import tasks, runs

clf = ensemble.RandomForestClassifier()
task = tasks.get_task(3954)
run = runs.run_model_on_task(clf, task, avoid_duplicate_runs=False)
print(run)
```

#### Expected Results
Some basic evaluation info. E.g. similar to whet the R API returns:
```
OpenML Run NA :: (Task ID = 3954, Flow ID = NA)

$bmr
task.id                    learner.id                        acc.test.join          timetrain.test.sum
1 MagicTelescope classif.randomForest     0.8831756            150.753
```

#### Actual Results
Only some basic info
```
OpenML Run
==========
Uploader Name: None
Metric.......: None
Run ID.......: None
Task ID......: 3954
Task Type....: None
Task URL.....: https://www.openml.org/t/3954
Flow ID......: None
Flow Name....: sklearn.ensemble._forest.RandomForestClassifier
Flow URL.....: https://www.openml.org/f/None
Setup ID.....: None
Setup String.: Python_3.7.13. Sklearn_1.0.2. NumPy_1.21.6. SciPy_1.4.1.
Dataset ID...: 1120
Dataset URL..: https://www.openml.org/d/1120
```

#### Partial solution
This returns the accuracy and runtime for a run (even if run only locally)
```
def summary(metric):
  mean = np.mean(list(run.fold_evaluations[metric][0].values()))
  var = np.std(list(run.fold_evaluations[metric][0].values()))
  return "{:.4f} +- {:.4f}".format(mean,var)
print("Accuracy: ",summary('predictive_accuracy'))
print("Time (ms): ",summary('usercpu_time_millis'))
```

#### Versions
Linux-5.4.188+-x86_64-with-Ubuntu-18.04-bionic
Python 3.7.13 (default, Apr 24 2022, 01:04:09) 
[GCC 7.5.0]
NumPy 1.21.6
SciPy 1.4.1
[...]

PR #1214, opened 2023-02-22
Add summary of locally computed metrics to representation of run  

LennartPurucker:
This addresses #1144 and might be a solution for the issue. 

It is important to note that the representation already included a "Result" field and thus could have reported results in the printout. However, this only happens if the `evaluation` of the run is filled. 

This implementation now also covers locally computed results. The locally computed results are only shown if no task/run-specific results already exist. 

### Example 
```python
from sklearn import ensemble
from openml import tasks, runs

clf = ensemble.RandomForestClassifier()
task = tasks.get_task(3954)
run = runs.run_model_on_task(clf, task, avoid_duplicate_runs=False)
print(run)
```

Old Output:

```
OpenML Run
==========
Uploader Name: None
Metric.......: None
Run ID.......: None
Task ID......: 3954
Task Type....: None
Task URL.....: https://www.openml.org/t/3954
Flow ID......: None
Flow Name....: sklearn.ensemble._forest.RandomForestClassifier
Flow URL.....: https://www.openml.org/f/None
Setup ID.....: None
Setup String.: Python_3.7.13. Sklearn_1.0.2. NumPy_1.21.6. SciPy_1.4.1.
Dataset ID...: 1120
Dataset URL..: https://www.openml.org/d/1120
```

New Output (ignore the difference in `Setup String`):
```
OpenML Run
==========
Uploader Name...................: None
Metric..........................: None
Local Result - Accuracy (+- STD): 0.8790 +- 0.0054
Local Runtime - ms (+- STD).....: 9662.5000 +- 2621.0648
Run ID..........................: None
Task ID.........................: 3954
Task Type.......................: None
Task URL........................: https://www.openml.org/t/3954
Flow ID.........................: None
Flow Name.......................: sklearn.ensemble._forest.RandomForestClassifier
Flow URL........................: https://www.openml.org/f/None
Setup ID........................: None
Setup String....................: Python_3.9.10. Sklearn_1.2.1. NumPy_1.23.5. SciPy_1.10.0.
Dataset ID......................: 1120
Dataset URL.....................: https://www.openml.org/d/1120
```

### Test
I am not sure if we need to write a test for this. I did not find any tests for the representation so far.

### DIFF
--- a/doc/progress.rst
+++ b/doc/progress.rst
@@ -13,7 +13,7 @@ Changelog
  * ADD#1144: Add locally computed results to the ``OpenMLRun`` object's representation.
  * FIX #1197 #559 #1131: Fix the order of ground truth and predictions in the ``OpenMLRun`` object and in ``format_prediction``.
  * FIX #1198: Support numpy 1.24 and higher.
- * ADD#1144: Add locally computed results to the ``OpenMLRun`` object's representation.
+ * ADD#1144: Add locally computed results to the ``OpenMLRun`` object's representation if the run was created locally and not downloaded from the server.
 
 0.13.0
 ~~~~~~
--- a/doc/progress.rst
+++ b/doc/progress.rst
@@ -11,6 +11,7 @@ Changelog
 
  * FIX #1197 #559 #1131: Fix the order of ground truth and predictions in the ``OpenMLRun`` object and in ``format_prediction``.
  * FIX #1198: Support numpy 1.24 and higher.
+ * ADD#1144: Add locally computed results to the ``OpenMLRun`` object's representation.
 
 0.13.0
 ~~~~~~
--- a/doc/progress.rst
+++ b/doc/progress.rst
@@ -9,6 +9,8 @@ Changelog
 0.13.1
 ~~~~~~
 
+ * Add new contributions here.
+ * ADD#1144: Add locally computed results to the ``OpenMLRun`` object's representation.
  * FIX #1197 #559 #1131: Fix the order of ground truth and predictions in the ``OpenMLRun`` object and in ``format_prediction``.
  * FIX #1198: Support numpy 1.24 and higher.
  * ADD#1144: Add locally computed results to the ``OpenMLRun`` object's representation.
--- a/openml/runs/run.py
+++ b/openml/runs/run.py
@@ -31,36 +31,55 @@ class OpenMLRun(OpenMLBase):
     Parameters
     ----------
     task_id: int
+        The ID of the OpenML task associated with the run.
     flow_id: int
+        The ID of the OpenML flow associated with the run.
     dataset_id: int
+        The ID of the OpenML dataset used for the run.
     setup_string: str
+        The setup string of the run.
     output_files: Dict[str, str]
-        A dictionary that specifies where each related file can be found.
+        Specifies where each related file can be found. 
     setup_id: int
+        An integer representing the ID of the setup used for the run.
     tags: List[str]
+        Representing the tags associated with the run.
     uploader: int
-        User ID of the uploader.
+        User ID of the uploader. 
     uploader_name: str
+        The name of the person who uploaded the run.
     evaluations: Dict
+        Representing the evaluations of the run.
     fold_evaluations: Dict
+        The evaluations of the run for each fold.
     sample_evaluations: Dict
+        The evaluations of the run for each sample.
     data_content: List[List]
         The predictions generated from executing this run.
     trace: OpenMLRunTrace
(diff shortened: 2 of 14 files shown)


openml/runs/run.py
```diff
@@ -139,8 +139,37 @@ def predictions(self) -> pd.DataFrame:
     def id(self) -> Optional[int]:
         return self.run_id
 
+    def evaluation_summary(self, metric: str):
```
mfeurer, line 142:
```suggestion
    def evaluation_summary(self, metric: str) -> str:
```

openml/runs/run.py
```diff
@@ -139,8 +139,37 @@ def predictions(self) -> pd.DataFrame:
...
+            A formatted string that displays the metric's evaluation summary.
+            The summary consists of the mean and std.
+        """
+        fold_score_lists = self.fold_evaluations[metric].values()
+
+        # Get the mean and std over all repetitions
+        rep_means = [np.mean(list(x.values())) for x in fold_score_lists]
+        rep_stds = [np.std(list(x.values())) for x in fold_score_lists]
+
+        return "{:.4f} +- {:.4f}".format(np.mean(rep_means), np.mean(rep_stds))
```
mfeurer, line 168: Does this work for all kinds of tasks?

LennartPurucker: Good question, as far as I can tell, the fold evaluations should be stored in such a "list of lists" container for all tasks. 

But I have not found an overview for all tasks and split settings to confirm this. I could a test for that. 
Do you know the conventions for this, or is this specified somewhere?

mfeurer: We store the task types `openml.tasks.task`, and we might store the evaluation protocols somewhere, too. Then you could query the server for runs that follow the cross-product of the two and check those.

LennartPurucker: I will change the design such that the current PR only enables the feature for supported task types (SUPERVISED_CLASSIFICATION, LEARNING_CURVE, SUPERVISED_REGRESSION). 

Moreover, I will add a test for the print and string representation.

LennartPurucker: This will now only be for local measures where we also have fold evaluations. Specifically, this now only aggregates the fold scores computed here: https://github.com/openml/openml-python/blob/develop/openml/runs/functions.py#L582

doc/progress.rst
```diff
@@ -11,6 +11,7 @@ Changelog
 
  * FIX #1197 #559 #1131: Fix the order of ground truth and predictions in the ``OpenMLRun`` object and in ``format_prediction``.
  * FIX #1198: Support numpy 1.24 and higher.
+ * ADD#1144: Add locally computed results to the ``OpenMLRun`` object's representation.
```
mfeurer, line 14: Maybe you could add a half sentence that this only applies if the run wasn't downloaded from the server?

tests/test_runs/test_run_functions.py
```diff
@@ -533,6 +533,14 @@ def determine_grid_size(param_grid):
 
         # todo: check if runtime is present
         self._check_fold_timing_evaluations(run.fold_evaluations, 1, num_folds, task_type=task_type)
+
+        # Check if run string and print representation do not run into an error
```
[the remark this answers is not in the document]
LennartPurucker: AFAIK both str(run) and print/logger.info(run) will result in calling OpenMLRun._get_repr_body_fields. The run's __repr__ will be called, which will call _get_repr_body_fields.  

Hence, if both do not crash across the many different calls to _run_and_upload, then we can be somewhat sure that the newly added fields in the representation are not a problem. 

Luckily, we also do not need to check if the state of the run is correct as _check_fold_timing_evaluations does that for us already.

doc/progress.rst
LennartPurucker: Done

openml/runs/run.py
mfeurer: Did you already do so?

LennartPurucker: Yes, it did not change the code here, but where the function is called. We could resolve this conversation if you are okay with it.

mfeurer: Yes, fine with me.

mfeurer:
Before merging this, could you please have a look at the pre-commit? That one is currently failing.

LennartPurucker:
I fixed the doc string for the run that was malformed as a result of another PR. The pre-commit should work now.

mfeurer (approved):
Since the remaining failures are due to the test server this looks ready to merge for me.

PR #1218, opened 2023-02-24
Fix CI Python 3.6

mfeurer:
by using Ubuntu 20.04 instead of the latest Ubuntu (22.04).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,15 +19,15 @@ jobs:
           - python-version: 3.6
             scikit-learn: 0.18.2
             scipy: 1.2.0
-            os: ubuntu-latest
+            os: ubuntu-20.04 
             sklearn-only: 'true'
           - python-version: 3.6
             scikit-learn: 0.19.2
-            os: ubuntu-latest
+            os: ubuntu-20.04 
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -8,7 +8,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8]
+        python-version: [3.7, 3.8]
         scikit-learn: [0.21.2, 0.22.2, 0.23.1, 0.24]
         os: [ubuntu-latest]
```
LennartPurucker, line 13: If we keep `ubuntu-latest` here, might the same problem not happen again in the future? 
Or do we want to keep `ubuntu-latest`?

PGijsbers: Two things for why I am ok with this:
 1. this legacy support will be factored out into a separate scikit-learn extension module, after which only testing on `ubuntu-latest` seems fine.
 2. the correct way (I think) would be first to have some bot notification about changes to `ubuntu-latest` and then decide what to do with it based on what the new ubuntu supports

PGijsbers (approved):

PR #1219, opened 2023-02-24
Refactor if-statements

PGijsbers:


openml/extensions/sklearn/extension.py
```diff
@@ -580,15 +580,13 @@ def _is_cross_validator(self, o: Any) -> bool:
 
     @classmethod
     def _is_sklearn_flow(cls, flow: OpenMLFlow) -> bool:
-        if getattr(flow, "dependencies", None) is not None and "sklearn" in flow.dependencies:
-            return True
-        if flow.external_version is None:
-            return False
-        else:
-            return (
+        return (isinstance(flow.dependencies, str) and "sklearn" in flow.dependencies) or (
```
mfeurer, line 583: I understand the change away from `getattr`, but I'm not sure if the overall outcome is more readable.

LennartPurucker: I think this also results in a crash for windows, as not every flow has this attribute: https://github.com/openml/openml-python/actions/runs/4261463411/jobs/7415818507#step:9:792

PGijsbers: Yes. DummyFlow is not actually an `OpenMLFlow` object, which is why it crashes:

```python
class DummyFlow:
    external_version = "DummyFlow==0.1"
```

But `flow.dependencies` is set on `__init__` so any proper `OpenMLFlow` has the attribute.

PGijsbers: Refactored the method by first explicitly assigning the individual expressions to a variable. Of course we lose short-cutting the `or`, but that shouldn't ever matter.

LennartPurucker (approved):

mfeurer (approved):

Issue #1180, opened 2022-11-18
The data w/ id 41949 and its checksum don't match.


yoichii:
#### Description

We get a checksum error every time we try to download the dataset w/ id 41949.

#### Steps/Code to Reproduce

```
import openml
openml.datasets.get_dataset(41949)
```
#### Expected Results

Successfully get the dataset.
#### Actual Results

The data and its checksum don't match.
```
OpenMLHashException: Checksum of downloaded file is unequal to the expected checksum b1a29cd9bed190cd066f750fb8247528 when downloading https://api.openml.org/data/v1/download/21378932/Elegibilidade.arff. Raised when downloading dataset 41949.
```
#### Versions

Linux-4.15.0-192-generic-x86_64-with-glibc2.27
Python 3.9.12 (main, Apr  5 2022, 06:56:58) 
[GCC 7.5.0]
NumPy 1.23.4
SciPy 1.9.3
Scikit-Learn 1.1.3
OpenML 0.12.2

PGijsbers:
Thanks for reporting the issue. It seems to be some issue with the file encoding (`openml-python` assumes it is `utf-8`, but it is not. The file reports it is `ISO-8859` but there seem to be characters that are non-compliant, so it might be some variant.), we are looking into this. In the meantime, it is still possible to download the data from openml.org/d/41949, but I don't have a clean workaround for loading it with `openml-python`. Stay tuned :)

LennartPurucker:
Heyho,

so I looked into this, and so far, these are my finding:

The checksum of the server's text response is not equal to the server's checksum.
_However_, the checksum of the sever's (binary) content response is equal to the stored checksum. 

```python
response = session.get(url, params=data)  # Get the response from the server for this dataset (called by get_dataset)

__is_checksum_equal(response.text, md5_checksum)  # False 

md5 = hashlib.md5()
md5.update(response.content)
md5_checksum_content = md5.hexdigest()
checksum_equal_to_content = md5_checksum_content == md5_checksum   # True 
```

How should we go forward with this @PGijsbers? 
We could handle this edge case for this specific encoding ("windows-1252") by updating the response.text parameter. 

But this would then need to support any encoding and this would not be UTF-8 anymore. 
Otherwise, we could still return the dataset as UTF-8 (i.e., the wrong encoding) and assume that the user of the dataset needs to handle the encoding during preprocessing.

mfeurer:
[...]

PR #1225, opened 2023-03-01
Better Error for Checksum Mismatch

LennartPurucker:
See #1180

### DIFF
--- a/openml/_api_calls.py
+++ b/openml/_api_calls.py
@@ -297,7 +297,7 @@ def __read_url(url, request_method, data=None, md5_checksum=None):
     )
 
 
-def __is_checksum_equal(downloaded_file_binary, md5_checksum=None):
+def __is_checksum_equal(downloaded_file_binary: bytes, md5_checksum: Optional[str] = None) -> bool:
     if md5_checksum is None:
         return True
     md5 = hashlib.md5()
--- a/openml/_api_calls.py
+++ b/openml/_api_calls.py
@@ -324,6 +324,24 @@ def _send_request(request_method, url, data, files=None, md5_checksum=None):
                     raise NotImplementedError()
                 __check_response(response=response, url=url, file_elements=files)
                 if request_method == "get" and not __is_checksum_equal(response.text, md5_checksum):
+
+                    # -- Check if encoding is not UTF-8 perhaps
+                    # Get checksum for binary response content
+                    md5 = hashlib.md5()
+                    md5.update(response.content)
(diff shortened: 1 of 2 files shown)


openml/_api_calls.py
```diff
@@ -324,6 +324,24 @@ def _send_request(request_method, url, data, files=None, md5_checksum=None):
...
                 if request_method == "get" and not __is_checksum_equal(response.text, md5_checksum):
+
+                    # -- Check if encoding is not UTF-8 perhaps
+                    # Get checksum for binary response content
+                    md5 = hashlib.md5()
+                    md5.update(response.content)
+                    md5_checksum_content = md5.hexdigest()
+
+                    # Check if checksum is equal to the expected checksum and raise better error
+                    if md5_checksum_content == md5_checksum:
```
PGijsbers, line 335:
```suggestion 
                    if __is_checksum_equal(response.content, md5_checksum):
```

LennartPurucker: This won't work as `__is_checksum_equal()` assumes UTF-8 encoding and text as input. `response.content` is already a binary. I could rewrite `__is_checksum_equal()` to allow both if you prefer that?

PGijsbers: Ah my bad, I did a quick patch locally yesterday (trying to read the files to convert to parquet), and forgot I actually also changed `__is_checksum_equal` itself. I think it would be cleaner to rewrite `__is_checksum_equal`. I'm impartial to wether that means the function taking both str and bytes, or if you accept only bytes and call it with `repsonse.text.encode('utf-8')` first.

LennartPurucker: I updated it accordingly

openml/_api_calls.py
```diff
@@ -297,11 +297,11 @@ def __read_url(url, request_method, data=None, md5_checksum=None):
     )
 
 
-def __is_checksum_equal(downloaded_file, md5_checksum=None):
+def __is_checksum_equal(downloaded_file_binary, md5_checksum=None):
```
PGijsbers, line 300:
```suggestion
def __is_checksum_equal(downloaded_file_binary: bytes, md5_checksum: Optional[str] = None) -> bool:
```

PGijsbers (changes requested):
I think it should be good (added type hints), but let's wait until the test server is operational again.

LennartPurucker:
>  but let's wait until the test server is operational again.

Ah sorry, I only saw this part of the message afterward.

PGijsbers (approved):

Issue #1210, opened 2023-02-21
Add warning to run_model_on_task with avoid duplicates if no authentication


PGijsbers:
see #1145

Prachi194agrawal:
Hi @PGijsbers,

I noticed that this issue is still open and the previous linked PRs (#1240 and #1246) seem to be inactive since early 2023.

I would like to pick this up. I plan to add the UserWarning in openml/runs/functions.py when avoid_duplicate_runs=True is passed without an authenticated API key.

I can submit a PR for this shortly!

PR #1239, opened 2023-03-27
Documented remaining Attributes of classes and functions

v-parmar:
#### Reference Issue
#716 

#### What does this PR implement/fix? Explain your changes.
I have added the missing documentation for attributes of classes and functions.

PR #1240, opened 2023-03-29
Added warning to run_model_on_task to avoid duplicates if no authentication 

v-parmar:
#### Reference Issue
#1210


LennartPurucker:
Hey @v-parmar, thank you for this PR. 

Could you perhaps remove the other commits/changes that are not related to the issue and the purpose of this PR? 
I think your other PR (#1239) covers these changes already.

LennartPurucker:
Hey, @v-parmar, I think you deleted a lot of files to roll back your changes. We can not merge it like this. 
I think it might be best if you reset your branch to the current dev branch and then add your specific change by hand again. 

This should allow us to have the PR only for the specific change it is supposed to fix.

Issue #1227, opened 2023-03-06
Refactor tests: decouple from server state


PGijsbers:
Decouple openml-python tests from server state by mocking server responses.

PR #1243, opened 2023-04-15
Fix TaskID for Example Custom Flow

### LINKED ISSUE #1241: Wrong references in 'Creating Custom Flow' tutorial
#### Description
In section **2. Using the flow** of https://openml.github.io/openml-python/main/examples/30_extended/custom_flow_.html#sphx-glr-examples-30-extended-custom-flow-py the comment for `task_id=1965` references `# Iris Task` but 1965 is actually https://test.openml.org/t/1965 which uses the **anneal** dataset. 

Further below in the tutorial, predictions for the iris dataset are prepared.

```
# random class probabilities (Iris has 150 samples and 3 classes):
r = np.random.rand(150 * n_repeats, 3)
```

The iris dataset contains 4 classes. It's nothing major, just wanted to mention it.

LennartPurucker:
Closes #1241 

Moreover, we might want to add or rework the process for the examples if the IDs can change, maybe in relation to #1227.

### DIFF
--- a/examples/30_extended/custom_flow_.py
+++ b/examples/30_extended/custom_flow_.py
@@ -124,7 +124,7 @@ parameters = [
     OrderedDict([("oml:name", "time"), ("oml:value", 120), ("oml:component", flow_id)]),
 ]
 
-task_id = 1965  # Iris Task
+task_id = 1200  # Iris Task
 task = openml.tasks.get_task(task_id)
 dataset_id = task.get_dataset().dataset_id
 


LennartPurucker:
> Patch coverage has no change and project coverage change: **`-0.10`** ⚠️

We could add tests that verify that our examples are not crashing.
I am thinking about very simple tests like: 

```python
def test_run_custom_flow_example():
    example_code = __import__("examples.30_extended.custom_flow_")
```

This would increase the code coverage and would allow us to see if our examples are actually working. 
However, I am not sure if this is really good practice (but I also have no better idea right now).
 I would be fine with it, but I feel like this depends on your preferences.

LennartPurucker:
We might want to connect this to #1070 and resolve both at the same time.

LennartPurucker:
@mfeurer the usual tests seem to pass. Does that suffice for you?

LennartPurucker:
@mfeurer I think this is ready for merging now. Or do you require any other changes?

mfeurer (approved):

PGijsbers:
Examples are tested, I don't really see a point for calculating coverage metrics on the example code itself. As far as I am aware, this is also not something we do. The difference in code coverage is more likely to be because of changes to the test server state (and thus different error/code paths). If there are parts of the code only covered by examples and not unit tests, then the right way to correct that would be to add unit tests.

Issue #1234, opened 2023-03-27
KeyError on list_tasks


Innixma:
#### Description

OpenML crashes during list_tasks call.

It was working correctly 1 week ago.
I first noticed the error on March 25th when running a script to generate a task metadata file locally.

I've reproduced the error both on an ubuntu EC2 instance and locally on my Mac laptop.

#### Steps/Code to Reproduce

```python3
import openml
tasks = openml.tasks.list_tasks()
```

#### Exception

```
Traceback (most recent call last):
  File "/home/ubuntu/.conda/envs/code/lib/python3.8/site-packages/openml/tasks/functions.py", line 273, in __list_tasks
    task[input["@name"]] = proc_dict[int(input["#text"])]["name"]
KeyError: 0

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/ubuntu/workspace/code/scratch/scratch2.py", line 2, in <module>
    tasks = openml.tasks.list_tasks()
  File "/home/ubuntu/.conda/envs/code/lib/python3.8/site-packages/openml/tasks/functions.py", line 180, in list_tasks
    return openml.utils._list_all(
  File "/home/ubuntu/.conda/envs/code/lib/python3.8/site-packages/openml/utils.py", line 272, in _list_all
    new_batch = listing_call(
  File "/home/ubuntu/.conda/envs/code/lib/python3.8/site-packages/openml/tasks/functions.py", line 222, in _list_tasks
    return __list_tasks(api_call=api_call, output_format=output_format)
  File "/home/ubuntu/.conda/envs/code/lib/python3.8/site-packages/openml/tasks/functions.py", line 291, in __list_tasks
    raise KeyError("Invalid xml for task %d: %s\nFrom %s" % (tid, e, task_))
[...]

PR #1244, opened 2023-04-16
Rework list_tasks: change raise error to warning for bad tasks to avoid crashing for bad server state 

LennartPurucker:
Closes #1234

### DIFF
--- a/openml/tasks/functions.py
+++ b/openml/tasks/functions.py
@@ -288,9 +288,10 @@ def __list_tasks(api_call, output_format="dict"):
             tasks[tid] = task
         except KeyError as e:
             if tid is not None:
-                raise KeyError("Invalid xml for task %d: %s\nFrom %s" % (tid, e, task_))
+                warnings.warn("Invalid xml for task %d: %s\nFrom %s" % (tid, e, task_))
             else:
-                raise KeyError("Could not find key %s in %s!" % (e, task_))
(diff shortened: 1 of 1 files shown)


mfeurer:
Hey, is this still an issue? I don't think we should add workaround code for uncommon bugs on the server (i.e. they happened only once so far) and rather fix things on the server instead.

LennartPurucker:
> Hey, is this still an issue? I don't think we should add workaround code for uncommon bugs on the server (i.e. they happened only once so far) and rather fix things on the server instead.

Mhm, I would say the current workflow (i.e., crashing) is an issue, and the alternative workflow (i.e., warning instead of crashing) would be better. But I see your point that we technically are only following the server specification. 

I would leave this to your judgment. Feel free to close the PR.

PGijsbers:
I vote +1 on merging this.

While philosophically I agree that openml-python should not accommodate to all kind of buggy server states. However specifically for this case, I think the pragmatic approach to issue a warning instead of an error leads to a much better user experience. My reason for preferring this is because the error is more likely than not caused by tasks in which the user had no interest to begin with (to contrast, I would not add work-arounds for actually downloading and instantiating such a corrupted task, for example).

mfeurer (approved):

mfeurer:
Merging this now as two people would like to have this merged.

PR #1246, opened 2023-04-18
Added warning to run_model_on_task to avoid duplicates if no authentication

v-parmar:
Added warning to run_model_on_task to avoid duplicates if no authentication

#### Reference Issue
#1210

### DIFF
--- a/openml/runs/functions.py
+++ b/openml/runs/functions.py
@@ -98,7 +98,14 @@ def run_model_on_task(
     flow : OpenMLFlow (optional, only if `return_flow` is True).
         Flow generated from the model.
     """
-
+    if avoid_duplicate_runs and not config.apikey:
+        warnings.warn(
+            "avoid_duplicate_runs is set to True, but no API key is set. "
+            "Please set your API key in the OpenML configuration file "
+            "or pass it directly to the function.",
+            DeprecationWarning
+        )
+        
     # TODO: At some point in the future do not allow for arguments in old order (6-2018).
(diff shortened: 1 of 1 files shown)


openml/runs/functions.py
```diff
@@ -98,6 +98,13 @@ def run_model_on_task(
     flow : OpenMLFlow (optional, only if `return_flow` is True).
         Flow generated from the model.
     """
+    if avoid_duplicate_runs and not config.apikey:
+        warnings.warn(
+            "avoid_duplicate_runs is set to True, but no API key is set. "
+            "Please set your API key in the OpenML configuration file "
+            "or pass it directly to the function.",
+            DeprecationWarning,
+        )
```
PGijsbers, line 107:
```suggestion
        warnings.warn(
            "avoid_duplicate_runs is set to True, but no API key is set. "
            "Please set your API key in the OpenML configuration file, see"
            "https://openml.github.io/openml-python/main/examples/20_basic/introduction_tutorial.html#authentication"
            " for more information on authentication.",
        )
```

Two changes:
 1. I don't think `DeprecationWarning` is appropriate here because we are not informing about changed behavior/signature of the function, I would favor the `UserWarning` (which is the default warning type, so I left that argument implicit).
 2. There is no way to pass the API key to the function, so that help text is incorrect. I updated the text to point to the documentation.

PGijsbers (changes requested):
Thanks for your contribution! Please see my suggestion

PGijsbers:
Sorry, missed the update. LGTM. Thanks for your contribution!

PR #1254, opened 2023-06-12
Fix 1124: provide clear naming for cache directories

### LINKED ISSUE #1124: Provide clear naming for the inconsistent ways to get the cache directory
```python
import openml
openml.config.get_cache_directory()   # is:   '~\\.openml\\org\\openml\\www'
openml.config.cache_directory         # is:   '~\\.openml'
``` 

`get_cache_directory`  gets the cache for the currently configured server, while `cache_directory` gets the root of the cache directory.
`set_cache_directory(str)` configures the root.

If we plan to maintain an external interface for both types, I propose we:
 - we make them both functions so they are accessed in the same way (and using a function will make a get\set pattern easier where a set can be optionally for the session or saved to the configuration file)
 - make it clear in the name what (part of) the cache directory is returned

mfeurer:
#### Reference Issue
Fixes #1124

* Introduce the concept of a `root_cache_directory`, i.e. the cache directory in which the concrete cache directories for the different servers reside
* Rename the function `set_root_directory` to `set_root_cache_directory`
* Improve the docstring of both functions

Regular unit tests.

#### Any other comments?
No.

### DIFF
--- a/openml/config.py
+++ b/openml/config.py
@@ -343,6 +343,17 @@ def get_config_as_dict():
 def get_cache_directory():
     """Get the current cache directory.
 
+    This gets the cache directory for the current server relative
+    to the base cache directory that can be set via
+    ``set_base_cache_directory()``. The cache directory is the
+    ``base_cache_directory`` with additional information on which
+    subdirectory to use based on the server name. By default it is
+    ``base_cache_directory / org / openml / www`` for the standard
+    OpenML.org server and is defined as
+    ``base_cache_directory / top-level domain / second-level domain /
+    hostname``
+    ```
+
     Returns
     -------
     cachedir : string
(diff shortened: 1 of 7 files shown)


openml/config.py
```diff
@@ -351,27 +362,32 @@ def get_cache_directory():
...
+def set_root_cache_directory(root_cache_directory):
+    """Set module-wide base cache directory.
 
-    Sets the cache directory into which to download datasets, tasks etc.
+    Sets the base cache directory that defines how the actual cache
+    directory for the server being used is derived. This is
+    ``root_cache_directory / top-level domain / second-level domain /
+    hostname``, and by default is set to
+    ``root_cache_directory / org / openml / www`` for the standard
+    OpenML.org server.
```
PGijsbers, line 377:
```suggestion
    Sets the root cache directory wherein the cache directories are created to store content
    from different openml servers. For example, by default, cached data for the standard openml.org
    server is stored in ``root_cache_directory / org / openml / www``.
```
I found the description a bit vague, due to inconsistent use of base vs. root and that it does not really define how the cache directory is derived, rather which directory the caches are placed in. This is just a suggestion, but any update the removes this ambiguity is fine with me.

openml/testing.py
```diff
@@ -93,7 +93,7 @@ def setUp(self, n_levels: int = 1):
         self.production_server = "https://openml.org/api/v1/xml"
         openml.config.server = TestBase.test_server
         openml.config.avoid_duplicate_runs = False
-        openml.config.cache_directory = self.workdir
+        openml.config._cache_directory = self.workdir
```
PGijsbers, line 96: We should probably use `set_root_cache_directory()` here.

PGijsbers (changes requested):
See comments.

openml/config.py
mfeurer: Changed this, thanks for spotting that I forgot to replace `base` here with `root`.

PGijsbers (approved):
LGTM

PR #1255, opened 2023-06-13
Doc/type def

Bilgecelik:
(dataset.py) annotate functions and classes, forbid use of "any". docstring not updated yet.

tests fail with pre-changes state too

#### Any other comments?
switched back to "cast" solution to handle optional variable errors in mypy, since tests failed for the function we excluded from the class. (also more cases occured other than that function)


.pre-commit-config.yaml
```diff
@@ -13,12 +13,19 @@ repos:
         additional_dependencies:
           - types-requests
           - types-python-dateutil
+        args: [ --disallow-untyped-defs, --disallow-any-generics,
```
mfeurer, line 16: I'm afraid that this is too strict, as it will test everything. Could you please create a new entry in which you can add the newly typed files?

openml/datasets/dataset.py
```diff
@@ -207,7 +217,7 @@ def find_invalid_characters(string, pattern):
         self.paper_url = paper_url
         self.update_comment = update_comment
         self.md5_checksum = md5_checksum
-        self.data_file = data_file
+        self.data_file = cast(str, data_file)
```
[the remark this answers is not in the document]
Bilgecelik: you are right, moved it to after data_file check and it solved cast requirement

openml/datasets/dataset.py
```diff
@@ -921,6 +891,63 @@ def _to_dict(self) -> "OrderedDict[str, OrderedDict]":
 
         return data_container
 
+    def _get_arff(self, format: str) -> Dict:  # type: ignore
```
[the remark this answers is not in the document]
Bilgecelik: Every type I tried threw an error=) I silenced it and moved on, but I will try one more time.

.pre-commit-config.yaml
Bilgecelik: I separated it and restricted to current changed file only, let me know if it is not what you wanted.

openml/datasets/dataset.py
```diff
@@ -891,7 +894,7 @@ def _to_dict(self) -> "OrderedDict[str, OrderedDict[str, str]]":
 
         return data_container
 
-    def _get_arff(self, format: str) -> Dict:  # type: ignore
+    def _get_arff(self, format: str) -> Dict[str, Union[arff.DENSE, arff.COO]]:
```
mfeurer, line 897: I think the return type of this needs to be the same as the inner `decode_arff`.

mfeurer (changes requested):

openml/datasets/dataset.py
Bilgecelik: Okay the issue here is since our Dict returned from __get_arff_ is a big Union, in _parse_data_from_arff_ function, all dict loads are complaining with all the type options dict indices / values can get. I added a typing ignore to that function for now, will go back to it later.

openml/datasets/dataset.py
```diff
@@ -311,6 +312,7 @@ def _download_data(self) -> None:
         if self._minio_url is not None:
             self.parquet_file = _get_dataset_parquet(self)
 
+    @typing.no_type_check
```
mfeurer, line 315: Why do we need this?

openml/datasets/functions.py
```diff
@@ -635,9 +643,11 @@ def create_dataset(
                     attributes_[attr_idx] = (attr_name, attributes[attr_name])
     else:
         attributes_ = attributes
+
+    # attributes: Union[List[Tuple[str, str]], List[Tuple[str, List[str]]], str]
+    # attributes_: List[Union[Tuple[str, str], Tuple[str, List[str]]]]
```
mfeurer, line 648: What are these comments for?

openml/datasets/data_feature.py
```diff
@@ -59,8 +59,8 @@ def __init__(
         self.nominal_values = nominal_values
         self.number_missing_values = number_missing_values
 
-    def __repr__(self):
+    def __repr__(self) -> str:
         return "[%d - %s (%s)]" % (self.index, self.name, self.data_type)
 
-    def _repr_pretty_(self, pp, cycle):
+    def _repr_pretty_(self, pp, cycle) -> None:  # type: ignore
```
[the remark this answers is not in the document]
Bilgecelik: Ah I see its use case now from pretty library but still not sure what variable type to define.

openml/datasets/dataset.py
```diff
@@ -117,8 +118,8 @@ def __init__(
         description: str,
         data_format: str = "arff",
         cache_format: str = "pickle",
-        dataset_id: Optional[int] = None,
-        version: Optional[int] = None,
+        dataset_id: Optional[str] = None,
```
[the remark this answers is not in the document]
Bilgecelik: Yes so dataset_id was sometimes used as a str and sometimes int, but in this function there was a line to convert it to int (line:184), so i think docstring was wrong about int. In function __parse_publish_response_ xml parser also returns a str and converted to int (line: 858). This combination of str and int's was the only one mypy didn't complain about mismatches at the end.

openml/datasets/data_feature.py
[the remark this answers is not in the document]
mfeurer: This function is for displaying things nicely for Jupyter notebooks and documented [here](https://ipython.readthedocs.io/en/stable/config/integrating.html#rich-display).

openml/datasets/dataset.py
mfeurer: Hm, but we would basically know what type of data we have when applying these functions. Therefore, I would assume that it's possible to be more strict here. But let's leave this for now and focus on getting the rest done :)

openml/base.py
```diff
@@ -46,7 +46,7 @@ def _entity_letter(cls) -> str:
         return cls.__name__.lower()[len("OpenML") :][0]
 
     @abstractmethod
-    def _get_repr_body_fields(self) -> List[Tuple[str, Union[str, int, List[str]]]]:
+    def _get_repr_body_fields(self) -> List[Tuple[str, Union[str, int, List[str], None]]]:
```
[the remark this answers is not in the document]
Bilgecelik: No class but I think the issue is with list comprehension. I get the following error without None in the options:
`List comprehension has incompatible type List[Tuple[str, Union[str, int, None]]]; expected List[Tuple[str, Union[str, int, List[str]]]]  [misc]`
I guess if none of the keys exist it returns a None?

openml/datasets/data_feature.py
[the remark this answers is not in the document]
Bilgecelik: pp is a prettyprinter class instance so this will require prettyprinter library to be installed, but not sure if it is in the requirements.

openml/datasets/functions.py
```diff
@@ -72,8 +73,8 @@ def list_datasets(
     status: Optional[str] = None,
     tag: Optional[str] = None,
     output_format: str = "dict",
-    **kwargs,
-) -> Union[Dict, pd.DataFrame]:
+    **kwargs: Optional[Union[str, int]],
```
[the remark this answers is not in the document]
Bilgecelik: It is a dict but mypy has issues with kwargs (see: https://github.com/python/typing/discussions/1406). I changed it to a union of dict and str, int for now so it doesn't complain and dict still works. But please feel free to adjust if you have a cleaner idea.

openml/datasets/functions.py
```diff
@@ -462,14 +469,16 @@ def get_dataset(
     finally:
         if remove_dataset_cache:
             _remove_cache_dir_for_id(DATASETS_CACHE_DIR_NAME, did_cache_dir)
-
-    dataset = _create_dataset_from_description(
-        description, features_file, qualities_file, arff_file, parquet_file, cache_format
-    )
+    if qualities_file:
```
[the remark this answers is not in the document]
Bilgecelik: mypy gives an arg-type error otherwise, since qualities_file can be None (in exception but mypy doesn't see that). So in run time behavior it is never None so line 472 is always true, since otherwise exception is thrown.

openml/datasets/functions.py
```diff
@@ -525,24 +534,24 @@ def attributes_arff_from_df(df):
...
-    default_target_attribute,
-    ignore_attribute,
-    citation,
-    row_id_attribute=None,
-    original_data_url=None,
-    paper_url=None,
-    update_comment=None,
-    version_label=None,
-):
+    name: str,
```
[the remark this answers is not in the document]
Bilgecelik: I changed the optional ones in OpenMLDataset to optional here too.

openml/datasets/functions.py
Bilgecelik: deleted

openml/datasets/dataset.py
```diff
@@ -561,12 +520,12 @@ def _load_data(self):
...
         logger.info(f"{self.cache_format} load data {self.name}")
         try:
-            if self.cache_format == "feather":
+            if self.cache_format == "feather" and self.feather_attribute_file:
                 data = pd.read_feather(self.data_feather_file)
                 fpath = self.feather_attribute_file
                 with open(self.feather_attribute_file, "rb") as fh:
                     categorical, attribute_names = pickle.load(fh)
-            else:
+            elif self.data_pickle_file:
```
mfeurer, line 528: Could you please add an `else` again to make sure we're not running into some edge case?

openml/datasets/dataset.py
```diff
@@ -697,7 +664,7 @@ def get_data(
         include_ignore_attribute: bool = False,
         dataset_format: str = "dataframe",
     ) -> Tuple[
-        Union[np.ndarray, pd.DataFrame, scipy.sparse.csr_matrix],
+        Union[np.ndarray, pd.DataFrame, pd.SparseDtype],
```
mfeurer, line 667: This looks wrong. Do we return spares Dtype instead of sparse scipy matrices now?

openml/datasets/dataset.py
```diff
@@ -1061,14 +1114,3 @@ def _check_qualities(qualities: List[Dict[str, str]]) -> Dict[str, float]:
             value = float(xmlquality["oml:value"])
         qualities_[name] = value
     return qualities_
-
-
-def _parse_qualities_xml(qualities_xml):
```
mfeurer, line 1066: This looks like it is reverting some changes from @PGijsbers, is this on purpose or rather a merge issue?

openml/datasets/dataset.py
```diff
@@ -992,6 +969,80 @@ def _to_dict(self) -> "OrderedDict[str, OrderedDict]":
 
         return data_container
 
+    def _get_arff(
```
mfeurer, line 972: Is there a reason you moved this function? This makes it hard to review, and I would appreciate if you could move it back to its original place.

openml/datasets/dataset.py
PGijsbers: I don't see any reason to revert the changes at least.

.pre-commit-config.yaml
```diff
@@ -19,14 +19,21 @@ repos:
...
+          - types-python-dateutil
       - id: mypy
         name: mypy top-level-functions
         files: openml/_api_calls.py
         additional_dependencies:
           - types-requests
           - types-python-dateutil
         args: [ --disallow-untyped-defs, --disallow-any-generics,
-                --disallow-any-explicit, --implicit-optional ]
+                --disallow-any-explicit, --implicit-optional, --allow-redefinition]
```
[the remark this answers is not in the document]
Bilgecelik: allow-redefinition allows redefining a variable with a different type within same block and nesting depth, there were a few cases that happened and i didn't want to change the code too much. Agreed with the second part, adding them to datasets part as well.

LennartPurucker:
Heyho, after looking at the changes, conflicts, and open questions, I propose closing this PR and manually adding its changes as part of the refactor in #1298. With the changes in #1298 and the linter of ruff, it will be easier to spot all the missing parts and have the same typing everywhere. 

I will update this PR once we decide on our next steps related to PR #1298.

PGijsbers:
@LennartPurucker could you please make Bilge co-author in one of the commits that migrate her changes over? That way she gets listed as a contributor when it eventually gets squash merged into develop.

LennartPurucker:
Will do!

Issue #1132, opened 2022-03-24
Allow forcing a download even if cached files are already present


PGijsbers:
Sometimes cache has to be updated, and manually removing the cache directory to force a download is required. It would be nice to simply force a cache refresh from code.

PGijsbers:
In the meanwhile, you can use this function (you can use the same arguments as `get_dataset`).
```python
import openml

def force_get_dataset(dataset_id=None, *args, **kwargs):
    """ Remove any existing local files about `dataset_id` and then download new copies. """
    did_cache_dir = openml.utils._create_cache_directory_for_id(openml.datasets.functions.DATASETS_CACHE_DIR_NAME, dataset_id, )
    openml.utils._remove_cache_dir_for_id(openml.datasets.functions.DATASETS_CACHE_DIR_NAME, did_cache_dir)
    return openml.datasets.get_dataset(dataset_id, *args, **kwargs)

if __name__ == "__main__":
    force_get_dataset(61)
```

LennartPurucker:
This was resolved by PR #1260.

Issue #1034, opened 2021-03-09
Don't download (large) datasets by default


joaquinvanschoren:
#### Description
In datasets.get_dataset(data_id) the default is currently to always download the dataset:
https://openml.github.io/openml-python/master/generated/openml.datasets.get_dataset.html#openml.datasets.get_dataset

This is problematic for large datasets - it takes a long time and may cause out-of-memory errors. Sometimes we need to look at the full meta-data (of many datasets) without downloading the data. We can do that now with the option download_data=False, but it feels like this should be the default. Some users may also be unaware of this option or the fact that get_dataset will actually download the data and consume resources.

A simple solution would be to make download_data=False the default.

#### Steps/Code to Reproduce
```python
import openml
openml.datasets.get_dataset(41081)
```

#### Expected Results
The dataset metadata within seconds

#### Actual Results
A long time waiting until the dataset has downloaded and parsed.

#### Versions
macOS-10.16-x86_64-i386-64bit
Python 3.8.5 (default, Sep  4 2020, 02:22:02) 
[Clang 10.0.0 ]
NumPy 1.19.5
SciPy 1.5.2
Scikit-Learn 0.23.2
OpenML 0.11.1dev

LennartPurucker:
We set this to be the default from 0.15.0 onwards in PR #1260.

PGijsbers:
Started working on this with the `add/1034` branch

PR #1256, opened 2023-06-13
Download updates

### LINKED ISSUE #1081: Make downloading dataset features optional
Currently the data features file is always downloaded [here](https://github.com/openml/openml-python/blob/3aee2e05186f2151e45c9ddc5bdd0709459bfce3/openml/datasets/functions.py#L412). I propose we make it optional, similar to `qualities` and the dataset itself. If not downloaded, it should follow the same lazy loading behavior that `qualities` does.

Having `features` immediately available is often not necessary. E.g. automated runs don't use them and when users download the data much of `features` info is already available in the built dataframe. But it does bring the usual downsides of longer user wait times, especially when downloading `features`  with high cardinality categorical features, e.g. [albert](https://www.openml.org/api/v1/xml/data/features/41147), and the additional server strain.

LennartPurucker:
Related to #1034 #1081 #1132 
Closes #1081 #1132 

In detail, in relation to #1034, I have added a deprecation warning for `get_dataset`, making clear that we will default to not downloading the data in the future (starting from version 1.4, but this is just my guess for now). 

In relation to #1081, I added an option to disable downloading features. Moreover, I adjusted the `repr` method and added semi-lazy loading for features by the use of the function `OpenMLDataset.load_metadata`. As part of this, I also added the option to load qualities after calling `get_dataset`, which seems to have been missing so far. 

In relation to #1132, I added the option `force_refresh_cache` to force refresh the cache by deleting the existing cache initially. 

Note, I have also changed the docstring of `get_dataset` and adjusted the parts about being threading/multiprocessing safe, which were incorrect from my previous experiences and observations w.r.t. the cache. This might open a discussion to make this threading/multiprocessing safe based on an option that only downloads the data and ignores the cache entirely. I advise against this, for now, see [here ](https://github.com/openml/openml-python/pull/1256/files#diff-9cc0e191426810b9a76098707c2ddc9ec73a7f1852024a72445ddcb08b52f09bR361-R367) for more details on this.

### DIFF
--- a/openml/datasets/data_feature.py
+++ b/openml/datasets/data_feature.py
@@ -62,5 +62,17 @@ class OpenMLDataFeature(object):
     def __repr__(self):
         return "[%d - %s (%s)]" % (self.index, self.name, self.data_type)
 
+    def __eq__(self, other):
+        if not isinstance(other, OpenMLDataFeature):
+            return False
+        else:
+            return (
+                self.index == other.index
+                and self.name == other.name
+                and self.data_type == other.data_type
+                and self.nominal_values == other.nominal_values
+                and self.number_missing_values == other.number_missing_values
+            )
+
     def _repr_pretty_(self, pp, cycle):
         pp.text(str(self))
--- a/openml/datasets/data_feature.py
+++ b/openml/datasets/data_feature.py
@@ -65,14 +65,14 @@ class OpenMLDataFeature(object):
     def __eq__(self, other):
         if not isinstance(other, OpenMLDataFeature):
             return False
-        else:
-            return (
-                self.index == other.index
-                and self.name == other.name
-                and self.data_type == other.data_type
-                and self.nominal_values == other.nominal_values
-                and self.number_missing_values == other.number_missing_values
-            )
+
+        return (
+            self.index == other.index
+            and self.name == other.name
+            and self.data_type == other.data_type
+            and self.nominal_values == other.nominal_values
+            and self.number_missing_values == other.number_missing_values
+        )
 
     def _repr_pretty_(self, pp, cycle):
         pp.text(str(self))
--- a/openml/datasets/data_feature.py
+++ b/openml/datasets/data_feature.py
@@ -66,13 +66,7 @@ class OpenMLDataFeature(object):
         if not isinstance(other, OpenMLDataFeature):
             return False
 
-        return (
-            self.index == other.index
-            and self.name == other.name
-            and self.data_type == other.data_type
-            and self.nominal_values == other.nominal_values
-            and self.number_missing_values == other.number_missing_values
-        )
+        return self.__dict__ == other.__dict__
 
     def _repr_pretty_(self, pp, cycle):
         pp.text(str(self))
--- a/openml/datasets/dataset.py
+++ b/openml/datasets/dataset.py
@@ -16,6 +16,7 @@ import scipy.sparse
 import xmltodict
 
 from openml.base import OpenMLBase
+from openml._api_calls import _perform_api_call
 from .data_feature import OpenMLDataFeature
 from ..exceptions import PyOpenMLError
 
@@ -787,6 +788,31 @@ class OpenMLDataset(OpenMLBase):
 
         return data, targets, categorical, attribute_names
 
+    def load_metadata(self, features: bool = False, qualities: bool = False):
+        """Load the missing medata information from the server and store it in the server.
+
+        The purpose of the function is to support lazy loading.
+
+        Parameters
+        ----------
+        features : bool (default=False)
+            If True, load the `self.features` data if not already loaded.
+        qualities: bool (default=False)
+            If True, load the `self.qualities` data if not already loaded.
+        """
+
+        if self.dataset_id is None:
+            raise ValueError(
+                """No dataset id specified. Please set the dataset id.
+                                Otherwise we cannot load metadata."""
+            )
+
+        if features and self.features is None:
+            self.features = _parse_features_xml(_get_features_xml(self.dataset_id))
+
+        if qualities and self.qualities is None:
+            self.qualities = _parse_qualities_xml(_get_qualities_xml(self.dataset_id))
+
     def retrieve_class_labels(self, target_name: str = "class") -> Union[None, List[str]]:
         """Reads the datasets arff to determine the class-labels.
 
@@ -936,6 +962,7 @@ class OpenMLDataset(OpenMLBase):
         return data_container
 
 
+# -- Code for Features Property
 def _read_features(features_file: str) -> Dict[int, OpenMLDataFeature]:
     features_pickle_file = _get_features_pickle_file(features_file)
     try:
@@ -944,27 +971,37 @@ def _read_features(features_file: str) -> Dict[int, OpenMLDataFeature]:
     except:  # noqa E722
         with open(features_file, encoding="utf8") as fh:
             features_xml_string = fh.read()
-        xml_dict = xmltodict.parse(
-            features_xml_string, force_list=("oml:feature", "oml:nominal_value")
-        )
-        features_xml = xml_dict["oml:data_features"]
 
-        features = {}
-        for idx, xmlfeature in enumerate(features_xml["oml:feature"]):
-            nr_missing = xmlfeature.get("oml:number_of_missing_values", 0)
-            feature = OpenMLDataFeature(
-                int(xmlfeature["oml:index"]),
-                xmlfeature["oml:name"],
-                xmlfeature["oml:data_type"],
-                xmlfeature.get("oml:nominal_value"),
-                int(nr_missing),
-            )
-            if idx != feature.index:
-                raise ValueError("Data features not provided in right order")
-            features[feature.index] = feature
+        features = _parse_features_xml(features_xml_string)
 
         with open(features_pickle_file, "wb") as fh_binary:
             pickle.dump(features, fh_binary)
+    return features
+
+
+def _get_features_xml(dataset_id):
+    url_extension = "data/features/{}".format(dataset_id)
+    return _perform_api_call(url_extension, "get")
+
+
+def _parse_features_xml(features_xml_string):
+    xml_dict = xmltodict.parse(features_xml_string, force_list=("oml:feature", "oml:nominal_value"))
+    features_xml = xml_dict["oml:data_features"]
+
+    features = {}
+    for idx, xmlfeature in enumerate(features_xml["oml:feature"]):
+        nr_missing = xmlfeature.get("oml:number_of_missing_values", 0)
+        feature = OpenMLDataFeature(
+            int(xmlfeature["oml:index"]),
+            xmlfeature["oml:name"],
+            xmlfeature["oml:data_type"],
+            xmlfeature.get("oml:nominal_value"),
+            int(nr_missing),
+        )
+        if idx != feature.index:
+            raise ValueError("Data features not provided in right order")
+        features[feature.index] = feature
+
     return features
 
 
@@ -973,6 +1010,7 @@ def _get_features_pickle_file(features_file: str) -> str:
     return features_file + ".pkl"
 
 
+# -- Code for Qualities Property
 def _read_qualities(qualities_file: str) -> Dict[str, float]:
     qualities_pickle_file = _get_qualities_pickle_file(qualities_file)
     try:
@@ -981,17 +1019,15 @@ def _read_qualities(qualities_file: str) -> Dict[str, float]:
     except:  # noqa E722
         with open(qualities_file, encoding="utf8") as fh:
             qualities_xml = fh.read()
-        xml_as_dict = xmltodict.parse(qualities_xml, force_list=("oml:quality",))
-        qualities = xml_as_dict["oml:data_qualities"]["oml:quality"]
-        qualities = _check_qualities(qualities)
+        qualities = _parse_qualities_xml(qualities_xml)
         with open(qualities_pickle_file, "wb") as fh_binary:
             pickle.dump(qualities, fh_binary)
     return qualities
 
 
-def _get_qualities_pickle_file(qualities_file: str) -> str:
-    """This function only exists so it can be mocked during unit testing"""
-    return qualities_file + ".pkl"
+def _get_qualities_xml(dataset_id):
+    url_extension = "data/qualities/{}".format(dataset_id)
+    return _perform_api_call(url_extension, "get")
 
 
 def _check_qualities(qualities: List[Dict[str, str]]) -> Dict[str, float]:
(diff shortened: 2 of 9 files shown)


LennartPurucker:
@mfeurer I have not found a good way to put these functionalities into a general function among all OpenML entity types. Hence, I would like your opinion on the current proposed way of doing it. Does it suffice? 
Feel free to remark on additional reference points (entities) that would need such download-related treatment so that I can better understand how to make these changes more abstract.

openml/datasets/data_feature.py
```diff
@@ -62,5 +62,17 @@ def __init__(
     def __repr__(self):
         return "[%d - %s (%s)]" % (self.index, self.name, self.data_type)
 
+    def __eq__(self, other):
+        if not isinstance(other, OpenMLDataFeature):
+            return False
+        else:
```
PGijsbers, line 68: You can drop the `else` and dedent the remainder.

openml/datasets/dataset.py
```diff
@@ -773,6 +788,31 @@ def get_data(
 
         return data, targets, categorical, attribute_names
 
+    def load_metadata(self, features: bool = False, qualities: bool = False):
+        """Load the missing medata information from the server and store it in the server.
```
PGijsbers, line 792:
```suggestion
        """Load the missing metadata information from the server and store it in the server.
```

openml/datasets/dataset.py
```diff
@@ -773,6 +788,31 @@ def get_data(
...
+        """
+
+        if self.dataset_id is None:
+            raise ValueError(
+                """No dataset id specified. Please set the dataset id.
+                                Otherwise we cannot load metadata."""
+            )
+
+        if features and self.features is None:
+            self.features = _parse_features_xml(_get_features_xml(self.dataset_id))
```
PGijsbers, line 811: Prefer one function call per line, it is much more readable if an (unexpected) error is raised.

PGijsbers (commented):
didn't have a close look, had a very small amount of time, still needs a review (not necessarily by me).

openml/datasets/data_feature.py
```diff
@@ -62,5 +62,17 @@ def __init__(
     def __repr__(self):
         return "[%d - %s (%s)]" % (self.index, self.name, self.data_type)
 
+    def __eq__(self, other):
+        if not isinstance(other, OpenMLDataFeature):
+            return False
+
+        return (
```
mfeurer, line 69: Could we simplify this code by comparing `self.__dict__` with `other.__dict__`?

openml/datasets/functions.py
```diff
@@ -383,21 +386,42 @@ def get_dataset(
...
-    cache_format : str (default='pickle')
+    cache_format : str (default='pickle') in {'pickle', 'feather'}
         Format for caching the dataset - may be feather or pickle
         Note that the default 'pickle' option may load slower than feather when
         no.of.rows is very high.
     download_qualities : bool (default=True)
         Option to download 'qualities' meta-data in addition to the minimal dataset description.
+        If True, download and cache the qualities file.
+        If False, create the OpenMLDataset without qualities metadata. The data may later be added
+        to the OpenMLDataset hrough the `OpenMLDataset.load_metadata(qualities=True)` method.
```
mfeurer, line 397:
```suggestion
        to the OpenMLDataset through the `OpenMLDataset.load_metadata(qualities=True)` method.
```

openml/datasets/functions.py
```diff
@@ -355,13 +352,19 @@ def get_dataset(
...
     """Download the OpenML dataset representation, optionally also download actual data file.
 
-    This function is thread/multiprocessing safe.
-    This function uses caching. A check will be performed to determine if the information has
-    previously been downloaded, and if so be loaded from disk instead of retrieved from the server.
+    This function is by default NOT thread/multiprocessing safe, as this function uses caching.
+    A check will be performed to determine if the information has previously been downloaded to a
+    cache, and if so be loaded from disk instead of retrieved from the server.
+
+    To make this function thread/multiprocessing safe initialize the cache first by calling
```
mfeurer, line 365: You could also extend the documentation that this becomes thread-safe when oslo.concurrency is installed (see the decorator).

openml/datasets/functions.py
```diff
@@ -383,21 +386,42 @@ def get_dataset(
...
+        because this creates a race condition to creating and deleting the cache; as in general with
+        the cache.
 
     Returns
     -------
     dataset : :class:`openml.OpenMLDataset`
         The downloaded dataset.
     """
+    if any([download_qualities, download_features_meta_data]):
+        warnings.warn(
```
mfeurer, line 418: I would suggest approaching this a bit differently:
1. We set the flag to `None` above and emit a `DeprecationWarning` then.
2. We ask the user to pass a `bool` to disable this warning.
3. We internally assume `True` if `None` is given.
4. And also do this for `download_dataset`.

openml/datasets/functions.py
```diff
@@ -419,6 +443,15 @@ def get_dataset(
             "`dataset_id` must be one of `str` or `int`, not {}.".format(type(dataset_id))
         )
 
+    # Note: we could also (quite heavily) re-implement the below to only download the
+    # data and do not cache the data at all. This would always be thread/multiprocessing
+    # safe. However, this would likely drastically increase the strain on the server.
```
mfeurer, line 448: And make any work with OpenML really slow. Please see my comment above on how to make this class thread-safe.

tests/test_datasets/test_dataset.py
```diff
@@ -262,6 +262,19 @@ def test_get_data_corrupt_pickle(self):
         self.assertIsInstance(xy, pd.DataFrame)
         self.assertEqual(xy.shape, (150, 5))
 
+    def test_load_metadata(self):
+        _compare_dataset = openml.datasets.get_dataset(
+            2, download_data=False, download_features_meta_data=True, download_qualities=True
+        )
+
+        _dataset = openml.datasets.get_dataset(
+            2, download_data=False, download_features_meta_data=False, download_qualities=False
```
mfeurer, line 271: Hm, interesting. So the data is already available on the local drive, so you would not download it again, so the argument name is wrong. This should actually be called `load_features_meta_data` then, I guess?

mfeurer (commented):
Hey, I left some comments. All in all this looks great and I'm optimistic that we can merge this soonish.

Just a side note: we'd also need to update the task loading to disable dataset loading by default.

openml/datasets/functions.py
PGijsbers: +1, we definitely want to cache data. In the vast amount of use cases it is more important than making this function thread safe.

tests/test_datasets/test_dataset.py
LennartPurucker: Not sure what you mean exactly. Calling `get_dataset(download_features_meta_data=True)` will download features meta_data or load it from the cache. This works similarly for `download_data` and `download_qualities`. Technically, all of these are a bit misleading, but considering that it is a cache I would say it is okay. 

If you referring to the function call below, `OpenMLDataset.load_metadata(features=True)`, the code will always download the data. I think I should also add a cache for that as `load_data` also caches the data or loads from the cache if available. I will change this accordingly.

openml/datasets/functions.py
LennartPurucker: Your idea sounds much better. I will change this accordingly.

One question, I noticed that Python, by default, suppresses deprecation warnings. Should we maybe switch to user warnings? Otherwise, users only see the warning if they enable deprecation warnings. 

We could also change the warning level in our code, but this would change settings in the outer scope unless we clean up properly.

tests/test_datasets/test_dataset.py
LennartPurucker: I have added cache usage to `OpenMLDataset.load_metadata`. It is now in line with `get_dataset()` and `OpenMLDataset.get_data`'s `_load_data`.

openml/datasets/functions.py
LennartPurucker: I added your suggestion and switched to user warnings for now.

LennartPurucker:
I have also adjusted task loading. This got a bit hacky so that it is backward compatible and supports all arguments by passing kwargs. I am not sure that this is a good solution. Hence,  I would be happy for feedback on how to handle the arguments best. 

Moreover, I added actually lazy loading based on properties such that the user does not have to call `load_metadata` by hand. This is now more in line with how `get_data` functions as far as I understand it.

openml/datasets/functions.py
mfeurer: scikit-learn appears to use `FutureWarning`, so we might use that instead?

openml/runs/functions.py
```diff
@@ -102,8 +102,8 @@ def run_model_on_task(
         warnings.warn(
             "avoid_duplicate_runs is set to True, but no API key is set. "
             "Please set your API key in the OpenML configuration file, see"
-            "https://openml.github.io/openml-python/main/examples/20_basic/introduction_tutorial.html#authentication"
-            "for more information on authentication.",
+            "https://openml.github.io/openml-python/main/examples/20_basic/introduction_tutorial"
+            + ".html#authenticatio for more information on authentication.",
```
mfeurer, line 106: Did you drop an `n` here?

openml/tasks/functions.py
```diff
@@ -327,27 +326,39 @@ def get_tasks(
...
-    task_id: int, download_data: bool = True, download_qualities: bool = True
+    task_id: int, *dataset_args, download_splits: Optional[bool] = None, **get_dataset_kwargs
 ) -> OpenMLTask:
     """Download OpenML task for a given task ID.
 
-    Downloads the task representation, while the data splits can be
-    downloaded optionally based on the additional parameter. Else,
-    splits will either way be downloaded when the task is being used.
+    Downloads the task representation, while the data splits can be downloaded optionally based on
+    the additional parameter that are passed to :meth:`openml.datasets.get_dataset`.
```
mfeurer, line 334: I don't quite understand this. If the `download_splits` is given, this has nothing to do with the additional parameter, or does it?

openml/datasets/dataset.py
```diff
@@ -831,10 +853,6 @@ def retrieve_class_labels(self, target_name: str = "class") -> Union[None, List[
         -------
         list
         """
-        if self.features is None:
```
mfeurer, line 793: Why can this be dropped?

tests/test_datasets/test_dataset.py
mfeurer: Nevermind, I'm not sure whether my comment makes too much sense.

openml/tasks/functions.py
LennartPurucker: I will update the doc string to make it more clear.

openml/datasets/dataset.py
LennartPurucker: To clarify, I dropped this because with the new lazy loading,` self.feature` cannot be `None`.
AFAIK, it would make no sense to call this function with the intention of `self.feature is None`, because the function requires values for `self.feature`. Hence, I think defaulting to lazy loading is appropriate here.

tests/test_datasets/test_dataset.py
LennartPurucker: It was a good comment. I have changed a few things to adjust to your comment and I the code is now better/clearer.

openml/tasks/functions.py
```diff
@@ -327,31 +326,54 @@ def get_tasks(
...
+        warnings.warn(
+            "Starting from Version 0.15.0 `download_splits` will default to ``False`` instead "
+            "of ``True`` and be independent from `download_data`. To disable this message until "
+            "version 0.15 explicitly set `download_splits` to a bool.",
+            FutureWarning,
+        )
+        download_splits = get_dataset_kwargs.get("download_data", True)
+
     if not isinstance(task_id, int):
+        # TODO(0.14): Remove warning
```
mfeurer, line 373:
```suggestion
        # TODO(0.15): Remove warning
```

mfeurer (changes requested):
1 more change, then we can merge this one!

openml/tasks/functions.py
LennartPurucker: Done!

mfeurer (approved):

openml/datasets/dataset.py
```diff
@@ -234,12 +239,36 @@ def find_invalid_characters(string, pattern):
             self.data_feather_file = None
             self.feather_attribute_file = None
 
+    @property
+    def features(self):
+        # Lazy loading of features
```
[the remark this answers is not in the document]
PGijsbers: I think this is a common enough pattern to not warrant a comment.

openml/datasets/dataset.py
```diff
@@ -922,6 +981,7 @@ def _to_dict(self) -> "OrderedDict[str, OrderedDict]":
         return data_container
 
 
+# -- Code for Features Property
```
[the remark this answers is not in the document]
PGijsbers: I'd rather open a new issue on Github as a note that we need to refactor our files later. Having code headers shouldn't really be necessary.

openml/tasks/functions.py
```diff
@@ -327,31 +326,54 @@ def get_tasks(
...
+    task_id: int, *dataset_args, download_splits: Optional[bool] = None, **get_dataset_kwargs
 ) -> OpenMLTask:
     """Download OpenML task for a given task ID.
 
-    Downloads the task representation, while the data splits can be
-    downloaded optionally based on the additional parameter. Else,
-    splits will either way be downloaded when the task is being used.
+    Downloads the task representation. By default, this will also download the data splits and
+    the dataset. From version 0.15.0 onwards, the splits will not be downloaded by default
+    nor the dataset.
```
PGijsbers, line 335:
```suggestion
    the dataset. From version 0.15.0 onwards, the splits nor the dataset will not be downloaded by default.
```

tests/test_datasets/test_dataset.py
```diff
@@ -262,6 +262,36 @@ def test_get_data_corrupt_pickle(self):
...
+        self.assertEqual(_dataset.qualities, _compare_dataset.qualities)
+
+        # -- Test without cache
+        openml.utils._remove_cache_dir_for_id(
+            openml.datasets.functions.DATASETS_CACHE_DIR_NAME, did_cache_dir
+        )
+
+        _dataset = openml.datasets.get_dataset(
+            2, download_data=False, download_features_meta_data=False, download_qualities=False
+        )
```
PGijsbers, line 290: Should check cache is empty after this call? A function which always loads the data should now pass this test.

tests/test_datasets/test_dataset_functions.py
```diff
@@ -546,8 +546,47 @@ def test__get_dataset_qualities(self):
...
+        openml.datasets.get_dataset(2, force_refresh_cache=True)
+        self.assertNotEqual(change_time, os.stat(did_cache_dir).st_mtime)
+
+        # Clean up
+        openml.utils._remove_cache_dir_for_id(
+            DATASETS_CACHE_DIR_NAME,
+            did_cache_dir,
+        )
+
+        # Test clean start
```
PGijsbers, line 581: This section should be a separate test.

PGijsbers (changes requested):
I didn't look too much at the correctness of the code (since Matthias already had a look), but I think there are several things that could be improved from a code quality point of view. Please have a look at the comments.

openml/datasets/dataset.py
```diff
@@ -773,6 +802,40 @@ def get_data(
...
+        features_file, qualities_file = _get_dataset_metadata(
+            self.dataset_id, features=features, qualities=qualities
+        )
+
+        if features_file is not None:
+            self._features = _read_features(features_file)
+
+        if qualities_file is not None:
+            self._qualities = _read_qualities(qualities_file)
+            self._no_qualities_found = self._qualities is None
```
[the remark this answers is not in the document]
LennartPurucker: This is required as far as I can tell, because `self._no_qualities_found` is initially False and only becomes True in this edge case after lazy loading the qualities for the first time.

PR #1257, opened 2023-06-15
Add mypy annotations for _api_calls.py

mfeurer:
This commit adds mypy annotations for _api_calls.py to make sure that we have no untyped classes and functions.

### DIFF
--- a/openml/_api_calls.py
+++ b/openml/_api_calls.py
@@ -11,7 +11,7 @@ import urllib.parse
 import xml
 import xmltodict
 from urllib3 import ProxyManager
-from typing import Dict, Optional, Union
+from typing import Dict, Optional, Tuple, Union
 import zipfile
 
 import minio
@@ -23,6 +23,9 @@ from .exceptions import (
     OpenMLServerNoResult,
     OpenMLHashException,
 )
+
+DATA_TYPE = Dict[str, Union[str, int]]
+FILE_ELEMENTS_TYPE = Dict[str, Union[str, Tuple[str, str]]]
 
 
 def resolve_env_proxies(url: str) -> Optional[str]:
@@ -54,7 +57,12 @@ def _create_url_from_endpoint(endpoint: str) -> str:
     return url.replace("=", "%3d")
 
 
-def _perform_api_call(call, request_method, data=None, file_elements=None):
+def _perform_api_call(
+    call: str,
+    request_method: str,
+    data: Optional[DATA_TYPE] = None,
+    file_elements: Optional[FILE_ELEMENTS_TYPE] = None,
+) -> str:
     """
     Perform an API call at the OpenML server.
 
@@ -76,8 +84,6 @@ def _perform_api_call(call, request_method, data=None, file_elements=None):
 
     Returns
     -------
-    return_code : int
-        HTTP return code
     return_value : str
         Return value of the OpenML server
     """
@@ -257,7 +263,7 @@ def _download_text_file(
         return None
 
 
-def _file_id_to_url(file_id, filename=None):
+def _file_id_to_url(file_id: str, filename: Optional[str] = None) -> str:
     """
     Presents the URL how to download a given file id
     filename is optional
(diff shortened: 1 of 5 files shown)


openml/_api_calls.py
```diff
@@ -306,10 +319,16 @@ def __is_checksum_equal(downloaded_file_binary: bytes, md5_checksum: Optional[st
...
+    request_method: str,
+    url: str,
+    data: DATA_TYPE,
+    files: Optional[FILE_ELEMENTS_TYPE] = None,
+    md5_checksum: Optional[str] = None,
+) -> requests.Response:
     n_retries = max(1, config.connection_n_retries)
 
-    response = None
+    response: requests.Response
```
LennartPurucker, line 331: Is this for the sake of typing? If yes, is this needed if we know the return type of the function already?

mfeurer: Yes. Not doing this will set the type of `response` to `Optional[...]` while it actually cannot be `None`.

PGijsbers: I am fine with this, but optionally I think it might work if you:
 - `return response` instead of `break`, and
 - use the `else` clause of the `for` loop to raise the last exception
 
 but that's a lot of rewriting, and I think the function probably ought to be refactored in the future anyway (it's kind of unwieldy).

openml/datasets/functions.py
```diff
@@ -4,7 +4,7 @@
 import logging
 import os
 from pyexpat import ExpatError
-from typing import List, Dict, Union, Optional, cast
+from typing import List, Dict, Optional, Tuple, Union, cast  # noqa: F401
```
PGijsbers, line 7: Why do we need a noqa here?

openml/study/functions.py
```diff
@@ -277,7 +277,7 @@ def update_study_status(study_id: int, status: str) -> None:
     legal_status = {"active", "deactivated"}
     if status not in legal_status:
         raise ValueError("Illegal status value. " "Legal values: %s" % legal_status)
-    data = {"study_id": study_id, "status": status}
+    data = {"study_id": study_id, "status": status}  # type: Dict[str, Union[str, int]]
```
PGijsbers, line 280: Should this not be `# type: openml._api_calls.DATA_TYPE`?

PGijsbers (changes requested):
OK, though I think there is one annotation that should probably be changed.

openml/datasets/functions.py
mfeurer: Indeed, that's not necessary. Will remove.

openml/datasets/functions.py
```diff
@@ -827,7 +827,7 @@ def edit_dataset(
         raise TypeError("`data_id` must be of type `int`, not {}.".format(type(data_id)))
 
     # compose data edit parameters as xml
-    form_data = {"data_id": data_id}
+    form_data = {"data_id": data_id}  # type: openml._api_calls.DATA_TYPE
```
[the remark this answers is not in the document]
mfeurer: Because from my understanding we would then get `Dict[str, int]` while the function expects `Dict[str, Union[str, int]]` as an argument.

openml/study/functions.py
mfeurer: Good catch.

openml/datasets/functions.py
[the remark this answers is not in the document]
PGijsbers: Since `int` is compatible with `Union[str, int]` I don't think that's an issue.

LennartPurucker (approved):

openml/datasets/functions.py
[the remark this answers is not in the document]
mfeurer: That's what I though as well, but no: `openml/datasets/functions.py:895: error: Argument "data" to "_perform_api_call" has incompatible type "Dict[str, int]"; expected "Optional[Dict[str, Union[str, int]]]"  [arg-type]`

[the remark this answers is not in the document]
PGijsbers: cannot reproduce MWE, it's fine for now and we revisit when the rest of the file gets annotated.

PGijsbers (approved):

PR #1258, opened 2023-06-15
Deprecate `output_format='dict'`

PGijsbers:
Deprecates `output_format='dict'`:
 - add a future warning with any of the `list_*` functions are called with `output_format='dict'`,
 - update all internal code and examples which used to rely on the default of `output_format='dict'`
 
 WIP: Going through unit tests I incidentally broke

### DIFF
--- a/openml/datasets/functions.py
+++ b/openml/datasets/functions.py
@@ -252,7 +252,8 @@ def check_datasets_active(
     Check if the dataset ids provided are active.
 
     Raises an error if a dataset_id in the given list
-    of dataset_ids does not exist on the server.
+    of dataset_ids does not exist on the server and
+    `raise_error_if_not_exist` is set to True (default).
 
     Parameters
     ----------
@@ -267,17 +268,11 @@ def check_datasets_active(
     dict
         A dictionary with items {did: bool}
     """
-    dataset_list = list_datasets(status="all", data_id=dataset_ids, output_format="dataframe")
-    active = {}
-
-    for did in dataset_ids:
-        dataset = dataset_list.get(did, None)
-        if dataset is None:
-            if raise_error_if_not_exist:
-                raise ValueError(f"Could not find dataset {did} in OpenML dataset list.")
-        else:
-            active[did] = dataset["status"] == "active"
-
+    datasets = list_datasets(status="all", data_id=dataset_ids, output_format="dataframe")
+    active = dict(datasets["status"] == "active")
+    missing = set(dataset_ids) - set(datasets["did"])
+    if raise_error_if_not_exist and missing:
+        raise ValueError(f"Could not find dataset(s) {missing} in OpenML dataset list.")
     return active
 
 
--- a/openml/datasets/functions.py
+++ b/openml/datasets/functions.py
@@ -131,6 +131,7 @@ def list_datasets(
             "Invalid output format selected. " "Only 'dict' or 'dataframe' applicable."
         )
 
+    # TODO: [0.15]
     if output_format == "dict":
         msg = (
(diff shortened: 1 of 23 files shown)


openml/datasets/functions.py
```diff
@@ -259,17 +269,11 @@ def check_datasets_active(
...
-    for did in dataset_ids:
-        dataset = dataset_list.get(did, None)
-        if dataset is None:
-            if raise_error_if_not_exist:
-                raise ValueError(f"Could not find dataset {did} in OpenML dataset list.")
-        else:
-            active[did] = dataset["status"] == "active"
-
+    datasets = list_datasets(status="all", data_id=dataset_ids, output_format="dataframe")
+    active = dict(datasets["status"] == "active")
```
mfeurer, line 273: Would it maybe be better / more readable to do `(datasets["status"] == "active")["did"])`

PGijsbers: That isn't the same, what I do here is create a `pd.Series` which indicates for each dataset if it is active or not: `datasets["status"] == "active"`, which gives something like:
```
>>> d["status"]=='active'
2        True
3        True
4        True
5        True
6        True
```
 
I then call `dict(pd.Series)`, which turns the index into keys, and the values into values: `{..., 45573: True, 45574: True}`.


(in fact, the proposed method would lead to an error as the series can not be indexed by `"did"`.

PGijsbers:
I _think_ I cleared up the tests I accidentally broke after refactoring, but there are a lot of failures due to the test server state (also on develop).

openml/datasets/functions.py
mfeurer: Thanks for the explanation. Sorry, I'm not good enough yet reading pandas :(

tests/test_runs/test_run.py
```diff
@@ -26,18 +26,19 @@ class TestRun(TestBase):
...
     def test_tagging(self):
-        runs = openml.runs.list_runs(size=1)
-        run_id = list(runs.keys())[0]
+        runs = openml.runs.list_runs(size=1, output_format="dataframe")
+        assert not runs.empty, "Test server state is incorrect"
+        run_id = runs["run_id"].iloc[0]
         run = openml.runs.get_run(run_id)
         tag = "testing_tag_{}_{}".format(self.id(), time())
-        run_list = openml.runs.list_runs(tag=tag)
+        run_list = openml.runs.list_runs(tag=tag, output_format="dataframe")
```
LennartPurucker, line 34: Is the name `run_list` on purpose here? Previously, you changed every `*_list` together with changing the output format.

tests/test_tasks/test_task_methods.py
```diff
@@ -17,14 +17,14 @@ def tearDown(self):
     def test_tagging(self):
         task = openml.tasks.get_task(1)  # anneal; crossvalidation
         tag = "testing_tag_{}_{}".format(self.id(), time())
-        task_list = openml.tasks.list_tasks(tag=tag)
+        task_list = openml.tasks.list_tasks(tag=tag, output_format="dataframe")
```
LennartPurucker, line 20: Same point as above regarding `task_list`.

tests/test_runs/test_run.py
PGijsbers: Oversight, I suppose. Fixed!

mfeurer (approved):

LennartPurucker (approved):

PGijsbers:
@mfeurer only `test_list_all_with_multiple_batches ` fails now, as expected. I think we can go ahead and merge this.

PR #1261, opened 2023-06-16
Add more type annotations

mfeurer:
This PR add type annotations for the remaining files in `openml` namespace.

### DIFF
--- a/openml/testing.py
+++ b/openml/testing.py
@@ -7,7 +7,7 @@ import pathlib
 import shutil
 import sys
 import time
-from typing import Dict, Union, cast
+from typing import Dict, List, Optional, Tuple, Union, cast  # noqa: F401
 import unittest
 import pandas as pd
 import requests
@@ -35,7 +35,8 @@ class TestBase(unittest.TestCase):
         "task": [],
         "study": [],
         "user": [],
-    }  # type: dict
+    }  # type: Dict[str, List[int]]
+    flow_name_tracker = []  # type: List[str]
     test_server = "https://test.openml.org/api/v1/xml"
     # amueller's read/write key that he will throw away later
     apikey = "610344db6388d9ba34f6db45a3cf71de"
@@ -44,7 +45,7 @@ class TestBase(unittest.TestCase):
     logger = logging.getLogger("unit_tests_published_entities")
     logger.setLevel(logging.DEBUG)
 
-    def setUp(self, n_levels: int = 1):
+    def setUp(self, n_levels: int = 1) -> None:
         """Setup variables and temporary directories.
 
         In particular, this methods:
@@ -100,7 +101,7 @@ class TestBase(unittest.TestCase):
         self.connection_n_retries = openml.config.connection_n_retries
         openml.config.set_retry_policy("robot", n_retries=20)
 
-    def tearDown(self):
+    def tearDown(self) -> None:
         os.chdir(self.cwd)
         try:
             shutil.rmtree(self.workdir)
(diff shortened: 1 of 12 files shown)


openml/testing.py
```diff
@@ -141,7 +147,9 @@ def _delete_entity_from_tracker(self, entity_type, entity):
             if entity_type == "flow":
                 delete_index = [
                     i
-                    for i, (id_, _) in enumerate(TestBase.publish_tracker[entity_type])
+                    for i, (id_, _) in enumerate(
+                        zip(TestBase.publish_tracker[entity_type], TestBase.flow_name_tracker)
```
LennartPurucker, line 151: Why did you add `TestBase.flow_name_tracker` here? 
Would a refactor to `for i, id_ in enumerate(TestBase.publish_tracker[entity_type])` be clearer?

mfeurer: Because we the flow_name in order to delete flows in the correct order (we delete flows with longer names first, because if a flow would be part of another flow, that flow needs to be deleted first, and is guaranteed to have a longer name). Speaking that out loud, it should be sufficient to delete flows by their ID, because a subflow needs to be present before the superflow is added and will therefore have a lower ID. What do you think?

mfeurer: @PGijsbers @janvanrijn @joaquinvanschoren would it work that we simplify the code here a bit?

PGijsbers: From my understanding the ID logic is correct, and sorting them based on IDs would be OK.

PGijsbers: Also I still don't understand what `TestBase.flow_name_tracker` does here, as its variable is not used. The only thing it does here is to make sure that it checks at most `len(TestBase.flow_name_tracker)` elements? It doesn't affect order at all?
And if you _are_ after the length cut-off, then it would be better to be more explicit, e.g.,:
```python
n_flows_to_check = len(TestBase.flow_name_tracker)
flows_to_check = TestBase.publish_tracker[entity_type][:n_flows_to_check]
... enumerate(flows_to_check) ...
```
Second, it looks like we are simply looking up the first element with `value` and then removing it from the list. There's a method for this (`list.remove`). But we can do better if we already have a set anyway:
```python
if entity_type not in TestBase.publish_tracker:
  return

if entity not in TestBase.publish_tracker[entity_type]:
  msg = f"Error deleting {entity=} of {entity_type=}: entity not found in publish tracker."
  raise ValueError(msg)

# Remove duplicates and entity:
remaining_entities = set(TestBase.publish_tracker[entity_type]) - {entity}
TestBase.publish_tracker[entity_type] = list(remaining_entities)
```

Note that the second check is only to preserve behavior that we expect the entity to be present (as indicated by the hard `[0]` index). If it doesn't matter whether or not it is present we can drop that.

PGijsbers: (and even then it probably makes sense to refactor further: it doesn't look like `TestBase.publish_tracker` needs to be a list in the first place, etc. But I would prefer that to be left for a separate PR)

LennartPurucker (approved):
Approving, to finalize the PR before the Hackathon. I will also resolve the current conflicts next.

openml/testing.py
LennartPurucker: I noted this down and will look into it as part of a future PR and test reworks

LennartPurucker:
The tests are now failing (they did not previously) due to server state. 
As far as I can tell, this is not related to the PR itself, and I will merge it now to prepare for the hackathon next week (where we will hopefully fix the tests).

PR #1269, opened 2023-07-08
Pytest/utils

PGijsbers:
Converts test_utils to pytest.
Adds some fixtures which will likely be moved to a common file later, but I figure I would generalize when needed.
Also added markers, this will make it easier to run tests e.g., without internet connection, or to later revisit the tests and see if we can rewrite them to not need the server (e.g., use a cached response).

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -8,3 +8,5 @@ markers=
   server: anything that connects to a server
   upload: anything that uploads to a server
   production: any interaction with the production server
+  cache: anything that interacts with the (test) cache
+
--- a/setup.cfg
+++ b/setup.cfg
@@ -4,3 +4,7 @@ description-file = README.md
 [tool:pytest]
 filterwarnings =
     ignore:the matrix subclass:PendingDeprecationWarning
+markers=
+  server: anything that connects to a server
+  upload: anything that uploads to a server
+  production: any interaction with the production server
--- a/tests/test_utils/test_utils.py
+++ b/tests/test_utils/test_utils.py
@@ -6,13 +6,13 @@ import openml
 from openml.testing import TestBase
 
 
+def mocked_perform_api_call(call, request_method):
+    url = openml.config.server + "/" + call
+    return openml._api_calls._download_text_file(url)
+
+
 class OpenMLTaskTest(TestBase):
     _multiprocess_can_split_ = True
-
-    def mocked_perform_api_call(call, request_method):
-        # TODO: JvR: Why is this not a staticmethod?
-        url = openml.config.server + "/" + call
-        return openml._api_calls._download_text_file(url)
 
     def test_list_all(self):
         openml.utils._list_all(listing_call=openml.tasks.functions._list_tasks)
--- a/tests/test_utils/test_utils.py
+++ b/tests/test_utils/test_utils.py
@@ -24,16 +24,26 @@ def with_test_server():
     openml.config.stop_using_configuration_for_example()
 
 
-def mocked_perform_api_call(call, request_method):
+def _mocked_perform_api_call(call, request_method):
     url = openml.config.server + "/" + call
     return openml._api_calls._download_text_file(url)
 
 
-def test_list_all(as_robot, with_test_server):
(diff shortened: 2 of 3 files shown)


tests/test_utils/test_utils.py
```diff
@@ -5,21 +5,38 @@
 import openml
 from openml.testing import TestBase
 
+import pytest
 
-class OpenMLTaskTest(TestBase):
-    _multiprocess_can_split_ = True
```
PGijsbers, line 10: @mfeurer as far as I can tell this is a nosetest directive. Git blame says you added it, so maybe you know more?

PGijsbers:
For most of these tests we really need to consider them together with the other test (e.g., `test_dataset_functions`). There seems to be a lot of overlap. But in general these tests seem to be testing server behavior and state as well, which we also need to separate. For now it made sense to just convert the tests to pytest and later take note of where to cut or re-organize our tests.

tests/test_utils/test_utils.py
```diff
@@ -1,118 +1,167 @@
...
-    def test_list_all_for_flows(self):
-        required_size = 15  # default test server reset value
-        flows = openml.flows.list_flows(
-            batch_size=25, size=required_size, output_format="dataframe"
-        )
-        self.assertEqual(len(flows), required_size)
-
-    def test_list_all_for_setups(self):
-        required_size = 50
-        # TODO apparently list_setups function does not support kwargs
```
PGijsbers, line 80: @janvanrijn what `kwargs` are missing here? The function (currently) accepts `flow`, `tag` and `setup` as well as `offset` and `size`. This seems to cover the API parameters. Is this outdated? Or was it a TODO to add more filters to the REST API itself?

tests/test_utils/test_utils.py
mfeurer: Yes, this is a nosetest directive and it basically allows nosetests to run tests in this class in parallel, which it would not do otherwise.

PGijsbers: But we don't use nosetest, so there is no reason to keep it around, right?

tests/test_utils/test_utils.py
janvanrijn: I presume this line of code was written more than 5 years ago, and it seems plausible that the code has evolved since then. 

kwargs usually indicates a set of arbitrary arguments that can be passed to a function (which I am no fan of), but a design pattern that we use throughout the openml-python code

PGijsbers: I understand what kwargs does, but all parameters are fully specified already. Any additional pass-through keyword arguments would not be supported by the REST API.

janvanrijn: Ok, I am no longer aware why I wrote this comment (note that this was several years ago). Based on what you say, I think it can be removed.

PGijsbers: Thanks!

tests/test_utils/test_utils.py
mfeurer: Yes, indeed, no need to keep this around. I don't recall when this was necessary, maybe before we added Oslo to ensure that OpenML-Python is thread safe?

setup.cfg
```diff
@@ -4,3 +4,9 @@ description-file = README.md
 [tool:pytest]
 filterwarnings =
     ignore:the matrix subclass:PendingDeprecationWarning
+markers=
+  server: anything that connects to a server
+  upload: anything that uploads to a server
+  production: any interaction with the production server
+  cache: anything that interacts with the (test) cache
+
```
LennartPurucker, line 12: Just going from this for documentation purposes: this would mean that if a test is marked with `server` but not with `production`, we know the test uses a test server?

LennartPurucker (approved):
Looks good to me, nice work!

setup.cfg
PGijsbers: Yes, but since that's default I figured it wouldn't need a marker.

PR #1305, opened 2024-01-08
Tagging constraints

janvanrijn:


LennartPurucker (approved):
Note to self: We might want to mark the `test_illegal_*` tests as server tests not only related to the Python API during the rework.

Issue #1083, opened 2021-05-18
Schedule stop for 3.6 support and discuss a policy going forward


PGijsbers:
Most scientific packages already have `Py3.7+` releases or are talking about it ([NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html#support-table), [scikit-learn issue](https://github.com/scikit-learn/scikit-learn/issues/20084)). This means we should make plans on when to drop `Py3.6` as well if we want to use the latest version of those packages. Additionally it allows us to use newer language features and ensures users are using a safer and faster Python. It's probably also a good idea to have a discussion on a policy going forward.

The NumPy follows NEP 29:

> When a project releases a new major or minor version, we recommend that they support at least all minor versions of Python introduced and released in the prior 42 months from the anticipated release date with a minimum of 2 minor versions of Python, and all minor versions of NumPy released in the prior 24 months from the anticipated release date with a minimum of 3 minor versions of NumPy.

This would mean that we should have dropped 3.6 last year, and 3.7 in December. Since (Python) minor versions are backwards compatible we don't actually have to make code changes, but we do have to:
 - update `setup.py` with new minimum requirement and removing a pypi classifier
 - update the continuous integration Python version matrix
 
And as a bonus it allows us to refactor old code with new language features, where applicable. These include:
 - 3.7: 
   - [Data classes](https://docs.python.org/3/library/dataclasses.html) might simplify our core OpenML entity classes, worth a closer look. 
   -  Module attributes could perhaps be interesting for the extensions as [described here](https://realpython.com/python37-new-features/#customization-of-module-attributes).
    - ordered dictionaries in the language specification (as opposed to CPython implementation detail)
 - 3.8: 
   - the notorious [walrus operator](https://docs.python.org/3/whatsnew/3.8.html#assignment-expressions).
   - [positional only parameters](https://docs.python.org/3/whatsnew/3.8.html#positional-only-parameters)
 - 3.10 (release later this year):
    - [match](https://www.python.org/dev/peps/pep-0635/) statement
 
 Of course actually using any of these features _would_ break backwards compatibility to older Python minor versions.

PGijsbers:
I propose we bump straight to `Py3.8` before the end of this year, and from then on follow the NEP guideline or stricter (i.e. drop minor versions faster).

mfeurer:
Sounds good to me. In addition to NEP29 we should always strive to support the python version of the latest Ubuntu LTS and Google Colab as a lot of people will be using those.

[...]

PR #1308, opened 2024-01-09
ci: Remove Python 3.6/7

eddiebergman:
We removed python 3.6 and python 3.7 in accordance to an undocumented policy #1083:

> We follow the Numpy end-of-life schedule: https://numpy.org/neps/nep-0029-deprecation_policy.html#support-table

We will also remove python 3.8 once we work on moving out the sklearn-extensions to their own dedicated repository.


LennartPurucker (approved):
LGTM

PR #1313, opened 2024-01-10
Add feature description PR

janvanrijn:


openml/datasets/data_feature.py
```diff
@@ -34,6 +36,7 @@ def __init__(  # noqa: PLR0913
         data_type: str,
         nominal_values: list[str],
         number_missing_values: int,
+        ontologies: list[str],
```
LennartPurucker, line 39: I think this needs to have a default value as it is otherwise a breaking change?

openml/datasets/dataset.py
```diff
@@ -1070,6 +1070,7 @@ def _parse_features_xml(features_xml_string: str) -> dict[int, OpenMLDataFeature
             xmlfeature["oml:data_type"],
             xmlfeature.get("oml:nominal_value"),
             int(nr_missing),
+            xmlfeature.get("oml:ontology"),
```
LennartPurucker, line 1073: Same as above, requires a default or assumes None as default

openml/datasets/functions.py
```diff
@@ -1061,6 +1061,61 @@ def fork_dataset(data_id: int) -> int:
     return int(data_id)
 
 
+def data_feature_add_ontology(data_id, index, ontology):
+    """
+    Adds an ontology (URL) to a given dataset feature (defined by a dataset id
```
LennartPurucker, line 1066: Please add a short description of the concept of ontology in the docstring.

openml/datasets/data_feature.py
```diff
@@ -23,6 +23,8 @@ class OpenMLDataFeature:
         list of the possible values, in case of nominal attribute
     number_missing_values : int
         Number of rows that have a missing value for this feature.
+    ontologies : list(str)
+        list of ontologies attached to this feature
```
LennartPurucker, line 27: Maybe add a short description of the concept of ontologies.

openml/datasets/functions.py
```diff
@@ -1061,6 +1061,61 @@ def fork_dataset(data_id: int) -> int:
     return int(data_id)
 
 
+def data_feature_add_ontology(data_id, index, ontology):
```
LennartPurucker, line 1064: Requires type hints and return type

openml/datasets/functions.py
```diff
@@ -1061,6 +1061,61 @@ def fork_dataset(data_id: int) -> int:
...
+    -------
+    True or throws an OpenML server exception
+    """
+    upload_data = {"data_id": data_id, "index": index, "ontology": ontology}
+    openml._api_calls._perform_api_call("data/feature/ontology/add", "post", data=upload_data)
+    # an error will be thrown in case the request was unsuccessful
+    return True
+
+
+def data_feature_remove_ontology(data_id, index, ontology):
```
LennartPurucker, line 1091: Add typing

openml/datasets/functions.py
```diff
@@ -1061,6 +1061,61 @@ def fork_dataset(data_id: int) -> int:
...
+    """
+    Removes an existing ontology (URL) to a given dataset feature (defined
```
LennartPurucker, line 1093:
```suggestion
    Removes an existing ontology (URL) from a given dataset feature (defined
```

openml/datasets/functions.py
```diff
@@ -1061,6 +1061,61 @@ def fork_dataset(data_id: int) -> int:
...
+    and index). The dataset has to exists on OpenML and needs to have been
+    processed by the evaluation engine.
+
+    Parameters
+    ----------
+    data_id : int
+        id of the dataset to which the feature belongs
+
+    index : int
+        index of the feature in dataset (0-based)
```
LennartPurucker, line 1076: Which list is this based on? I.e., which 0-based list of a dataset object is used? It might help the user to orientate with the dataset.

LennartPurucker (changes requested):
Added multiple comments. Otherwise LGTM!

LennartPurucker:
Also, try to fill the PR template so we can document the change for later; thanks!

openml/datasets/data_feature.py
janvanrijn: while technically it would make break code, I find it unlikely that people use this constructor outside our library. In the library, I think I have updated all constructors

openml/datasets/dataset.py
janvanrijn: can you elaborate what you mean? This is the usage of the constructor, right?

LennartPurucker (approved):

eddiebergman:
Closing and we cherry picked these into a new PR for a simpler times, @janvanrijn you were added to the contributors  of that PR!

PR #1317, opened 2024-01-15
ci: Test for 3.9, 3.10, 3.11

eddiebergman:
This PR adds testing for 3.9-3.12. Some of these may fail due to scikit-learn. Not sure what to do about this since this will eventually be refactored out into its own package

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -91,6 +91,7 @@ jobs:
           fi
       - name: Store repository status
         id: status_before
+        shell: bash
         run: |
           status="$(git status --porcelain -b)"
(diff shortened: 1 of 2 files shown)


LennartPurucker:
Let us wait until we decide on the future of the sklearn extension to proceed with this PR.

eddiebergman:
> Let us wait until we decide on the future of the sklearn extension to proceed with this PR.

Just going to set it up in the meantime

eddiebergman:
This PR now works. As expected, there's scikit-learn errors with 3.10, 3.11 and 3.12 that still need to be handled.

PR #1323, opened 2024-01-18
New Release 0.14.2

LennartPurucker:


PGijsbers:
There are some problems with the history between `main` and `develop`. We can clean this up (for example, by branching off `main` again), but Lennart proposed we simply abandon the two branch system in favor of using tags for releases.
I personally also would prefer this option, as it's less error prone. We could keep the `main` branch around for archival reasons. The documentation pages would need an update.

PR #1335, opened 2024-04-02
Feat/progress

PGijsbers:
* Adds a progress bar to MinIO downloads
* When `download_all_files=True`, only download the files which do not yet exist in cache. Overwriting is no longer necessary, as it was a crutch that was built before `force_refresh_cache` was added to the `get_dataset` method. This means we now can also avoid re-downloading image datasets over and over again.

WIP:
 - [x] add/update test for cache behavior
 - [x] update release notes


openml/_api_calls.py
```diff
@@ -161,12 +163,12 @@ def _download_minio_file(
     proxy_client = ProxyManager(proxy) if proxy else None
 
     client = minio.Minio(endpoint=parsed_url.netloc, secure=False, http_client=proxy_client)
-
     try:
         client.fget_object(
             bucket_name=bucket,
             object_name=object_name,
             file_path=str(destination),
+            progress=ProgressBar() if config.show_progress else None,
```
eddiebergman, line 171: I'm not entirely sure how you parse the config file but just ensure it's a boolean as if `config.show_progress == "false"`, this would still evaluate to `True`

openml/config.py
```diff
@@ -255,6 +259,7 @@ def _setup(config: _Config | None = None) -> None:
     avoid_duplicate_runs = config["avoid_duplicate_runs"]
     apikey = config["apikey"]
     server = config["server"]
+    show_progress = config.get("show_progress", _defaults["show_progress"])
```
eddiebergman, line 262: Should this explicitly cast to bool somewhere? I found this line in the parsing of the Config which seems to suggest it needs explicit handling to convert to booleans.

https://github.com/openml/openml-python/blob/8d0324374733795ca05975b2427757c12e24d949/openml/config.py#L336-L340

PGijsbers: Yikes, can't believe I missed that. Thanks. 
It turns out that this didn't work either. Setting avoid_duplicate_runs would actually give me a TypeError (wasn't allowed to set a boolean value to the Config object), so I rewrote this and added a regression test.

eddiebergman:
Did a brief look at `configparser` and `get_boolean()` for common methods of specifying True/False in raw text. Covered my concerns, never knew about it, seems super useful!

PGijsbers:
Honestly if we did it again I would probably just use toml files, which are typed.

PGijsbers:
@eddiebergman the review request is still open. It needs explicit approval.

eddiebergman (approved):

eddiebergman:
Ahh sorry, thought it was just a thing of getting some feedback, not explicit approval, my bad

Issue #1333, opened 2024-03-23
Dataset download progress bar


joaquinvanschoren:
#### Description
Downloading large datasets can take an unpredictable amount of time. It would be nice if a progress bar could be shown. 

#### Steps/Code to Reproduce
openml_dataset = openml.datasets.get_dataset(ID)
X, y, _, _ = openml_dataset.get_data()

#### Expected Results
A progress bar is shown (either in standard out or in a notebook)

#### Actual Results
Nothing is shown

#### Versions
All

PGijsbers:
It seems that there is a `progress` parameter to the `fget_object` call when downloading from minio: https://github.com/minio/minio-py/blob/74e8e5200d4dbf48ae9bbd6a8a2f54614e1958b3/minio/api.py#L1066
it's not in the main doc pages for whatever reason. I think it would be reasonable to only implement this for downloads from minio as otherwise files should generally be small (especially once we drop arff).

PGijsbers:
This is implemented in #1335, which can hopefully be included in 0.15, whenever that may be.

PGijsbers:
Added in #1335, so it will be in the next release. 
Enable it by setting `openml.config.show_progress = True`. 

Might rename it to the more verbose but clearer "show_download_progress" before PyPI release - just something that came into my mind now.

PR #1336, opened 2024-04-23
Implemented progress bar in dataset download

megansu13:
- Modified datasets/dataset.py to add progress bar using tqdm
- Added new script run_openml.py for example usage of OpenML with progress bar

Implemented progress bar using tqdm. Within datasets/dataset.py, I imported requests and tqdm, 
modified _download_data() and _parse_data_from_arff() methods, as well as added run_openml.py.  

Here is what run_openml.py does:
1. Fetching Dataset: It loads a dataset from OpenML with a specific ID (in your example, it's ID 1471, which is often the Iris dataset). The get_dataset function has the ability to download the dataset, its associated qualities, and feature metadata if not already present.
2. Extracting Data: It retrieves the features X and the target variable y from the dataset. The target is specified by the dataset's default target attribute.
3. Data Exploration: It prints out the first few entries of the feature set X using X.head(), gives a statistical summary of X with X.describe(), and shows the distribution of different classes/targets in y with y.value_counts().
4. Data Splitting: It splits the data into training and testing sets with an 80-20 ratio using train_test_split. The random_state parameter ensures that the split is reproducible.
5. Model Training: It initializes a Random Forest classifier, trains it on the training set, and then predicts on the testing set.
6. Evaluation: It evaluates the predictions by computing the accuracy score, comparing the predicted target values y_pred against the true values y_test.
7. Progress Bar: It demonstrates a simple progress bar using tqdm, which iterates through a range of 10, pausing for 0.5 seconds at each step.

This PR should be tested by running:
cd openml
python3 run_openml.py

#### Any other comments?
Let me know if there is anything else I should provide, I am happy to help!

### DIFF
--- a/openml/datasets/dataset.py
+++ b/openml/datasets/dataset.py
@@ -21,6 +21,9 @@ from openml.exceptions import PyOpenMLError
 
 from .data_feature import OpenMLDataFeature
 
+import requests
+from tqdm import tqdm
+
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


PGijsbers:
Hi! 👋 First off, thanks so much for taking the effort to contribute to OpenML! It's always a big step to make a first contribution, so we really appreciate it.

Unfortunately, it seems that our efforts got crossed a little bit. A progress bar for MinIO downloads has been added in https://github.com/openml/openml-python/pull/1335. However, this PR does also create a progress bar when downloading ARFF files, which is nice (at least for the larger arff files). 

There are also a few things about the PR that would need changing regardless (I don't advise you take this step until we see how we handle the multiple PRs, but I wanted to include it for any future work):
 
  * The PR contains commented out code - we don't want that in our code base. Comments are used to explain why certain code exists and, in rare cases, to explain what code does. Commented out code just makes the source code harder to navigate.
  * The behavior of the `_download_data` function has been changed. `self.data_file` and `self.parquet_file` should be given a value in this function that correspond to the paths on disk. It is likely other parts of the API break if this is not done. 
  * The functions `_get_dataset_arff` and `_get_dataset_parquet` don't just download the files (like your new function to replace them does), they also take care of caching and calculating checksums to verify the downloads were successful. These functionalities are crucial for our users, so we cannot simply replace them by this new function.
  * The new function which loads arff does not take into account the metadata when deciding the types of data in the dataframe.  This will almost certainly lead to datatypes being inferred incorrectly (or at least, inconsistently).
  * A usage example to demonstrate the PR can be provided directly on Github as a code listing. Or, if appropriate, should be added to the documentation. We do not want extra example scripts in other parts of our project structure.

To avoid this from happening again, if you want to contribute, the best way to go about this is to either comment on an open issue to let us know you are interested in it, or open a new issue indicating you want to set up a PR for this. That way, we can first check whether we are (still) interested in the feature, ensure that no one else has been working on it, and provide some pointers on how we think it should best be implemented.

I will go ahead and close this PR, as it would need major rewrites. We would be open to having a progress bar for (large) arff file downloads to accompany #1335, but the changes should be much less invasive. I think a good place to start would be to look at [this](https://github.com/openml/openml-python/blob/90380d45d190a9d40535353d232bb0909630faa5/openml/_api_calls.py#L212) function and the functions it calls to actually download the file. If you are still interested in this, I would propose that you have a look at that, and maybe share your findings and make a suggestion in the related issue #1333 before working on a new PR. Looking forward to hearing from you :)

PR #1338, opened 2024-05-12
read file in read mode

### LINKED ISSUE #1337: Config file being overwritten at import
#### Description
The config file located at .../.config/openml/config is being overwritten with the default values when importing openml.

#### Steps/Code to Reproduce
1 - Create a config file in .../.config/openml/config
2 - ```import openml``` in any python script
3 - check that the file is now empty and the config values being used are the default ones

#### Expected Results
Do not overwrite the config file

#### Actual Results
Config file is overwritten

#### Versions
Linux-6.5.0-28-generic-x86_64-with-glibc2.35
Python 3.11.9 | packaged by conda-forge | (main, Apr 19 2024, 18:36:13) [GCC 12.3.0]
NumPy 1.26.4
SciPy 1.13.0
Scikit-Learn 1.4.2
OpenML 0.14.2

#### Possible solution
I think that the reason is the following code under config.py

```
def _parse_config(config_file: str | Path) -> _Config:
    """Parse the config file, set up defaults."""
    config_file = Path(config_file)
    config = configparser.RawConfigParser(defaults=_defaults)  # type: ignore

    # The ConfigParser requires a [SECTION_HEADER], which we do not expect in our config file.
    # Cheat the ConfigParser module by adding a fake section header
    config_file_ = StringIO()
    config_file_.write("[FAKE_SECTION]\n")
    try:
        with config_file.open("w") as fh:
            for line in fh:
                config_file_.write(line)
    except FileNotFoundError:
        logger.info("No config file found at %s, using default configuration.", config_file)
    except OSError as e:
        logger.info("Error opening file %s: %s", config_file, e.args[0])
    config_file_.seek(0)
    config.read_file(config_file_)
    return dict(config.items("FAKE_SECTION"))
```
The problem is that the file is being opened in write mode 'w', when it should be open in read mode 'r'. I will see if I can provide a PR to fix it.

BrunoBelucci:
#### Reference Issue
Fixes #1337 

It changes the mode that config_file is being opened, so it will not be overwritten. It also casts the variables of the config_file to the expected types, otherwise, everything is read as str.

1 - Create a new config file .../.config/openml/config
2 - do ```import openml```
3 - check that the config file is preserved

### DIFF
--- a/openml/config.py
+++ b/openml/config.py
@@ -252,11 +252,11 @@ def _setup(config: _Config | None = None) -> None:
     if config is None:
         config = _parse_config(config_file)
 
-    avoid_duplicate_runs = config.get("avoid_duplicate_runs", False)
+    avoid_duplicate_runs = bool(config.get("avoid_duplicate_runs", False))
     apikey = config["apikey"]
     server = config["server"]
-    short_cache_dir = config["cachedir"]
-    n_retries = config["connection_n_retries"]
+    short_cache_dir = Path(config["cachedir"])
+    n_retries = int(config["connection_n_retries"])
 
     set_retry_policy(config["retry_policy"], n_retries)
 
--- a/openml/config.py
+++ b/openml/config.py
@@ -252,7 +252,7 @@ def _setup(config: _Config | None = None) -> None:
     if config is None:
         config = _parse_config(config_file)
 
-    avoid_duplicate_runs = bool(config.get("avoid_duplicate_runs", False))
+    avoid_duplicate_runs = config.get("avoid_duplicate_runs", "").lower() == "true"
     apikey = config["apikey"]
     server = config["server"]
     short_cache_dir = Path(config["cachedir"])
--- a/openml/config.py
+++ b/openml/config.py
@@ -252,7 +252,7 @@ def _setup(config: _Config | None = None) -> None:
     if config is None:
         config = _parse_config(config_file)
 
-    avoid_duplicate_runs = config.get("avoid_duplicate_runs", "").lower() == "true"
+    avoid_duplicate_runs = config["avoid_duplicate_runs"]
     apikey = config["apikey"]
     server = config["server"]
     short_cache_dir = Path(config["cachedir"])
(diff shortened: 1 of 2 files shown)


openml/config.py
```diff
@@ -252,11 +252,11 @@ def _setup(config: _Config | None = None) -> None:
     if config is None:
         config = _parse_config(config_file)
 
-    avoid_duplicate_runs = config.get("avoid_duplicate_runs", False)
+    avoid_duplicate_runs = bool(config.get("avoid_duplicate_runs", False))
```
PGijsbers, line 255: This is not going to work, `bool("False")` will resolve to `True`.

PGijsbers (changes requested):
Thanks for catching this issue and setting up a pull request! There is an issue with casting `avoid_duplicate_runs` into a boolean, but otherwise it looks good! Would you be able to resolve that? Something like `config.get("avoid_duplicate_runs", "").lower() == "true"` is good enough for now I think.

openml/config.py
BrunoBelucci: changed to ```avoid_duplicate_runs = config.get("avoid_duplicate_runs", "").lower() == "true"```

BrunoBelucci:
@PGijsbers it is done =)

PGijsbers:
Thanks! That I had missed something, so I changed it a little after all. Also added a test.

@LennartPurucker I know this isn't very pretty. But I don't think it will be without overwriting more of the config module which I would be in favor of but shouldn't be in this PR. Since I touched this PR now, could you sign off on it if you agree that we should apply the patch?

PGijsbers (approved):
Changing my review status to indicate my earlier concerns are addressed and to make sure it can be merged later. Still need approval of a 2nd reviewer before merging as I now contributed to this PR.

LennartPurucker (approved):
LGTM! 

The change from `avoid_duplicate_runs = config.get("avoid_duplicate_runs", False)` to `avoid_duplicate_runs = config["avoid_duplicate_runs"]` should not brake due to the config object needs to have `avoid_duplicate_runs` set anyhow.

LennartPurucker:
The tests seem to be failing due to a (test) server error.

PGijsbers:
> The change from avoid_duplicate_runs = config.get("avoid_duplicate_runs", False) to avoid_duplicate_runs = config["avoid_duplicate_runs"] should not break since the config object must have avoid_duplicate_runs set anyhow.

Yes this was deliberate.
As for the tests failing... I know :( I don't know how easy it is to resolve that. Despite that, I would still be in favor of merging it. These changes are easy to understand and solve a bug which breaks a core feature.

LennartPurucker:
I agree, feel free to merge!

PGijsbers:
Thanks again @BrunoBelucci 👏

LennartPurucker:
Thank you @BrunoBelucci!

PR #1340, opened 2024-07-03
Fix/sklearn test compatibility

PGijsbers:
This PR updates tests to be compatible with newer versions of scikit-learn. In the process, fixes some bugs and adds numpy 2.0 support. I updated the CI matrix to include newer versions of scikit-learn. For the most part, openml-python works fine on newer versions, it was just that the unit tests themselves had hard dependencies on specifics. I extended/updated in the most cases, but in a few it was a little unreasonable (needs to be refactored to make it less sensitive to small scikit-learn api changes). I already spent quite a bit of time on this, so I hope that the proposed changes are good enough for now, so we can work on a 0.15.0 release.

The docs fail in the check-links step. I propose to fix this in a separate PR.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -89,8 +89,8 @@ jobs:
         if [ ${{ matrix.code-cov }} ]; then codecov='--cov=openml --long  --cov-report=xml'; fi
         # Most of the time, running only the scikit-learn tests is sufficient
         if [ ${{ matrix.sklearn-only }} = 'true' ]; then sklearn='-m sklearn'; fi
-        echo pytest -n 4 --durations=20 --timeout=600 --timeout-method=thread --dist load -sv $codecov $sklearn --reruns 5 --reruns-delay 1 -o log_cli=true
-        pytest -n 4 --durations=20 --timeout=600 --timeout-method=thread --dist load -sv $codecov $sklearn --reruns 5 --reruns-delay 1 -o log_cli=true
+        echo pytest -n 4 --durations=20 --dist load -sv $codecov $sklearn -o log_cli=true
+        pytest -n 4 --durations=20 --dist load -sv $codecov $sklearn -o log_cli=true
     - name: Run tests on Windows
       if: matrix.os == 'windows-latest'
       run: |  # we need a separate step because of the bash-specific if-statement in the previous one.
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -76,10 +76,10 @@ jobs:
       if: ${{ matrix.python-version == '3.8' && matrix.scikit-learn == '0.23.1' }}
       run: |
         pip install numpy==1.23.5
-    - name: Install NumPy 1.x for scikit-learn < 1.4
+    - name: "Install NumPy 1.x for scikit-learn < 1.4"
       if: ${{ contains(fromJSON('["1.0", "1.1", "1.2", "1.3"]'), matrix.scikit-learn) }}
       run: |
-        pip install numpy<=2.0
+        pip install "numpy<=2.0"
     - name: Install scipy ${{ matrix.scipy }}
       if: ${{ matrix.scipy }}
       run: |
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -89,6 +89,8 @@ jobs:
       id: status-before
       run: |
         echo "::set-output name=BEFORE::$(git status --porcelain -b)"
+    - name: Show installed dependencies
+      run: python -m pip list
     - name: Run tests on Ubuntu
       if: matrix.os == 'ubuntu-latest'
       run: |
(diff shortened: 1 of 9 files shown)


PGijsbers:
Test failures seem to be caused by exceeding the 10 minute time limit we have set. I am not sure why these tests need as long, but it looks like they fail and are repeated. It's likely their failure is deterministic however. So I will go ahead and disable retries so the test does a hard fail instead with a more meaningful error message (hopefully).

PGijsbers:
Tests are incredibly slow, I am not sure why. This also happens locally, so it's not just the runners.

PGijsbers:
On Numpy 2.0 support/scikit-learn support (from #1341):

Numpy2.0 cleaned up their namespace, so we have to alias for the new version.

The problem is that this is not covered in CI (tested versions do not support numpy 2.0), and in general tests need to be updated to support more recent scikit-learn versions that support 2.0 (1.5, maybe 1.4?). I don't know if I really have the time for that right now. Options:

Put a pin in the dependencies
Accept this as a best-effort, and hope that the tests that don't fail due to scikit-learn incompatibility sufficiently cover the code to raise any numpy1/2 issues
Wait until I find time to make tests support scikit-learn 1.5

openml/runs/functions.py
```diff
@@ -270,8 +270,7 @@ def run_flow_on_task(  # noqa: C901, PLR0912, PLR0915, PLR0913
             raise PyOpenMLError(
                 "Flow does not exist on the server, but 'flow.flow_id' is not None."
             )
-
-        if upload_flow and flow_id is None:
+        if upload_flow and flow_id is False:
```
PGijsbers, line 273: `flow_exists` does not return `None` but `False` when a flow does not exist.
I find `flow_id is False` easier to parse than `not flow_id` since `flow_id` can be both integer or boolean. While both are correct, the check is clearer to me this way.

PGijsbers: This bug would fail a unit test only the first time around, unit tests would later upload the flow anyway which means the second time this path is no longer taken and tests can pass. Because there are other tests which update server state that need to be ran a few times, this error was never caught as it was just in the initial sea of reds after a server reset.

.github/workflows/test.yml
```diff
@@ -25,32 +25,34 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: ["3.8"]
```
PGijsbers, line 28: I rewrote the matrix to be effectively the same, but also testing each 1.x release.
The only difference is that the default is now 3.9, since that is supported for all scikit-learn versions (except 0.23) so it makes things easier.

.github/workflows/test.yml
```diff
@@ -72,9 +74,15 @@ jobs:
         pip install scikit-learn==${{ matrix.scikit-learn }}
     - name: Install numpy for Python 3.8
       # Python 3.8 & scikit-learn<0.24 requires numpy<=1.23.5
-      if: ${{ matrix.python-version == '3.8' && contains(fromJSON('["0.23.1", "0.22.2", "0.21.2"]'), matrix.scikit-learn) }}
+      if: ${{ matrix.python-version == '3.8' && matrix.scikit-learn == '0.23.1' }}
```
PGijsbers, line 77: older versions were (no longer) included in the run matrix anyway.

tests/test_runs/test_run_functions.py
```diff
@@ -119,14 +119,14 @@ def _wait_for_processed_run(self, run_id, max_waiting_time_seconds):
...
         while time.time() - start_time < max_waiting_time_seconds:
-            run = openml.runs.get_run(run_id, ignore_cache=True)
 
             try:
                 openml.runs.get_run_trace(run_id)
             except openml.exceptions.OpenMLServerException:
                 time.sleep(10)
                 continue
 
+            run = openml.runs.get_run(run_id, ignore_cache=True)
```
PGijsbers, line 129: Sometimes test fail below because `run.evaluations` is `None`. But I am not able to reproduce this locally. However, the fails do seem fewer now that I move this line (which makes sense, initially there is a race condition where perhaps the trace isn't yet processed in get `get_run` call, but is by the time the `get_run_trace` is called.

PGijsbers:
Note that the CI results are [here](https://github.com/openml/openml-python/actions/runs/9794921128/job/27045923855)... the `[skip ci]` doesn't work well with pre-commit.ci apparently.
pre-commit fails there because it was aborted by the newer commit.

PGijsbers:
@LennartPurucker I pinged you for a review since at this point I am unsure what to do. The work is done, kind of. Sometimes the tests still fail sporadically because the pytest workers get killed on GitHub actions. I have not had this happen locally. It's also not consistent - they can also pass. At this point, there are already large improvements to the tests and CI. It's probably already worth merging. From there, I can work on both new features/updates as well as debug, but then at least that can be in parallel and be discussed in separate smaller PRs. What do you think?

LennartPurucker (approved):
LGTM! 

I agree and think we should merge this. The changes are all reasonable and important improvements.

PR #1343, opened 2024-07-10
Superseded by pre-commit.ci

PGijsbers:
No longer needed since we have pre-commit.ci active. I suppose the latter only works on the delta and not all files, but that should not be an issue.


eddiebergman:
I've never used the standalone github plugin. Looks pretty good in terms of speed, probably better than a naive setup through github actions. Fire ahead!

One caveat to consider is that if you ever want to consider `pre-commit` as a requirement before a release, i.e. in an automated fashion, you may need to re-introduce that it runs through github actions _somewhere_

eddiebergman (approved):

PGijsbers:
It's definitely faster, often finishes before the GHA job is even set up. 
I do think the concern is valid, but I think that would be clearer to do that in whatever dedicated release pipeline we build and not as a standalone workflow anyway.

PR #1344, opened 2024-08-23
Pin NumPy to be lower than 2.0

knyazer:
Fixes ```AttributeError: `np.sctypes` was removed in the NumPy 2.0 release.``` on newer NumPy versions and poetry installation

#### Reference Issue

None

Fixes the 
```
AttributeError: `np.sctypes` was removed in the NumPy 2.0 release. Access dtypes explicitly instead.. Did you mean: 'dtypes'?
```
in the ```sklearn``` extension. (line 53, ```sklearn/extension.py```)

You might say "but nobody has ever experienced this issue!". True. For some reason, pip can figure out that the dependency should be lower than 2.0; I don't how it does it. However, I use poetry as a package manager, and if installing openml via poetry, it uses numpy 2.1 as a dependency, which breaks openml.

Install poetry. Then:
```
mkdir test && cd test && poetry init -q && poetry add openml && poetry run python -m openml
```
This would throw before the changes, but would not after the changes.

#### Any other comments?

Should be pretty easy to merge.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,7 +16,7 @@ dependencies = [
   "python-dateutil",  # Installed through pandas anyway.
   "pandas>=1.0.0",
   "scipy>=0.13.3",
-  "numpy>=1.6.2",
+  "numpy>=1.6.2,<2.0.0",
   "minio",
   "pyarrow",
 ]


PGijsbers:
Did you try yet to install from GitHub? The develop branch should already be numpy 2.0 compatible: https://github.com/openml/openml-python/pull/1341

knyazer:
No, I actually did not :(

Yep, it works on the development version. Annoying that it is not yet on the pypi.

Sorry for the noise, then.

PR #1345, opened 2024-08-23
Pass kwargs through task to ```get_dataset```

knyazer:
Allows to follow the directions in the FutureWarning ```Starting from Version 0.15 `download_data`, `download_qualities`, and `download_features_meta_data` will all be ``False`` instead of ``True`` by default to enable lazy loading.```

#### Reference Issue

None

When doing something like 
```
dataset = task.get_dataset()
```
which takes no parameters one gets the warning. Now it takes any parameters that the ```datasets.get_dataset``` takes, thus allowing to set the optional parameters.

Test that the suite passes, I guess; there is not really a short way to test.

#### Any other comments?

Maybe kwargs-passing behaviour should be added to all the other methods in the ```OpenMLTask```?

(this PR might be noise again, though I checked that the issue is present on dev, sorry if I am wrong though)

### DIFF
--- a/openml/tasks/task.py
+++ b/openml/tasks/task.py
@@ -145,7 +145,7 @@ class OpenMLTask(OpenMLBase):
         ]
         return [(key, fields[key]) for key in order if key in fields]
 
-    def get_dataset(self, /, **kwargs) -> datasets.OpenMLDataset:
+    def get_dataset(self, **kwargs) -> datasets.OpenMLDataset:
         """Download dataset associated with task.
 
         Accepts the same keyword arguments as the `openml.datasets.get_dataset`.
--- a/openml/tasks/task.py
+++ b/openml/tasks/task.py
@@ -145,8 +145,11 @@ class OpenMLTask(OpenMLBase):
         ]
         return [(key, fields[key]) for key in order if key in fields]
 
-    def get_dataset(self, **kwargs) -> datasets.OpenMLDataset:
-        """Download dataset associated with task."""
+    def get_dataset(self, /, **kwargs) -> datasets.OpenMLDataset:
+        """Download dataset associated with task.
+
+        Accepts the same keyword arguments as the `openml.datasets.get_dataset`.
+        """
         return datasets.get_dataset(self.dataset_id, **kwargs)
 
     def get_train_test_split_indices(
--- a/openml/tasks/task.py
+++ b/openml/tasks/task.py
@@ -145,9 +145,9 @@ class OpenMLTask(OpenMLBase):
         ]
         return [(key, fields[key]) for key in order if key in fields]
 
-    def get_dataset(self) -> datasets.OpenMLDataset:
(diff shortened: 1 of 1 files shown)


openml/tasks/task.py
```diff
@@ -145,9 +145,9 @@ def _get_repr_body_fields(self) -> Sequence[tuple[str, str | int | list[str]]]:
         ]
         return [(key, fields[key]) for key in order if key in fields]
 
-    def get_dataset(self) -> datasets.OpenMLDataset:
+    def get_dataset(self, **kwargs) -> datasets.OpenMLDataset:
```
PGijsbers, line 148: Please add a note to the documentation that it accepts arguments for the `datasets.get_dataset` call so the caller knows what kwargs are allowed.

PGijsbers (changes requested):
Thanks for the contribution! I think it's fair to allow this, but it needs a small improvement for the documentation. I don't think it is necessary to add tests for this functionality.

openml/tasks/task.py
knyazer: Sure, will do in a few hours. Thanks for taking a look!

knyazer:
Btw, are you in need of any maintainers/collaborators? I'm happy to be spending quite a bit of time supporting openml, since I work on automl, and frequently use the library, and I have a bit of free time this year (part-time job, so I can work on random projects quite a bit).

And yeah, it seems like there are a lot of low-ish priority feature requests in the issues, and you are effectively the only maintainer..

So yeah, if you are happy with that I can just start implementing/fixing random stuff pretty actively :wink:

PGijsbers:
Absolutely! It would be best if you could join our Slack server (see https://openml.org/about) for general contact and maybe discussing the kind of issues we currently prioritize. For discussing specific issues/prs, GitHub is the main platform. We also do monthly engineering calls that are open. Every first monday of the month at 17.00 CEST (when daylight saving is applicable in Europe, otherwise CET).
We always appreciate the help!

openml/tasks/task.py
```diff
@@ -145,9 +145,12 @@ def _get_repr_body_fields(self) -> Sequence[tuple[str, str | int | list[str]]]:
         ]
         return [(key, fields[key]) for key in order if key in fields]
 
-    def get_dataset(self) -> datasets.OpenMLDataset:
-        """Download dataset associated with task."""
-        return datasets.get_dataset(self.dataset_id)
+    def get_dataset(self, /, **kwargs) -> datasets.OpenMLDataset:
```
PGijsbers, line 148:
```suggestion
    def get_dataset(self, **kwargs) -> datasets.OpenMLDataset:
```

knyazer:
Alright, disallowing positional arguments while there are zero of them is useless; I just put slash there to be more explicit, as a sort of self-documentation :)

knyazer:
And yeah, thanks for fixing it, the back and forth through the GitHub is a lot slower than changing two chars in the commit

PGijsbers (approved):

PR #1350, opened 2024-09-13
Fix/1349

### LINKED ISSUE #1349: Stop relying on distutils for Py3.12 compatibility
https://docs.python.org/3.10/whatsnew/3.10.html#distutils

to be replaced with a third-party package, potentially.

PGijsbers:
Closes #1349 

Distutils is removed in Py3.12, so we have to rely on external packages.
`packaging` is already a dependency of one of our dependencies, so it doesn't add to the stack (though I made it implicit, in case the dependency chain gets broken at some point for some reason).

### DIFF
--- a/openml/extensions/sklearn/extension.py
+++ b/openml/extensions/sklearn/extension.py
@@ -237,11 +237,9 @@ class SklearnExtension(Extension):
         -------
         str
         """
-        # the variable is named incorrectly: major.minor.micro/patch
-        openml_major_version = Version(openml.__version__).minor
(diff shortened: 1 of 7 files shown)


PGijsbers:
There are unfortunately minor server issues after all. They didn't happen earlier in the week in a local test, currently don't have the internet connection bandwidth to track it down (travelling). But looking at the failures, they can't be related to the PR.

eddiebergman (approved):

eddiebergman:
Not super familiar with this domain of Python but from a quick read it seems mostly superficial in terms of changes and didn't spot anything off

PR #1351, opened 2024-09-13
Remove archive after it is extracted to save disk space

### LINKED ISSUE #1348: Add clean up function for downloaded image archive
Currently the image archive sticks around, but we probably don't want that by default.

cf. https://github.com/openml/openml-python/blob/de983ac5a313319ff0d2f3528f609a49d767382e/openml/_api_calls.py#L172

PGijsbers:
Closes #1348

I don't really know that we need to make this configurable at this point. Let's not add more options for now, and see if we get requests.

### DIFF
--- a/openml/_api_calls.py
+++ b/openml/_api_calls.py
@@ -172,6 +172,7 @@ def _download_minio_file(
         if destination.is_file() and destination.suffix == ".zip":
             with zipfile.ZipFile(destination, "r") as zip_ref:
                 zip_ref.extractall(destination.parent)
+            destination.unlink()
 
     except minio.error.S3Error as e:
         if e.message is not None and e.message.startswith("Object does not exist"):
--- a/openml/_api_calls.py
+++ b/openml/_api_calls.py
@@ -6,6 +6,7 @@ import hashlib
 import logging
 import math
 import random
+import shutil
 import time
 import urllib.parse
 import xml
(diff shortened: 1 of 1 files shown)


PGijsbers:
@prabhant This is what you are looking for, right?

prabhant:
Looks good to me

PGijsbers:
It just sprang to my mind that removing the file does conflict with the caching mechanism, so it will likely always download the archive now... What would be a good way to resolve that? Leaving some kind of empty marker with the same name is a quick hack. @LennartPurucker @eddiebergman

eddiebergman:
Interesting, I came across this exact problem recently where I don't know the contents of tarfiles directly and needed to check if their content were already present. I didn't implement a solution other than `force: bool` but your idea about a marker sounds promising. It would assume that the any new tarfiles that might be downloaded are uniuly named, such that you can one-to-one map them to a marker file.

I don't have a good solution but please keep me posted if you come up with one!

LennartPurucker (changes requested):
Yea, I agree. A marker file for the file being already extracted would be needed and an option to still force the download (force bool or cleaning up the cache). 

Ideally, we would have a uuid for the content of the zip file on the server that, if it changes, would prompt us to re-download. Then, we would just name the marker file using this uuid and check for a match before downloading and extracting again.

PGijsbers:
In this case we can apparently use the metadata which contains a hash (at least, as far as I can tell, that's what it is). For a force refresh, we already support that at a get_dataset level, so I didn't think we need to do anything special here.

LennartPurucker (approved):
Very nice, LGTM, thanks!

PR #1353, opened 2024-09-26
Make test insensitive to OrderedDict stringification

PGijsbers:
Sometime between 3.9 and 3.12 the stringification of ordered dicts changed from using a list of tuples to a dictionary.
 

---
was setting a new environment and noticed this little change (apparently 3.11->3.12):
```
➜  omlp git:(fix/regex) ✗ pyenv shell 3.12
➜  omlp git:(fix/regex) ✗ python -c "from collections import OrderedDict; print(OrderedDict(a=2))"
OrderedDict({'a': 2})
➜  omlp git:(fix/regex) ✗ pyenv shell 3.11
➜  omlp git:(fix/regex) ✗ python -c "from collections import OrderedDict; print(OrderedDict(a=2))"
OrderedDict([('a', 2)])
```

### DIFF
--- a/tests/test_flows/test_flow_functions.py
+++ b/tests/test_flows/test_flow_functions.py
@@ -5,6 +5,8 @@ import copy
 import functools
 import unittest
 from collections import OrderedDict
+from multiprocessing.managers import Value
+
 from packaging.version import Version
 from unittest import mock
 from unittest.mock import patch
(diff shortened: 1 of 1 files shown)


eddiebergman (approved):
Looks good, interesting behavioural change, wonder what their motivation for the change was?

PGijsbers:
If I had to guess, it's because of the 3.6 implementation/3.7 spec change to dictionaries where they are now ordered by default. For old representations a list made more sense since that was always ordered. I wonder if they missed this earlier or it has something to do with internal deprecation cycles or something, but I imagine it has to be related to that.

PR #1354, opened 2024-09-28
Change defaults for `get_task` to be lazy

### LINKED ISSUE #1034: Don't download (large) datasets by default

PGijsbers:
Unfortunately, we missed a warning last patch, so it still won't close #1034 

edit: letting CI run since local unit tests are slow for some reason.

### DIFF
--- a/openml/tasks/functions.py
+++ b/openml/tasks/functions.py
@@ -347,8 +347,8 @@ def __list_tasks(  # noqa: PLR0912, C901
 # TODO(eddiebergman): Maybe since this isn't public api, we can make it keyword only?
 def get_tasks(
     task_ids: list[int],
-    download_data: bool = True,  # noqa: FBT001, FBT002
-    download_qualities: bool = True,  # noqa: FBT001, FBT002
+    download_data: bool | None = None,  # noqa: FBT001, FBT002
+    download_qualities: bool | None = None,  # noqa: FBT001, FBT002
 ) -> list[OpenMLTask]:
     """Download tasks.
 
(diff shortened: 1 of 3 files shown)


openml/tasks/functions.py
```diff
@@ -367,79 +367,65 @@ def get_tasks(
...
 
 
 @openml.utils.thread_safe_if_oslo_installed
 def get_task(
     task_id: int,
-    *dataset_args: Any,
-    download_splits: bool | None = None,
+    download_splits: bool = False,  # noqa: FBT001, FBT002
     **get_dataset_kwargs: Any,
 ) -> OpenMLTask:
```
eddiebergman, line 399: Not recommending to do anything but just pointing out this is technically a backwards breaking change, given the *args was removed.

eddiebergman (approved):

openml/tasks/functions.py
PGijsbers: While no deprecation warning was given, it was included in the docs:

> The args are only present for backwards compatibility and will be removed from version 0.15.0 onwards.

Since those `*args` can still be set ( through `**kwargs`) and it's pretty niche, I chose to let it slide and not add a more explicit deprecation cycle after discovering this.

PR #1355, opened 2024-10-03
Release/0.15.0

PGijsbers:
According to new release cycle, as I understand it:
 - merge this into develop
 - release on GitHub (adding tag to develop commit)
 - build and release on PyPI

### DIFF
--- a/doc/progress.rst
+++ b/doc/progress.rst
@@ -9,11 +9,22 @@ Changelog
 next
 ~~~~~~
 
+0.15.0
+~~~~~~
+
  * ADD #1335: Improve MinIO support.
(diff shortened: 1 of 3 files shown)


eddiebergman:
Nothing that stands out but I would ask more clarification on the release/workflow. Is it the case of `develop` is the main branch to make branches off of, and then main just exists to hold the currently released version seperate from develop?

eddiebergman (approved):

eddiebergman:
Ran into the same issue with Auto-sklearn. The benefits of the main/develop was being able to do hot-fixes while a bigger feature was being worked on for develop, but I havn't found it super helpful otherwise.

PR #1359, opened 2024-10-14
fix(config): Fix XDG_X_HOME env vars, add OPENML_CACHE_DIR env var

### LINKED ISSUE #1324: [Feature] Environment variable for openml cache
This could be useful in certain scenarios where you may have programmatic access before running something but during the execution of the python script, you do not have access to it.

I would recommend something like `"OPENML_CACHE_DIRECTORY"`?

eddiebergman:
Fixes two issues related to `XDG_CONFIG_DIR` and `XDG_CACHE_DIR`, in which we were not adding the `"openml"` suffix to the path, i.e. we had:

* `${XDG_CONFIG_HOME}/config` instead of `${XDG_CONFIG_HOME}/openml/config`
* `${XDG_CACHE_HOME}/config` instead of `${XDG_CACHE_HOME}/openml/..`

To keep things backwards compatible:
* For the first one, we use the new correct path if we do not detect that the old one exists. If the file does exist at the old path, we attempt to parse it. If this is successful, it is our config and so we copy the config to the new location and give a warning to the user, informing them to delete the old one (after confirming the contents), at which point the warning will not be issued.
* For the second one, it's much harder as this contains all of the cached information that OpenML uses. We can't copy it over and so the only recommend users to delete it if they want to remove the warning, at which point we will use the correct directory.

... and also:
* Closes #1324

You can now specify `OPENML_CACHE_DIR` as an environment variable

### DIFF
--- a/openml/config.py
+++ b/openml/config.py
@@ -119,7 +119,7 @@ def _resolve_default_cache_dir() -> Path:
         return Path("~", ".cache", "openml")
 
     # This is the proper XDG_CACHE_HOME directory, but
-    # we unfortunatly had a problem where we used XDG_CACHE_HOME/org,
+    # we unfortunately had a problem where we used XDG_CACHE_HOME/org,
     # we check heuristically if this old directory still exists and issue
     # a warning if it does. There's too much data to move to do this for the user.
 
--- a/openml/config.py
+++ b/openml/config.py
@@ -111,37 +111,37 @@ def _resolve_default_cache_dir() -> Path:
     if user_defined_cache_dir is not None:
         return Path(user_defined_cache_dir)
 
-    if platform.system().lower() == "linux":
-        xdg_cache_home = os.environ.get("XDG_CACHE_HOME")
-        if xdg_cache_home is None:
-            return Path("~", ".cache", "openml")
+    if platform.system().lower() != "linux":
+        return _user_path / ".openml"
 
-        # This is the proper XDG_CACHE_HOME directory, but
-        # we unfortunatly had a problem where we used XDG_CACHE_HOME/org,
-        # we check hueristically if this old directory still exists and issue
-        # a warning if it does. There's too much data to move to do this for the user.
+    xdg_cache_home = os.environ.get("XDG_CACHE_HOME")
+    if xdg_cache_home is None:
+        return Path("~", ".cache", "openml")
 
-        # The new cache directory exists
-        cache_dir = Path(xdg_cache_home) / "openml"
-        if cache_dir.exists():
-            return cache_dir
+    # This is the proper XDG_CACHE_HOME directory, but
+    # we unfortunatly had a problem where we used XDG_CACHE_HOME/org,
+    # we check heuristically if this old directory still exists and issue
+    # a warning if it does. There's too much data to move to do this for the user.
 
-        # The old cache directory *does not* exist
-        hueristic_dir_for_backwards_compat = Path(xdg_cache_home) / "org" / "openml"
-        if not hueristic_dir_for_backwards_compat.exists():
-            return cache_dir
+    # The new cache directory exists
+    cache_dir = Path(xdg_cache_home) / "openml"
+    if cache_dir.exists():
+        return cache_dir
 
-        root_dir_to_delete = Path(xdg_cache_home) / "org"
-        openml_logger.warning(
-            "An old cache directory was found at '%s'. This directory is no longer used by "
-            "OpenML-Python. To silence this warning you would need to delete the old cache "
-            "directory. The cached files will then be located in '%s'.",
-            root_dir_to_delete,
-            cache_dir,
-        )
-        return Path(xdg_cache_home)
+    # The old cache directory *does not* exist
+    heuristic_dir_for_backwards_compat = Path(xdg_cache_home) / "org" / "openml"
+    if not heuristic_dir_for_backwards_compat.exists():
+        return cache_dir
 
-    return _user_path / ".openml"
+    root_dir_to_delete = Path(xdg_cache_home) / "org"
+    openml_logger.warning(
+        "An old cache directory was found at '%s'. This directory is no longer used by "
+        "OpenML-Python. To silence this warning you would need to delete the old cache "
+        "directory. The cached files will then be located in '%s'.",
+        root_dir_to_delete,
+        cache_dir,
+    )
+    return Path(xdg_cache_home)
 
 
 _defaults: _Config = {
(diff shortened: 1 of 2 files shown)


tests/test_openml/test_config.py
```diff
@@ -29,15 +29,29 @@ def test_non_writable_home(self, log_handler_mock, warnings_mock):
         assert not log_handler_mock.call_args_list[0][1]["create_file_handler"]
         assert openml.config._root_cache_directory == Path(td) / "something-else"
 
-    @unittest.mock.patch("os.path.expanduser")
-    def test_XDG_directories_do_not_exist(self, expanduser_mock):
+    def test_XDG_directories_do_not_exist(self):
         with tempfile.TemporaryDirectory(dir=self.workdir) as td:
+            # Save previous state
+            _prev = os.environ.get("XDG_CONFIG_HOME")
```
PGijsbers, line 35: You shouldn't need to save this, right? `os.environ` shouldn't persist to the environment variables.

openml/config.py
```diff
@@ -101,14 +104,50 @@ def set_file_log_level(file_output_level: int) -> None:
 
 # Default values (see also https://github.com/openml/OpenML/wiki/Client-API-Standards)
 _user_path = Path("~").expanduser().absolute()
+
+
+def _resolve_default_cache_dir() -> Path:
+    user_defined_cache_dir = os.environ.get(OPENML_CACHE_DIR_ENV_VAR)
+    if user_defined_cache_dir is not None:
+        return Path(user_defined_cache_dir)
```
PGijsbers, line 112: 🌶️  Probably want to verify it's a directory, because at least at this point you know that the source is the environment variable so you can give more directed instructions to the user.

openml/config.py
```diff
@@ -101,14 +104,50 @@ def set_file_log_level(file_output_level: int) -> None:
...
+
+    if platform.system().lower() == "linux":
```
PGijsbers, line 114: 🚒 invert the statement to make it a guard and dedent the rest of the code.

openml/config.py
```diff
@@ -101,14 +104,50 @@ def set_file_log_level(file_output_level: int) -> None:
...
+        xdg_cache_home = os.environ.get("XDG_CACHE_HOME")
+        if xdg_cache_home is None:
+            return Path("~", ".cache", "openml")
+
+        # This is the proper XDG_CACHE_HOME directory, but
+        # we unfortunatly had a problem where we used XDG_CACHE_HOME/org,
+        # we check hueristically if this old directory still exists and issue
```
PGijsbers, line 121:
```suggestion
        # we check heuristically if this old directory still exists and issue
```
🫑

openml/config.py
```diff
@@ -101,14 +104,50 @@ def set_file_log_level(file_output_level: int) -> None:
...
+        # a warning if it does. There's too much data to move to do this for the user.
+
+        # The new cache directory exists
+        cache_dir = Path(xdg_cache_home) / "openml"
+        if cache_dir.exists():
+            return cache_dir
+
+        # The old cache directory *does not* exist
+        hueristic_dir_for_backwards_compat = Path(xdg_cache_home) / "org" / "openml"
+        if not hueristic_dir_for_backwards_compat.exists():
```
PGijsbers, line 131:
```suggestion
        heuristic_dir_for_backwards_compat = Path(xdg_cache_home) / "org" / "openml"
        if not heuristic_dir_for_backwards_compat.exists():
```

openml/config.py
```diff
@@ -218,11 +257,66 @@ def stop_using_configuration_for_example(cls) -> None:
...
 
+def _handle_xdg_config_home_backwards_compatibility(
+    xdg_home: str,
+) -> Path:
+    # NOTE(eddiebergman): A previous bug results in the config
+    # file being located at `${XDG_CONFIG_HOME}/config` instead
+    # of `${XDG_CONFIG_HOME}/openml/config`. As to maintain backwards
+    # compatibility, where users may already may have had a configuration,
+    # we copy it over an issue a warning until it's deleted.
+    # As a huerisitic to ensure that it's "our" config file, we try
```
PGijsbers, line 268:
```suggestion
    # As a heurisitic to ensure that it's "our" config file, we try
```

openml/config.py
```diff
@@ -218,11 +257,66 @@ def stop_using_configuration_for_example(cls) -> None:
...
+    # parse it first.
+    config_dir = Path(xdg_home) / "openml"
+    backwards_compat_config_file = config_dir / "config"
```
PGijsbers, line 271:
```suggestion
    backwards_compat_config_file = Path(xdg_home) / "config"
```

PGijsbers: 🌶️ Isn't this supposed to check xdg_home/config?

openml/config.py
```diff
@@ -218,11 +257,66 @@ def stop_using_configuration_for_example(cls) -> None:
...
+    if not backwards_compat_config_file.exists():
+        return config_dir
+
+    correct_config_location = config_dir / "config"
+    # If it errors, that's a good sign it's not ours and we can
+    # safely ignore it, jumping out of this block. This is a huerisitc
```
PGijsbers, line 277:
```suggestion
    # safely ignore it, jumping out of this block. This is a heurisitc
```
🧑‍🏫

PGijsbers (changes requested):
Thanks 🙏 for your 👆   contribution 🙇 
Unfortunately 👿 there are some issues 💥 that you probably want to have a look 👀 at.
Good luck 🤞

LennartPurucker:
Cool 💯

tests/test_openml/test_config.py
eddiebergman: Unfortunately it persists between functions calls. Technically I shouldn't need to save it but just wanted to reset the state to where it was exactly. I would leave this as is

openml/config.py
eddiebergman: Good point, will do

openml/config.py
eddiebergman: Good catch

openml/config.py
eddiebergman: :red_envelope:

openml/config.py
eddiebergman: :chile:

openml/config.py
```diff
@@ -101,14 +104,50 @@ def set_file_log_level(file_output_level: int) -> None:
...
+
+    if platform.system().lower() != "linux":
+        return _user_path / ".openml"
+
+    xdg_cache_home = os.environ.get("XDG_CACHE_HOME")
+    if xdg_cache_home is None:
+        return Path("~", ".cache", "openml")
+
+    # This is the proper XDG_CACHE_HOME directory, but
+    # we unfortunatly had a problem where we used XDG_CACHE_HOME/org,
```
PGijsbers, line 122:
```suggestion
    # we unfortunately had a problem where we used XDG_CACHE_HOME/org,
```
🔥 💥

tests/test_openml/test_config.py
PGijsbers: Then it should be good to create it as a context manager instead so that it resets it when an error happens in between the save and restore.

PGijsbers (approved):
🌈 🏆

PR #1362, opened 2024-10-15
fix: Avoid Random State and Other Test Bug

LennartPurucker:
See last failure of tests.


eddiebergman:
I had to kill the windows runner as it was running for over 40 minutes without stopping. This seems to happen occasionally, will re-run failed jobs now

eddiebergman (commented):
lgtm, assuming no more sporadic test failures show up

eddiebergman (approved):

PR #1363, opened 2024-10-15
fix: Strip Whitespace New PR

LennartPurucker:
Rel: #1136, #1125

@PGijsbers I still have not found a good way to fix the edge case where this change would require a cache reset.

I am also trying to figure out how to see if this breaks anything we do not know about (all tests pass).

## Cache Reset 

Relevant code:

```python
def _read_features(features_file: Path) -> dict[int, OpenMLDataFeature]:
    features_pickle_file = Path(_get_features_pickle_file(str(features_file)))
    try:
        with features_pickle_file.open("rb") as fh_binary:
            return pickle.load(fh_binary)  # type: ignore  # noqa: S301

    except:  # noqa: E722
        with Path(features_file).open("r", encoding="utf8") as fh:
            features_xml_string = fh.read()

        features = _parse_features_xml(features_xml_string)

        with features_pickle_file.open("wb") as fh_binary:
            pickle.dump(features, fh_binary)

        return features
```

Let us discuss this in person.

### DIFF
--- a/openml/_api_calls.py
+++ b/openml/_api_calls.py
@@ -460,7 +460,7 @@ def __parse_server_exception(
         raise OpenMLServerError(f"URI too long! ({url})")
 
     try:
-        server_exception = xmltodict.parse(response.text)
+        server_exception = xmltodict.parse(response.text, strip_whitespace=False)
     except xml.parsers.expat.ExpatError as e:
         raise e
     except Exception as e:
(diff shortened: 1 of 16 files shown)


LennartPurucker:
Closing this as a result of recent developments: https://github.com/openml/openml-python/issues/1125

LennartPurucker:
New: #1368

PR #1367, opened 2024-10-15
doc: make all examples use names instead of IDs as reference.

LennartPurucker:
See https://github.com/openml/openml-python/pull/1070

Closes https://github.com/openml/openml-python/pull/1070, https://github.com/openml/openml-python/issues/1066, https://github.com/openml/openml-python/issues/1063


eddiebergman (approved):
good to go

PR #1369, opened 2024-10-16
fix: workaround for git test workflow for Python 3.8 

LennartPurucker:
Our Python 3.8 tests kept failing due to temporary files not being deleted in our pytest setup.

This was because, for some unexplainable reason, oslo concurrency lock files were created after a pytest fixture closed. This pytext fixture was used in a parameterized pytest and even deleted the lock file itself. However, the lock files were re-created after the fixture finished. 

As this only affects our git workflow and only for Python 3.8, I added a skip to the test that causes this problem, as far as I know.

### DIFF
--- a/openml/testing.py
+++ b/openml/testing.py
@@ -115,7 +115,7 @@ class TestBase(unittest.TestCase):
         """Tear down the test"""
         os.chdir(self.cwd)
         try:
-            shutil.rmtree(self.workdir, ignore_errors=True)
+            shutil.rmtree(self.workdir)
         except PermissionError as e:
             if os.name != "nt":
                 # one of the files may still be used by another process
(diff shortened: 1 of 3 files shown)


LennartPurucker:
The previous point is still the root cause but the fix did not work. Thus, we now skip it in the workflow for 3.8.

openml/testing.py
```diff
@@ -115,7 +115,7 @@ def tearDown(self) -> None:
         """Tear down the test"""
         os.chdir(self.cwd)
         try:
-            shutil.rmtree(self.workdir, ignore_errors=True)
```
LennartPurucker, line 118: I introduced this in another PR to fix this issue but it did not help. Hence, I removed it here again.

eddiebergman (approved):
lit

PR #1372, opened 2024-10-16
refactor: Deprecate array formats and default to dataframe

eddiebergman:
WIP


SubhadityaMukherjee (approved):
Okay looks good to me now. Merge conflicts were fixed, target_names was renamed back to target since I could not find a reason to change it everywhere.

SubhadityaMukherjee:
I am not confident in trying to merge it to develop until the tests are up and running again.

LennartPurucker (commented):
Did one full pass, LGTM, now looking into tests

tests/test_datasets/test_dataset_functions.py
```diff
@@ -113,24 +113,27 @@ def test_tag_untag_dataset(self):
...
     def test_list_datasets_paginate(self):
         size = 10
         max = 100
         for i in range(0, max, size):
             datasets = openml.datasets.list_datasets(offset=i, size=size)
-            assert size == len(datasets)
-            self._check_datasets(datasets)
+            assert len(datasets) == size
+            assert len(datasets.columns) >= 2
+            # Maybe index?
```
PGijsbers: Make the `did` the index to the dataframe rows?

openml/setups/functions.py
```diff
@@ -148,94 +149,87 @@ def list_setups(  # noqa: PLR0913
...
-        - If 'dict' the output is a dict of dict
         - If 'dataframe' the output is a pandas DataFrame
+        - If 'object' the output is a dictionary of OpenMLSetup objects
 
     Returns
     -------
     dict or dataframe
     """
-    if output_format not in ["dataframe", "dict", "object"]:
+    if output_format not in ["dataframe", "object"]:
```
PGijsbers, line 159: Is there a reason we don't deprecate `'object'` here?

openml/utils.py
```diff
@@ -352,14 +316,15 @@ def _list_all(  # noqa: C901, PLR0912
             # check if the number of required results has been achieved
             # always do a 'bigger than' check,
             # in case of bugs to prevent infinite loops
-            if len(result) >= LIMIT:
+            n_recieved = sum(len(result) for result in results)
+            if n_recieved >= LIMIT:
```
PGijsbers:
```suggestion
            n_received = sum(len(result) for result in results)
            if n_received >= LIMIT:
```

openml/utils.py
```diff
@@ -352,14 +316,15 @@ def _list_all(  # noqa: C901, PLR0912
...
                 break
 
             # check if there are enough results to fulfill a batch
-            if LIMIT - len(result) < BATCH_SIZE_ORIG:
-                batch_size = LIMIT - len(result)
+            if LIMIT - n_recieved < BATCH_SIZE_ORIG:
+                batch_size = LIMIT - n_recieved
```
PGijsbers:
```suggestion
            if LIMIT - n_received < BATCH_SIZE_ORIG:
                batch_size = LIMIT - n_received
```

openml/setups/setup.py
```diff
@@ -7,6 +7,7 @@
 import openml.flows
 
 
+# TODO(eddiebergman): All functionality handled by dataclass
```
PGijsbers:
```suggestion
```

openml/setups/setup.py
```diff
@@ -57,6 +67,7 @@ def __repr__(self) -> str:
         return header + body
 
 
+# TODO(eddiebergman): All functionality handled by dataclass
```
PGijsbers:
```suggestion
```

openml/setups/setup.py
PGijsbers: We can open an issue for this if desired. But dataclasses don't handle type validation, have a different representation, so in the end you would have most of this anyway. Pydantic may be a nicer way to handle some of this, but then we would add an extra dependency.

PGijsbers (changes requested):
Pretty much fine to merge

tests/test_datasets/test_dataset_functions.py
LennartPurucker: I do not get this point, please elaborate.

openml/setups/functions.py
LennartPurucker: Not sure, seems from some older use case. Uniquely different from dict. I would keep it for now and maybe refactor it later.

PR #1373, opened 2024-10-16
fix: github workflows and pytest issue

LennartPurucker:
This fixes three things:

1. First, when pushing to develop, the CI will not fail if we do not need to update the gh docs.
2. Second, it rolls back a previous change to the workflow that did not fix an issue.
3. We fix the original issue by removing leftover dirs.


eddiebergman (approved):
The gpt enshitification has encroached further, plunging us further into the demise of deep dark despairs. The light has faded, shadow has gained and the sky weeps a woeful tear upon this dry earth. however... LGTM :rocket: !

PR #1374, opened 2024-10-17
fix/maint: avoid exit code (which kills the docs building)

LennartPurucker:
If you exit an example, it will exist building docs...


eddiebergman (approved):
`exit(0)`

PR #1378, opened 2024-10-18
test: pytest scope to be function as tests might change server

LennartPurucker:
To make sure that each test runs on the test server by default, we need to change the scope to "function," as another test might overwrite this otherwise.

@eddiebergman Was there a reasons to set this to session?

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -263,7 +263,7 @@ def as_robot() -> Iterator[None]:
     openml.config.set_retry_policy(policy, n_retries)
 
 
-@pytest.fixture(autouse=True, scope="session")
+@pytest.fixture(autouse=True, scope="function")
 def with_test_server():
     openml.config.start_using_configuration_for_example()
     yield


LennartPurucker:
So it seems there was a reason because the tests are now failing, but whatever the reason, it broke other tests. So we need to re-implement the way this works

eddiebergman:
I think I did session because when run in parallel with "function", where it will keep entering and exiting the fixture which sets and unsets the test server config, a global variable. This caused the tests to fail.

It's actually really hard to get around this since they all use the same config in parallel, where some tests want production and some want test server. In retro-spect, session doesn't make huge sense either, but it would essentially set the default to test server at the start which is most of them. Any production server tests however could reset this global.

PGijsbers:
re 2,5,6: basically just use a pytest marker and run one with "not production" and one with "production", and that is a fairly transparent way of dealing with this. That said, it's still not ideal but from the top of my head I am not aware of a good solution.

PGijsbers:
I am a bit confused about this. Previously (with unittest) this was never an issue, yet they still referred to the same global variable.

PGijsbers:
but tests pass for me locally and as far as I can tell the original issue this PR tries to address also doesn't occur when run on CI/CD: https://github.com/openml/openml-python/actions/runs/12274682764/job/34248109797?pr=1384

PGijsbers:
I still don't fully understand by what miracle it _does_ work though. Since all access is `openml.config.server` one really would expect changing this from anywhere should affect its usage everywhere else..

LennartPurucker:
Resolved in #1411 as far as we know

PR #1379, opened 2024-10-21
Docs/mkdoc

PGijsbers:
Work converting the Python docs to mkdocs. Left to do before it's "functional":

 - [ ] Make sure notebooks run
 - [ ] Add the test server warning back, preferably in a reusable way as was before.

Otherwise, most of the files have been converted.

That said, the documentation _really_ needs a rewrite. I refrained from changing any content as much as I could, with rare exceptions. But the main problems, to me, are:

 - Information is all over the place. A few cohesive examples with a proper ToC would be better (I think).
 - The formatting is bad.
 - There is a mix of styles. Some examples even talk about "exercises" (e.g., [here](http://openml.github.io/openml-python/main/examples/30_extended/datasets_tutorial.html#sphx-glr-examples-30-extended-datasets-tutorial-py)).
 - Some of the content is outdated.

How should we move forward with this PR? I don't mind rewriting (some of) this documentation, but I think it would be better to merge this at one point so it's easier to collaborate. WDYT? @eddiebergman @LennartPurucker


LennartPurucker:
IMO: merge first and focus on user stories/use cases afterward to refactor the documentation. That is, likely move all the old documentation to an "extended example" list and only keep a set of user stories / use cases in the new main documentation.

SubhadityaMukherjee (commented):
Merged all the changes, they seem correct and run correctly. Next I will test the notebooks/scripts to verify if everything runs.

SubhadityaMukherjee (approved):

SubhadityaMukherjee:
Looks like mkdocs-jupyter does not yet support admonitions. 
I added it in manually for now since it was only used in 3 places but was unable to find any other way of doing it. (aka it works, but is not reusable without copy pasting a block of text)

SubhadityaMukherjee:
Seems I forgot to update this. These examples dont work. Im not entirely sure how to fix them. They are from examples/


20_basic/
simple_flows_and_runs_tutorial : the flow seems to be wrong. `expected flow parameters: []
model parameters: ['bootstrap', 'ccp_alpha', 'class_weight', 'criterion', 'max_depth', 'max_features', 'max_leaf_nodes', 'max_samples', 'min_impurity_decrease', 'min_samples_leaf', 'min_samples_split', 'min_weight_fraction_leaf', 'monotonic_cst', 'n_estimators', 'n_jobs', 'oob_score', 'random_state', 'verbose', 'warm_start’]`

30_extended/
run_setup_tutorial : IndexError: positional indexers are out-of-bounds

flows_and_runs_tutorial : No reg-eye-state, or dataset 68 seems to be something else. 

study_tutorial: the flow seems to be wrong. ```ValueError: Parameters of the model do not match the parameters expected by the flow:
expected flow parameters: []
model parameters: ['bootstrap', 'ccp_alpha', 'class_weight', 'criterion', 'max_depth', 'max_features', 'max_leaf_nodes', 'max_samples', 'min_impurity_decrease', 'min_samples_leaf', 'min_samples_split', 'min_weight_fraction_leaf', 'monotonic_cst', 'n_estimators', 'n_jobs', 'oob_score', 'random_state', 'verbose', 'warm_start’]```

40_paper/
2018_kdd_rijn_example.py: This example is deprecated, remove the `if False` in this code to use it manually.

SubhadityaMukherjee:
Removed linkcheck for now since the one I picked apparently is not maintained. 
Also an option for versioning using "mike" which requires some extra work. - https://squidfunk.github.io/mkdocs-material/setup/setting-up-versioning/

SubhadityaMukherjee:
Ah @PGijsbers I think we will also need to point the gh-pages option in settings to a different page. I dont have access to it so I cant check if its the same as sphinx.
It should be Pages -> Deploy from branch : gh-pages -> root

PGijsbers:
> Ah @PGijsbers I think we will also need to point the gh-pages option in settings to a different page. I dont have access to it so I cant check if its the same as sphinx. It should be Pages -> Deploy from branch : gh-pages -> root

That's already set. Sorry for the slow reply. Next time just ask me directly if I don't respond :)

.github/workflows/docs.yaml
```diff
@@ -22,43 +22,21 @@ jobs:
...
-        git commit -am "$last_commit"
-        git diff --quiet @{u} HEAD || git push
+      - name: Make docs
+        run: |
+          mkdocs build
+      - name: Deploy to GitHub Pages
+        if: (contains(github.ref, 'develop') || contains(github.ref, 'main')) && github.event_name == 'push'
+        run: |
+          mkdocs gh-deploy --force
+        env:
```
PGijsbers: Does this take into account the branch? For example, if we publish version 0.16 and its documentation, to me it reads like our `develop` documentation will overwrite it as soon as we push to develop.

PGijsbers: I know I initially said we didn't need it but.. maybe it might be a good idea to look into `mike`? https://squidfunk.github.io/mkdocs-material/setup/setting-up-versioning/

it seems to handle the thing we'd want later and it would also solve this problem.
Then we can simply build docs when:
 - push to `develop` (as `dev`)
 - a release is made, i.e., a tag of format `v*`

eddiebergman: Some reference for what you might need from `mike` if you want versioned docs.
https://github.com/automl/amltk/blob/5221d80fb4fec92f2d6f2fc73fce4eb16139c09f/.github/workflows/release.yml#L90-L110

It's also good to be aware that all `mike` does for you is work on the `gh-pages` branch and so you can always go nuke it or manually edit it if testing goes wrong. Took me a while before realising that, but it means you can really test out the workflow locally before doing the annoying dance with GH actions to get it to work.

eddiebergman: The root of this branch is pretty instructive for what it's doing, i.e. with the symlink, the `index.html` pointig to `latest` and the `versions.json` it's using to store metadata
https://github.com/automl/amltk/tree/gh-pages

LennartPurucker: A wild Eddie appears!

eddiebergman: You can just use regular pokeball, no need for an ultra

SubhadityaMukherjee: *throws a master ball*

SubhadityaMukherjee:
all the tests pass in reality but the added files check fails, maybe mike adds them. Im looking into it

examples/30_extended/datasets_tutorial.py
```diff
@@ -1,24 +1,29 @@
...
 #
+<<<<<<< docs/mkdoc -- Incoming Change
+# * List datasets
+
+# %%
+datalist = openml.datasets.list_datasets(output_format="dataframe")
+=======
 # * List datasets and return a dataframe
 datalist = openml.datasets.list_datasets()
+>>>>>>> develop -- Current Change
```
LennartPurucker: Merge bug?

tests/files/misc/features_with_whitespaces.xml.pkl
LennartPurucker, line 1: What is this file?

.github/workflows/docs.yaml
eddiebergman: _*struggle, struggle*, pokémon Eddie breaks free_

uv.lock
LennartPurucker, line 1: this file is not needed

PR #1385, opened 2024-12-11
fix: fixed the documentation of split.py

### LINKED ISSUE #332: OpenMLSplit.__init__ not documented or tested
Any instantiation of ``OpenMLSplit`` is done via a *private function* in the code. That seems really weird. And the class is not really documented.

VHemanth45:
#### Reference Issue
Fixes #332

Basically I added the some documentation to understand the split.py for better understanding

visit the split.py file which is present in task folder

### DIFF
--- a/openml/tasks/split.py
+++ b/openml/tasks/split.py
@@ -1,64 +1,92 @@
 # License: BSD 3-Clause
-from __future__ import annotations
 
+from collections import namedtuple, OrderedDict
+import os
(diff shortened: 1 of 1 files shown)


LennartPurucker:
I created a new PR with changes to resolve this in #1415 

The PR here has a lot of changes not related to the problem, or adding documentation that is not needed (e.g., for private functions), or not really correct.

PR #1388, opened 2025-01-25
Hotfix/arff

PGijsbers:
This introduces a backdoor for disabling attempting to download parquet files through the `OPENML_SKIP_PQ` variable.
If `OPENML_SKIP_PQ` is set to `true` (case insensitive), then parquet files will not be downloaded in the `get_dataset` and `OpenMLDataset.get_data` calls.

The PR also fixes a bug where an error would be raised if the parquet file failed to download. The `str(_get_dataset_parquet_file(self))` would return `None` if it failed to download, but by converting it to a string the next check is always false and thus it would not fall back to arff.

The reason for using an environment variable:
 - It's a bit quicker to implement than a configuration option, and it makes it easier to turn it on for a single call.
 - Compared to adding function arguments, the environment variable (or configuration option) don't need changes to existing scripts. Making it easier for people to start using it now, and stop using it later. We can issue a warning if the environment variable remains set in a later release.

### DIFF
--- a/openml/datasets/dataset.py
+++ b/openml/datasets/dataset.py
@@ -18,6 +18,7 @@ import scipy.sparse
 import xmltodict
 
 from openml.base import OpenMLBase
+from openml.config import OPENML_SKIP_PARQUET_ENV_VAR
 from openml.exceptions import PyOpenMLError
 
 from .data_feature import OpenMLDataFeature
@@ -359,7 +360,7 @@ class OpenMLDataset(OpenMLBase):
         # import required here to avoid circular import.
         from .functions import _get_dataset_arff, _get_dataset_parquet
 
-        skip_parquet = os.environ.get("OPENML_SKIP_PQ", "false").casefold() == "true"
+        skip_parquet = os.environ.get(OPENML_SKIP_PARQUET_ENV_VAR, "false").casefold() == "true"
         if self._parquet_url is not None and not skip_parquet:
             parquet_file = _get_dataset_parquet(self)
             self.parquet_file = None if parquet_file is None else str(parquet_file)
--- a/openml/datasets/dataset.py
+++ b/openml/datasets/dataset.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 
 import gzip
 import logging
+import os
 import pickle
 import re
 import warnings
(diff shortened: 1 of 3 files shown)


openml/datasets/dataset.py
```diff
@@ -358,8 +359,10 @@ def _download_data(self) -> None:
         # import required here to avoid circular import.
         from .functions import _get_dataset_arff, _get_dataset_parquet
 
-        if self._parquet_url is not None:
-            self.parquet_file = str(_get_dataset_parquet(self))
+        skip_parquet = os.environ.get("OPENML_SKIP_PQ", "false").casefold() == "true"
```
josvandervelde, line 362: 1) Should a `OPENML_SKIP_PQ_ENV_VAR` be defined, similar to https://github.com/openml/openml-python/blob/develop/openml/config.py#L25, so that it's easier to see which env vars are used? Should it be documented somewhere?

2) For me, "OPENML_SKIP_PARQUET" would be slightly clearer.

openml/datasets/functions.py
```diff
@@ -560,20 +561,22 @@ def get_dataset(  # noqa: C901, PLR0912
         if download_qualities:
             qualities_file = _get_dataset_qualities_file(did_cache_dir, dataset_id)
 
-        if "oml:parquet_url" in description and download_data:
+        parquet_file = None
+        skip_parquet = os.environ.get("OPENML_SKIP_PQ", "false").casefold() == "true"
```
josvandervelde, line 565: Alternatively: `distutils.util.strtobool(os.environ.get("OPENML_SKIP_PQ", "false"))` so that 
"True values are y, yes, t, true, on and 1; false values are n, no, f, false, off and 0. Raises if val is anything else."

openml/datasets/dataset.py
PGijsbers: Agreed, done!

LennartPurucker (approved):
LGTM with comments from Jos

openml/datasets/functions.py
PGijsbers: `distutils` isn't 3.12+, I'm sure we can use something else, but in the interest of getting the patch out, I'll leave it as is for now.

Issue #1115, opened 2021-10-05
Default `output_format`  to  `"dataframe"`


PGijsbers:
I propose that the several functions that take the `ouput_format`  parameter and produce lists change their defaults to return dataframes instead of dictionaries (e.g. `list_datasets`).
I think originally we had loose integration with pandas and it was an optional dependency, but since it's a required dependency now I don't see a reason to provide `dicts`  as default. I'd argue most people prefer dataframes in interactive sessions, and it makes sense to minimize the amount of writing people have to do in those (also plenty of people might prefer dataframes in scripts too).

amastruserio:
I took a look at this to see how the proposed changes would behave and it appeared to lead to a few issues with some functions not fully integrated with dataframes yet. Many of these errors were Attribute errors centering around flow_id, task_id, and run_id, and appear to be the result of some functions expecting objects other than dataframes.
It also lead to repeated unit test failures involving Key_Errors, namely with flow_id, task_id, and uploader. 
There were also some assertion errors that wanted to assert some object was a dict, but I imagine these assertions would be irrelevant with dataframes as the default.
I have a complete list of unit tests that failed after implementing these changes that I can provide as well, but I think there is still more integration with dataframes needed from the codebase before being able to totally implement this change.

varshneydevansh:
Hi!

I am not sure how I ended up here, but I looked at what you guys are doing and It's pretty remarkable. I find this challenge to be my starter in this project and would love to contribute more. =)

PGijsbers:
Hi @varshneydevansh, thank you! We already started working on this, as can be seen in #1258. It's already assigned to me since it's pretty involved. However, if you would like to do this instead, here is basically what needs to be done:
The next step is to actually make the change and make `output_format='dataframe'` the default as well as removing `output_format='dict'` support wherever the message indicated it. You should be able to find all instances of this in `# TODO: [0.15]`. Then, make sure to update any unit tests necessary and update the documentation.

If you'd like something smaller, I could recommend you to work on #1142 or #1213 (the latter requires some Github Actions knowledge, but no knowledge of our code-base).

PGijsbers:
We missed this in the 0.15 release, adding it to 0.16

Roshangoli:
Hi! I’m new to open source and was excited to contribute to this issue as my first task, but I see it’s already being handled in PR #1390. Could I please be recommended a similar beginner-friendly issue to get started with? Thank you! 🙌

Omswastik-11:
[...]

PR #1390, opened 2025-02-06
defaults are set to dataframe as part of first step towards issue #1115

SubhadityaMukherjee:
#### Reference Issue
 First step towards issue #1115 as mentioned in the discussion. 

Here, only the defaults have been changed from either dict or array to dataframe.

Since only the defaults were changed. Might not need changes to tests. Although that remains to be seen.

#### Any other comments?
There are some issues with minio imports that might cause the test to fail. (Not part of changes done here)

### DIFF
--- a/openml/datasets/functions.py
+++ b/openml/datasets/functions.py
@@ -120,7 +120,7 @@ def list_datasets(
     size: int | None = None,
     status: str | None = None,
     tag: str | None = None,
-    output_format: Literal["dataframe", "dict"] = "dict",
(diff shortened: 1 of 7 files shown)


SubhadityaMukherjee:
Contrary to what the default github message says, this will not close the issue. This is step 1 out of 2 in removing the dict/array and only keeping dataframe.

PGijsbers:
Oops.. looks like #1372 forgot to mention #1115. Looking at the work done so far, we should build on #1372 instead. I don't believe @eddiebergman is currently planning contributions to OpenML as he is moving to a new job.

PR #1393, opened 2025-04-09
Change default server, fix bug writing config to file

PGijsbers:
#### Reference Issue

None, recent issues with proxy.

Sets the default url to use the `api` subdomain, which avoids a layer of indirection and subsequently should be stable.
It also fixes an issue with the CLI tool so that it's possible again to set variables through it.

Affects all endpoints. Not really that suitable for manual testing.

### DIFF
--- a/openml/config.py
+++ b/openml/config.py
@@ -148,7 +148,7 @@ def _resolve_default_cache_dir() -> Path:
 
 _defaults: _Config = {
     "apikey": "",
-    "server": "https://www.openml.org/api/v1/xml",
+    "server": "https://api.openml.org/api/v1/xml",
     "cachedir": _resolve_default_cache_dir(),
     "avoid_duplicate_runs": True,
     "retry_policy": "human",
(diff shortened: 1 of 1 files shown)


LennartPurucker (approved):
This looks correct to me and the code works with "https://api.openml.org/api/v1/xml" for me  bug free. 

I cannot see yet how it affects the CLI

openml/config.py
```diff
@@ -400,10 +402,9 @@ def set_field_in_config_file(field: str, value: Any) -> None:
             # There doesn't seem to be a way to avoid writing defaults to file with configparser,
             # because it is impossible to distinguish from an explicitly set value that matches
             # the default value, to one that was set to its default because it was omitted.
-            value = config.get("FAKE_SECTION", f)  # type: ignore
-            if f == field:
-                value = globals()[f]
-            fh.write(f"{f} = {value}\n")
+            value = globals()[f] if f == field else config.get(f)  # type: ignore
+            if value is not None:
+                fh.write(f"{f} = {value}\n")
```
PGijsbers, line 407: Sorry for not being clear about the CLI fix. That's this bit of code.. it's used only by the CLI (I think). I understand that's not particularly clear since the other edit is in the `cli.py` file which you might reasonably expect  to also have the fix that was mentioned.

Basically, if `f!=field` then it would write the name of the variable to itself (e.g., `apikey = apikey`), instead of fetching the corresponding value from the configuration and re-writing that.

openml/cli.py
```diff
@@ -110,7 +110,7 @@ def replace_shorthand(server: str) -> str:
         if server == "test":
             return "https://test.openml.org/api/v1/xml"
         if server == "production":
-            return "https://www.openml.org/api/v1/xml"
+            return "https://api.openml.org/api/v1/xml"
```
PGijsbers:
```suggestion
            return "https://www.openml.org/api/v1/xml"
```

openml/config.py
```diff
@@ -148,7 +148,7 @@ def _resolve_default_cache_dir() -> Path:
 
 _defaults: _Config = {
     "apikey": "",
-    "server": "https://www.openml.org/api/v1/xml",
+    "server": "https://api.openml.org/api/v1/xml",
```
PGijsbers:
```suggestion
    "server": "https://www.openml.org/api/v1/xml",
```

PR #1395, opened 2025-04-14
Tasks from sever incorrectly uses default estimation procedure ID

LennartPurucker:
I noticed that the tasks use the default estimation procedure ID and not the correct one.

I think this PR fixes this. Am still testing and thinking about it but wanted to share the thought already.


PGijsbers (approved):
Very good, here's a parrot 🦜

PR #1396, opened 2025-04-19
Feature/fix1296

### LINKED ISSUE #1296: [CI] Linting of docs
Currently there is no linter for doc and docstrings.

As a follow up, this should really be linted. I would advocate using [`ruff`](https://github.com/astral-sh/ruff) but theoretically that's blocked on still supporting python 3.6 (#1083).

Should work if working in an environment that's 3.7+ though.

The alternative is to use [`pydocstyle`](http://www.pydocstyle.org/en/stable/).

This would prevent issues like #716 from appearing. (which I guess can be closed now?)

samplecatalina:
#### Reference Issue
Fixes #1296 

Added GitHub Actions workflow to run `Ruff` linting in CI pipeline; added caching for `Ruff`, which will speed up the workflow by caching the `Ruff` cache directory between runs.

Non-regression tests for the GitHub Actions workflow has been added in .github/workflows/tests directory.

To run these tests:
1. Go to the "Actions" tab in the GitHub repository
2. Select "Test Lint Workflow"
3. Click "Run workflow"
4. Choose the test case you want to run from the dropdown

These tests will verify proper code formatting, docstring compliance, caching functionality, error detection, workflow configuration, and each test case creates temporary test files in a test_files directory to avoid interfering with the actual codebase.

#### Any other comments?
No.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,41 @@
+name: Lint
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  lint:
+    name: Run Ruff
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.8'
+          cache: 'pip'
+      
+      - name: Cache Ruff
+        uses: actions/cache@v3
+        with:
+          path: |
+            .ruff_cache
+            .ruff_cache.db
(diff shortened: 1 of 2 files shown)


PGijsbers (changes requested):
I know the original issue mentioned it being a workflow, but I think linting should be part of pre-commit instead. It's already where the general linting for updates to tests and source code works. But the pre-commit is currently not set up for docs. So we should simply update the pre-commit file so it lints docs also. Sorry for the double work.

LennartPurucker:
Closing this for now, as the PR itself is too far off from our vision (via pre-commit file).

PR #1398, opened 2025-04-20
fix#1126: Benchmarking Suites export to LaTeX

### LINKED ISSUE #1126: Support for Exporting Benchmarking Suites to LaTeX
The best way to export basic benchmarking suite information to LaTeX is current using pandas' [`to_latex`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.to_latex.html) function. Something like:

```python
import openml
suite = openml.study.get_suite(271)
tasks = [openml.tasks.get_task(tid, download_data=False, download_qualities=False) for tid in suite.tasks]
metadata = openml.datasets.list_datasets(data_id=[t.dataset_id for t in tasks], output_format="DataFrame")
metadata.style.to_latex("my-table.tex")
```

It would be nice if we can make this easily available. However, I don't think exporting latex exports natively is the right call. It will produce a lot of overhead (e.g., selecting columns, forwarding arguments to `to_latex` which may be deprecate in the future, etc.).
I think our primary mode of support for this feature should be to include an example in the documentation on how to achieve this (including some simple styling examples). However, I do think we could add the function (or lazy property) of `suite.metadata` which returns the dataframe that is generated above (with some added task information). This allows for quick generation of LaTeX tables on the one hand while minimizing the support we would need to give for output customization on the other.

 @mfeurer we discussed the functionality before, thoughts?

samplecatalina:
#### Reference Issue

Fixes #1126 

1. Added a to_latex method to the OpenMLDataset class that:
- Uses pandas 2.0's LaTeX export functionality
- Supports multi-page tables using the longtable package
- Allows customization of columns, caption, and label
- Can write output to a file or return as a string
- Makes jinja2 dependency optional (only needed if using custom templates)

2. Added comprehensive tests that verify:
- Basic LaTeX export functionality
- Column filtering
- Caption and label support
- File output functionality

3. Updated requirements to ensure pandas 2.0+ is available

A `to_to_latex()` has been added to tests/test_datasets/latex.py

#### Any other comments?
This function could be tested locally with this snippet below. Although it is not exactly benchmarking suites as in the discussion of issue #1126, it indicates this `to_latex` function is a working solution.

```python3
import openml

# Get a dataset
dataset = openml.datasets.get_dataset(1)  # Replace with your dataset ID

# Export to LaTeX
latex = dataset.to_latex(
    caption="Dataset Information",
    label="tab:dataset",
    columns=["Name", "Creator", "Description"]  # Optional: specify columns
)

# Or write directly to a file
dataset.to_latex(output_file="dataset.tex")
```


requirements.txt
```diff
@@ -0,0 +1 @@
+pandas>=2.0.0
```
PGijsbers, line 1: Requirements are specified in `pyproject.toml`

PGijsbers (commented):
Thanks for taking the time to contribute! Unfortunately, this PR does not address the linked issue. You added a function to create a latex table based on a dataset, whereas the issue was about a benchmarking suite. At this time, I do not think that we should provide support for latex tables for datasets -- such a feature hasn't ever been requested and I've also not seen any similar tables in papers.

I'll close this PR for that reason, but we would be happy to help you get started on another issue (even #1126) :)

PR #1401, opened 2025-06-16
maint: switch local install description to use uv

LennartPurucker:
Make onboarding easier by switching to uv for the install.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -181,15 +181,24 @@ following rules before you submit a pull request:
 However it is also possible to use the [openml-python docker image](https://github.com/openml/openml-python/blob/main/docker/readme.md) for testing and building documentation.
 This can be useful for one-off contributions or when you are experiencing installation issues.
 
-First install openml with its test dependencies by running
-  ```bash
-  $ pip install -e .[test]
-  ```
+First install Python 3.8 or higher, pip, and git.
+Then clone the repository:
+```bash
+pip install uv # Install uv via pip (or see https://docs.astral.sh/uv/getting-started/installation/)
+uv venv --seed --python 3.8 ~/.venvs/openml-python
+source ~/.venvs/openml-python/bin/activate
+pip install uv # Install uv within the virtual environment
+```
+
+Then install openml with its test dependencies by running
+```bash
+uv pip install -e .[test]
+```
 from the repository folder.
 Then configure pre-commit through
- ```bash
- $ pre-commit install
- ```
+```bash
+pre-commit install
+```
 This will install dependencies to run unit tests, as well as [pre-commit](https://pre-commit.com/).
 To run the unit tests, and check their code coverage, run:
   ```bash
@@ -247,7 +256,7 @@ information.
 
 For building the documentation, you will need to install a few additional dependencies:
 ```bash
-$ pip install -e .[examples,docs]
+$ uv pip install -e .[examples,docs]
 ```
 When dependencies are installed, run
 ```bash


CONTRIBUTING.md
```diff
@@ -181,15 +181,24 @@ following rules before you submit a pull request:
...
-  $ pip install -e .[test]
-  ```
+First install Python 3.8 or higher, pip, and git.
+Then clone the repository:
+```bash
+pip install uv # Install uv via pip (or see https://docs.astral.sh/uv/getting-started/installation/)
+uv venv --seed --python 3.8 ~/.venvs/openml-python
+source ~/.venvs/openml-python/bin/activate
+pip install uv # Install uv within the virtual environment
+```
```
PGijsbers: I would at least mention why we recommend `uv` and indicate that it's not a requirement and standard `pip/venv` will also work. Someone starting contributing to a new project already has information overload, so if they now have to figure out a new package manager we just make the problem worse (even if it's pretty much entirely compatible).

We could also use [Content Tabs](https://squidfunk.github.io/mkdocs-material/reference/content-tabs/) and default to `uv` instructions. Downside is that the plain markdown is cluttered for people not visiting the built documentation pages.

PGijsbers (changes requested):

CONTRIBUTING.md
LennartPurucker: Mhm, I think for anyone not familiar with pip, uv is less of an information overload (besides always working when following the setps above).

LennartPurucker: And we should reduce the option that ppl use different tools for dev

SubhadityaMukherjee: I second mentioning why it's better but not mandatory. I honestly think everyone who will contribute has atleast used pip once but probably not uv (even I dont always use it).

PGijsbers: I don't really understand how uv is less (or more) information than pip. They are the same commands? It's just that `uv` is not default and people can use whatever manager they like?

And while I agree that we should have one recommended way to do things (and clearly indicate that), it really shouldn't make a difference if people use their own thing? It may even uncover some issues early, as our users will also use different package managers.

LennartPurucker: Using uv provides less information overload than pip as its output, documentation, and usability are meaningful enough for newcomers to be able to understand it. 

I think the documentation here is way too verbose, and adding any more text does not add any useful information. I do not think we need to provide reasons here at all. 

I will do a full pass over the .md and streamline it. My general goal would be only to mention what a developer actually needs to know. No reason to provide verbose information that no one cares about or reads because it is too long.

SubhadityaMukherjee: My only problem with uv is the company that makes it is well, vc backed and Im not sure if we should just remove pip instructions since thats the _only_ thing thats default python and is unlikely to go away. 

We can always just add both instructions and the user can pick what they want. 
Yes, streamlining the README is a good idea anyway. But I dont think we should "decide" for the user based off recent trends.

LennartPurucker: I have changed the documentation a lot, so resolving this now. Let us talk about it more in person to come to a final decision.

LennartPurucker:
Closes #1409

LennartPurucker:
@PGijsbers @SubhadityaMukherjee I recommend reading the new documentation files as markdown in the branch instead of via the file differences.

CONTRIBUTING.md
```diff
@@ -15,66 +15,112 @@ in Python, [scikit-learn](http://scikit-learn.org/stable/index.html).
...
 
-General git workflow
---------------------
+## Contributing Workflow Overview 
+To contribute to the openml-python package, follow these steps:
+
+0. Determine how you want to contribute (see above).
+1. Set up your local development environment.
+   1. Fork and clone the `openml-python` repository. Then, create a new branch from the ``develop`` branch. If you are new to `git`, see our [detailed documentation](#basic-git-workflow), or rely on your favorite IDE.   
+   2. [Install the local dependencies](#install-local-dependencies) to run the tests your contribution.
```
PGijsbers:
```suggestion
   2. [Install the local dependencies](#install-local-dependencies) to run the tests for your contribution.
```

CONTRIBUTING.md
```diff
@@ -15,66 +15,112 @@ in Python, [scikit-learn](http://scikit-learn.org/stable/index.html).
...
+   3. [Test your installation](#testing-your-installation) to ensure everything is set up correctly.
+4. Implement your contribution. If contributing to the documentation, see [here](.
```
PGijsbers: link dropped

CONTRIBUTING.md
```diff
@@ -15,66 +15,112 @@ in Python, [scikit-learn](http://scikit-learn.org/stable/index.html).
...
+You may then run a specific module, test case, or unit test respectively:
+```bash
+pytest tests/test_datasets/test_dataset.py
+pytest tests/test_datasets/test_dataset.py::OpenMLDatasetTest
+pytest tests/test_datasets/test_dataset.py::OpenMLDatasetTest::test_get_data
+```
+
+To test your new contribution, add [unit tests](https://github.com/openml/openml-python/tree/develop/tests), and, if needed, [examples](https://github.com/openml/openml-python/tree/develop/examples) for any new functionality being introduced. Some notes on unit tests and examples:
+* If a unit test contains an upload to the test server, please ensure that it is followed by a file collection for deletion, to prevent the test server from bulking up. For example, `TestBase._mark_entity_for_removal('data', dataset.dataset_id)`, `TestBase._mark_entity_for_removal('flow', (flow.flow_id, flow.name))`.
+* Please ensure that the example is run on the test server by beginning with the call to `openml.config.start_using_configuration_for_example()`.
```
PGijsbers: This is actually default for all `TestBase` derived tests and also with an auto-use pytest fixture.

PGijsbers (changes requested):
Minor mistakes

CONTRIBUTING.md
```diff
@@ -15,66 +15,112 @@ in Python, [scikit-learn](http://scikit-learn.org/stable/index.html).
...
+4. Implement your contribution. If contributing to the documentation, see [here](.
+5. [Create a pull request](#pull-request-checklist). 
+
+### Install Local Dependencies
+
+We recommend following the instructions below to install all requirements locally.
+However, it is also possible to use the [openml-python docker image](https://github.com/openml/openml-python/blob/main/docker/readme.md) for testing and building documentation. Moreover, feel free to use any alternative package managers, such as `pip`.
+
+
+1. To ensure a smooth development experience, we recommend using the `uv` package manager. Thus, first install `uv`. If any Python version already exists on your system, follow the steps below, otherwise see [here](https://docs.astral.sh/uv/getting-started/installation/).
```
SubhadityaMukherjee, line 59: Why is it smooth for anyone who hasnt used uv? Maybe link to something that justifies using it would be a little nicer here I think. Guess probably something like [this](https://blog.kusho.ai/uv-pip-killer-or-yet-another-package-manager/) but more formal?

PR #1404, opened 2025-06-16
[CI] Try to fix race condition

LennartPurucker:
Ensure we have one process per file such that autouse fixture and TestBase class do not clash.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 <div id="user-content-toc">
   <ul align="center" style="list-style: none;">
     <summary>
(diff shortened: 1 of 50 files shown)


LennartPurucker:
Wait for #1372 to merge, rebase this, and try agian.

LennartPurucker:
Together with the fix regarding MP spawn in #1372, this should resolve https://github.com/openml/openml-python/pull/1378

LennartPurucker:
This does not seem to solve the problem. So we need to refactor the tests in a different way.

PR #1411, opened 2025-06-18
Automatically connect to production server based on mark

PGijsbers:
Idea on how to configure the usage of the server in one place.
~Problem is that it still doesn't really work with the existing `unittest` cases which also execute their own setup/teardown.
... or we also modify the setup/teardown to avoid setting server configuration. brb.~

### DIFF
--- a/tests/test_tasks/test_classification_task.py
+++ b/tests/test_tasks/test_classification_task.py
@@ -2,6 +2,7 @@
 from __future__ import annotations
 
 import pandas as pd
+import pytest
 
 from openml.tasks import TaskType, get_task
 
@@ -29,6 +30,7 @@ class OpenMLClassificationTaskTest(OpenMLSupervisedTaskTest):
         assert task.class_labels == ["tested_negative", "tested_positive"]
 
 
+@pytest.mark.server()
 def test_get_X_and_Y():
     task = get_task(119)
     X, Y = task.get_X_and_y()


.github/workflows/test.yml
```diff
@@ -96,14 +96,22 @@ jobs:
...
+        echo pytest -n 4 --durations=20 --dist load -sv $codecov -o log_cli=true -m "$marks"
+        pytest -n 4 --durations=20 --dist load -sv $codecov -o log_cli=true -m "$marks"
+    - name: Run tests on Ubuntu Production
+      if: matrix.os == 'ubuntu-latest'
+      run: |
+        if [ ${{ matrix.code-cov }} ]; then codecov='--cov=openml --long  --cov-report=xml'; fi
+        # Most of the time, running only the scikit-learn tests is sufficient
+        if [ ${{ matrix.sklearn-only }} = 'true' ]; then marks='sklearn and production'; else marks='production'; fi
+        echo pytest -n 4 --durations=20 --dist load -sv $codecov -o log_cli=true -m "$marks"
+        pytest -n 4 --durations=20 --dist load -sv $codecov -o log_cli=true -m "$marks"
```
PGijsbers, line 114: This really should be parallelized in the matrix, but the special jobs make this very verbose, so breaking it up into two workflows _may_ be better. In any case we could do it as a separate PR?

tests/test_tasks/test_classification_task.py
```diff
@@ -34,4 +26,14 @@ def test_download_task(self):
...
+
+
+def test_get_X_and_Y():
+    task = get_task(119)
+    X, Y = task.get_X_and_y()
+    assert X.shape == (768, 8)
+    assert isinstance(X, pd.DataFrame)
+    assert Y.shape == (768,)
+    assert isinstance(Y, pd.Series)
+    assert pd.api.types.is_categorical_dtype(Y)
```
PGijsbers, line 41: Has not much to do with the PR per se, was an artifact of experimenting with how/when fixtures are applied to `unittest` tests. Figured might as well leave it, one less test to move to pytest

PGijsbers:
@LennartPurucker 
https://www.youtube.com/watch?v=4XeWngFZaa8

tests/test_tasks/test_classification_task.py
```diff
@@ -34,4 +26,14 @@ def test_download_task(self):
 
     def test_class_labels(self):
         task = get_task(self.task_id)
-        assert task.class_labels == ["tested_negative", "tested_positive"]
\ No newline at end of file
+        assert task.class_labels == ["tested_negative", "tested_positive"]
+
+
+def test_get_X_and_Y():
```
LennartPurucker, line 34: Pytest.mark missing?

LennartPurucker (changes requested):
LGTM, minor

PR #1414, opened 2025-06-18
convert static_cache_dir and workdir to fixture and test to pytest

Taniya-Das:
#### Metadata
* Reference Issue: 
* New Tests Added: 
* Documentation Updated: 
* Change Log Entry:

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -291,29 +291,15 @@ def with_test_cache(test_files_directory, request):
         shutil.rmtree(tmp_cache)
         
         
-def find_test_files_dir(start_path: Path, max_levels: int = 1) -> Path:
-    """
-    Starting from start_path, climb up to max_levels parents looking for 'files' directory.
-    Returns the Path to the 'files' directory if found.
-    Raises FileNotFoundError if not found within max_levels parents.
-    """
-    current = start_path.resolve()
-    for _ in range(max_levels):
-        candidate = current / "files"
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -288,3 +289,31 @@ def with_test_cache(test_files_directory, request):
...
+        if candidate.is_dir():
+            return candidate
+        current = current.parent
+    raise FileNotFoundError(f"Cannot find 'files' directory within {max_levels} levels up from {start_path}")
+
+@pytest.fixture
+def static_cache_dir():
+
+    start_path = Path(__file__).parent
+    return find_test_files_dir(start_path)
```
LennartPurucker: Not sure we need this. 

I think we could just return `Path(__file__).parent / "files"` and this should always point to the `/tests/files` path.

tests/conftest.py
```diff
@@ -288,3 +289,31 @@ def with_test_cache(test_files_directory, request):
...
+
+@pytest.fixture
+def workdir(tmp_path):
+    original_cwd = os.getcwd()
```
LennartPurucker: Let's use pathlib here as well instead of os

tests/conftest.py
Taniya-Das: right. the find method is not needed.

LennartPurucker (approved):
LGTM

PR #1415, opened 2025-06-18
maint: docu for split object

LennartPurucker:
Add more documentation to the split object.

Rel.: #1385


LennartPurucker:
Closes #332

PR #1418, opened 2025-06-18
Maint/to pytest test dataset sparse dataset

Taniya-Das:
#### Metadata
* Reference Issue: 
* New Tests Added: 
* Documentation Updated: 
* Change Log Entry: 

#### Details 

The downloaded sparse data file is trimmed by removing some rows and columns to decrease the file size for testing.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -308,24 +308,28 @@ def workdir(tmp_path):
     yield tmp_path
     os.chdir(original_cwd)
 
+@pytest.fixture
+def mock_data_base_path(test_files_directory):
+    """Fixture to provide the base path for mock data files."""
+    return  test_files_directory / "mock_responses" 
+    
     
 @pytest.fixture
-def mock_sparse_categorical_395(requests_mock, test_files_directory):
-    base_path = test_files_directory / "mock_responses" / "datasets" / "sparse_categorical_395"
+def mock_sparse_categorical_395(requests_mock, mock_data_base_path):
     
-    description_file = base_path / "description.xml"
+    description_file = mock_data_base_path / "datasets" / "sparse_categorical_395" / "description.xml"
     requests_mock.get(
         "https://www.openml.org/api/v1/xml/data/395",
         text=description_file.read_text()
     )
     
-    data_file = base_path / "dataset.arff"
+    data_file = mock_data_base_path / "datasets" / "sparse_categorical_395" / "dataset.arff"
     requests_mock.get(
         "https://api.openml.org/data/v1/download/52507/re1.wc.sparse_arff",
         text=data_file.read_text()
     )
     
-    feature_file = base_path / "features.xml"
+    feature_file = mock_data_base_path / "datasets" / "sparse_categorical_395" / "features.xml"
     requests_mock.get(
         "https://www.openml.org/api/v1/xml/data/features/395",
         text=feature_file.read_text()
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -292,3 +292,54 @@ def with_test_cache(test_files_directory, request):
...
+def static_cache_dir():
+
+    return Path(__file__).parent / "files" 
+
+@pytest.fixture
+def workdir(tmp_path):
+    original_cwd = os.getcwd()
+    os.chdir(tmp_path)
+    yield tmp_path
+    os.chdir(original_cwd)
```
LennartPurucker: I think these changes are from a different PR, make sure to remove them and merge it into the branch once we merged the other one if needed

tests/test_datasets/test_dataset.py
```diff
@@ -337,65 +338,75 @@ def test_add_illegal_url_ontology(self):
...
-        openml.config.server = self.production_server
 
-        self.sparse_dataset = openml.datasets.get_dataset(4136, download_data=False)
-
-    def test_get_sparse_dataset_dataframe_with_target(self):
-        X, y, _, attribute_names = self.sparse_dataset.get_data(target="class")
-        assert isinstance(X, pd.DataFrame)
-        assert isinstance(X.dtypes[0], pd.SparseDtype)
-        assert X.shape == (600, 20000)
+@pytest.mark.production
```
LennartPurucker, line 345: If we mock the test, do we still need the mark here? I think we can remove it as long as we do not connect anymore to any server

tests/conftest.py
```diff
@@ -292,3 +292,54 @@ def with_test_cache(test_files_directory, request):
...
+    
+@pytest.fixture
+def mock_sparse_categorical_395(requests_mock, test_files_directory):
+    base_path = test_files_directory / "mock_responses" / "datasets" / "sparse_categorical_395"
+    
+    description_file = base_path / "description.xml"
+    requests_mock.get(
+        "https://www.openml.org/api/v1/xml/data/395",
```
LennartPurucker, line 322: maybe make the API base path a fixture as well

LennartPurucker (changes requested):
Minor comments

tests/conftest.py
Taniya-Das: ok, but it doesn't make a lot of difference. the generic base is just `test_files_directory / "mock_responses"`.

tests/test_datasets/test_dataset.py
Taniya-Das: By default it tries to connect to test server `https://test.openml.org/` otherwise. 
As it is just a mock, I can mock the files to test server. But this might not a be a good thing, as these datasets don't exist on test server.

PGijsbers: We technically also have the `@pytest.mark.server()` marker for things that actually connect to the server. So in that way it makes sense if we update this to be consistent, @pytest.mark.production just means a production configuration, and @pytest.mark.server means an actual network operation is performed (not everything is mocked).

PGijsbers: So what I am saying is that @pytest.mark.production() is needed for any production server configuration, even if it does not actually access the production (it's about how URLs are formed internally). Otherwise we would still end up with that race condition again. Either that or modify the URLs and the files to use the test server constants -- but that's not particularly clear either because the mocks are based on production data.

tests/test_datasets/test_dataset.py
```diff
@@ -339,65 +340,75 @@ def test_add_illegal_url_ontology(self):
...
+def test_get_sparse_categorical_data_id_395(mock_sparse_categorical_395):
+    
+    dataset = openml.datasets.get_dataset(395, download_data=True)
+    feature = dataset.features[3758]
+    assert isinstance(dataset, OpenMLDataset)
+    assert isinstance(feature, OpenMLDataFeature)
+    assert dataset.name == "re1.wc"
+    assert feature.name == "CLASS_LABEL"
+    assert feature.data_type == "nominal"
+    assert len(feature.nominal_values) == 25
```
PGijsbers, line 355: It looks like this is the only test which uses `mock_sparse_categorical_395`, is that correct?
If so, we can remove the data file from the repository, and remove `download_data=True` since it looks like we are only interested in accessing features.
And on that note, we could also remove most of the features xml file, and only keep the feature we are interested in analysing. Let me know if you have any questions about it.

PR #1431, opened 2025-06-20
Default to not checking for duplicates

PGijsbers:
#### Metadata
* Reference Issue: #1143
* New Tests Added: No
* Documentation Updated: Documentation is being rewritten in other PRs, will check back later.
* Change Log Entry: The `avoid_duplicate_runs` parameter of `openml.runs.run_model_on_task` and `openml.runs.run_flow_on_task` now defaults to `False` instead of `True` and this default is now configurable from your configuration file.

#### Details 
The main reason for doing this is so that it doesn't require an API key by default. This API key was necessary to check for duplicate runs, because that uses an authenticated flow endpoint (`flow/exists`).

### DIFF
--- a/openml/runs/functions.py
+++ b/openml/runs/functions.py
@@ -80,7 +80,7 @@ def run_model_on_task(  # noqa: PLR0913
     avoid_duplicate_runs : bool, optional (default=None)
         If True, the run will throw an error if the setup/task combination is already present on
         the server. This feature requires an internet connection.
-        If not set, it will use the default from your openml configuration.
+        If not set, it will use the default from your openml configuration (False if unset).
     flow_tags : List[str], optional (default=None)
         A list of tags that the flow should have at creation.
     seed: int, optional (default=None)
@@ -201,7 +201,7 @@ def run_flow_on_task(  # noqa: C901, PLR0912, PLR0915, PLR0913
     avoid_duplicate_runs : bool, optional (default=None)
         If True, the run will throw an error if the setup/task combination is already present on
         the server. This feature requires an internet connection.
-        If not set, it will use the default from your openml configuration.
+        If not set, it will use the default from your openml configuration (False if unset).
     flow_tags : List[str], optional (default=None)
         A list of tags that the flow should have at creation.
     seed: int, optional (default=None)
--- a/openml/runs/functions.py
+++ b/openml/runs/functions.py
@@ -59,7 +59,7 @@ ERROR_CODE = 512
 def run_model_on_task(  # noqa: PLR0913
     model: Any,
     task: int | str | OpenMLTask,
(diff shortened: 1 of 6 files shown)


tests/test_openml/test_config.py
```diff
@@ -175,13 +175,14 @@ def test_configuration_file_not_overwritten_on_load():
 
 def test_configuration_loads_booleans(tmp_path):
     config_file_content = "avoid_duplicate_runs=true\nshow_progress=false"
-    with (tmp_path / "config").open("w") as config_file:
+    tmp_file = tmp_path / "config"
+    with tmp_file.open("w") as config_file:
         config_file.write(config_file_content)
-        read_config = openml.config._parse_config(tmp_path)
+    read_config = openml.config._parse_config(tmp_file)
```
PGijsbers, line 181: This was passed a directory... which lead to a silent error and an empty config, which went unnoticed because the values below happened to coincide with the old defaults :D

openml/runs/functions.py
```diff
@@ -198,6 +201,7 @@ def run_flow_on_task(  # noqa: C901, PLR0912, PLR0915, PLR0913
     avoid_duplicate_runs : bool, optional (default=True)
```
PGijsbers:
```suggestion
    avoid_duplicate_runs : bool, optional (default=None)
```

openml/runs/functions.py
```diff
@@ -195,9 +198,10 @@ def run_flow_on_task(  # noqa: C901, PLR0912, PLR0915, PLR0913
         all supervised estimators of scikit learn follow this definition of a model.
     task : OpenMLTask
         Task to perform. This may be an OpenMLFlow instead if the first argument is an OpenMLTask.
-    avoid_duplicate_runs : bool, optional (default=True)
+    avoid_duplicate_runs : bool, optional (default=None)
         If True, the run will throw an error if the setup/task combination is already present on
         the server. This feature requires an internet connection.
+        If not set, it will use the default from your openml configuration.
```
PGijsbers:
```suggestion
        If not set, it will use the default from your openml configuration (False if unset).
```

openml/runs/functions.py
```diff
@@ -77,9 +77,10 @@ def run_model_on_task(  # noqa: PLR0913
     task : OpenMLTask or int or str
         Task to perform or Task id.
         This may be a model instead if the first argument is an OpenMLTask.
-    avoid_duplicate_runs : bool, optional (default=True)
+    avoid_duplicate_runs : bool, optional (default=None)
         If True, the run will throw an error if the setup/task combination is already present on
         the server. This feature requires an internet connection.
+        If not set, it will use the default from your openml configuration.
```
PGijsbers:
```suggestion
        If not set, it will use the default from your openml configuration (False if unset).
```

LennartPurucker (approved):
LGTM!

PR #1432, opened 2025-06-20
Big Rework of the Documentation

LennartPurucker:
#### Metadata
* Reference Issue: N/A
* New Tests Added: N/A
* Documentation Updated: Yes
* Change Log Entry: Reworked and streamlined all of the documentation.

#### Details 
See changes.


LennartPurucker:
@PGijsbers

PGijsbers:
Please fix the regex in `tests/test_openml/test_api_calls.py::test_authentication_endpoints_requiring_api_key_show_relevant_help_link`

docs/index.md
```diff
@@ -1,65 +1,79 @@
 # OpenML
```
PGijsbers, line 1: We should just re-use the top-level README instead of introducing duplication, right? A `docs/README.md` is automatically shown on GitHub if there is no README.md in the root.

docs/index.md
```diff
@@ -1,65 +1,79 @@
...
+You can install OpenML-Python with:
 
 ```bash
-python -m pip install openml
+pip install openml
 ```
 
 For more advanced installation information, please see the
-["Introduction"](../examples/20_basic/introduction_tutorial.py) example.
+["Introduction"](../examples/Basics/introduction_tutorial.py) example.
```
PGijsbers:
```suggestion
["Introduction"](../examples/Basics/introduction_tutorial) example.
```

PGijsbers: There are a number of warnings when serving the docs, including this broken link.

examples/introduction.py
```diff
@@ -0,0 +1,22 @@
+# %% [markdown]
```
PGijsbers, line 1: Is there a reason this is a Python file instead of markdown?

examples/Basics/introduction_tutorial.py
```diff
@@ -0,0 +1,58 @@
+# %% [markdown]
+# An example how to set up OpenML-Python followed up by a simple example.
+
+# %% [markdown]
+# ## Installation
+# Installation is done via ``pip``:
+#
+# ```bash
+# pip install openml
+# ```
```
PGijsbers:
```suggestion
# %% [markdown]
# ## Installation
# Installation is done via ``pip``:
#
# ```bash
# pip install openml
# ```
```

Addresses:
 - "followed up by a simple example" is no longer accurate.
 - the different cells introduce a lot of additional vertical whitespace

PGijsbers (commented):
I started reviewing but I wasn't entirely sure what to review. 
Are there particular pointers, e.g., pages/tutorials that changed significantly? Things I really need to see before we merge? There are a number of things I noted that could be improved, e.g., using more common mkdocs-material features, such as admonitions and tabs. Those could also be iterated over in future PRs, which might be an easier way for me to give feedback as well (I could just set up a PR with my suggested improvements).

Other notes:

 - The favicon is interestingly only applied to subpages only. I am guessing this is some caching artifact on my end, though even an in-cognito browser seems to exhibit this behavior.

LennartPurucker:
I guess you can review all the changes and documentation as much as you like as it is generally very outdaed and I mostly moved it or deleted stuff

docs/index.md
LennartPurucker: Mostly but not fully, here we can have more details.

 Maybe generally, be aware that I mostly took the state that already existed, which did not have a lot of native mkdocs structure and some sub-optimal code choices. I mostly refactored the content.

docs/index.md
LennartPurucker: When I build locally, it also throws warnings for non-broken links, so I do not know how to judge which one is correct

examples/introduction.py
LennartPurucker: It did not show up on the docs as .md. 

Looking into it, it is because with markdown this file is not copied into the docs folder from the pre-existing script, I think. (scripts/gen_ref_pages.py)

docs/index.md
LennartPurucker: I think this link is marked as broken, as it does not exist at the time the link checker comes in. The notebooks are only built later. It works locally

LennartPurucker:
Fun fact: the regrex test crashing only happened because of PyCharms refactor logic also changing one but not the other value. Without it, I would not have even known that the link to the documentation in the code is totally wrong already.

docs/index.md
PGijsbers: > here we can have more details

We can, but think it's fine not to in order to allow re-use and avoid desync.

docs/index.md
PGijsbers: It does for you? I tried this one locally and it did not work for me.

LennartPurucker: Yes (with the committed fix from you)

SubhadityaMukherjee:
Alright, I fixed some minor things and the rest looks good. I still need some more clarification on these : @LennartPurucker 

In Advanced/create_upload_tutorial.py, the title is “Dataset is a numpy array” instead of Creating and uploading datasets. I am unable to figure out how to change this :/

/Advanced/tasks_tutorial/ : resampling link leads to an empty search

Details : this doesnt seem like a good title :/ details of what exactly? - the page has information about config and docker and stuff

Openml.org/cite doesnt exist in contributing

LennartPurucker:
1. You need to remove this: https://github.com/openml/openml-python/blob/maint/rework_docu/examples/Advanced/create_upload_tutorial.py#L27 and likely other level one headings and `===`. All headings must be level two, I think, so at most `##`. 
2. I have no idea where the new WebUI shows the potential estimation procedures. I am always using the REST API to figure this out. Maybe someone else knows a new link for this?
3. I think the title of the page is already `Advanced User Guide`, no? And the nav button is just a summary / prompt to click on this for more in-depth notes. I would not go over a one-word value for this. But I also don't have any better ones in my mind right now.
4. Funny, I think it did during the Hackathon (?) I guess it is now: https://www.openml.org/terms

SubhadityaMukherjee:
Okay cool, Ill do that

SubhadityaMukherjee:
Okay its good to go from my side

LennartPurucker:
Did you update the citation link?

SubhadityaMukherjee:
Yep!

LennartPurucker:
Supi, LGTM from me, feel free to merge! 
(we need to check if it gets deployed correctly)

Issue #1252, opened 2023-06-12
Migrate all unittests to pytest


Taniya-Das:
[edited by @PGijsbers] Currently a lot of unit tests are written with the `unittest` module, but we want to use `pytest` instead. Generally speaking, this means extracting the test methods into pytest functions, optionally adding things like [pytest fixtures](https://docs.pytest.org/en/6.2.x/fixture.html) to replace setup/tear-down logic and using plain `assert` instead of the `unittest`-based assert methods. Below is an overview of which files have already been migrated:

openml-python/tests/
- [ ] test_datasets/
  - [ ] *.py  (added as we go)
- [ ] test_evaluations/
  - [ ] *.py   (added as we go)
- [ ] test_extensions/
  - [ ] *.py   (added as we go)
- [ ] test_flows/
  - [ ] *.py   (added as we go)
- [ ] test_openml/
  - [ ] *.py   (added as we go)
- [ ] test_runs/
  - [ ] *.py   (added as we go)
- [ ] test_setups/
  - [ ] *.py   (added as we go)
- [ ] test_study/
  - [ ] *.py   (added as we go)
- [ ] test_tasks/
  - [ ] *.py   (added as we go)
- [ ] test_utils/
  - [ ] *.py   (added as we go)

PGijsbers:
Working on test_utils: #1269

SubhadityaMukherjee:
This was closed in #1269

PGijsbers:
That's a slight misunderstanding. Only the tests for `test_utils` were converted to pytest style. The remainder of the submodules still have (many) unittest-style tests.

Omswastik-11:
Hi @PGijsbers and @Taniya-Das 👋
I am Om , a B.Tech undergrad from India . I have a past experiece of witting tests for libraries . I want to contribute to this issue . Is this issue is open ? Can you assign me ?

ravjot07:
Hi @PGijsbers  and @Taniya-Das , @naman9271  

I’d also like to contribute to this issue. I can pick up the tests that haven’t been worked on yet. I have prior experience writing and maintaining end-to-end and integration tests during my LFX project, so I’m comfortable handling test coverage.

PR #1439, opened 2025-11-14
Converting test_study to pytest

prady0t:
I've made some conversions to the `test_study_functions.py` file, and it now only uses pytest. List of changes made:

- Removed the unittest.TestCase class structure
- Converted setUp / tearDown into pytest fixtures
- Replaced helper methods with fixtures
- Removed all uses of `self`
- Replaced unittest.assert* with plain assert
- Removed test wrapper methods & replaced with parametrization
- Moved shared utilities into conftest.py

#### Metadata
* Reference Issue: #1252 
* New Tests Added: 
* Documentation Updated: 
* Change Log Entry:

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -33,7 +33,7 @@ import os
 import shutil
 from pathlib import Path
 import pytest
-# import openml_sklearn
+import openml_sklearn
 
 import openml
 from openml.testing import TestBase
--- a/tests/test_study/test_study_functions.py
+++ b/tests/test_study/test_study_functions.py
@@ -2,7 +2,6 @@
 from __future__ import annotations
 
 import pytest
-# import unittest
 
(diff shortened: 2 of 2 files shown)


tests/conftest.py
```diff
@@ -308,3 +311,45 @@ def workdir(tmp_path):
     os.chdir(tmp_path)
     yield tmp_path
     os.chdir(original_cwd)
+
+@pytest.fixture
+def test_env(tmp_path, request):
+    # locate static cache folder
+    abspath_this_file = Path(request.node.fspath).absolute()
+    static_dir = abspath_this_file.parent.parent  # same as n_levels = 1
```
PGijsbers:
```suggestion
    static_dir = abspath_this_file.parent.parent
```

PR #1446, opened 2025-11-15
feature: upload additional files in run object

### LINKED ISSUE #1332: Upload additional files in run object
#### Description
Implement upload of additional files  (such as model in binary format etc.) in the run object. 

#### Steps
To upload additional files, each file should be added as a tuple `("file_name", "file_content")` to `file_elements` in https://github.com/openml/openml-python/blob/develop/openml/base.py#L132.

This has been implemented as a patch in this commit -  https://github.com/openml/openml-tensorflow/commit/259f356ff9ef2ae96c47510e2ee06101afc021d2

#### Example XML with additional file
Example: https://www.openml.org/api/v1/xml/run/10594197

uddhavphatak:
#### Metadata
* Reference Issue: Fixes #1332
* New Tests Added: No
* Documentation Updated: No
* Change Log Entry: feature: upload additional files in run object

#### Details 

* ## What does this PR implement/fix? Explain your changes.

This PR adds first-class support for attaching arbitrary additional files (e.g., ONNX models, binary representations, logs, custom metadata) to an OpenMLRun object and having them uploaded to the OpenML server when the run is published.

The PR introduces:

A new internal storage container:
```
self._additional_files: dict[str, tuple[str, bytes]]
```

A public method:
```
run.add_file("filename.ext", file_content)
```

Updated _get_file_elements() so that these files are included in the upload payload sent via _perform_api_call().

This matches the API design discussed in issue #1332 and the extension use-case described in openml-tensorflow@259f356.

* ## Why is this change necessary? What is the problem it solves?

OpenML extensions (e.g., openml-tensorflow, ONNX flow exporters, model provenance modules) need a standard mechanism to attach arbitrary files to a run.

Until now:

- Only predictions and trace files were supported.

-There was no generic user-facing API to upload additional artifacts.

-Extensions had to patch _get_file_elements() or depend on private methods.

This PR introduces a clean, extensible, documented mechanism for extensions and users to attach files to a run without modifying low-level upload code.

### DIFF
--- a/openml/runs/run.py
+++ b/openml/runs/run.py
@@ -6,11 +6,11 @@ import time
 from collections import OrderedDict
 from pathlib import Path
 from typing import (
+    IO,
     TYPE_CHECKING,
     Any,
     Callable,
     Sequence,
-    IO,
 )
 
 import arff


uddhavphatak:
Hi, This is my first PR to this repo, should I write test cases as well to support this enhancement addition?

PR #1484, opened 2025-11-18
Fix: Use UserWarning for unauthenticated duplicate run checks (Fixes #1210)

### LINKED ISSUE #1210: Add warning to run_model_on_task with avoid duplicates if no authentication

Prachi194agrawal:
### Description
This PR fixes #1210 by improving the warning message when `avoid_duplicate_runs=True` is used without an API key.

**Changes:**
- Replaced the generic `RuntimeWarning` with a `UserWarning`.
- Updated the warning message to explicitly inform the user that duplicate runs cannot be checked server-side without authentication.
- Changed the condition check to explicitly look for `openml.config.apikey is None`.

### Type of change
- [x] Bug fix (non-breaking change which fixes an issue)

### How Has This Been Tested?
I verified this locally by mocking `openml.runs.run_model_on_task` with `apikey=None` and asserting that the new `UserWarning` is triggered correctly.

### DIFF
--- a/openml/runs/functions.py
+++ b/openml/runs/functions.py
@@ -111,7 +111,8 @@ def run_model_on_task(  # noqa: PLR0913
         warnings.warn(
             "The 'avoid_duplicate_runs' parameter is set to True, but no API key is configured. "
             "Duplicate runs cannot be checked server-side without authentication. "
-            "Please set your API key (see http://openml.github.io/openml-python/latest/examples/Basics/introduction_tutorial/). "
+            "Please set your API key "
+            "http://openml.github.io/openml-python/latest/examples/Basics/introduction_tutorial/."
             "The run will proceed, but duplicates may be created.",
             UserWarning,
             stacklevel=2,


Prachi194agrawal:
Hi @PGijsbers,

I have submitted this fix for #1210. I updated the warning to a UserWarning and added the specific check for the missing API key as discussed.

Could you please review this when you have a moment? It looks like the GitHub workflows (CI) are waiting for approval to run as well.

Thanks!

openml/runs/functions.py
```diff
@@ -107,13 +107,12 @@ def run_model_on_task(  # noqa: PLR0913
...
+    if avoid_duplicate_runs and openml.config.apikey is None:
         warnings.warn(
-            "avoid_duplicate_runs is set to True, but no API key is set. "
-            "Please set your API key in the OpenML configuration file, see"
-            "https://openml.github.io/openml-python/main/examples/20_basic/introduction_tutorial"
-            ".html#authentication for more information on authentication.",
-            RuntimeWarning,
+            "The 'avoid_duplicate_runs' parameter is set to True, but no API key is configured. "
+            "Duplicate runs cannot be checked server-side without authentication. "
+            "The run will proceed, but duplicates may be created.",
```
PGijsbers, line 116: Please keep the reference to the documentation on how users can acquire and configure a REST API key (now at http://openml.github.io/openml-python/latest/examples/Basics/introduction_tutorial/).

PGijsbers (changes requested):

openml/runs/functions.py
Prachi194agrawal: "Thanks! I've updated the warning message to include the link to the documentation as requested."

Issue #1486, opened 2025-11-18
Enhancement Add CLI commands for browsing and searching OpenML flows (models)


Jitmisra:
#### Description

Currently, users can only browse and search OpenML flows (models) through the Python API...... This requires writing Python code even for simple tasks like listing available models or searching for a specific model.... adding cli commands would significantly improve the user experience and make the model catalogue more accessible.

#### Proposed Solution

Add three new CLI subcommands under `openml models`:

1. `openml models list` - List flows with filtering options (tag, uploader, pagination)
2. `openml models info <flow_id>` - Display detailed information about a specific flow
3. `openml models search <query>` - Search flows by name

#### Expected Results

Users should be able to:
- Browse available models from the command line without writing Python code
- Search for models by name
- Get detailed information about specific models
- Filter models by tags, uploader etc.......

#### Use Cases

# List first 10 models
openml models list --size 10

# Search for RandomForest models
openml models search RandomForest

# Get detailed info about a model
openml models info 12345

# List models with a specific tag
openml models list --tag sklearn#### Additional Context

This would complement the existing `openml configure` CLI command and improve accessibility of the OpenML model catalogue. The implementation would use existing `openml.flows.list_flows()` and `openml.flows.get_flow()` functions, so no changes to core API are needed.

Jitmisra:
if you have any feedback let me know

KanishqGandharv219:
I have now also implemented CLI commands for Tasks in PR #1693, providing a complete solution for browsing OpenML flows, datasets, and tasks. All commands include full filtering and integration tests.

PR #1487, opened 2025-11-18
feat: Add CLI commands for browsing and searching OpenML flows (models)

Jitmisra:
#### Metadata

* Reference Issue :https://github.com/openml/openml-python/issues/1486

* New Tests Added: Yes

* Documentation Updated: No (CLI help text serves as documentation)

* Change Log Entry: "Add CLI commands for browsing and searching OpenML flows (models): `openml models list`, `openml models info`, and `openml models search`"

#### Details 

**What does this PR implement/fix?**

This PR adds three new CLI subcommands under `openml models` to improve the user experience of the model catalogue:

1. `openml models list` - List flows (models) with optional filtering (tag, uploader, pagination, output format)
2. `openml models info <flow_id>` - Display detailed information about a specific flow
3. `openml models search <query>` - Search flows by name with case-insensitive matching

**Why is this change necessary? What is the problem it solves?**

Currently, users must write Python code to browse or search OpenML flows, even for simple tasks like listing available models or finding a specific model. This creates a barrier to entry and makes the model catalogue less accessible. Adding CLI commands allows users to interact with the model catalogue directly from the command line without writing code.

This directly addresses the ESoC 2025 goal of "Improving user experience of the model catalogue in AIoD and openML".

**How can I reproduce the issue this PR is solving and its solution?**

**Before (requires Python code):**
import openml
flows = openml.flows.list_flows(size=10)
for _, row in flows.iterrows():
    print(row['name'])

**After (CLI commands):**
# List first 10 flows
openml models list --size 10

# Search for RandomForest models
openml models search RandomForest

# Get detailed info about a model
openml models info 12345

# List models with a specific tag
openml models list --tag sklearn --format table --verbose**Implementation Details:**

- Added three new functions in `openml/cli.py`: `models_list()`, `models_info()`, and `models_search()`
- Integrated into main CLI parser with proper argument handling
- Added comprehensive test suite (6 test cases) in `tests/test_openml/test_cli.py`
- Uses existing `openml.flows.list_flows()` and `openml.flows.get_flow()` functions - no changes to core API
- Follows existing CLI patterns (similar to `configure` command)
- All tests use mocked API calls to avoid requiring server connections

**Any other comments?**

- All pre-commit hooks pass (ruff, mypy, formatting)
- No breaking changes
- Follows project code style and patterns
- Ready for review

### DIFF
--- a/openml/cli.py
+++ b/openml/cli.py
@@ -1,4 +1,4 @@
-"""Command Line Interface for `openml` to configure its settings."""
+"""Command Line Interface for `openml` to configure settings and browse resources."""
 
 from __future__ import annotations
 
@@ -9,6 +9,7 @@ from pathlib import Path
 from typing import Callable
 from urllib.parse import urlparse
 
+import openml
 from openml import config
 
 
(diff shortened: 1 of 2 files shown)


joaquinvanschoren:
Thanks! Can you please update it to use the name 'flow' for now? We might introduce a new concept 'model' that is somewhat different from a flow, and we should avoid any confusion here.

Also, would it be possible to generalize this to other entities, e.g. datasets?  

So far, we were only using this cli for managing configurations. It makes sense to add a CLI, but longer term it would be better to have this as a seperate repo, so that people don't have to install openml-python and developers don't have to run the full openml-python test suite. It would also keep the openml-python repo more focussed.

For now, it would be ok to merge it here (after updating the naming) until it is more complete (with other entities like datasets) and then spin it out as a separate library.

Jitmisra:
> Thanks! Can you please update it to use the name 'flow' for now? We might introduce a new concept 'model' that is somewhat different from a flow, and we should avoid any confusion here.
> 
> Also, would it be possible to generalize this to other entities, e.g. datasets?
> 
> So far, we were only using this cli for managing configurations. It makes sense to add a CLI, but longer term it would be better to have this as a seperate repo, so that people don't have to install openml-python and developers don't have to run the full openml-python test suite. It would also keep the openml-python repo more focussed.
> 
> For now, it would be ok to merge it here (after updating the naming) until it is more complete (with other entities like datasets) and then spin it out as a separate library.

Thanks for the review!...... I have updated the CLI to use “flow” terminology everywhere and added a flows namespace..... I also added a datasets namespace (list/info/search) so the CLI is extensible to other entities. openml/cli.py now exposes openml flows ... and openml datasets ... and the parser is organized so more entities can be added later......
i would be happy if you have any other things in mind on which i can work upon.....

KanishqGandharv219:
Hi @Jitmisra ! I saw this PR has been stale for a few months with failing CI checks. Since I needed these features, I've opened a fresh PR (#1693) that implements both the flows and datasets CLI commands and passes all current CI checks. Thank you for the initial work outlining the approach!

Issue #1497, opened 2025-11-23
feat: add progress tracking for suite benchmarking


anujagrawal699:
## Problem
When benchmarking models across OpenML suites (e.g., OpenML-CC18 with multiple tasks), users have no visibility into:
- How many tasks have completed
- How long each task takes
- Estimated time remaining
- Whether the process is still running or has stalled

## Proposed Solution
Add a `run_suite_with_progress()` function that wraps suite benchmarking with a progress bar using tqdm.

## Benefits
- Better UX: Users can monitor long-running benchmarks
- Debugging: Easy to identify slow/stuck tasks
- Planning: ETA helps users schedule experiments
- AI-on-Demand Platform: Improves platform usability for researchers

anujagrawal699:
> Run this and try again. [24f7f45cc921.zip](https://github.com/user-attachments/files/23697794/24f7f45cc921.zip)

Is GitHub a right place to performing hacking?

Aymuos22:
Hi @anujagrawal699 ,
Can you assign this to me??

PGijsbers:
Sorry for the slow response. I think at this time we are not looking to increase the surface area of the API for features which are easily taken care of by a user (such as writing a loop with tqdm progress bar). If we were to do something like this in the future, I think it would be more likely that we would consider expanding run_model_on_task to take a task list as opposed to adding a new separate function. 

I am closing this issue for now. For users that want to see a feature like this, please indicate by "upvoting" the original proposal with something like a 👍

PR #1498, opened 2025-11-23
feat(suite benchmarking): add progress tracking

anujagrawal380:
### Description
Adds `run_suite_with_progress()` function to provide real-time progress feedback when benchmarking models across OpenML suites.

### Motivation
When running benchmarks on large suites (e.g., OpenML-CC18), users currently have no visibility into progress, making it difficult to:
- Monitor long-running experiments
- Estimate completion time
- Debug stuck/slow tasks

This is particularly important for the AI-on-Demand Platform where researchers run multi-hour benchmarks.

### Related Issue
#1497 

### Changes
- New function: `openml.study.run_suite_with_progress()`
- Progress bar with tqdm showing completion, ETA, and speed
- Full backwards compatibility (no breaking changes)
- Comprehensive tests (3 test cases)
- Tutorial with examples

### Example Usage

```python
from openml.study import run_suite_with_progress

clf = RandomForestClassifier(n_estimators=10, random_state=42)
runs = run_suite_with_progress(suite_id=334, model=clf)
```

### Output
```python
Benchmarking OpenML-Tiny: 100%|████████████| 7/7 [01:35<00:00, 13.6s/task]
```

### DIFF
--- a/openml/study/__init__.py
+++ b/openml/study/__init__.py
@@ -20,8 +20,8 @@ from .functions import (
 from .study import OpenMLBenchmarkSuite, OpenMLStudy
 
 __all__ = [
-    "OpenMLStudy",
     "OpenMLBenchmarkSuite",
+    "OpenMLStudy",
     "attach_to_study",
     "attach_to_suite",
     "create_benchmark_suite",
@@ -35,6 +35,6 @@ __all__ = [
     "list_studies",
     "list_suites",
     "run_suite_with_progress",
-    "update_suite_status",
     "update_study_status",
+    "update_suite_status",
 ]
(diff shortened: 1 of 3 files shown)


PGijsbers:
Sorry for the slow response. I think at this time we are not looking to increase the surface area of the API for features which are easily taken care of by a user (such as writing a loop with tqdm progress bar). If we were to do something like this in the future, I think it would be more likely that we would consider expanding `run_model_on_task` to take a task list as opposed to adding a new separate function.

But for now I will close this PR, but thanks for your effort.
If you are still looking to contribute, we are starting to have some more availability to engage in discussions on the issues, which is advised before moving on the create a PR.

PR #1513, opened 2025-11-25
Modernize _delete_entity with Literal typing and f-string formatting

axif0:
Modernizes _to_dict by replacing temporary variables with walrus operator assignments and removing outdated Python 3.7 comments.

- Replaced content = getattr(...) + if with walrus operator for brevity.
- Removed legacy TODO notes about dropping Python 3.7.
- Improved readability and reduced repetition in both simple and complex property handling.


PGijsbers (approved):
Thank you! We are currently doing server maintenance so CI/CD might be a bit flaky. Things should stabilize by next week, but we might not be able to merge it before then.

PR #1551, opened 2025-12-23
[ENH] Top-Level Convenience Aliases for Common Operations

Omswastik-11:
## Redundant syntax for getters

In getters, syntax is repeated and redundant, mainly through
the submodule having to be imported or addressed.

```python
import openml

# List all datasets and their properties
openml.datasets.list_datasets(output_format="dataframe")

# Get dataset by ID
dataset = openml.datasets.get_dataset(61)

# Get dataset by name
dataset = openml.datasets.get_dataset('Fashion-MNIST')

# This is similar for flows, runs, studies, such as

study = openml.study.get_study(42)
flow = openml.flows.get_flow(42)

```

## API

```python
import openml

# List all datasets with minimal friction
datasets_df = openml.list_datasets(output_format="dataframe")

# Get dataset by ID
dataset = openml.get_dataset(61)

# Get dataset by name
dataset = openml.get_dataset("Fashion-MNIST")

# Similarly for other types (no submodule import needed)
task = openml.get_task(31)
flow = openml.get_flow(10)
run = openml.get_run(20)

tasks_df = openml.list_tasks(task_type=TaskType.SUPERVISED_CLASSIFICATION)
flows_df = openml.list_flows(size=5)
runs_df = openml.list_runs(size=10)

```

## Implementation Details

* Added **re-exports in `__init__.py`** for:

  * `list_datasets`, `get_dataset`
  * `list_flows`, `get_flow`
  * `list_tasks`, `get_task`
  * `list_runs`, `get_run`

* Functions are imported directly from their respective submodules and exposed via `__all__`.

* All existing submodule-level imports remain fully functional.

* No changes to the underlying implementations — these are **pure convenience aliases**.

* Added unit tests to confirm that the aliases reference the correct underlying functions.

PR #1552, opened 2025-12-23
[ENH] Simplified Unified Get/List API

Omswastik-11:
## Redundant syntax for getters

In getters, syntax is repeated and redundant, mainly through
the submodule having to be imported or addressed.

```python
import openml

# List all datasets and their properties
openml.datasets.list_datasets(output_format="dataframe")

# Get dataset by ID
dataset = openml.datasets.get_dataset(61)

# Get dataset by name
dataset = openml.datasets.get_dataset('Fashion-MNIST')

# This is similar for flows, runs, studies, such as

study = openml.studies.get_study(42)
flow = openml.flows.get_flows(42)

```

## API implementation
```python
import openml

# List all datasets
datasets_df = openml.list_all("dataset", output_format="dataframe")

# Get dataset by ID
dataset = openml.get("dataset", 61)

# Get dataset by name
dataset = openml.get("dataset", "Fashion-MNIST")

# Get task
task = openml.get("task", 31)

# Get flow
flow = openml.get("flow", 10)

# Get run
run = openml.get("run", 20)

# Shortcut: infer dataset from name when no type specified
dataset = openml.get("Fashion-MNIST")

```

## Implementation Details

* Added **`openml.list_all(object_type: str, **kwargs) -> Any`**, a dispatcher that forwards to:

  * `list_datasets`
  * `list_tasks`
  * `list_flows`
  * `list_runs`

* Added **`openml.get(object_type_or_name, identifier=None, **kwargs) -> Any`**, a unified getter with support for:

  * **Type-based lookup**

    ```python
    openml.get("dataset", 61)
    openml.get("dataset", "dataset_name")
    ```
  * **Name-only shortcut for datasets**

    ```python
    openml.get("Fashion-MNIST")
    ```

* Exported both functions via `__all__` and documented them with docstrings.

* Preserved full backward compatibility:

  * Existing submodule APIs (e.g., `openml.datasets.get_dataset`) remain unchanged.

* Added unit tests to validate dispatcher behavior without requiring network access.


openml/__init__.py
```diff
@@ -49,12 +52,74 @@
...
+    ``list_*`` functions. Existing imports remain available for backward compatibility.
+    """
+    dispatch: ListDispatcher = {
+        "dataset": datasets.functions.list_datasets,
+        "task": tasks.functions.list_tasks,
+        "flow": flows.functions.list_flows,
+        "run": runs.functions.list_runs,
+    }
+
+    try:
```
fkiraly: you should really stop abusing try/except for case distinctions.
This is not good style, since you cannot distinguish actual exceptions from the try-block with the intended exception.

Instead, use if/else with a precise condition. In this case, you can also:
* use `dict.get`, and then check if `None` was retrieved.
* do an input check on `object_type`

openml/__init__.py
```diff
@@ -49,12 +52,74 @@
...
+        func = dispatch[object_type.lower()]
+    except KeyError as exc:  # pragma: no cover - defensive branch
+        raise ValueError(
+            "Unsupported object_type for list; expected one of 'dataset', 'task', 'flow', 'run'.",
+        ) from exc
+
+    return func(**kwargs)
+
+
+def get(object_type_or_name: Any, identifier: Any | None = None, /, **kwargs: Any) -> Any:
```
fkiraly: the first arg can be two different things, I would avoid that - instead, I would do one of two things:

* use `*`, the argument syntax
* make the `identifier` first, and `object_type` second

openml/__init__.py
```diff
@@ -49,12 +52,74 @@
...
+
+    object_type = str(object_type_or_name).lower()
+    dispatch: GetDispatcher = {
+        "dataset": datasets.functions.get_dataset,
+        "task": tasks.functions.get_task,
+        "flow": flows.functions.get_flow,
+        "run": runs.functions.get_run,
+    }
+
+    try:
```
fkiraly: again a try/except that you should avoid.

openml/__init__.py
```diff
@@ -49,12 +52,74 @@
...
+
+    return func(identifier, **kwargs)
+
 
 def populate_cache(
-    task_ids: list[int] | None = None,
-    dataset_ids: list[int | str] | None = None,
-    flow_ids: list[int] | None = None,
-    run_ids: list[int] | None = None,
+    task_ids: builtins.list[int] | None = None,
```
fkiraly: why are you changing this?

openml/__init__.py
```diff
@@ -49,12 +52,74 @@
     OpenMLTask,
 )
 
+ListDispatcher = Dict[str, Callable[..., Any]]
+GetDispatcher = Dict[str, Callable[..., Any]]
+
+
+def list(object_type: str, /, **kwargs: Any) -> Any:  # noqa: A001
```
fkiraly: `list` is not a good name, as it overloads python `list` - we should avoid that!

fkiraly: what other good names are there?

fkiraly: maybe `list_all`

openml/__init__.py
Omswastik-11: due to the list method  it overloads python list

openml/__init__.py
jgyasu, line 1: Is `openml/__init__.py` the best place to implement these functions? I would say let's keep it for the imports and add these utilities in another file?

Omswastik-11: Yes , It would be better if I will shift them to a new file .

geetu040 (changes requested):
Copilot left a few solid review comments above, please take a look and address them.

fkiraly:
@PGijsbers, what do you think of this suggestion?

openml/dispatchers.py
```diff
@@ -0,0 +1,113 @@
...
+    "flow": list_flows,
+    "run": list_runs,
+}
+
+_GET_DISPATCH: GetDispatcher = {
+    "dataset": get_dataset,
+    "task": get_task,
+    "flow": get_flow,
+    "run": get_run,
+}
```
geetu040, line 29: non-blocker but it could simply be like this, since we are not using them elsewhere

```suggestion
_LIST_DISPATCH: dict[str, Callable[..., Any]] = {
    "dataset": list_datasets,
    "task": list_tasks,
    "flow": list_flows,
    "run": list_runs,
}

_GET_DISPATCH: dict[str, Callable[..., Any]] = {
    "dataset": get_dataset,
    "task": get_task,
    "flow": get_flow,
    "run": get_run,
}
```

geetu040 (approved):
LGTM.

@PGijsbers could you please take a look and approve if you agree with the design?

PGijsbers (commented):
I'm sorry to only see this so late, but I don't really see the added value for this. Please do push back if you think I am wrong in this.

I much prefer the design of #1551. I know they are technically not mutually exclusive, but I feel that PR pretty much provides the same convenience as this one in terms of removing duplication from the calls. However, unlike this PR, the #1551 method also provides:

- (more) accurate type annotation, e.g., `get_task` returns a task whereas `get("task", ..` returns `Any`. This is useful for editors that give hints/autocomplete suggestions based on type annotations (common LSP functionality).
- call-specific docstrings and parameters (e.g., `task_type` is for `list_tasks` and not other list calls), again useful for the user that they do not have to pick and choose what is the relevant documentation for their entity type.
- less prone to typos: mistyping `list_task` (instead of plural) is caught by most LSPs because the function doens't exist, whereas `list_all("task", ...)` is likely not because its validity is delegated to runtime checks.

Granted, these could be remedied by:
 - providing different annotations and signatures with `typing.overload` to provide input-specific signatures and return types
 - providing an enum for allowed values

but at that point the #1551 seems easier to understand for users, and easier to maintain for us.

Unless I am mistaken, I suppose the main argument for this design would be "extensibility" in that adding a new type of object to retrieve doesn't change the function to call but only an argument. However, this requires either forgoing more detailed annotation (as outlined above). If you do want to provide the developer with good annotations then you would still need to document the new overload of the signature. That seems just as "much" work as providing it as a dedicated function.
We also do not expect to be adding or changing the entities on OpenML at such velocity that this makes a significant difference.

geetu040:
@fkiraly what do you think about this: https://github.com/openml/openml-python/pull/1552#pullrequestreview-3943962993

PR #1553, opened 2025-12-24
[MNT] Update ruff and mypy version, and format files to match latest ruff checks

### LINKED ISSUE #1550: [MNT] Updating `ruff` version for `pre-commit`
When I run `ruff format .` (ruff version being 0.14.10) for openml-python it ends up formatting 11 files, which I found weird. The only difference between the pre-commit ruff hooks in openml-python and server-api (passes `ruff format .` with no files changed) is the ruff version (0.14.10 for  server-api, 0.7.3 for openml-python). 

I assume from the above evidence that we probably just need to change the `ruff` version and it will be working fine after that. I'll make a PR to update the `ruff` version in `.pre-commit-config.yaml` to 0.14.10, after I run `ruff format .` to format all the files and avoid CI failures for the latest `ruff` version.

satvshr:
#### Metadata
* Stacks on #1547 (for ignoring `.ruff_cache`)
* Reference Issue: fixes #1550 
* New Tests Added: No
* Documentation Updated: No

#### Details 
What does this PR implement/fix? Explain your changes.
* Updates the ruff version in .pre-commit-config.yaml to 0.14.10
* Runs `ruff format .` to align the codebase with the formatting rules of the updated Ruff version
* Fixes also added to pass `ruff check .` checks
* Add `noqa` tags in places that will end up changing the architecture of the function/class if I try fixing it
* Only changes from my end to the actual code would be changing small things like:
  * the print statements to be compatible with check [UP031](https://docs.astral.sh/ruff/rules/printf-string-formatting/)
  * Changing variable names to `_` to be compatible with [RUF059](https://docs.astral.sh/ruff/rules/unused-unpacked-variable/)

This PR is going to be a bigger one in size but in my opinion, we should be compatible with the latest ruff version and get it over with sooner rather than later.

On a separate note, there are already a significant number of `noqa` tags in the codebase. We should consider revisiting the architecture of the functions and classes that rely on them to better align with Ruff’s best practices. Where alignment isn’t appropriate, we should at least discuss why those components don’t need to be Ruff-compatible.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -98,9 +98,9 @@ jobs:
         fi
 
         if [ "${{ matrix.sklearn-only }}" = "true" ]; then
-          marks="sklearn and not production"
+          marks="sklearn and not production and not uses_test_server"
         else
-          marks="not production"
+          marks="not production and not uses_test_server"
         fi
 
         pytest -n 4 --durations=20 --dist load -sv $codecov -o log_cli=true -m "$marks"
@@ -113,9 +113,9 @@ jobs:
         fi
 
         if [ "${{ matrix.sklearn-only }}" = "true" ]; then
-          marks="sklearn and production"
+          marks="sklearn and production and not uses_test_server"
         else
-          marks="production"
+          marks="production and not uses_test_server"
         fi
 
         pytest -n 4 --durations=20 --dist load -sv $codecov -o log_cli=true -m "$marks"
@@ -123,7 +123,7 @@ jobs:
     - name: Run tests on Windows
       if: matrix.os == 'windows-latest'
       run: |  # we need a separate step because of the bash-specific if-statement in the previous one.
-        pytest -n 4 --durations=20 --dist load -sv --reruns 5 --reruns-delay 1
+        pytest -n 4 --durations=20 --dist load -sv --reruns 5 --reruns-delay 1 -m "not uses_test_server"
 
     - name: Check for files left behind by test
       if: matrix.os != 'windows-latest' && always()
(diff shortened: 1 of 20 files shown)


examples/Basics/introduction_tutorial.py
```diff
@@ -12,7 +12,7 @@
 # For certain functionality, such as uploading tasks or datasets, users have to
 # sign up. Only accessing the data on OpenML does not require an account!
 #
-# If you don’t have an account yet, sign up now.
+# If you dont have an account yet, sign up now.
```
fkiraly: typo, this should be "don't"

openml/evaluations/evaluation.py
```diff
@@ -123,7 +123,7 @@ def __repr__(self) -> str:
             "Run ID",
             "OpenML Run URL",
             "Task ID",
-            "OpenML Task URL" "Flow ID",
+            "OpenML Task URLFlow ID",
```
fkiraly: typo, missing comma?

fkiraly (changes requested):
Great, thanks.

* please go through the files and check for problems or typos. I marked a few issue above.
* While we are at it, can you also ensure the target version for the checks is 3.10 or higher, so the `from __future__ import annotations` is not always added by `pre-commit`?

satvshr:
fixed typos and bumped target version to 3.10 (previously 3.8)

satvshr:
@fkiraly I also updated the mypy version to python 3.10 (previously 3.8), I know this is out of scope for this PR, but I would have had to made 44 changes if I continued with the older version. Hope that is okay?

fkiraly:
@satvshr, that is great! I was already wondering where the setting was that kept adding the `from __future__ import` in `precommit`.

fkiraly (approved):
Looks good to me, but perhaps someone else should also look at whether there is anything they spot in the files.

satvshr:
> I noticed quite a few `noqa` comments, some could be fixed easily, but addressing all of them at once would require manual effort. I suggest we handle them gradually over time.

Yeah I did not want to make changes that could be debated, hence delaying a pretty huge PR.

PR #1554, opened 2025-12-24
[ENH] Simplified Publish API with Automatic Type Recognition

Omswastik-11:
initially 

```python
from openml_sklearn.extension import SklearnExtension
clf = KNeighborsClassifier(n_neighbors=3)
extension = SklearnExtension()# User instantiates the extension object
knn_flow = extension.model_to_flow(clf) # User manually converts the model (estimator instance) to an OpenMLFlow object
knn_flow.publish()

```

## API
```python
import openml_sklearn  # Register the extension
import openml

clf = KNeighborsClassifier(n_neighbors=3)

openml.publish(clf)

```


examples/Basics/simple_flows_and_runs_tutorial.py
```diff
@@ -48,6 +48,13 @@
 clf = KNeighborsClassifier(**knn_parameters)
 clf.fit(X_train, y_train)
 
+# Option A: auto-publish the estimator via unified helper (requires openml-sklearn extension).
+try:
```
fkiraly: try/except for case distinction is not good. This does not allow to distinguish the intended case from a genuine failure in the "try" branch.

Plus, in the examples, there should be no try/except.

openml/__init__.py
```diff
@@ -50,6 +53,37 @@
 )
 
 
+def publish(obj: Any, *, name: str | None = None, tags: Sequence[str] | None = None) -> Any:
+    """Publish a common object (flow/model/run/dataset) with minimal friction.
+
+    If ``obj`` is already an OpenML object (``OpenMLBase``) it will call its ``publish`` method.
```
fkiraly: Please write a proper docstring.

fkiraly (changes requested):
I get this is a draft still, some early comments.

* works for flows only, I would recommend to try for at least two different object types to see the dispatching challenge there.
* do the extension checking inside `publish` and not in the usage example

Omswastik-11:
Thanks @fkiraly !! 
I checked on `flow` , `datset` , `task` . it is working correctly but in `run` it is getting some server side issues.

```
Task 1 failed: https://test.openml.org/api/v1/xml/data/features/1 returned code 274: No features found. Additionally, dataset processed with error - None
```

jgyasu:
The PR description is not entirely correct. This is how the interface looks currently:
```python
from openml_sklearn.extension import SklearnExtension
clf = KNeighborsClassifier(n_neighbors=3)
extension = SklearnExtension()# User instantiates the extension object
knn_flow = extension.model_to_flow(clf) # User manually converts the model (estimator instance) to an OpenMLFlow object
knn_flow.publish()
```
But I like the idea of a unified `publish`. I am currently working on a design document for refactoring `Extension` and this design coincides mine as well, which is a good thing.

Omswastik-11:
> The PR description is not entirely correct. This is how the interface looks currently:
> 
> ```python
> from openml_sklearn.extension import SklearnExtension
> from sklearn.neighbors import KNeighborsClassifier
> clf = KNeighborsClassifier(n_neighbors=3)
> extension = SklearnExtension()# User instantiates the extension object
> knn_flow = extension.model_to_flow(clf) # User manually converts the model (estimator instance) to an OpenMLFlow object
> knn_flow.publish()
> ```
> 
> But I like the idea of a unified `publish`. I am currently working on a design document for refactoring `Extension` and this design coincides mine as well, which is a good thing.

Thanks for the correction I used the syntax example used in example script . this unified publish was Franz's idea . https://github.com/gc-os-ai/openml-project-dev/issues/8

openml/__init__.py
```diff
@@ -50,6 +54,122 @@
...
+      it directly calls their ``publish()`` method.
+    - For external models (e.g., scikit-learn estimators), it uses registered
+      extensions to convert them to ``OpenMLFlow`` objects before publishing.
+
+    Parameters
+    ----------
+    obj : Any
+        The object to publish. Can be:
+        - An OpenML object (OpenMLDataset, OpenMLFlow, OpenMLRun, OpenMLTask)
+        - A machine learning model from a supported framework (e.g., scikit-learn)
```
jgyasu: > A machine learning model

I would call it an estimator instance, here and elsewhere

openml/__init__.py
```diff
@@ -50,6 +54,122 @@
...
+    ----------
+    obj : Any
+        The object to check.
+    """
+    obj_module = type(obj).__module__
+
+    # Check for scikit-learn models
+    if obj_module.startswith("sklearn"):
+        with contextlib.suppress(ImportError):
+            import openml_sklearn  # noqa: F401
```
jgyasu: What is the use of this function? `get_extension_by_model` already takes care of this logic.

jgyasu (changes requested):

jgyasu:
I have added some comments. I also feel we should not populate `__init__.py` with these functions, we can have them in a seperate file and use `__init__.py` only for imports.

openml/__init__.py
Omswastik-11: Thanks for pointing . I missed that part

geetu040 (changes requested):
I am fine with most of the core logic, just have to double-check the extension work and tests.
There are some reviews by Copilot you would want to address.
I've left a comment for discussion.

openml/publishing.py
```diff
@@ -0,0 +1,102 @@
...
+from __future__ import annotations
+
+from collections.abc import Sequence
+from typing import Any
+
+from . import extensions
+from .base import OpenMLBase
+
+
+def publish(obj: Any, *, name: str | None = None, tags: Sequence[str] | None = None) -> Any:
```
geetu040, line 11: What's the reason for adding `name` and `tags` as a parameter to `publish`? I'm not necessarily against it, I would actually prefer it this way, but I am just curious about the motivation or use case behind this. Are there other parameters we can introduce like these? possibly the common attributes between resources that are used in `_to_dict` to create artifacts for upload.

geetu040 (changes requested):
I am fine with most of the core logic, just have to double-check the extension work and tests.
There are some reviews by Copilot you would want to address.
I've left a comment for discussion.

openml/publishing.py
Omswastik-11: Thanks for the question — the main motivation for `name` and `tags` is to keep `openml.publish(obj)` as a single high-level entry point while still allowing lightweight metadata customization without requiring users to manually mutate objects before publish. This helps especially for extension-backed estimators, where users often want to set a readable flow name and add experiment tags at publish time. I limited it to name and tags because they are common, low-risk metadata across publishable artifacts; adding more kwargs would increase API ambiguity and make validation harder across object types.

openml/publishing.py
```diff
@@ -0,0 +1,114 @@
...
+    # Case 1: Object is already an OpenML entity
+    if isinstance(obj, OpenMLBase):
+        if tags is not None and hasattr(obj, "tags"):
+            existing = list(getattr(obj, "tags", []) or [])
+            if all(isinstance(tag, str) for tag in existing):
+                merged = list(existing)
+                for tag in tags:
+                    if tag not in merged:
+                        merged.append(tag)
+                obj.tags = merged
```
geetu040: this can be simplified as above, also see the comment on tests, we could simplify this even further.
```suggestion
        if tags and hasattr(obj, "tags"):
            existing = obj.tags or []
            if all(isinstance(tag, str) for tag in existing):
                obj.tags = list(dict.fromkeys([*existing, *tags]))
```

openml/publishing.py
```diff
@@ -0,0 +1,114 @@
...
+    flow = extension.model_to_flow(obj)
+
+    if name is not None:
+        flow.name = name
+
+    if tags is not None:
+        existing_tags = list(getattr(flow, "tags", []) or [])
+        flow.tags = list(dict.fromkeys([*existing_tags, *tags]))
+
+    return flow.publish()
```
geetu040: we can recursively call `publish(flow)` to avoid duplicate logic
also this part of code can be moved to top under an if clause and the main logic for `OpenMLBase` object can be written outside the if-block to save an indentation.

tests/test_openml/test_openml.py
```diff
@@ -41,3 +41,85 @@ def test_populate_cache(
         assert task_mock.call_count == 2
         for argument, fixture in zip(task_mock.call_args_list, [(1,), (2,)]):
             assert argument[0] == fixture
+
+    def test_publish_with_openml_object_merges_tags_and_name(self):
+        class Dummy(openml.base.OpenMLBase):
```
geetu040: this dummy class should be moved outside this test function, and used in both functions

tests/test_openml/test_openml.py
```diff
@@ -41,3 +41,85 @@ def test_populate_cache(
...
+        result = openml.publish(obj, name="new", tags=["b", "a"])
+        assert result is obj
+        assert obj.published is True
+        assert obj.name == "new"
+        assert obj.tags == ["a", "b"]  # dedup and preserve order from original
+
+    def test_publish_with_openml_object_non_string_tags_does_not_crash(self):
+        class Dummy(openml.base.OpenMLBase):
+            def __init__(self) -> None:
+                self.tags = [{"tag": "legacy"}]
```
geetu040: This is confusing, is it possible to have tags like this? I don't think so. We can always expect tags to be `list[str]` therefore this case is not possible, and the above logic can be simplified as well.

tests/test_openml/test_openml.py
```diff
@@ -41,3 +41,85 @@ def test_populate_cache(
...
+
+        obj = Dummy()
+        result = openml.publish(obj, name="new", tags=["x"])
+        assert result is obj
+        assert obj.published is True
+        assert obj.name == "new"
+        assert obj.tags == [{"tag": "legacy"}]
+
+    @mock.patch("openml.publishing.extensions.get_extension_by_model")
+    def test_publish_with_extension(self, get_ext_mock):
```
geetu040: this is heavily mocked test, instead, write 2 tests `test_openml_publish` and `test_openml_publish_ext`, where you mock the `OpenMLBase.publish` and assert that it was called, also mark `test_openml_publish_ext` with `sklearn` marker, so the extension should work fine.

geetu040 (changes requested):

PGijsbers (commented):
I have to give this some more consideration. Here are my thoughts so far.

Currently, all OpenML entities that are publishable have a `publish()` method. In that sense, the API is already unified. In my view, the benefit of this is therefor mainly that it can extend to new object types, such as estimators for which an extension is registered that can serialize it to a Flow. So it seems there are now two alternatives:

```python
dataset = OpenMLDataset(name="foo", tags=["bar"], ...)
dataset.publish()
```
or
```python
dataset = OpenMLDataset(...)
openml.publish(dataset, name="foo", tags=["bar"])
# of course, name and tags could also have been provided during initialisation of the OpenMLDataset
```

So for anything that is already an OpenML object, I do not really see the benefit. It just introduces two different ways to do things, which I would generally be against. (I assume here the intention is for the `publish` method to remain on the OpenML objects as well.)

For estimators that are to be converted to flows, this is of course significantly shorter as shown in the original proposal. However, I think it would also be worth considering an alternative. Consider that perhaps instead an OpenMLFlow could be initialised with an arbitrary object which would be attempted to be resolved with extensions. Then it could also provide a similarly smooth experience:
```python
estimator = sklearn.tree.DecisionTreeClassifier()
flow = OpenMLFlow(estimator, name="foo", tags=["bar"])
flow.publish()
```
Sure, it introduces an extra line of code, but it does make it explicit to the user what kind of OpenML object they are publishing.

I am trying to think of other categories of objects that would be parsed into OpenML objects that would also benefit from this general publish function, but I can't really think of any. In most cases, I think it would be far more useful to have one dedicated function to create the object and thus communicate the metadata schema. E.g., a dataset can have a name, description, author, and so on, and I do not think this is something that would translate well into a general publish call (thinking of e.g., `openml.publish(dataframe, name="...", description="...", ...)` but where the type hints cannot provide information as so what metadata is valid).

In any case, the coupling of object creation with publication to the platform is problematic in the case where users do not have an internet connection. This can be the case where e.g., a user prepopulates their cache when they have a connection and then executes experiments in an offline setting (most commonly some compute server setups, but potentially also something like a regular outage). While we do provide some utility functions that do this (like `run_model_on_task`, though there too we made sure it could run offline with the right arguments), I am hesitant to introduce that as the main way to create/share objects.

geetu040:
I think the main motivation for the `publish` API originally came from the model indexing and sharing feature that was planned in OpenML, which has since moved to aiondemand. Given that, we don't currently have additional object types in mind beyond the existing resources that might benefit from extending the `publish` API.

So for now, I don't strongly disagree with your point. It probably makes sense to park this idea for the time being. If new object types or workflows come up later where a more general `publish` function clearly improves the UX, we can revisit it then.

That said, even if extensibility is not a strong motivation right now, I still think the `openml.publish(...)` syntax reads quite nicely and feels intuitive. Personally, I would still lean slightly in favor of this design and the PR.

@fkiraly what do you think about the situation?

PR #1556, opened 2025-12-24
[BUG] Fix Sklearn Models detection by safely importing openml-sklearn

### LINKED ISSUE #1542: [BUG] Fix Sklearn Models detection by safely registering SklearnExtension
Running a scikit-learn `Pipeline` with `openml.runs.run_model_on_task` fails even when `openml-sklearn` is installed.

Reproduction:
```py
import openml
from sklearn import impute, tree, pipeline

clf = pipeline.Pipeline(
    steps=[
        ('imputer', impute.SimpleImputer()),
        ('estimator', tree.DecisionTreeClassifier(max_depth=2))
    ]
)

task = openml.tasks.get_task(32)
run = openml.runs.run_model_on_task(clf, task, avoid_duplicate_runs=False)

print(run)
```

This raises:
```
ValueError: No extension registered which can handle model: Pipeline(...).
But it looks related to scikit-learn. Please install the OpenML scikit-learn extension (openml-sklearn) and try again.
```
This happens even if `openml-sklearn` is already installed.

If `openml-sklearn` is installed, sklearn models (including `Pipeline`) should be detected automatically without requiring any manual imports.

`SklearnExtension` is only registered if it is imported somewhere during execution. In normal usage, this does not happen, so the extension is never added to `openml.extensions.extensions`.

Possible fixes:

* Safely import and register `SklearnExtension` when `openml-sklearn` is available. This can be done by adding a safe import statement or moving the `register_extension` out of `openml_sklearn/__init__.py`
* Improve the error message when `openml-sklearn` is not installed. The current link does not lead to an installation guide.

EmanAbdelhaleem:
#### Metadata

* Reference Issue: Fixes #1542
 
#### Details 
Fixed sklearn models detection by safely importing openml-sklearn at `openml/runs/__init__.py`


openml/extensions/functions.py
```diff
@@ -16,8 +17,7 @@
 SKLEARN_HINT = (
     "But it looks related to scikit-learn. "
     "Please install the OpenML scikit-learn extension (openml-sklearn) and try again. "
-    "For more information, see "
-    "https://github.com/openml/openml-sklearn?tab=readme-ov-file#installation"
+    "You can use `pip install openml-sklearn` for installation."
 )
```
geetu040, line 23: Keep the "for more information" lines at the very end. Perhaps we can refer to one of the following links for more information
- https://github.com/openml/openml-sklearn
- https://docs.openml.org/python/extensions/

I would prefer the second link, but let me know if you think otherwise.

EmanAbdelhaleem: Since we are going to move openml-sklearn back to openml-python, I think the second link would be better.

fkiraly (commented):
Quick question - why does this work, given that `openml_sklearn` is not actually directly used? Is there some background magic that does something non-explicit, in case the module gets imported?

That feels very dangerous design, if it were so (fully aware that it is not yours).

EmanAbdelhaleem:
> Is there some background magic that does something non-explicit, in case the module gets imported?

Yes, the `SklearnExtension` gets registered, the issue was that it wasn't registered, and actually with the current design, it only gets registered when `openml-sklearn` is imported.
Check here: https://github.com/openml/openml-sklearn/blob/main/src/openml_sklearn/__init__.py#L15C1-L16C1

fkiraly (approved):

PR #1557, opened 2025-12-24
[BUG] fix docstring style for the API

### LINKED ISSUE #1548: [BUG] fix docstring style for the API
The structure of the parameters section for the API ref in the documentation is not right. Here is [an example](https://openml.github.io/openml-python/latest/reference/datasets/functions/#openml.datasets.functions.create_dataset). The parameter section is just a wall of text. There is probably just a mismatch between the docstring style in the code and the expected docstring style in the mkdocs config.

EmanAbdelhaleem:
#### Metadata
* Reference Issue: Fixes #1548 

#### Details 
Our docstrings are written in NumPy docstring style, however in `mkdocs.yml` we used `docstring_style: google` which led to having a wall of text for the parameter sections in the API ref in the documentation.

### DIFF
--- a/.github/workflows/dist.yaml
+++ b/.github/workflows/dist.yaml
@@ -23,7 +23,7 @@ jobs:
   dist:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v6
     - name: Setup Python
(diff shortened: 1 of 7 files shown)


EmanAbdelhaleem:
How it looked:
image


How it now looks:
image

PR #1560, opened 2025-12-25
[ENH] Extend `Extension` class test suite

### LINKED ISSUE #1545: [ENH] Extension Module Tests (or Examples?)
The extensions module has only 2 test methods, which cover the non-abstract methods: `get_extension_by_model` and `get_extension_by_flow`. The extension-interface is an abstract class, so testing it would require to implement the child-class in form of a Dummy. I still think this would be an improvement, because it shows and tests the way this abstract class is used and what needs to be implemented. Please comment if this makes sense or should we go straight for examples of extension usage? 

Here are some suggestions for test-methods, for this issue:

**Extension Interface Methods:**
   ```python
   def test_flow_to_model_with_defaults():
       """Test flow_to_model with initialize_with_defaults=True."""

   def test_flow_to_model_strict_version_false():
       """Test flow_to_model with strict_version=False."""

   def test_model_to_flow_conversion():
       """Test converting a model back to flow representation."""
   ```

**Error Handling:**
   ```python
   def test_invalid_flow_raises_error():
       """Test that invalid flow raises appropriate error."""

   def test_extension_not_found_error_message():
       """Test error message contains helpful information."""
   ```

And for the functions.py:

**Edge Cases:**
   ```python
   def test_register_same_extension_twice():
       """Test behavior when registering same extension twice."""

   def test_extension_priority_order():
       """Test that extensions are checked in registration order."""
   ```

satvshr:
#### Metadata
* Reference Issue: fixes #1545 
* New Tests Added: Yes
* Documentation Updated: No
* Change Log Entry: Add tests for extension interface contract and extension registry edge cases

#### Details 
This PR adds unit tests for the OpenML Extension interface and for extension registry behavior. The tests added are the 7 tests mentioned in #1545 
Previously, only the non-abstract registry helpers (`get_extension_by_model`, `get_extension_by_flow`) were covered. The abstract `Extension` interface itself was not tested.

### DIFF
--- a/openml/utils.py
+++ b/openml/utils.py
@@ -2,12 +2,20 @@
 from __future__ import annotations
 
 import contextlib
+import re
 import shutil
 import warnings
-from collections.abc import Callable, Mapping, Sized
+from abc import ABC, abstractmethod
+from collections.abc import Callable, Iterable, Mapping, Sequence, Sized
 from functools import wraps
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Literal, TypeVar, overload
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Literal,
+    TypeVar,
+    overload,
+)
 from typing_extensions import ParamSpec
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


satvshr:
Error seems unrelated to the PR given all this PR does is extend the `Extension` class. The failing test:
```
    @pytest.mark.sklearn()
    @unittest.skipIf(
        Version(sklearn.__version__) == Version("0.19.1"),
        reason="Requires scikit-learn!=0.19.1, because target flow is from that version.",
    )
    @pytest.mark.production()
    def test_get_flow_with_reinstantiate_strict_with_wrong_version_raises_exception(self):
        self.use_production_server()
        flow = 8175
        expected = "Trying to deserialize a model with dependency sklearn==0.19.1 not satisfied."
        self.assertRaisesRegex(
            ValueError,
            expected,
            openml.flows.get_flow,
            flow_id=flow,
            reinstantiate=True,
            strict_version=True,
        )
```

Edit: The tests pass locally.

satvshr:
@jgyasu the problem in fact was related to this PR, if `openml.extensions.extensions.clear()` runs before `test_get_flow_with_reinstantiate_strict_with_wrong_version_raises_exception`, and `import openml_sklearn` is called before `openml.extensions.extensions.clear()` once, it will be cleared from the `openml.extensions.extensions` list, and further imports won't add it back given `sys.modules` will identify that it was imported previously and hence won't add the data it is supposed to to `openml.extensions.extensions`, hence causing failures!

satvshr:
Test failures discussed [here](https://github.com/openml/openml-python/pull/1577#issuecomment-3782915130)

PR #1565, opened 2025-12-27
[ENH] Add `get_cache_size` Utility Function

### LINKED ISSUE #1561: [ENH] Add Cache Size Utility Function
Users have no easy way to check how much disk space the OpenML cache is using. The cache can grow large with many datasets. The solution would be to add a utility function to calculate the current cache size. 

Much of the relevant code is in `openml/config.py` and `openml/utils.py`. We could put the function into `openml/utils.py` and expose it to the public API from there.

JATAYU000:
#### Metadata
* Reference Issue: Fixes #1561 
* New Tests Added: Yes
* Documentation Updated: Yes (Doc string)
* Change Log Entry: Add new function `get_cache_size()` in `utils`

#### Details 
* What does this PR implement/fix? 
Implements a `get_cache_size()` function which returns the total size of the `openml` cache directory in bytes.


fkiraly:
@geetu040 please review

Issue #1529, opened 2025-12-12
Hide test server admin key


PGijsbers:
Currently the key for an admin user of the test server is baked into the code base (also of other repositories). While no real harm can be done with this (the test server resets completely regularly and is completely isolated from production), it would still be better to delegate those tests to only run in CI (with secrets) and/or core contributors (through environment variables).

rohansen856:
I would to work on this issue.
I see the admin key for the OpenML test server was hardcoded directly in the source code (`openml/testing.py`).
I would like to do the following:

- add a new environment variable constant for the test server admin key (`openml/config.py`):

```python
OPENML_TEST_SERVER_ADMIN_KEY_ENV_VAR = "OPENML_TEST_SERVER_ADMIN_KEY"
```

- then use the env var instead of hardcoding the secret (`openml/testing.py`):

```python
admin_key = os.environ.get(openml.config.OPENML_TEST_SERVER_ADMIN_KEY_ENV_VAR)
```

- also update the tests in `tests/test_openml/test_config.py` and `tests/test_datasets/test_dataset_functions.py`
- also updating the workflow in `.github/workflows/test.yml` is needed to use the env var as secret

```yaml
env:
  OPENML_TEST_SERVER_ADMIN_KEY: ${{ secrets.OPENML_TEST_SERVER_ADMIN_KEY }}
```

PR #1568, opened 2025-12-29
[ENH] replaced hardcoded test server admin key with env variable and secrets

rohansen856:
#### Metadata
* Reference Issue: #1529 
* New Tests Added: Yes
* Documentation Updated: No
* Change Log Entry: 

#### Details 

this PR is made to remove the hardcoded test server admin key from the codebase and replace it with environment variable-based authentication.

## Summary

- in `openml/config.py` Added a new environment variable constant for the test server admin key:

```python
OPENML_TEST_SERVER_ADMIN_KEY_ENV_VAR = "OPENML_TEST_SERVER_ADMIN_KEY"
```

- Testing Base Class Updated in `openml/testing.py`. Modified the `TestBase` class to read the admin key from an environment variable instead of using a hardcoded value:

**Before**:
```python
admin_key = "abc"
```

**After**:
```python
admin_key = os.environ.get(openml.config.OPENML_TEST_SERVER_ADMIN_KEY_ENV_VAR)
```

**Note**:
- The admin key is now `None` by default when the environment variable is not set
- Tests requiring the admin key will fail gracefully if the key is not available

Also in the tests, Added `pytest.skipif` decorators to tests that require admin privileges in the following test files:

#### `tests/test_openml/test_config.py`
**Test**: `test_switch_to_example_configuration`

**Added decorator**:
```python
@pytest.mark.skipif(
    not os.environ.get(openml.config.OPENML_TEST_SERVER_ADMIN_KEY_ENV_VAR),
    reason="Test requires admin key. Set OPENML_TEST_SERVER_ADMIN_KEY environment variable.",
)
```

#### `tests/test_datasets/test_dataset_functions.py`
**Test**: `test_data_status`

**Added decorator**:
```python
@pytest.mark.skipif(
    not os.environ.get(openml.config.OPENML_TEST_SERVER_ADMIN_KEY_ENV_VAR),
    reason="Test requires admin key. Set OPENML_TEST_SERVER_ADMIN_KEY environment variable.",
)
```

**Note**:
- These tests will be automatically skipped if the admin key is not provided
- Clear skip reason is displayed when tests are skipped
- No failures or errors when running tests locally without the admin key

### 4. CI Configuration Update (`.github/workflows/test.yml`)
Added the environment variable to all test execution steps in the GitHub Actions workflow:

**Steps updated**:
- Run tests on Ubuntu Test
- Run tests on Ubuntu Production  
- Run tests on Windows

**Added to each step**:
```yaml
env:
  OPENML_TEST_SERVER_ADMIN_KEY: ${{ secrets.OPENML_TEST_SERVER_ADMIN_KEY }}
```

**Impact**:
- CI will use the secret stored in GitHub repository settings
- Tests requiring admin privileges will run in CI
- The actual key value is never exposed in logs or code

@PGijsbers this requires someone to put the admin key in the github secrets which would be a critical step.

# Update on reviews
the configurations should be done from openml config files in `./.openml/config` for directory level configurations, instead of the added responsibility of a new `.env` file and dependencies. in case of local testing the concerned tests would be skipped in case no key is provided.

### DIFF
--- a/tests/test_extensions/test_functions.py
+++ b/tests/test_extensions/test_functions.py
@@ -1,12 +1,14 @@
 # License: BSD 3-Clause
 from __future__ import annotations
 
+from collections import OrderedDict
+
 import inspect
-
+import numpy as np
 import pytest
-
+from unittest.mock import patch
 import openml.testing
-from openml.extensions import get_extension_by_flow, get_extension_by_model, register_extension
+from openml.extensions import Extension, get_extension_by_flow, get_extension_by_model, register_extension
 
 
 class DummyFlow:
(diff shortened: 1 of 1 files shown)


rohansen856:
> This successfully removes the secret from the code base.
> 
> My only question, is there a documentation location (e.g., developer documentation) that also needs to get updated about this? Since developers wishing to use the test server now have to:
> 
> * obtain credentials (from whom or where?)
> * set the environment variable in python

I think if any documentation update has to be done it should be in Advanced user guide (`https://openml.github.io/openml-python/latest/details/`). As for local testing, there is no `.env` file in the codebase so it needs to be provided by the developer before testing (ex.: `export OPENML_TEST_SERVER_ADMIN_KEY="your-admin-key" && pytest tests/`) But even if a new developer does not have the context about this, The tests would automatically skip if the developer does not provide the variable so tests wont fail because of this.

also this is the admin API key for the OpenML test server (https://test.openml.org`) (even though we used very weak credentials for testing purposes: `abc`). so it should be behind a env var. So instead of having it exposed, using it as a secret in github actions would be better and i can update the documentation for any developer who want to actually run the two tests that would be skipped any way in case of local testing if they dont provide the value.

Please do let me know if you have other suggestions on this @fkiraly

fkiraly:
@jgyasu, does this need to be added to the developer guide you are writing?

.github/workflows/test.yml
```diff
@@ -98,6 +98,8 @@ jobs:
       run: python -m pip list
     - name: Run tests on Ubuntu Test
       if: matrix.os == 'ubuntu-latest'
+      env:
+        OPENML_TEST_SERVER_ADMIN_KEY: ${{ secrets.OPENML_TEST_SERVER_ADMIN_KEY }}
```
geetu040, line 110: we should make sure these are added in the github secrets before merging this

openml/config.py
```diff
@@ -24,6 +24,7 @@
 
 OPENML_CACHE_DIR_ENV_VAR = "OPENML_CACHE_DIR"
 OPENML_SKIP_PARQUET_ENV_VAR = "OPENML_SKIP_PARQUET"
+OPENML_TEST_SERVER_ADMIN_KEY_ENV_VAR = "OPENML_TEST_SERVER_ADMIN_KEY"
```
geetu040, line 28: in case of using a `.env` file, some code is required to load it and may add the extra dependency `python-dotenv`

tests/test_openml/test_config.py
```diff
@@ -106,6 +106,10 @@ def test_setup_with_config(self):
 
 
 class TestConfigurationForExamples(openml.testing.TestBase):
+    @pytest.mark.skipif(
+        not os.environ.get(openml.config.OPENML_TEST_SERVER_ADMIN_KEY_ENV_VAR),
+        reason="Test requires admin key. Set OPENML_TEST_SERVER_ADMIN_KEY environment variable.",
+    )
     @pytest.mark.production()
     def test_switch_to_example_configuration(self):
         """Verifies the test configuration is loaded properly."""
```
geetu040, line 111: btw this test passes anyways even if you use any random `admin_key`, right?

geetu040 (changes requested):
Looks fine, just need to add code for loading `.env` if that is required. And we should also mention this in the internal developer setup.

tests/test_openml/test_config.py
rohansen856: Yes, `test_switch_to_example_configuration` doesn't make any API calls, it only tests that `start_using_configuration_for_example()` correctly swaps the apikey and server values. So it passes with any value (even `None` is ok)

But `test_data_status` (tests\test_datasets\test_dataset_functions.py) does require a valid admin key because it makes actual API calls to `openml.datasets.status_update()` which the server validates, so if a user passes invalid value it would return authentication error: 
```py
FAILED tests/test_datasets/test_dataset_functions.py::TestOpenMLDataset::test_data_status - openml.exceptions.OpenMLServerException: https://test.openml.org/api/v1/xml/data/status/update ...
```

openml/config.py
rohansen856: yes definitely! Will go ahead with the new dep if approved.

.github/workflows/test.yml
PGijsbers: secret set

tests/test_openml/test_config.py
PGijsbers: Yes, I think the more reasonable approach here is to update this test to use some other sentinel value.

rohansen856:
> Mostly LGTM, but I am not yet sold on introducing the `.env` file.
> 
> Regular user settings are currently configured through a dedicated configuration file already. I'm not yet convinced that adding a `.env` file for (developer) configuration is more logical than extending the existing configuration file. A decision should probably be made in the larger context on how to handle configurations?
> 
> Something that seems reasonable is to support openml config files in `./.openml/config` for directory level configurations.
> 
> In my opinion we can merge this as is (except for the note on `test_switch_to_example_configuration`), and then discuss ideas on how we should have user and developer configurations after.

Sure thing! scratching the plan for .env file then. Also can you please elaborate what do you mean by "some other sentinel value"...?

geetu040:
> Sure thing! scratching the plan for .env file then.

I think "handling the configuration" is out of scope for this PR. We should target an immediate solution given the current options.

> Regular user settings are currently configured through a dedicated configuration file already. I'm not yet convinced that adding a `.env` file for (developer) configuration is more logical than extending the existing configuration file.

I agree that the `.env` should not be introduced if there is a better alternate with the existing config setting, but I cannot really think of one where we can use the key without showing it in code except for the `.env` method.

geetu040:
Please mention in description
Fixes https://github.com/openml/openml-python/issues/1529

rohansen856:
@geetu040 updated the PR description accordingly.

rohansen856:
> > Sure thing! scratching the plan for .env file then.
> 
> I think "handling the configuration" is out of scope for this PR. We should target an immediate solution given the current options.
> 
> > Regular user settings are currently configured through a dedicated configuration file already. I'm not yet convinced that adding a `.env` file for (developer) configuration is more logical than extending the existing configuration file.
> 
> I agree that the `.env` should not be introduced if there is a better alternate with the existing config setting, but I cannot really think of one where we can use the key without showing it in code except for the `.env` method.

Would it be a solution if when a core developer has the actual key they set it directly from the terminal, ex.:
```bash
$env:MY_KEY = "hello"
# or
export MY_KEY="hello"
```

and pdating the contributing.md to keep track of this?

Ex.:

#### Running Tests That Require Admin Privileges

Some tests require admin privileges on the test server and will be automatically skipped unless you provide an admin API key. For regular contributors, the tests will skip gracefully. For core contributors who need to run these tests locally, first set the proper env var from terminal. ex.:

```bash
# For windows
$env:MY_KEY = "hello"
# For linux/mac
export MY_KEY="hello"
```

tests/test_openml/test_config.py
PGijsbers: > Also can you please elaborate what do you mean by "some other sentinel value"...?

Just anything that helps verify that `openml.config.apikey` is updated appropriately. So any literal that isn't identical to `TestBase.user_key`.

PGijsbers:
I think it's fine to expect the environment variable to be set for now. Later, this could be provided through either the normal configuration file? The configuration file can be updated before tests are executed (or we could still allow it to be overriden by an environment variable).

rohansen856:
> I think it's fine to expect the environment variable to be set for now. Later, this could be provided through either the normal configuration file? The configuration file can be updated before tests are executed (or we could still allow it to be overriden by an environment variable).

sure... we can expect the contributors having the actual key to set the env var from terminal if they want. updated the contributiong doc with a short instruction to keep track.

geetu040 (approved):
LGTM!
CC: @PGijsbers for final approval.

tests/test_openml/test_config.py
PGijsbers: The suggestion I meant to make (but GitHub isn't letting me) is to remove the check for the admin key, and instead update the key used in this test since admin privileges are irrelevant to the actual test (in fact, whether or not the original api key is valid is irrelevant):

```diff
-       openml.config.apikey = TestBase.admin_key
+       openml.config.apikey = "any-api-key"
```

CONTRIBUTING.md
```diff
@@ -96,6 +96,20 @@ To test your new contribution, add [unit tests](https://github.com/openml/openml
 * Please ensure that the example is run on the test server by beginning with the call to `openml.config.start_using_configuration_for_example()`, which is done by default for tests derived from `TestBase`.
 * Add the `@pytest.mark.sklearn` marker to your unit tests if they have a dependency on scikit-learn.
 
+#### Running Tests That Require Admin Privileges
+
+Some tests require admin privileges on the test server and will be automatically skipped unless you provide an admin API key. For regular contributors, the tests will skip gracefully. For core contributors who need to run these tests locally:
+
+**Set up the key** by exporting the variable:
+run this in the terminal before running the tests:
```
PGijsbers:
```suggestion
Some tests require admin privileges on the test server and will be automatically skipped unless you provide an admin API key. For regular contributors, the tests will skip gracefully. For core contributors who need to run these tests locally, you can set up the key by exporting the variable as below before running the tests:
```

tests/test_openml/test_config.py
```diff
@@ -106,6 +106,10 @@ def test_setup_with_config(self):
 
 
 class TestConfigurationForExamples(openml.testing.TestBase):
+    @pytest.mark.skipif(
+        not os.environ.get(openml.config.OPENML_TEST_SERVER_ADMIN_KEY_ENV_VAR),
+        reason="Test requires admin key. Set OPENML_TEST_SERVER_ADMIN_KEY environment variable.",
+    )
```
PGijsbers:
```suggestion
```
I think we can skip this check and make the test not require the admin key (see other comment I left below). This is what I meant to communicate with https://github.com/openml/openml-python/pull/1568/changes#r2698514445

CONTRIBUTING.md
rohansen856: Got it! updating with next commit.

tests/test_openml/test_config.py
rohansen856: Got it now! implementing with the next commit.

tests/test_openml/test_config.py
rohansen856: Done! Removed the `skipif` decorator and changed to use "any-api-key" since this test only verifies config swapping without making API calls.

rohansen856:
> Prior concerns weren't addressed, I made the suggested changese more explicit.

Thank you for the clear feedback! All suggestions have been implemented in the latest commit. Review would be appreciated!

PGijsbers (approved):
LGTM

fkiraly (approved):
Thanks!

Issue #1538, opened 2025-12-23
[DOC] Add Usage Examples to Core Function Docstrings


SimonBlanke:
Examples in docstrings provide various benefits for users and devs: You can see these examples during IDE hover tooltips, can generate examples for your auto generated documentation and can quickly see the usage pattern as a dev going through the code. Core public functions lack `Examples` sections in their docstrings. 
Here are some of the public methods that would benefit from examples in docstrings:
 
- `openml/datasets/functions.py:get_dataset()`
- `openml/datasets/functions.py:get_datasets()`
- `openml/tasks/functions.py:get_task()`
- `openml/runs/functions.py:run_model_on_task()`
- `openml/study/functions.py:get_suite()`
- `openml/study/functions.py:get_study()`
- `openml/study/functions.py:create_study()`

fkiraly:
I also noticed a few other issues with docstrings:

* missing defaults and types
* missing reference to retrieval from DB backend, e.g., what gets accessed
* missing reference to changes in database backend, e.g., what gets added, deleted

Kashish564:
Hi @SimonBlanke and @fkiraly 👋  
I’d like to work on this issue if it’s still available.

My plan is to:
- Add clear, minimal usage examples to the listed public functions’ docstrings
- While doing so, also improve missing types/defaults where obvious
- Clarify (briefly) what is retrieved from or written to the OpenML backend, where relevant

I’ll start with a small subset (e.g. get_dataset, get_task) and expand based on feedback.  
Please let me know if this approach sounds good. Thanks!

fkiraly:
feel free to go ahead, @Kashish564!
But make sure to pick something that is not currently being worked on - I think #1569 already deals with a number of these if not all.

siddharthbaleja7:
@fkiraly can you assign this issue to me?

krishajogia256:
Hi! I’d like to work on this issue if it’s still open.

I noticed that PR #1569 already improves docstrings for some core functions.
I’ll first review that PR and then focus on adding usage examples to functions that are not yet covered.

My plan is to:
- Identify remaining public core functions without usage examples
- Add minimal, clear examples following the existing docstring style used in the repository
- Ensure consistency with the current documentation format

Please let me know if this approach sounds good.

Abhishek9639:
Hi @SimonBlanke, @fkiraly,
[...]

PR #1569, opened 2025-12-29
[DOC] Enhance Docstrings of Flows Core Public Functions

EmanAbdelhaleem:
#### Metadata
* Reference Issue: #1538

#### Details 
enhance the docstrings of flows core public functions, add examples, parameter default, parameter type..etc


openml/flows/functions.py
```diff
@@ -138,32 +174,47 @@ def list_flows(
...
+        ``name``, ``version``, ``external_version``, and ``uploader``.
+
+    Raises
+    ------
+    OpenMLServerException
+        When the API call fails.
+
+    Side Effects
+    ------------
+    - None: results are fetched and returned; no local state is modified.
```
SimonBlanke: Is this true? Are there local connection states? And what about caching? 'no local state is modified' is a strong claim.

openml/flows/functions.py
```diff
@@ -71,23 +71,59 @@ def _get_cached_flow(fid: int) -> OpenMLFlow:
...
-
-    strict_version : bool, default=True
-        Whether to fail if version requirements are not fulfilled.
+    reinstantiate : bool, optional (default=False)
+        If True, convert the flow description into a concrete model instance
+        using the flow's extension (e.g., sklearn). If conversion fails and
+        ``strict_version`` is True, an exception will be raised.
+    strict_version : bool, optional (default=True)
+        When ``reinstantiate`` is True, whether to enforce exact version
+        requirements for the extension/model. If False, a fallback flow may
```
SimonBlanke: It is not a fallback flow though, is it? It is a new flow: https://github.com/openml/openml-python/blob/de22ee773c7911696d297a80997afd19aefe8217/openml/flows/functions.py#L137

Fallback would suggest some sort of 'backup' flow. Is this the case here (it might be)?

openml/flows/functions.py
EmanAbdelhaleem: I meant that it's a read-only operation that doesn't change variables, database or files. I am not sure about flows cuz I haven't gone very deep with its implementation, However, in evaluations, the list function has no caching as far as I know, and I think mostly this is the case for flows too.

Would you suggest modifying this to 
`None: read-only operation` ?

openml/flows/functions.py
EmanAbdelhaleem: nope it's not the case, I meant by `fallback` that it's an alternative flow, but I think `fallback` might be misleading, I think I will just make it `new flow`

openml/flows/functions.py
SimonBlanke: 'Read only' would be clear. It is easier to understand

openml/flows/functions.py
SimonBlanke: Okay new flow then. It is better to be more precise.

PR #1570, opened 2025-12-29
[ENH] Add `OpenMLAuthenticationError` for clearer API key error handling

### LINKED ISSUE #1562: [ENH] Add `OpenMLAuthenticationError` and improve Error Message When API Key is Invalid
Just a minor thing: Errors for authentication and authorization is currently both handled by `OpenMLNotAuthorizedError`. Here is the difference:

  | Concept        | Question           | Example                                        |
  |----------------|--------------------|------------------------------------------------|
  | Authentication | "Who are you?"     | No API key, invalid key, expired key           |
  | Authorization  | "Are you allowed?" | Valid key, but you didn't upload this resource |

So we could handle this case with a new error-class: https://github.com/openml/openml-python/blob/main/openml/_api_calls.py#L520
I would suggest a new class called `OpenMLAuthenticationError` that gives the user more information when the API key is invalid. It could look somewhat like this:

```python
class OpenMLAuthenticationError(...):
    """Exception raised when API authentication fails.

    This typically occurs when:
    - No API key is configured
    - The API key is invalid or expired
    - The API key lacks permissions for the requested operation
    - ...
    """

    def __init__(self, message: str):
        help_text = (
            "\n\nTo fix this:\n"
            "Get your API key at ...\n"
            "Run ... to configure apikey\n"
        )
        super().__init__(message + help_text)
```

Omswastik-11:
## Overview

This PR introduces a new **`OpenMLAuthenticationError`** exception to clearly distinguish **authentication errors** (invalid or missing API key) from **authorization errors** (valid API key without sufficient permissions).

---

### **New Exception**

* Added **`OpenMLAuthenticationError`** in `exceptions.py`
* Inherits from `OpenMLServerError` for consistency
* Automatically appends helpful guidance with links to:

  * Getting an API key: [https://www.openml.org/](https://www.openml.org/)
  * OpenML authentication documentation
* Includes a clear docstring explaining the difference between authentication and authorization errors

---

### **Updated Error Handling**

* Updated `_api_calls.py` to:

  * Import and raise `OpenMLAuthenticationError` for authentication failures

---

### **Tests Updated**

* Updated `test_authentication_endpoints_requiring_api_key_show_relevant_help_link`

  * Now expects `OpenMLAuthenticationError` instead of `OpenMLNotAuthorizedError`
  * Continues to assert that helpful guidance is included in the error message

---

Fixes #1562

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -114,9 +114,9 @@ jobs:
         fi
 
         if [ "${{ matrix.sklearn-only }}" = "true" ]; then
-          marks="sklearn and not production and not uses_test_server"
+          marks="sklearn and not production_server and not test_server"
         else
-          marks="not production and not uses_test_server"
+          marks="not production_server and not test_server"
         fi
 
         pytest -n 4 --durations=20 --dist load -sv $codecov -o log_cli=true -m "$marks"
@@ -131,9 +131,9 @@ jobs:
         fi
 
         if [ "${{ matrix.sklearn-only }}" = "true" ]; then
-          marks="sklearn and production and not uses_test_server"
+          marks="sklearn and production_server and not test_server"
         else
-          marks="production and not uses_test_server"
+          marks="production_server and not test_server"
         fi
 
         pytest -n 4 --durations=20 --dist load -sv $codecov -o log_cli=true -m "$marks"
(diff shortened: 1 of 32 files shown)


fkiraly (changes requested):
Waitasec, could you kindly explain why the `OpenMLNotAuthorizedError` gets removed?

Omswastik-11:
@fkiraly it is not completely removed but replaced . I have done it according to this :
https://github.com/openml/openml-python/issues/1562#issue-3763122986

fkiraly:
if i understand @SimonBlanke correctly, he does not want it replaced, but a new class added?

geetu040:
> if i understand @SimonBlanke correctly, he does not want it replaced, but a new class added?

@fkiraly the new class is added and the previous one is still there: [openml/exceptions.py](https://github.com/openml/openml-python/pull/1570/changes#diff-1dcbd766319414a0e485ee8df5ac1f919e720a57586f5567cb999cd0c182e2a6)

the confusion is probably coming from [openml/_api_calls.py](https://github.com/openml/openml-python/pull/1570/changes#diff-a894f3f1f23071f2589a09d3a906b3e738d7070195c3b5ba17a23e9b4fcc353f) where `OpenMLNotAuthorizedError` return is replaced with `OpenMLAuthenticationError`

PR #1571, opened 2025-12-29
[MNT] extend CI to newer python versions, deprecate python versions 3.8, 3.9 after EOL

fkiraly:
It seems the CI runs only on python versions 3.8 and 3.9 both of which have already reached end of life.

This PR updates the python versions, deprecating any logic that runs tests on python versions 3.8 and 3.9, or `scikit-learn` versions of that age.

The test matrix is updated to the currently supported python versions, 3.10-3.14.

### DIFF
--- a/.github/workflows/dist.yaml
+++ b/.github/workflows/dist.yaml
@@ -23,7 +23,7 @@ jobs:
   dist:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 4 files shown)


fkiraly:
FYI @PGijsbers

fkiraly:
superseded by https://github.com/openml/openml-python/pull/1579

PR #1577, opened 2025-12-30
[ENH] Improve Global Config Architecture

### LINKED ISSUE #1564: [ENH] Improve Global Config Architecture
The config module uses extensive globals and I agree with this comment: https://github.com/openml/openml-python/blob/main/openml/config.py#L395

This is problematic for the following reasons:
- Any code can modify `openml.config.apikey = "x"`. So bad or no encapsulation
- Global state persists between tests, causing flaky tests
- `openml.config.apikey = 123` silently accepts wrong type. So no validation

And I am sure there are more reasons.

I think it was done to have this kind of API instead of function-call-like syntax:

```python
openml.config.apikey = "my-key"
openml.config.server = "https://test.openml.org/api/v1/xml"
```

We can preserve this API and still get rid of most of these globals()/global, by defining a module level `__getattr__` and `__setattr__` and use a dataclass for the data encapsulation and validation:

```python
  from dataclasses import dataclass, replace

  @dataclass
  class OpenMLConfig:
      apikey: str = ""
      server: str = "https://www.openml.org/api/v1/xml"
      # ... 

  _config = OpenMLConfig()

  def __getattr__(name: str):
      if hasattr(_config, name):
          return getattr(_config, name)
      raise AttributeError(f"module 'openml.config' has no attribute '{name}'")

  def __setattr__(name: str, value):
      global _config
      if hasattr(_config, name):
          _config = replace(_config, **{name: value})
      else:
          raise AttributeError(f"module 'openml.config' has no attribute '{name}'")
```

This is still not a solution I am completely content with, but it is an improvement and not too difficult to implement.

satvshr:
#### Metadata
* Reference Issue: fixes #1564 
* New Tests Added: No
* Documentation Updated: No
* Change Log Entry: Replaces global variables with an instance of a dataclass `OpenMLConfig`.

#### Details 
  * Mainly Introduces an `OpenMLConfig` dataclass and an `OpenMLConfigManager` class to accumulate all config functions and varialbles into one class, and a single module-level instance `_config` in `openml.config`.
  * Reads/parses configuration defaults from `OpenMLConfig`.
  * Adds `__getattr__` so attribute reads on openml.config forward to the dataclass (preserving read compatibility).
  * Changes `openml/__init__.py` to route to the instance of `OpenMLConfigManager` when `openml.config` is called.
  * Moved `ConfigurationForExamples` inside `OpenMLConfigManager` to preserve API.
  
  * consolidates defaults and runtime state into one place,
  * simplifies consistent parsing and writing of the config file,
  * removes the use of global variables.

### DIFF
--- a/openml/tasks/task.py
+++ b/openml/tasks/task.py
@@ -1,13 +1,15 @@
 # License: BSD 3-Clause
 from __future__ import annotations
 
+import logging
 import warnings
 from abc import ABC
 from collections.abc import Sequence
 from enum import Enum
-from pathlib import Path
 from typing import TYPE_CHECKING, Any, ClassVar
 from typing_extensions import TypedDict
+
+import arff
 
 import openml._api_calls
 from openml import datasets
@@ -19,6 +21,9 @@ from .split import OpenMLSplit
 if TYPE_CHECKING:
     import numpy as np
     import pandas as pd
+
+
+logger = logging.getLogger(__name__)
 
 
 # TODO(eddiebergman): Should use `auto()` but might be too late if these numbers are used
@@ -177,18 +182,6 @@ class OpenMLTask(OpenMLBase):
 
         return self.split.get(repeat=repeat, fold=fold, sample=sample)
 
-    def _download_split(self, cache_file: Path) -> None:
-        # TODO(eddiebergman): Not sure about this try to read and error approach
-        try:
-            with cache_file.open(encoding="utf8"):
-                pass
-        except OSError:
-            split_url = self.estimation_procedure["data_splits_url"]
-            openml._api_calls._download_text_file(
-                source=str(split_url),
-                output_path=str(cache_file),
-            )
-
     def download_split(self) -> OpenMLSplit:
         """Download the OpenML split for a given task."""
         # TODO(eddiebergman): Can this every be `None`?
@@ -198,9 +191,23 @@ class OpenMLTask(OpenMLBase):
 
         try:
             split = OpenMLSplit._from_arff_file(cached_split_file)
-        except OSError:
+            logger.debug("Loaded file from cache: %s", str(cached_split_file))
+        except (OSError, arff.BadDataFormat):
+            logger.info("Failed to load file from cache: %s", str(cached_split_file))
+            if cached_split_file.exists():
+                logger.debug("Cleaning up old file")
+                cached_split_file.unlink()
             # Next, download and cache the associated split file
-            self._download_split(cached_split_file)
+            split_url = self.estimation_procedure["data_splits_url"]
+            openml._api_calls._download_text_file(
+                source=str(split_url),
+                output_path=str(cached_split_file),
+            )
+            if cached_split_file.exists():
+                logger.info("New file created of size %d", cached_split_file.stat().st_size)
+            else:
+                logger.info("Failed to create new file")
+
             split = OpenMLSplit._from_arff_file(cached_split_file)
 
         return split
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -34,15 +34,27 @@ jobs:
         sklearn-only: ["true"]
 
         exclude:
-          # incompatible version combinations
+          # (python, sklearn) combinations for which there is no PyPI release
+          # scikit-learn 1.3
           - python-version: "3.13"
             scikit-learn: "1.3.*"
+          - python-version: "3.14"
+            scikit-learn: "1.3.*"
+          # scikit-learn 1.4
           - python-version: "3.13"
             scikit-learn: "1.4.*"
           - python-version: "3.14"
-            scikit-learn: "1.3.*"
-          - python-version: "3.14"
             scikit-learn: "1.4.*"
+          # scikit-learn 1.5
+          - python-version: "3.14"
+            scikit-learn: "1.5.*"
+          # scikit-learn 1.6
+          - python-version: "3.14"
+            scikit-learn: "1.6.*"
+          # scikit-learn 1.7 is installed with pandas 3
+          - python-version: "3.10"
+            scikit-learn: "1.7.*"
+
 
         include:
           # Full test run on ubuntu, 3.14
@@ -64,14 +76,6 @@ jobs:
             sklearn-only: "false"
             code-cov: true
 
-          # Pandas 2 run
-          - os: ubuntu-latest
-            python-version: "3.12"
-            scikit-learn: "1.5.*"
-            sklearn-only: "false"
-            pandas-version: "2.*"
-            code-cov: false
-
     steps:
     - uses: actions/checkout@v6
       with:
@@ -82,15 +86,21 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
 
-    - name: Install test dependencies, scikit-learn, and optional pandas
+    - name: Install test dependencies, scikit-learn, and pandas
       shell: bash
       run: |
         python -m pip install --upgrade pip
         pip install -e .[test] scikit-learn==${{ matrix.scikit-learn }}
-        
-        if [ "${{ matrix.pandas-version }}" != "" ]; then
-          echo "Installing specific pandas version: ${{ matrix.pandas-version }}"
-          pip install "pandas==${{ matrix.pandas-version }}"
+
+        # scikit-learn 1.7+ requires pandas 3.x, earlier versions use pandas 2.x
+        version="${{ matrix.scikit-learn }}"
+        major=$(echo "$version" | cut -d. -f1)
+        minor=$(echo "$version" | cut -d. -f2)
+
+        if [[ "$major" -gt 1 ]] || { [[ "$major" -eq 1 ]] && [[ "$minor" -ge 7 ]]; }; then
+          pip install "pandas==3.*"
+        else
+          pip install "pandas==2.*"
         fi
 
     - name: Store repository status
@@ -103,20 +113,26 @@ jobs:
 
     - name: Clone Services
       if: matrix.os == 'ubuntu-latest'
+      id: clone-services
       run: |
         git clone --depth 1 https://github.com/openml/services.git
 
     - name: Start Docker Services
+      id: start-services
       if: matrix.os == 'ubuntu-latest'
       working-directory: ./services
       run: |
-        docker compose --profile rest-api --profile minio up -d
+        chmod -R a+rw ./data
+        chmod -R a+rw ./logs
+        docker compose --profile rest-api --profile minio --profile evaluation-engine up -d
 
         echo "Waiting for PHP API to boot..."
         timeout 60s bash -c 'until [ "$(docker inspect -f {{.State.Health.Status}} openml-php-rest-api)" == "healthy" ]; do sleep 5; done'
 
         echo "Final Verification: Gateway Connectivity..."
         curl -sSfL http://localhost:8000/api/v1/xml/data/1 | head -n 15
+
+        docker container ls
 
     - name: Show installed dependencies
       run: python -m pip list
@@ -173,8 +189,13 @@ jobs:
         fail_ci_if_error: true
         verbose: true
 
+    - name: Dump server logs
+      if: always() && steps.start-services.outcome == 'success'
+      run: |
+        docker logs openml-php-rest-api -t
+
     - name: Cleanup Docker setup
-      if: matrix.os == 'ubuntu-latest' && always()
+      if: always() && steps.clone-services.outcome == 'success'
       run: |
         sudo rm -rf services
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -126,6 +126,7 @@ version = {attr = "openml.__version__.__version__"}
 
 # https://docs.pytest.org/en/7.2.x/reference/reference.html#ini-options-ref
 [tool.pytest.ini_options]
+log_level="DEBUG"
 testpaths = ["tests"]
 minversion = "7.0"
 xfail_strict = true
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -286,12 +286,19 @@ def with_server(request):
 
 @pytest.fixture(autouse=True)
 def with_test_cache(test_files_directory, request):
+    # Skip this fixture for TestBase subclasses - they manage their own cache directory
+    # in setUp()/tearDown(). Having both mechanisms fight over the global config
+    # causes race conditions.
+    if request.instance is not None and isinstance(request.instance, TestBase):
+        yield
+        return
+
     if not test_files_directory.exists():
         raise ValueError(
             f"Cannot find test cache dir, expected it to be {test_files_directory!s}!",
         )
     _root_cache_directory = openml.config._root_cache_directory
-    tmp_cache = test_files_directory / request.node.name
+    tmp_cache = test_files_directory / request.node.nodeid.replace("/", ".").replace("::", ".")
     openml.config.set_root_cache_directory(tmp_cache)
     yield
     openml.config.set_root_cache_directory(_root_cache_directory)
--- a/tests/test_datasets/test_dataset_functions.py
+++ b/tests/test_datasets/test_dataset_functions.py
@@ -530,10 +530,6 @@ class TestOpenMLDataset(TestBase):
         datasets_cache_dir = os.path.join(openml.config.get_cache_directory(), "datasets")
         assert len(os.listdir(datasets_cache_dir)) == 0
 
-    @pytest.mark.skipif(
-        os.getenv("OPENML_USE_LOCAL_SERVICES") == "true",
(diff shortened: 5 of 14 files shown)


jgyasu:
Can you please add more details in your PR description?

satvshr:
It's not ready for review yet! Hence a draft.

satvshr:
@SimonBlanke if we make the dataclass `OpenMLConfig` frozen, we will have to change everything from the current `openml.config._config.cachedir = _root_cache_directory` to `_config = replace(_config, cachedir=_root_cache_directory)`, it is your call. In my opinion, it is a good idea to freeze it and use `_config = replace(_config, cachedir=_root_cache_directory)` everywhere (hence making it immutable) while I try to find a better fix to the problem.

For anyone else trying to help out, kindly refer to the issue attached to the PR (#1564) to get into the loop on the discussion.

ps: Even though this is a temporary fix it seems like a good step in the right direction for having a proper config. I think I would like to add and remove a few things to build on top of this once ive a better idea though (and once this PR is closed).

fkiraly (changes requested):
You changed the public interface to the config and apparently users have to write to a private module. That is not good design. Please leave the public APIs inteacts, otherwise this would be a breaking change.

Also, in the summary for this PR, please explain what you change and why.

satvshr:
> You changed the public interface to the config and apparently users have to write to a private module. That is not good design. Please leave the public APIs inteacts, otherwise this would be a breaking change.

I know that the public interface was changed, and that it shouldnt be changed, but the interface HAS to be changed if I follow Simon's proposed design in the issue description in #1564 given modular `__setattr__` is not allowed by python, which was part of the design proposed by @SimonBlanke in the issue, hence I went forward with the "minimum" changes required for the proposed architecture, for him to take a look at and make further decisions.

fkiraly:
> I know that the public interface was changed, and that it shouldnt be changed, 

It is a key requirement that it does not change. Or it will be a breaking change.

> but the interface HAS to be changed if I follow Simon's proposed design in the issue description in https://github.com/openml/openml-python/issues/1564

this is wrong. I would appreciate if you would not assert this with such certainty.

It is true that modular `__setattr__` does not work, but that does not mean that the design is impossible.

Also, you should be able to catch that a change in design that touches 13 files is perhaps not the right way to go.

>I went forward with the "minimum" changes required for the proposed architecture, for him to take a look at and make further decisions.

No, the minimal changes imo would be the ones that preserve the current API and stay close to @SimonBlanke´s proposal, instead of modifying 13 files. Changing 13 files is not "minimal" in the sense of scope and cohesion.

You should have pointed out the issue with `__setattr__` in the issue, btw.

Anyway, the way to go imo (in order to be faithful to @SimonBlanke´s design) would be:

* turn `config` into a private module
* in root `__init__`, modify `__getattr__`, so that `config` points to a global instance of the config class
* in that class, you can override `__setattr__` as well as `__getattr__` (or directly have attributes written).

fkiraly:
In terms of communication:

* you identified a problem with the design, great!
* you tried to propose a solution to the problem, great!
* however, it would have been better if you had immediately highlighted and communicated where you are at, in issue #1564:
    * that `__setattr__` does not work as designed, and that you will attempt to fix the design - you communicated that, great!
    * but it would be good if you spell out the design that you had in mind, on issue #1564, once worked out. Plus pros/cons on alternatives you considered.
    * then, if no alternatives you considered, fit the key requirement (do not change public API), highlight that explicitly, since that is reaching a blocker!

satvshr:
> Also, you should be able to catch that a change in design that touches 13 files is perhaps not the right way to go.

I definitely knew it was wrong, I was going to bring it up in tomorrow's meeting, and had no intention of getting this PR merged in the current state, hence should have probably kept it in draft. Though I think I should have said that more explicity in the PR itself.


> No, the minimal changes imo would be the ones that preserve the current API and stay close to @SimonBlanke´s proposal, instead of modifying 13 files. Changing 13 files is not "minimal" in the sense of scope and cohesion.

Now that there is a way around this, which involves not changing 13 files I do agree I was wrong, but just to give an idea of what was going through my head, by "minimal" I meant at that point of time to get all globals into 1 dataclass, I (felt I) had to change the 13 files to continue with Simon's design.

> * it would have been better if you had immediately highlighted and communicated where you are at

I was doing that throughout no? I didn't drop an update about a better design, because I am still working on it, so there was no update about that.

> * highlight that explicitly

I should have definitely done that.

openml/utils.py
```diff
@@ -18,8 +18,6 @@
 import openml._api_calls
 import openml.exceptions
 
-from . import config
```
satvshr: removed since it would end up importing the file and not the instance of `OpenMLConfigManager`

openml/__init__.py
```diff
@@ -18,9 +18,11 @@
 # License: BSD 3-Clause
 from __future__ import annotations
 
+from typing import TYPE_CHECKING
+
 from . import (
     _api_calls,
-    config,
+    config as _config_module,
```
SimonBlanke: As @fkiraly suggested, making the config into a private module would be better here. I could see it 'clashing' with the instance further down. This could lead to these two being different:
```python
print(type(openml.config))              # <class 'OpenMLConfigManager'>
print(type(sys.modules['openml.config']))  # <class 'module'>
```

Better to avoid this :-)

openml/cli.py
```diff
@@ -9,6 +9,8 @@
 from pathlib import Path
 from urllib.parse import urlparse
 
+from attr import fields
```
SimonBlanke: Here you are using the fields from 'attr' and use it for OpenMLConfig further down. I think you would get something like this here: 'NotAnAttrsClassError'
OpenMLConfig is a dataclass, so you should use the 'fields' from dataclass instead.

The tests all pass for your PR, but this part has no test-coverage:
https://app.codecov.io/gh/openml/openml-python/pull/1577?dropdown=coverage&src=pr&el=continue&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=openml

So this could have been dangerous to merge ;-)

openml/config.py
```diff
@@ -148,378 +53,416 @@ def _resolve_default_cache_dir() -> Path:
...
-    -------
-    cachedir : string
-        The current cache directory.
-
-    """
-    url_suffix = urlparse(server).netloc
-    reversed_url_suffix = os.sep.join(url_suffix.split(".")[::-1])  # noqa: PTH118
-    return os.path.join(_root_cache_directory, reversed_url_suffix)  # noqa: PTH118
 
+        self.set_retry_policy(config["retry_policy"], self._config.connection_n_retries)
```
SimonBlanke: It is difficult to see, because of the git-diff, but further up we already set these fields in `replace`. So this should be redundant. Do you agree?

satvshr: Good spot! I agree.

openml/__init__.py
satvshr: I do not know how tests were passing earlier, but once I renamed `config.py` to `_config.py` all imports had errors because (I assume) `import openml.config` was calling the module not the variable (the instance of `OpenMLConfigManager`). Made those changes too.

SimonBlanke:
@satvshr Could you comment on the test fails?

satvshr:
> @satvshr Could you comment on the test fails?

Tests were passing before merging with main (I requested a review before that), let me have a look

Edit: `test_cli` was recently added by Om, fixed it, cli.py had an import error, which was not tested since it never had a test before Om's PR.

SimonBlanke:
@satvshr This looks good to me. I am a bit worried, because of the test-fails from before: https://github.com/openml/openml-python/actions/runs/21062979947
It is strange to me, that only some jobs failed. I'll approve this PR, but I would like you to look into this again just to be sure.

satvshr:
I answered it above :D `test_cli` was recently added by Om. I fixed the bug, `cli.py` had an import error, which was not tested since it never had a test before Om's PR. In `cli.py` the import was: `from openml import config`, which was wrong and needed a fix.

fkiraly:
tests are failing on this - FYI @geetu040

geetu040:
> tests are failing on this - FYI @geetu040

I am not sure about these, they come from main. They pass for me on local. Are they observed on other PRs? Since it uses the production server, could it be that a faulty flow/dataset was there at the time of run?

satvshr:
@fkiraly @geetu040 Same tests fail in [this PR](http://github.com/openml/openml-python/actions/runs/21228880416/job/61082770429) where all I am doing is removing a skip test from a single test.

Edit: [Here](https://github.com/openml/openml-python/actions/runs/21228673827/job/61082120329?pr=1560) too.

geetu040:
yes that test skip doesnot look relevant to the failures

geetu040 (changes requested):
@satvshr please resolve merge conflicts.

openml/_config.py
```diff
@@ -0,0 +1,466 @@
...
+openml_logger = logging.getLogger("openml")
+
+
+def _resolve_default_cache_dir() -> Path:
+    user_defined_cache_dir = os.environ.get("OPENML_CACHE_DIR")
+    if user_defined_cache_dir is not None:
+        return Path(user_defined_cache_dir)
+
+    if platform.system().lower() != "linux":
+        return Path("~", ".openml")
```
geetu040:
```suggestion
        return Path("~", ".openml").expanduser()
```
so `"~"` is not considered a literal directory name

openml/_config.py
```diff
@@ -0,0 +1,466 @@
...
+
+    xdg_cache_home = os.environ.get("XDG_CACHE_HOME")
+    if xdg_cache_home is None:
+        return Path("~", ".cache", "openml")
```
geetu040:
```suggestion
        return Path("~", ".cache", "openml").expanduser()
```
same as above

openml/_config.py
```diff
@@ -0,0 +1,466 @@
...
+        cache_dir,
+    )
+    return Path(xdg_cache_home)
+
+
+@dataclass
+class OpenMLConfig:
+    """Dataclass storing the OpenML configuration."""
+
+    apikey: str = ""
```
geetu040: we should have it such that `apikey` can be `None`

openml/_config.py
```diff
@@ -0,0 +1,466 @@
...
+
+    def __init__(self) -> None:
+        self.console_handler: logging.StreamHandler | None = None
+        self.file_handler: logging.handlers.RotatingFileHandler | None = None
+
+        self.OPENML_CACHE_DIR_ENV_VAR = "OPENML_CACHE_DIR"
+        self.OPENML_SKIP_PARQUET_ENV_VAR = "OPENML_SKIP_PARQUET"
+        self._TEST_SERVER_NORMAL_USER_KEY = "normaluser"
+
+        self._user_path = Path("~").expanduser().absolute()
```
geetu040: you can remove this, it's not used anywhere

openml/_config.py
```diff
@@ -0,0 +1,466 @@
...
+
+    _FIELDS = {  # noqa: RUF012
+        "apikey",
+        "server",
+        "cachedir",
+        "avoid_duplicate_runs",
+        "retry_policy",
+        "connection_n_retries",
+        "show_progress",
+    }
```
geetu040: why do we need extra `_FIELDS`, can we not use keys in `OpenMLConfig`?

openml/_config.py
```diff
@@ -0,0 +1,466 @@
...
+    def overwrite_config_context(self, config: dict[str, Any]) -> Iterator[dict[str, Any]]:
+        """Overwrite the current configuration within a context manager."""
+        existing_config = self.get_config_as_dict()
+        merged_config = {**existing_config, **config}
+
+        self._setup(merged_config)
+        yield merged_config
+        self._setup(existing_config)
+
+    class ConfigurationForExamples:
```
geetu040: can we not keep this out of `OpenMLConfigManager`?

openml/_config.py
```diff
@@ -0,0 +1,466 @@
...
+
+            self._manager._config = replace(
+                self._manager._config,
+                server=cast("str", self._last_used_server),
+                apikey=cast("str", self._last_used_key),
+            )
+            type(self)._start_last_called = False
+
+
+_config = OpenMLConfigManager()
```
geetu040: this object is called `_config` and the file is also called `_config`, though it's handled in `__init__.py`, but it doesn't look good, please change one of them

openml/_config.py
satvshr: We can...will try doing that.

openml/_config.py
satvshr: We cannot because this is happening during `__init__` before `_config` exists

geetu040: would this not work?
```py
from dataclasses import fields

[f.name for f in fields(OpenMLConfig)]
```

satvshr: Oh! This should work, let me give it a try!

openml/_config.py
```diff
@@ -0,0 +1,459 @@
...
+    server: str = "https://www.openml.org/api/v1/xml"
+    cachedir: Path = field(default_factory=_resolve_default_cache_dir)
+    avoid_duplicate_runs: bool = False
+    retry_policy: Literal["human", "robot"] = "human"
+    connection_n_retries: int = 5
+    show_progress: bool = False
+
+    def __setattr__(self, name: str, value: Any) -> None:
+        if name == "apikey" and value is not None and not isinstance(value, str):
+            raise ValueError("apikey must be a string or None")
```
geetu040: Non-Blocker: Do we really need this? I don't think so, it wasn't here before either.

fkiraly:
@PGijsbers, could you kindly review? This touches a lot of files so it would be great if you could have a look.

openml/evaluations/evaluation.py
```diff
@@ -3,7 +3,7 @@
 
 from dataclasses import asdict, dataclass
 
-import openml.config
+import openml
```
PGijsbers:
```suggestion
```
If only this line changed and nothing else, is it an unused import?

openml/setups/setup.py
```diff
@@ -4,7 +4,7 @@
 from dataclasses import asdict, dataclass
 from typing import Any
 
-import openml.config
+import openml
```
PGijsbers:
```suggestion
```

openml/study/functions.py
```diff
@@ -8,8 +8,8 @@
 import pandas as pd
 import xmltodict
 
+import openml
```
PGijsbers:
```suggestion
```

openml/tasks/task.py
```diff
@@ -9,8 +9,8 @@
 from typing import TYPE_CHECKING, Any, ClassVar
 from typing_extensions import TypedDict
 
+import openml
```
PGijsbers:
```suggestion
```

openml/base.py
```diff
@@ -8,8 +8,8 @@
 
 import xmltodict
 
+import openml
```
PGijsbers:
```suggestion
```

openml/_config.py
PGijsbers:
```suggestion
        if name == "apikey" and not isinstance(value, (type(None), str)):
            raise TypeError("apikey must be a string or None")
```
Inappropriate types should raise a TypeError. ValueErrors are for values of the correct type but outside a legal range (e.g., if we would require a positive integer but a negative one is supplied).

openml/_config.py
```diff
@@ -0,0 +1,459 @@
...
+
+        if name in self._FIELDS:
+            # write into dataclass, not manager (prevents shadowing)
+            if name == "cachedir":
+                object.__setattr__(self, "_root_cache_directory", Path(value))
+            object.__setattr__(self, "_config", replace(self._config, **{name: value}))
+            return None
+
+        object.__setattr__(self, name, value)
+        return None
```
PGijsbers, line 131:
```suggestion
        if name in self._FIELDS:
            # write into dataclass, not manager (prevents shadowing)
            if name == "cachedir":
                object.__setattr__(self, "_root_cache_directory", Path(value))
            return object.__setattr__(self, "_config", replace(self._config, **{name: value}))

        return object.__setattr__(self, name, value)
```
I am probably missing something, but why would this not work? The names don't overlap. If it is a precaution against accidentally introducing a name in ConfigManager that is already in Config, then I think it would be wiser to raise an error.

satvshr: It isn't about name overlaps, it's to prevent infinite recursion during initialization. Since `__setattr__` is overridden to route values into `self._config`, we have to use `object.__setattr__` to create the `_config` object itself. Without this 'bypass' list, `__init__` would trigger `__setattr__`, which would try to access `self._config` before it exists, causing it to crash.

geetu040:
@fkiraly we have @PGijsbers's approval in https://github.com/openml/openml-python/pull/1577#pullrequestreview-3876017449
This should be good to go then.

Issue #1544, opened 2025-12-23
[MNT] Test failures in CI and local


jgyasu:
**Update - 29 Dec, 2025**: Now the tests are failing in local as well as on CI as pointed by @fkiraly 

While following the testing-your-installation guide, running pytest locally results in multiple test failures.

We should investigate the failing tests and debug them.

Todos after resolving tests:

* remove the `pytest.xfail`s
* set `sklearn-only: "false"` in the `include` jobs in the `test.yml` `test` job

Note: originally it was thought that only local tests fail while `main` CI passes. This was a confusion that was caused by a state of the repo in which `main` was no longer used and `develop` used as the main branch, and for releases. This has since been rectified (`main` is now the main branch), so remote CI and local failures agree.

rohansen856:
I would like to work on this issue. Looking into it now.

PGijsbers:
Quite a few tests rely on connections to the servers (both production and test). In some cases they even rely on server state. This means that it's possible for tests to fail "randomly". I'm on mobile so I can't easily search but I do believe there is already a different issue about making tests rely on connections less by a combination of reducing the surface each individual test tests, and using mocked responses.

rohansen856:
Did some digging and tried to backtrack the test calls and found this:

**Call Stack:**
```
openml.tasks.get_task(115)
  → dataset.retrieve_class_labels(task.target_name)
    → self.features.values()
      → self._load_features()
        → _get_dataset_features_file()
          → _get_features_xml(dataset_id=20)
            → OpenMLServerException (code 274)
```

The 20th dataset throws error when called from local: `https://test.openml.org/api/v1/xml/data/features/20`

Got this when i checked the url myself:

<img width="1919" height="971" alt="Image" src="https://github.com/user-attachments/assets/0d9586a1-1a40-4fe2-b037-d38bb006c69d" />

<img width="1919" height="970" alt="Image" src="https://github.com/user-attachments/assets/29d181d7-8189-43df-9667-5c244aafb238" />


Although i am still looking as to the root cause of the failures i have these doubts about why it locally fails but CI Passes:**
- CI configuration in `/.github/workflows/test.yml` runs tests in two separate steps:
  1. **Test server tests** (line 106-113): Excludes `production` marker ->`marks='sklearn and not production'`
  2. **Production server tests** (line 114-121): Only runs `production` marker ->`marks='sklearn and production'`
[...]

PR #1579, opened 2025-12-30
[MNT] extend CI to newer python versions, deprecate python versions 3.8, 3.9 after EOL, marking further failing tests as `xfail`

JATAYU000:
The CI runs only on python versions 3.8 and 3.9 both of which have already reached end of life.

This PR updates the python versions, deprecating any logic that runs tests on python versions 3.8 and 3.9, or `scikit-learn` versions of that age.

#### Metadata
Reference Issue: #1544 

Depends on https://github.com/openml/openml-python/pull/1584 fofr a fix, which should be merged first.

#### Details 

* The test matrix is updated to python versions, 3.10-3.13.
* Further failing tests are skipped using `mark.xfail` with  `reason="failures_issue_1544" ` for all the remaining failed tests (after #1572) in issue: #1544

### DIFF
--- a/tests/test_datasets/test_dataset.py
+++ b/tests/test_datasets/test_dataset.py
@@ -303,7 +303,6 @@ def test_get_feature_with_ontology_data_id_11():
     assert len(dataset.features[2].ontologies) >= 1
     assert len(dataset.features[3].ontologies) >= 1   
 
-@pytest.mark.xfail(reason="failures_issue_1544")
 def test_add_remove_ontology_to_dataset():
     did = 1
     feature_index = 1
@@ -311,7 +310,6 @@ def test_add_remove_ontology_to_dataset():
     openml.datasets.functions.data_feature_add_ontology(did, feature_index, ontology)
     openml.datasets.functions.data_feature_remove_ontology(did, feature_index, ontology)    
 
-@pytest.mark.xfail(reason="failures_issue_1544")
 def test_add_same_ontology_multiple_features():
     did = 1
     ontology = "https://www.openml.org/unittest/" + str(time())
--- a/tests/test_datasets/test_dataset_functions.py
+++ b/tests/test_datasets/test_dataset_functions.py
@@ -244,7 +244,6 @@ class TestOpenMLDataset(TestBase):
         assert len(datasets) == 2
         _assert_datasets_retrieved_successfully([1, 2])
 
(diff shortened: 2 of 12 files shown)


fkiraly:
FYI @JATAYU000, I have merged #1571 into this for simplicity.

This also allows us to track down any further failing tests.

JATAYU000:
> There are still a few failing tests, and also fixture construction errors - could you look at those too, @JATAYU000?

Yes will look into it

JATAYU000:
@fkiraly #1584 needs to be merged first for fixing 
```
AsertionError
assert False
 +  where False = exists()
 ```
 
 Would like your inputs on this fix

fkiraly:
ok, I will also check whether that resolves the failures here.

fkiraly:
this fixed the issue, so you should remove the `xfail`s from the fixed tests (see CI runs)

Not in this PR, but in #1584, and I will merge the changes to here as well for more extensive testing.

JATAYU000:
@fkiraly I am a bit lost on whats happening here, How are the all 48 tests (marked as xfail) passing under `sk-only:False` ?, #1584 only fixes a few of those

fkiraly:
A PR by @PGijsbers got merged which seems to fix these.
https://github.com/openml/openml-python/pull/1585

Unfortunately, this was unexpected since I thought he was on holiday until Jan - but it seems to be a fix, so we should adapt this PR to the new repository state.

I would suggest to revert the tests that are now fixed, could you quickly do that?

JATAYU000:
> I would suggest to revert the tests that are now fixed, could you quickly do that?

Sure, Thank you

fkiraly:
@JATAYU000, you still need to deal with the `sk-only: False` tests

JATAYU000:
> @JATAYU000, you still need to deal with the `sk-only: False` tests

Yes those tests marks are removed, rest are passing till now the check is still in progress

JATAYU000:
> @JATAYU000, you still need to deal with the `sk-only: False` tests

@fkiraly Can you point me which tests under `sk-only: False` are not handled yet? because I have updated all of them as far as I know [reference workflow fail](https://github.com/openml/openml-python/actions/runs/20616780464/job/59210985892).

fkiraly (approved):
Looks like we have the failing tests under control now.

PR #1583, opened 2025-12-31
[ENH] experimental design for model indexing framework

fkiraly:
This PR is an experimental draft showcasing:

* a light-weight packaging layer for objects of any kind, using python code serialization rather than "translation"
* a new taxon "model" which corresponds to (uninstantiated) classes
* an example implementation for tabular classifiers, which adds records for two examples: `XGBClassifier` from `xgboost`, and `AutoSklearnClassifier` from `auto-sklearn`

Note: this is currently not using any backend - the idea is that the light-weight packaging layer can be easily serialized by the backend.

For a broader design of this and adjacent features, see the architecture document https://github.com/gc-os-ai/openml-project-dev/pull/14

Depends on:

* https://github.com/openml/openml-python/pull/1612

### Usage pattern for end users

End users would interact with the layer as follows:

```python
import openml

clf1 = openml.get("XGBClassifier")
clf2 = openml.get("AutoSklearnClassifier")
clf3 = openml.get("RandomForestClassifier")
```
Either line will:
* if the required soft dependencies (e.g., `xgboost`, `auto-sklearn`, `scikit-learn`) are present, directly import the class from the package.
* if the required soft dependencies are not present, raise an informative error message

The strings live in a unique namespace for objects - usually, for widely known packages (but not necessarily), they will correspond to class names

This could be further extended to:
* a `deps` utility that, for a namespace ID provides required dependencies - the tags already are inspectable here
* tracking of dependencies on a more granular level for reproducibility, e.g., the exact version with which a benchmark in which a model figures was carried out, via `pip freeze`.

Versioning inside `openml` is consciously avoided in this design, as versioning already happens in the dependencies which are tracked by the tag system.

The logic internally works through:

* a registry lookup mechanism
* a unified, but openml-private `materialize` interface that obtains the class from the package layer

### Usage patterns for estimator contributors

Currently, an extender would interact with this by manually adding a class in `openml.models.classifiers`, in the simplest case they are thin wrappers pointing to a third party location:

```python
from openml.models.apis import _ModelPkgClassifier

class OpenmlPkg__AutoSklearnClassifier(_ModelPkgClassifier):
    _tags = {
        "pkg_id": "AutoSklearnClassifier",
        "python_dependencies": "auto-sklearn",
    }

    _obj = "autosklearn.classification.AutoSklearnClassifier"
```

But this class could also be reliant on a full python definition of a class itself, with `_obj` pointing to a python file, or the method `_materialize` implementing explicit python code.

This can also be automated further in the context of a database backend:

* creating a dynamic packaging class from an in-memory object
* call of this from a `publish` method that automatically extracts tags etc, similar to `OpenMLFlow.publish`
* or, a `publish` method applying to an entire package, crawling it, and extracting pointer records to all `sklearn` estimators

### Intended backend interaction pattern

A database backend would be inserted as follows.

#### for publish / populate

* `publish` creates a dynamic package class inheriting from the new `_BasePkg`
* calls `serialize` on it to obtain a string
* calls `get_tags` to obtain a `dict` of packaging metadata
* optionally, adds data about publish process metadata (author, timestamp, etc)
* publishes both to the database - DB API call

#### for query

* DB API call for lookup mechanism
* retrieve previously published record - also DB API call
* invert `serialize` by `zlib` decompress and `exec` to obtain a dynamic python package class
* `materialize` from that class


openml/base/_base_pkg.py
```diff
@@ -0,0 +1,118 @@
...
+
+class _BasePkg(BaseObject):
+    _tags = {
+        "python_dependencies": None,
+        "python_version": None,
+        # package register and manifest
+        "pkg_id": None,  # object id contained, "__multiple" if multiple
+        "pkg_obj": "reference",  # or "code"
+        "pkg_obj_type": None,  # openml API type
+        "pkg_compression": "zlib",  # compression
```
SimonBlanke, line 22: This just selects the compression algorithm, right?

openml/base/_base_pkg.py
```diff
@@ -0,0 +1,118 @@
...
+        raise RuntimeError("abstract method")
+
+    def serialize(self):
+        cls_str = class_to_source(type(self))
+        compress_method = self.get_tag("pkg_compression")
+        if compress_method in [None, "None"]:
+            return cls_str
+
+        cls_str = cls_str.encode("utf-8")
+        exec(f"import {compress_method}")
```
SimonBlanke, line 54: I guess this seems save, because the tag is not user-supplied code. But it still makes me nervous to see this. We could just add a simple mapping layer to restrict what can be executed in here. Something like this:
```
COMPRESSION_MODULES = {
  "zlib": "zlib",
  "gzip": "gzip",
  "bz2": "bz2",
  "lzma": "lzma",
}
```

openml/base/_base_pkg.py
```diff
@@ -0,0 +1,118 @@
...
+        return eval(f"{compress_method}.compress(cls_str)")
```
SimonBlanke, line 55: Couldn't we just use importlib for this? Something like this:

```python
module = importlib.import_module(compress_method)
cls_bytes = cls_str.encode("utf-8")
return module.compress(cls_bytes)
```

openml/models/base/_base.py
```diff
@@ -0,0 +1,65 @@
...
+        if pkg_obj == "reference":
+            from skbase.utils.dependencies import _safe_import
+
+            obj_loc = self._obj
+            pkg_name = self.get_tag("pkg_pypi_name")
+
+            return _safe_import(obj_loc, pkg_name=pkg_name)
+
+        if pkg_obj == "code":
+            exec(self._obj)
```
SimonBlanke, line 54: Is this line for importing a module again? Can't we use something different then 'exec'?

openml/models/base/_base.py
```diff
@@ -0,0 +1,65 @@
...
+
+            return obj
```
SimonBlanke, line 56: What is 'obj' where is it defined? I do not see it in this scope.

PGijsbers (commented):
Hi Franz,

I was writing up a longer response at first, but in the process I got more unsure about the proposal, so I'd like to clarify a few things first.
Some of these remarks reference the design document you provided separately.
Ultimately, it will be best to provide a cohesive proposal on our [Design and Feature Request](https://github.com/orgs/openml/discussions/categories/openml-design-and-feature-requests) discussion board, as I think that this proposal affects the core infrastructure and/or responsibilities of OpenML, which would mean it will ultimately also need to be discussed with all core contributors (also outside of openml-python).
But I would be ok first continuing the discussion here.

For the public, I will use some terms that Franz introduced there, most importantly:
 - Estimator Classes (e.g, sklearn's `RandomForestClassifier`, autosklearns `AutoSklearnClassifier`)
 - Estimator Instances (e.g., sklearn's `RandomForestClassifier(n_trees=42)`)
 - Fitted Instances (e.g., `RandomForestClassifier(n_trees=42).fit(X, y)`)

The _most important_ ideas of this proposal are to: 
 - introduce a standardized interface to which each Estimator Class must adhere (not explicitly part of this PR). The interface would depend on the class of problems the Estimator supports.
 - provide a repository to easily share these standardized estimators in a way that makes loading them in a uniform manner possible.
 - replace the serialization of Estimator Classes into distinct OpenML concepts (e.g., separate schemas for parameters) with serialization into a single string, reducing the complexity of working with Flows.

Do you agree?

For completeness, these ideas are also part of the proposal, but they rely on the above ideas:
- maintaining a built-in estimator repository that is shipped with the package (possibly based on some leaderboard)
- the standardized interface can be used to significantly simplify "run model on task"-like functions which currently program to implementations.

 **Creating Estimator Instances**
The idea is to facilitate an easier, unified way to create Estimator Instances.
They could then create estimator instances with a reference to the name, so instead of:

```python
rf = RandomForestClassifier()
```

they would write:

```python
import openml
rf = openml.get("RandomForestClassifier")
```

I reckon that the main advantage here is that we would know the latter object adheres to whatever estimator interface we define. If the imported estimator does not normally adhere to the shared interface, but the package manager saved a reference to some wrapped version of the estimator, it is now easier to use it out of the box. In a world where the packages already would provide class definitions that adhere to the common interface (e.g., xgboost added their `XGBoostClassifier` sklearn interface - and let's assume that's good enough), the only benefit would be that if the user only knows the name they would get an informative error message on what package to install (or openml could install it for them). In cases where the estimator is a pipeline with multiple components, multiple import statements may be saved.
Is that correct?

**Separation of Estimator Class and Estimator Instance**
I mention this because I think it was explicitly in one of the design documents, and I think it points to a misconception (either of my understanding of your proposal/terminology, or of your understanding of OpenML terminology). I believe this is the current function of Flows (Estimator Classes) and Setups (Estimator Instance).
The Flow defines the algorithm (e.g., sklearn 1.6 RandomForestClassifier) and its hyperparameters (`n_trees`) and defaults, but _not_ any specific configuration.
When an experiment is ran with e.g., `RandomForestClassifier(n_trees=40, ...)` a Flow is made for `RandomForestClassifier` and the run result will point to that Flow alongside the Setup which specifies the actual configured hyperparameters (`n_trees=40, ...)`. The learned _parameters_ (of a fitted instance) are not saved in principle (but can technically still be stored as a "run attachment", but we don't really support this in a systematic way).
Would you agree that this mapping is correct? Whether or not the specifications of Flows/Setups should be changed to be better/easier/... is a separate notion.

**Serialization of Estimators**
Currently, OpenML allows structured storage about information of the hyperparameters of each algorithm (Flow).
This includes each hyperparameter's type, default value, and recommended range ([definition](https://github.com/openml/OpenML/blob/b1265bda1b52780a6f10a7d239e3f54914517c2f/openml_OS/views/pages/api_new/v1/xsd/openml.implementation.upload.xsd#L58)).
From my understanding, this proposal does away with this.
The only information about the Estimator that is retained is what is currently in the tags alongside the import path.
Is that correct?

The other part of the current flow structure is that it can be defined in terms of subflows (or components). 
As such, on OpenML currently `Pipeline([Imputer, RandomForest])` would result in three flows: the overarching flow and a separate subflow for `Imputer` and `RandomForest` each.
This proposal does away with this, and only maintains the top-level flow. Is that correct?

**Versioning**
The deferral of versioning to PyPI isn't entirely clear to me, especially in conjunction with the ability to instantiate Estimator Instances from a string.
How would the following code example work?

```python
import openml
rf = openml.get("RandomForestClassifier(hyperparameter_a=23)")
```
Assume RandomForestClassifier comes from a package with is currently on version 1.8 and given that:
 - parameter defaults may change from version to version (e.g., `hyperparameter_b`'s default was updated in version 1.5)
 - parameters may be introduced or removed from version to version (`hyperparameter_a` was removed in version 1.7)
 - behavior may change from version to version (e.g., a bug was fixed in version 1.6)
 - not all scikit-learn versions are available on all Python version (assuming there is no support for local builds)

**Unique Namespace**
Flows currently have two unique identifiers: the integer identifier _as well as_ the (name, external version) tuple.
Either can be used to uniquely identify the flow. 
While this is surely not the original vision, a specific version of a package is effectively just a namespace.
Keeping the "external version" to a constant would allow you to create a lookup-by-name for flows (`/flow/exists/{name}/{version}` REST API endpoint even supports indexed lookups).
It might be interesting to explore that direction as it would allow the implementation (at least of that aspect of the proposal) without changes to server infrastructure.


Thanks.

fkiraly:
> The _most important_ ideas of this proposal are to: 
> Do you agree?

Yes, with minor comments

> - introduce a standardized interface to which each Estimator Class must adhere (not explicitly part of this PR). The interface would depend on the class of problems the Estimator supports.

Yes - of course the same task can map to multiple APIs ultimately. I think the "best" design structures openML around managing unified APIs primarily, packages (e.g., sklearn, torch, etc) second.

> - provide a repository to easily share these standardized estimators in a way that makes loading them in a uniform manner possible.

Yes, plus using them in downstream services like benchmarks or leaderboards.

> - replace the serialization of Estimator Classes into distinct OpenML concepts (e.g., separate schemas for parameters) with serialization into a single string, reducing the complexity of working with Flows.

Yes - I think that is key to manage the crawling, creeping complexity that makes the schemas so hard to maintain or extend. It also removes the onus entirely from openml to manage serialization, if we implement this smartly.

> For completeness, these ideas are also part of the proposal, but they rely on the above ideas:
> - maintaining a built-in estimator repository that is shipped with the package (possibly based on some leaderboard)

Yes, although I think this would be "the" extremely attractive user facing feature that drops out of. The above are more core design elements that a user may not see or consciously perceive.

> - the standardized interface can be used to significantly simplify "run model on task"-like functions which currently program to implementations.

Yes. Again, it would remove much complexity and case distinctions which imo are quite unnecessary (if the right design were adopted).

> to the shared interface, but the package manager saved a reference to some wrapped version of the estimator, it is now easier to use it out of the box. In a world where the packages already would provide class definitions that adhere to the common interface (e.g., xgboost added their `XGBoostClassifier` sklearn interface - and let's assume that's good enough),

Yes

> the only benefit would be that if the user only knows the name they would get an informative error message on what package to install (or openml could install it for them). 

No, not the only benefit. Further benefits:

* single entry point for distribution, similar to pypi managing packages
* the ability to perform registry operations, such as tagging, search and retrieval, on top

> In cases where the estimator is a pipeline with multiple components, multiple import statements may be saved.

Yes.

>**Separation of Estimator Class and Estimator Instance**

> The Flow defines the algorithm (e.g., sklearn 1.6 RandomForestClassifier) and its hyperparameters (`n_trees`) and defaults, but _not_ any specific configuration.

Is that so? It is not the conceptual model I got from reading the docs. It may be a misconception on my side, or a design inaccuracy on side of `openml`.

The issue may come from flows being used as both, so not clearly separated. In benchmarks or reports, flows are being treated - and being referred to - as estimator instances, not estimator classes:
https://docs.openml.org/concepts/flows/

Maybe the subtle point is as follows. there is design inconsistency in `openml`, namely: that user facing workflows treat "flow" as an estimator instance - e.g., `model_to_flow` or benchmark runs, while the backend treats flows as estimator classes, and "extracts" the setup? At least in the docs, I do not see an explicit point in user workflows where flows and setups are separate (since a user does not generate flows in the first place but they are created for him).

The "separation" that I find important is the separation *for the user*, i.e., apparent in user facing workflows.

>Would you agree that this mapping is correct?

As said, I think it depends on the conceptual model, and there seem to be two that are not entirely congruent:

* the conceptual model in the backend 
* the user facing conceptual modle (or at least what a user will understand as such when interacting with the package)

> **Serialization of Estimators**
> Currently, OpenML allows structured storage about information of the hyperparameters of each algorithm (Flow).
> This includes each hyperparameter's type, default value, and recommended range From my understanding, this proposal does away with this.

It does not *need* to. We can still keep the richer metadata.

My main concern is - as already above - with the user workflows.

I do not see a lot of meaningful user interaction with the aforementioned things - hyperparameter type, default, recommended range, etc. If at all, it feels like unnecessary ballast in how a user specifies models.

Having said that, the backend can still keep track of this, but I am really wondering: why would we want to?

> The only information about the Estimator that is retained is what is currently in the tags alongside the import path.
> Is that correct?

Yes, although: in the tags, a lot of metadata could be tracked; and there is no a-priori need to also kill the concept of flows entirely.

> As such, on OpenML currently `Pipeline([Imputer, RandomForest])` would result in three flows: the overarching flow and a separate subflow for `Imputer` and `RandomForest` each.
This proposal does away with this, and only maintains the top-level flow. Is that correct?

No, I think it is more subtle:

* in the vanilla proposal, *only the lowest level components* get tracked, i.e., `Pipeline`, `Imputer`, `RandomForest` - so-called "atoms" or atomic objects, in line with propositional logic based specification language
* the composite `Pipeline([Imputer, RandomForest])` is automatically indexed by its unique string once the components are! It will not be stored as a separate flow unless expressly requested by the user.
* optionally, we can introduce something similar to aliasing, e.g., have "myflow" alias the `Pipeline([Imputer, RandomForest])`. If we track parameters etc, the parameters of "myflow" would be implied by those of the atomic objects.

This way, we avoid putting the onus on the database - and its schemas - to explicitly model composition logic and mechanisms.

> **Versioning**
> The deferral of versioning to PyPI isn't entirely clear to me, especially in conjunction with the ability to instantiate Estimator Instances from a string.

We do not entirely defer to pypi, but each string is "versioned" as opposed to having an intrinsic version. For instance, `RandomForestClassifier` would exit in versions 1.5.0, 1.5.1, 1.6.0, etc, corresponding to the versions of its principal containing package.

> How would the following code example work?

```python
import openml
rf = openml.get("RandomForestClassifier(hyperparameter_a=23)")
```
> Assume RandomForestClassifier comes from a package with is currently on version 1.8 and given that:
>  - parameter defaults may change from version to version (e.g., `hyperparameter_b`'s default was updated in version 1.5)
>  - parameters may be introduced or removed from version to version (`hyperparameter_a` was removed in version 1.7)
>  - behavior may change from version to version (e.g., a bug was fixed in version 1.6)
>  - not all scikit-learn versions are available on all Python version (assuming there is no support for local builds)

Simple: the estimator that gets constructed is equivalent to the execution of the string after import from sklearn.

in terms of versioning, that is implicitly equivalent to

1. we check the local python environment for the version of `scikit-learn`. Let's say that is 1.5
2. we "materialize" the `RandomForestClassifier` of version 1.5. Which is simply the one available in the already installed `scikit-learn`.

(although no explicit checking takes place, since we simply defer to the package in the context environment!)

To handle some of the quirks of parameters being there or not, internally we could track:

* on which versions of `scikit-learn` the estimator is available
* on which versions which parameters are available

If we want more sugar, we can catch before `ModuleNotFoundError` etc and tell the user that the estimator or the parameter is not available on current version, but in a certain range only.

For the other problems you mentioned:

* Re changing defaults, I think it is the burden of the user to be aware of the defaults in the version they use. Again, we can provide sugar that links to the correct version of the docs when the estimator is pretty-printed (but sklearn actually does that by itself, since a while, already)
* re "sklearn version not available on all py versions", I do not think this is the business of `openml` at all. The user would notice this when they try to set up their environment - namely, that they cannot install the desired sklearn version.


**Unique Namespace**

> Keeping the "external version" to a constant would allow you to create a lookup-by-name for flows (`/flow/exists/{name}/{version}` REST API endpoint even supports indexed lookups).

That is nice - and we wouldn't have to change much even!

> It might be interesting to explore that direction as it would allow the implementation (at least of that aspect of the proposal) without changes to server infrastructure.

Yes!

fkiraly:
> [Design and Feature Request](https://github.com/orgs/openml/discussions/categories/openml-design-and-feature-requests) discussion board

Regarding this - what is your experience on commenting in specific sections of designs? We have found that using PR for discussion allows to attach comments to individual sections of design documents more easily.

PGijsbers:
> Regarding this - what is your experience on commenting in specific sections of designs? We have found that using PR for discussion allows to attach comments to individual sections of design documents more easily.

That can work, though I find that if the document also gets revised in the process discussions can be easily lost. I can't say that I am aware of a system that allows both easy threads off specific sections of the proposal while at the same time supporting revisions gracefully (though I've also not looked very hard). We've also used Google Docs in the past, but that has its own downsides (comments remain, but you cannot easily identify the text the comment thread originally was based off).

fkiraly:
@PGijsbers, any further thought about the design direction?

PGijsbers:
First off, sorry for not actually responding to the main reply. That wasn't intentional.

> I think the "best" design structures openML around managing unified APIs primarily, packages (e.g., sklearn, torch, etc) second.

Agreed.

> Yes, plus using them in downstream services like benchmarks or leaderboards.

That was certainly on my mind too, thanks for making it explicit.

> Yes - I think that is key to manage the crawling, creeping complexity that makes the schemas so hard to maintain or extend. It also removes the onus entirely from openml to manage serialization, if we implement this smartly.

The OpenML team has been unhappy with Flows for a while now, and has also discussed using more package specific approaches. E.g., `str` for scikit-learn or ONNX for neural networks. In fact, I believe `openml-pytorch` saves ONNX files. It's also important to keep in mind that OpenML is a platform accessible from multiple languages. I think you'll find support for this.

> the ability to perform registry operations, such as tagging, search and retrieval, on top

This means updates to the Flow schema and REST API endpoints, which is fine (if approved), but should be made explicit in the proposal.

> [about setup vs flow] Is that so?

Yes, but I agree that the documentation is ambiguous and the `openml-python` function names are misleading. I have updated the docs (PR under review) to make the distinction more explicit. Even then, we can improve it further and address the naming in openml-python. I agree that it should be completely obvious to the user and we have failed at that.

> Having said that, the backend can still keep track of this, but I am really wondering: why would we want to?

Maybe. I think it's fair to question that. The recommended range sounds nice in theory but meaningless in practice (no one has or provides that data), I would support dropping that. Hyperparameter names and types can still be useful, for example to deal with the issues discussed around changes to them, but it would mean that either the Flow/Estimator Class is only specified for a specific version (or version range), or the way in which the hyperparameter metadata is structured contains information about the versions. 

For completeness, when it comes to estimator instances I think we definitely want to keep the structured metadata about what configurations are used. One of the envisioned use cases for OpenML is to be able to perform large scale meta-learning on the uploaded experiments (i.e., learn patterns of which configurations work well for which estimators and/or datasets). Not having a structured way to access this metadata will really hurt in the possiblity for OpenML to fulfill that role, and I would expect strong resistance from the community.

> This way, we avoid putting the onus on the database - and its schemas - to explicitly model composition logic and mechanisms.

That's great because it's one of the things that we also found to be rather unproductive. I would probably prefer not to store composite flows as their own flow, and see if we run into any issues with that. I do believe this will require changes in how runs/setups are stored (currently runs point to one flow, setups to one run).

> re "sklearn version not available on all py versions", [...] The user would notice this when they try to set up their environment [...]

But if the user only knows they want to reinstantiate `RandomForest(hyperparameter_a=42)` they might not know up front what scikit-learn and python version they want/need. That said, I agree.

> [about batteries included] Yes, although I think this would be "the" extremely attractive user facing feature that drops out of. The above are more core design elements that a user may not see or consciously perceive.

Personally, I am not entirely sure (I'll come back to that). The reason why I phrased it this way is that I think that the batteries-included and the maintenance of a leaderboard features may be harder to find support for, as they require indefinite upkeep. But for the purposes of a streamlined discussion and the likelihood of getting acceptance or progress on individual items, I think it would make sense to treat those as separate proposals from the changes to the Flow infrastructure.

Coming back to the point, the research and practicioner communities often already share leaderboards/benchmark studies. We should absolutely allow them to use OpenML for that, but it's not obvious to me that we should maintain our own. And, if my understanding is correct, _if_ machine learning libraries already adhere to the interface OpenML expects, then the advantage of the local repository is "merely" avoids a server roundtrip?

Re: security
Personally, I think we should initially only work with PyPI packages. While our availability to work on OpenML is slowly improving, I think our effort is better spent elsewhere than reviewing code submissions in the near future.

PGijsbers:
GitHub discussions does allow threads within a post. I think we should try to use that, but you can always point to external resources (for example, a PR with exact schema changes on the [XSD files](https://github.com/openml/OpenML/tree/develop/openml_OS/views/pages/api_new/v1/xsd)). Community members can then start threads about specific sections.

We should also plan a breakout session during the hackathon in February.

PR #1584, opened 2025-12-31
[BUG] `get_task` removes the dir even if was already existing

JATAYU000:
#### Metadata
* Reference Issue: Refer failures in #1579 
* New Tests Added: No
* Documentation Updated: No
* Change Log Entry: Checks if the directory was created newly else doesn't remove.

### Details
`get_task` checks if the `tid_cache_dir` was already existing before removing it on `Exception`
`OpenMLServerException` causes `get_task` to remove the entire directory even if the directory was already existing and is used by other tests
* How can I reproduce the issue this PR is solving and its solution?
observe `exists assertion` errors for files under `tests/files/org/openml/test/task/1/` after running `pytest` 
or look at failures in #1579

### DIFF
--- a/tests/test_datasets/test_dataset.py
+++ b/tests/test_datasets/test_dataset.py
@@ -294,6 +294,7 @@ def test_tagging():
     datasets = openml.datasets.list_datasets(tag=tag)
     assert datasets.empty
 
+@pytest.mark.xfail(reason="failures_issue_1544")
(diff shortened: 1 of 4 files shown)


fkiraly (changes requested):
This fixes one of the issues present, so you need to remove the `xfail`s now.

fkiraly (approved):
Fixed the one discrepancy

PR #1585, opened 2025-12-31
[MNT] Update xfail for new test server state

PGijsbers:
#### Metadata
* Reference Issue: #1544 
* New Tests Added: No
* Documentation Updated: No

#### Details 
I investigated the failures and the root cause was incorrect test server state.
This still remains an issue for one test, but I can look into that later (after I return from my vacation).


PGijsbers:
> Does this completely, or only partially resolve the failures?

I suppose partially on account that there remains one test that's not working based on what seems like the test server state.

PR #1588, opened 2026-01-01
[MNT] Remove redundant `__init__`s in `OpenMLTask` descendants by adding ClassVar

### LINKED ISSUE #1578: [MNT] Reduce Task Constructor Duplication
I agree with this comment on code duplication: https://github.com/openml/openml-python/blob/main/openml/tasks/task.py#L2
All 6 task classes have nearly identical `__init__` methods with 8-10 parameters each. This creates duplicated code.

The differences between the task-like classes can be handled by class attributes.

Omswastik-11:
Fixes #1578

### DIFF
--- a/openml/_api_calls.py
+++ b/openml/_api_calls.py
@@ -22,8 +22,8 @@ from urllib3 import ProxyManager
 from . import config
 from .__version__ import __version__
 from .exceptions import (
+    OpenMLAuthenticationError,
     OpenMLHashException,
-    OpenMLNotAuthorizedError,
     OpenMLServerError,
     OpenMLServerException,
     OpenMLServerNoResult,
@@ -515,11 +515,7 @@ def __parse_server_exception(
         400,  # run/42 delete
         460,  # task/42 delete
     ]:
-        msg = (
-            f"The API call {url} requires authentication via an API key.\nPlease configure "
-            "OpenML-Python to use your API as described in this example:"
-            "\nhttps://openml.github.io/openml-python/latest/examples/Basics/introduction_tutorial/#authentication"
-        )
-        return OpenMLNotAuthorizedError(message=msg)
+        msg = f"The API call {url} requires authentication via an API key."
+        return OpenMLAuthenticationError(message=msg)
 
     return OpenMLServerException(code=code, message=full_message, url=url)
--- a/openml/cli.py
+++ b/openml/cli.py
@@ -109,7 +109,7 @@ def configure_server(value: str) -> None:
 
     def replace_shorthand(server: str) -> str:
         if server == "test":
-            return "https://test.openml.org/api/v1/xml"
+            return f"{config.TEST_SERVER_URL}/api/v1/xml"
         if server == "production_server":
             return "https://www.openml.org/api/v1/xml"
         return server
--- a/openml/config.py
+++ b/openml/config.py
@@ -28,6 +28,8 @@ OPENML_SKIP_PARQUET_ENV_VAR = "OPENML_SKIP_PARQUET"
 OPENML_TEST_SERVER_ADMIN_KEY_ENV_VAR = "OPENML_TEST_SERVER_ADMIN_KEY"
 _TEST_SERVER_NORMAL_USER_KEY = "normaluser"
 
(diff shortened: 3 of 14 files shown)


fkiraly (changes requested):
This changes a key part of the public API.

Please do not introduce breaking changes in refactors.

Omswastik-11:
Hi @fkiraly !! 
Thanks for suggestion . I reverted the changes that affect public API . 

```python
from openml.tasks.task import (
    OpenMLClassificationTask,
    OpenMLRegressionTask,
    OpenMLClusteringTask,
    OpenMLLearningCurveTask,
    TaskType,
)

def main():
    # Classification (defaults estimation_procedure_id=1)
    cls_task = OpenMLClassificationTask(
        task_id=None,
        task_type_id=TaskType.SUPERVISED_CLASSIFICATION,
        task_type="Supervised Classification",
        data_set_id=1,
        target_name="class",
    )
    print("Classification est proc:", cls_task.estimation_procedure_id)

    # Regression (defaults estimation_procedure_id=7)
    reg_task = OpenMLRegressionTask(
        task_id=None,
        task_type_id=TaskType.SUPERVISED_REGRESSION,
        task_type="Supervised Regression",
        data_set_id=2,
        target_name="target",
    )
    print("Regression est proc:", reg_task.estimation_procedure_id)

    # Clustering (defaults estimation_procedure_id=17)
    clu_task = OpenMLClusteringTask(
        task_id=None,
        task_type_id=TaskType.CLUSTERING,
        task_type="Clustering",
        data_set_id=3,
    )
    print("Clustering est proc:", clu_task.estimation_procedure_id)

    # Learning curve (defaults estimation_procedure_id=13)
    lc_task = OpenMLLearningCurveTask(
        task_id=None,
        task_type_id=TaskType.LEARNING_CURVE,
        task_type="Learning Curve",
        data_set_id=4,
        target_name="class",
    )
    print("Learning curve est proc:", lc_task.estimation_procedure_id)

if __name__ == "__main__":
    main()
```

openml/tasks/task.py
```diff
@@ -325,64 +349,13 @@ class OpenMLClassificationTask(OpenMLSupervisedTask):
 
     Parameters
     ----------
-    task_type_id : TaskType
```
fkiraly, line 334: why are you removing this from the docstring?

openml/tasks/task.py
```diff
@@ -249,32 +262,43 @@ class OpenMLSupervisedTask(OpenMLTask, ABC):
...
+            estimation_procedure_id=resolved_estimation_procedure_id,
             estimation_procedure_type=estimation_procedure_type,
             estimation_parameters=estimation_parameters,
             evaluation_measure=evaluation_measure,
             data_splits_url=data_splits_url,
+            target_name=target_name,
         )
 
-        self.target_name = target_name
+        self.class_labels = class_labels
```
fkiraly: why are you moving the task specific variables here? A regression task does not have class labels, etc.

Omswastik-11: I thought it would reduce the code size as we don't need to write a `__init__` for classification but I will go with your suggestion .

openml/tasks/task.py
Omswastik-11: I thought it would duplicate same variables multiple times . but I got it it would be less understandable

openml/tasks/task.py
```diff
@@ -250,33 +263,37 @@ class OpenMLSupervisedTask(OpenMLTask, ABC):
...
         target_name: str,
-        estimation_procedure_id: int = 1,
+        estimation_procedure_id: int | None = None,
         estimation_procedure_type: str | None = None,
         estimation_parameters: dict[str, str] | None = None,
         evaluation_measure: str | None = None,
         data_splits_url: str | None = None,
         task_id: int | None = None,
     ):
+        resolved_estimation_procedure_id = self._resolve_estimation_procedure_id(
```
fkiraly: why is this needed? Does not seem DRY.

Omswastik-11: ahh thanks for pointing . I will fix this .

PR #1589, opened 2026-01-02
[ENH] Replace `asserts` with proper `if else` Exception handling

### LINKED ISSUE #1581: [MNT] Replace Assert Statements with Proper Exceptions
In the 'runs'-module there are assert statements used for validation. I would rather see exceptions (ValueError, TypeError, etc.) there to have more explicit validation and error messages.

The relevant code can be found here:
- https://github.com/openml/openml-python/blob/main/openml/runs/run.py
- https://github.com/openml/openml-python/blob/main/openml/runs/functions.py

Omswastik-11:
Fixes #1581

### DIFF
--- a/tests/test_tasks/test_split.py
+++ b/tests/test_tasks/test_split.py
@@ -3,6 +3,8 @@ from __future__ import annotations
 
 import inspect
 import os
+import shutil
+import tempfile
 from pathlib import Path
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


Omswastik-11:
Hi @fkiraly !! I would like to know you suggestion on this .

Screenshot 2026-01-02 133722

In `openml\runs\run.py` the `_generate_arff_dict` function is causing the above issue . so I created a new function `_get_arff_attributes_for_task` to reduce complexity is it ok ? or should I think of a alternate solution ?

geetu040 (approved):
Approved this PR previously in https://github.com/openml/openml-python/pull/1589#pullrequestreview-3797228877

Only changes since then, are related to the reviews by Copilot, which looked fine and have been addressed.

@fkiraly please merge.

PR #1590, opened 2026-01-02
[ENH] Refactor `Extension`

jgyasu:
## Summary

This PR refactors the existing `Extension` abstraction into three focused base classes:
* `ModelSerializer`
* `ModelExecutor`
* `OpenMLAPIConnector`

In addition, it introduces a registry-based API resolver that automatically determines the appropriate connector for a given estimator instance or OpenML flow. The resolver inspects the input at runtime and returns the matching connector, removing the need for users to know about or manually import estimator/flow-specific extension.

This significantly simplifies the user-facing API, users can work directly with estimators or flows without needing to know which extension is responsible for handling them.

And, the existing `Extension` class is kept intact, ensuring that all current user-facing APIs remain fully backward compatible.

As part of this change, existing OpenML extensions need to be refactored to adopt the new abstractions and take advantage of the simplified API.

#### Before
```python
from openml_sklearn.extension import SklearnExtension
clf = KNeighborsClassifier(n_neighbors=3)
extension = SklearnExtension()
knn_flow = extension.model_to_flow(clf)
knn_flow.publish()
```

#### After (co-exists with the _Before_ API for now)

```python
from openml.flows import estimator_to_flow
clf = KNeighborsClassifier(n_neighbors=3)
knn_flow = estimator_to_flow(clf)
knn_flow.publish()
```

#### Before
```python
from openml_sklearn.extension import SklearnExtension
extension = SklearnExtension()
estimator_instance = extension.flow_to_model(knn_flow)
```

#### After (co-exists with the _Before_ API for now)
```python
from openml.flows import flow_to_estimstor
estimator_instance = flow_to_estimator(knn_flow)
```


jgyasu:
> Reason being since the current concept couples "executor" to "serializer" strongly. The same serialized object could be executed in multiple contextx.

I agree.

I have added the tests for the base classes and registry. I have a question:

This refactor means nothing and the public API change as mentioned in the PR description won't work unless we also refactor the openml extensions such as `openml-sklearn`, and `openml-pytorch`. What do you and other developers (if reading) suggest, 
1. Should we merge the extensions back into `openml-python` right now and keep the dependencies such as `sklearn`, and `pytorch` as soft dependencies so that `openml-python` is not dependent on them?
2. Should we keep them as standalone package and refactor them respectively to work with the refactored classes from `openml-python` for now?

Note: We are planning to move the extensions to `openml-python` mid-term or long-term anyway because we believe the API should be task-first rather than library-first, libraries should be soft dependencies that users can install additionally.

fkiraly:
> This refactor means nothing

I think you do not give yourself enough credit.

> the public API change as mentioned in the PR description won't work unless we also refactor the openml extensions such as `openml-sklearn`, and `openml-pytorch`. What do you and other developers (if reading) suggest, 

In a pure refactor, the behaviour of the class does not change. So you would replace the current extension class with instances of your new design, just to keep current behaviour exactly as it is.

Mid-term, I agree that `Extension` - as a class that has serializers and executors (or however you call them) both, needs to go. I think the two concepts need to be decoupled completely, since it is not an 1:1 relationship:

* different packages under a joint object type API can have different serialization strategies. For instance, `skops` for `scikit-learn` native estimators, or `torch`-based serialization for `torch` based tabular estimators following a weak `scikit-learn` API specification.
* estimators in a given object type API can live in multiple execution contexts. Even for tabular estimators, there are different ways to benchmark them, and hard-coding any one, and furthermore tying it to a specific serialization structure, seems like a very bad idea. Hence, "executors" should be first class citizens, and be inspectable as to which API type goes into them.

> 1. Should we merge the extensions back into `openml-python` right now and keep the dependencies such as `sklearn`, and `pytorch` as soft dependencies so that `openml-python` is not dependent on them?

Yes, I think that is a good next step (in a separate PR) - we ought to ensure that we strictly test:

* 100% isolation in terms of coupling and dependencies. No coupling should be re-introduced, no dependencies should arise in addition.
* current integration API, stability of the public API.

As per the consensus from discussion on 2026-01-05, we can merge back but these are two things we *must* ensure.

> 2. Should we keep them as standalone package and refactor them respectively to work with the refactored classes from `openml-python` for now?

How would that look like? I am not sure if I understand the pattern here. Refactoring across GitHub repositories seems like a lot more work to me.

I can see how the first steps of a refactor would work purely within `openml-python`, but once we start redesign, it might get hairy.

> Note: We are planning to move the extensions to `openml-python` mid-term or long-term anyway because we believe the API should be task-first rather than library-first, libraries should be soft dependencies that users can install additionally.

Yes, although the end state would definitely not be in the current design (which I think has a lot of problems like internal coupling, lack of flexibility, etc).

openml/extensions/base/_executor.py
```diff
@@ -0,0 +1,151 @@
...
+        ----------
+        model : Any
+
+        Returns
+        -------
+        bool
+        """
+
+    @abstractmethod
+    def obtain_parameter_values(
```
PGijsbers: Why is this in the executor instead of the serializer? This method serializes the hyperparameter configuration to the OpenML specifications (well, a dict which is later converted to xml).

openml/extensions/base/_executor.py
```diff
@@ -0,0 +1,151 @@
...
+if TYPE_CHECKING:
+    import numpy as np
+    import scipy.sparse
+
+    from openml.flows import OpenMLFlow
+    from openml.runs.trace import OpenMLRunTrace, OpenMLTraceIteration
+    from openml.tasks.task import OpenMLTask
+
+
+class ModelExecutor(ABC):
```
PGijsbers, line 19: Should this now also not have a `supports`  or `can_handle_model` function?

openml/extensions/registry.py
```diff
@@ -0,0 +1,49 @@
...
+from typing import TYPE_CHECKING, Any
+
+from openml.exceptions import PyOpenMLError
+
+if TYPE_CHECKING:
+    from openml.extensions.base import OpenMLAPIConnector
+
+API_CONNECTOR_REGISTRY: list[type[OpenMLAPIConnector]] = [
+    # Add OpenMLAPIConnector subclasses here to register them
+]
```
PGijsbers: What's the expected access pattern?
And why is this now at the `APIConnector` level instead of the `Executor` and `Serializer` levels? Wouldn't I expect a list of each now?

jgyasu: > Wouldn't I expect a list of each now?

No. There only needs to be a list of connectors. The `resolve_api_connector` resolves the connector by calling the `supports` method of the connectors and returns a compatible connector instance, from which the serializer and executor can be accessed. The `supports` method of the connectors will call the `can_handle_model` of serializer and executor.

If you look at the `flow_to_estimator` function that I have implemented, this is how it goes:
```python
def flow_to_estimator(flow: OpenMLFlow, initialize_with_defaults: bool = False, strict_version: bool = True) -> Any:

    connector = resolve_api_connector(flow)

    return connector.serializer().flow_to_model(
        flow,
        initialize_with_defaults=initialize_with_defaults,
        strict_version=strict_version,
    )
```

openml/extensions/base/_executor.py
jgyasu: I see! I didn't go through this method in detail, but `extension_interface.py` file has separated into serialzation and execution logic with comments and this method is in the execution section. I will fix it.

openml/extensions/registry.py
PGijsbers: I understand with the current situation where the Serializer and Executor are always bundled into one connector you can simply check the connectors. But later it would make sense that Serializers and Executors are registered independently, no? I thought that was the point. For `flow_to_estimator` you would only need a Serializer, not an Executor, but `resolve_api_connector` does not preserve that information.

openml/extensions/base/_connector.py
```diff
@@ -0,0 +1,28 @@
...
+        """Return the serializer for this API."""
+
+    @abstractmethod
+    def executor(self) -> ModelExecutor:
+        """Return the executor for this API."""
+
+    @classmethod
+    @abstractmethod
+    def supports(cls, estimator: Any) -> bool:
+        """High-level check if this connector supports the estimator instance or flow."""
```
PGijsbers: Like here I would expect something like `return self.serializer().can_handle_model(estimator) and self.executor().can_handle_model(estimator)` because the current "OpenMLAPIConnector" and all code that relies on them expect them to support both. Later in refactoring, the goal is to essentially remove the `OpenMLAPIConnector` and instead rely directly on `Serializer` and `Executor`, is it not?

PGijsbers: So it would ultimately become something _like_ (simplified):

```python
def run_model_on_task(estimator, task):
    serializer = registry.resolve_serializer(estimator)
    flow = serializer.to_flow(estimator)
    executor = registry.resolve_executor(estimator)
    run = executor.run(estimator, task)
    return flow, run
```

Then if I publish some new ML module that follows a common interface that an existing executor works with, but I have something bespoke for serializing my model as a flow, I can simply only provide a new `Serializer` for my estimator and it should work.

jgyasu: Yes, this makes sense!

geetu040 (commented):
@jgyasu what's the status of this PR? Are we planning to merge it or await review from OpenML team?

jgyasu:
@geetu040 I made changes as suggested by @PGijsbers, let's wait for his review. I will re-request a review from him.

PR #1595, opened 2026-01-03
[ENH] Added `ReprMixin` to share `__repr__` formatting

### LINKED ISSUE #1591: [ENH] Add `ReprMixin` to share `__repr__` formatting across OpenML classes
Add a ReprMixin class (in openml/utils.py) that provides:
 * `__repr__` which calls `self._get_repr_body_fields()`
* `_apply_repr_template`  to produce the formatted string

Refer Discussion
#1567

JATAYU000:
#### Metadata
* Reference Issue:  Fixes #1591 
* New Tests Added: No
* Documentation Updated: Yes (docstring)
* Change Log Entry: Adds `ReprMixin` in `utils`

#### Details 
* Any other comments?
Had a question, Can I make the classes use the `mixin` in this pr or should I open a new one?


geetu040:
@fkiraly the CI failures are coming from the production server - specifically database connection - probably because of overloading. Should we do something about these in short-term or just leave it for long-term test refactor?

geetu040:
There is one unrelated sporadic failure coming from the production server, due to high server load, which we are not skipping and should be fixed if rerun.

PR #1596, opened 2026-01-04
[ENH] Reduce complexity of `run_flow_on_task` func

### LINKED ISSUE #1580: [MNT] Refactor `run_flow_on_task`-function
The `run_flow_on_task` is ~160 lines of code with high complexity: https://github.com/openml/openml-python/blob/main/openml/runs/functions.py#L178

I would suggest to refactor it to reduce its complexity. We can keep the this main-function as the orchestrator and separate the responsibilities into:
  - input validation
  - parallel processing
  - aggregation of predictions and evaluations
  - creating the run from the `res`

Omswastik-11:
## Summary

This PR refactors `run_flow_on_task`, which had grown to ~160 lines with high cyclomatic complexity, by extracting small helper functions with clear, single responsibilities. The main function is now a readable orchestrator with clearly defined steps.

---

### Extracted helper functions

* **`_validate_flow_and_task_inputs`**
  Handles input validation and backward-compatible argument handling

* **`_sync_flow_with_server`**
  Synchronizes the flow with the server and checks for duplicate runs

* **`_prepare_run_environment`**
  Prepares environment information and run tags

* **`_create_run_from_results`**
  Builds the `OpenMLRun` object from execution results

### Internal structure improvements

* Introduced a `_RunResults` `NamedTuple` to bundle execution outputs
  (`data_content`, `trace`, `evaluations`) and reduce long parameter lists

---

## Type Safety Improvements

* Replaced `assert` statements with explicit `ValueError` / `TypeError` exceptions
* Added type guards before accessing attributes that may be `None`
* Made boolean parameters keyword-only where appropriate

---

Fixes #1580


openml/runs/functions.py
```diff
@@ -54,6 +54,242 @@
...
+    Parameters
+    ----------
+    flow : OpenMLFlow or OpenMLTask
+        The flow object (may be swapped with task for backward compatibility).
+    task : OpenMLTask or OpenMLFlow
+        The task object (may be swapped with flow for backward compatibility).
+    flow_tags : List[str] or None
+        A list of tags that the flow should have at creation.
+    avoid_duplicate_runs : bool or None
+        Whether to check for duplicate runs on the server.
```
geetu040: it actually checks if setup/task combination is already present, but this should be removed see the next comment.

openml/runs/functions.py
```diff
@@ -54,6 +54,242 @@
...
+    Raises
+    ------
+    ValueError
+        If flow_tags is not a list or task is not published.
+    """
+    if flow_tags is not None and not isinstance(flow_tags, list):
+        raise ValueError("flow_tags should be a list")
+
+    if avoid_duplicate_runs is None:
+        avoid_duplicate_runs = openml.config.avoid_duplicate_runs
```
geetu040: I would suggest to keep this in the main function, since it's out of scope given the current function's name

openml/runs/functions.py
```diff
@@ -54,6 +54,242 @@
...
+    Parameters
+    ----------
+    flow : OpenMLFlow
+        The flow to synchronize.
+    task : OpenMLTask
+        The task to check for duplicate runs.
+    upload_flow : bool
+        Whether to upload the flow if it doesn't exist.
+    avoid_duplicate_runs : bool
+        Whether to check for duplicate runs.
```
geetu040, line 136: it actually checks if setup/task combination is already present, you can see this from main function's docstring

openml/runs/functions.py
```diff
@@ -54,6 +54,242 @@
...
+    run_environment = flow.extension.get_version_information()
+    tags = ["openml-python", run_environment[1]]
+    return tags, run_environment
+
+
+class _RunResults(NamedTuple):
+    data_content: list[list]
+    trace: OpenMLRunTrace | None
+    fold_evaluations: OrderedDict[str, OrderedDict]
+    sample_evaluations: OrderedDict[str, OrderedDict]
```
geetu040: I don't really see much benefit of having this, can you tell me how could it be improving code?

without using `_RunResults`, the code looks like this:
```py
    # 7. Create run from results
    run = _create_run_from_results(
        task=task,
        flow=flow,
        flow_id=flow_id,
        data_content=data_content,
        trace=trace,
        fold_evaluations=fold_evaluations,
        sample_evaluations=sample_evaluations,
        tags=tags,
        run_environment=run_environment,
        upload_flow=upload_flow,
        avoid_duplicate_runs=avoid_duplicate_runs,
    )
```

with `_RunResults`, the code looks like this:
```py
    results = _RunResults(
        data_content=data_content,
        trace=trace,
        fold_evaluations=fold_evaluations,
        sample_evaluations=sample_evaluations,
    )

    # 7. Create run from results
    run = _create_run_from_results(
        task=task,
        flow=flow,
        flow_id=flow_id,
        tags=tags,
        run_environment=run_environment,
        upload_flow=upload_flow,
        avoid_duplicate_runs=avoid_duplicate_runs,
    )
```
is it really an improvement?

openml/runs/functions.py
```diff
@@ -54,6 +54,242 @@
...
+    OpenMLRun
+        The created run object.
+    """
+    dataset = task.get_dataset()
+    fields = [*run_environment, time.strftime("%c"), "Created by run_flow_on_task"]
+    generated_description = "\n".join(fields)
+
+    task_id = task.task_id
+    if task_id is None:
+        raise ValueError("The task should be published at OpenML")
```
geetu040: isn't this already validated in `_validate_flow_and_task_inputs`?

openml/runs/functions.py
```diff
@@ -222,116 +458,64 @@ def run_flow_on_task(  # noqa: C901, PLR0912, PLR0915, PLR0913
...
+        tags=tags,
+        run_environment=run_environment,
+        upload_flow=upload_flow,
+        avoid_duplicate_runs=avoid_duplicate_runs,
+    )
 
+    # 8. Log completion message
     if flow_id:
         message = f"Executed Task {task.task_id} with Flow id:{run.flow_id}"
     else:
```
geetu040, line 520: Looks much better and readable now.

geetu040 (changes requested):
Looks really nice, I have left a few comments with only minor changes requested.

openml/runs/functions.py
Omswastik-11: I thought it would be more readable and big functions with high number of arguments causes ruff errors . but I agree with your opinion as well.

openml/runs/functions.py
Omswastik-11: Yes !! Thanks For pointing out . It seems it was duplicated in old code also , I forgot to remove it .

Omswastik-11:
added `cast(int , task_id)` because of mypy pre-commit error . 

image

geetu040 (approved):
Nicely refactored, LGTM.

CC: @fkiraly, @SimonBlanke for review/merge.

openml/runs/functions.py
```diff
@@ -54,6 +54,225 @@
...
+            flow.publish()
+            flow_id = flow.flow_id
+        elif flow_id:
+            flow_from_server = get_flow(flow_id)
+            _copy_server_fields(flow_from_server, flow)
+            if avoid_duplicate_runs:
+                flow_from_server.model = flow.model
+                setup_id = setup_exists(flow_from_server)
+                task_id = task.task_id
+                ids = run_exists(cast("int", task_id), setup_id)
```
SimonBlanke: I would expect `cast(int, task_id)` here (without the doublequotes). This might be a problem with the pre-commit. @Omswastik-11 What do you think. Does "int" make sense here?

SimonBlanke (changes requested):
@Omswastik-11 Do you see a way to increase the test coverage here? This is not a hard requirement.

openml/runs/functions.py
geetu040: agreed, but I didn't point this out cause I thought it was a requirement for `pre-commit`, if that's not the case, `int` should definitely be preferred.

Omswastik-11: I think `ruff` is converting it to "int" automatically

geetu040: can you confirm it, in that case let's leave it as it is

Omswastik-11: Yes , Even if I revert that during pre-commit it automatically changed

geetu040: ok makes sense then

SimonBlanke: Okay, we leave it converted. Seems to be safer apparently

geetu040 (changes requested):
Actually there is no unit test for the function `openml.runs.run_flow_on_task`, could you please add one. There are some tests that use `openml.runs.run_flow_on_task` internally, but it would be nice to have an independent test that only checks this functionality. You can add this test in `tests/test_runs`.
Also if the helper functions in `openml.runs.run_flow_on_task` can be tested at unit (suggested in https://github.com/openml/openml-python/pull/1596#pullrequestreview-3673672157), that would be nice, but again, it's not a hard requirement.

geetu040 (approved):
LGTM!
@fkiraly please review/merge.

openml/runs/functions.py
```diff
@@ -54,6 +54,225 @@
...
+            "The old argument order (Flow, model) is deprecated and "
+            "will not be supported in the future. Please use the "
+            "order (model, Flow).",
+            DeprecationWarning,
+            stacklevel=3,
+        )
+        task, flow = flow, task
+
+    if not isinstance(flow, OpenMLFlow):
+        raise TypeError("Flow must be OpenMLFlow after validation")
```
fkiraly: please include error location, and correct reference to variable

openml/runs/functions.py
```diff
@@ -54,6 +54,225 @@
...
+    """
+    if flow_tags is not None and not isinstance(flow_tags, list):
+        raise ValueError("flow_tags should be a list")
+
+    # TODO: At some point in the future do not allow for arguments in old order (changed 6-2018).
+    # Flexibility currently still allowed due to code-snippet in OpenML100 paper (3-2019).
+    if isinstance(flow, OpenMLTask) and isinstance(task, OpenMLFlow):
+        # We want to allow either order of argument (to avoid confusion).
+        warnings.warn(
+            "The old argument order (Flow, model) is deprecated and "
```
fkiraly: please capitalize variable names correctly. Include source of warning.

openml/runs/functions.py
```diff
@@ -54,6 +54,225 @@
...
+    Raises
+    ------
+    PyOpenMLError
+        If flow_id mismatch or flow doesn't exist when expected.
+    OpenMLRunsExistError
+        If duplicate runs exist and avoid_duplicate_runs is True.
+    """
+    # We only need to sync with the server right now if we want to upload the flow,
+    # or ensure no duplicate runs exist. Otherwise it can be synced at upload time.
+    flow_id = None
```
fkiraly, line 152: code smell, too many indentations. Please simplify this further

openml/runs/functions.py
```diff
@@ -54,6 +54,225 @@
...
+    if upload_flow or avoid_duplicate_runs:
```
fkiraly: for example, this could be `if not upload_flow and not avoid_duplicate_runs: return flow_id`, then one indentation less below.

openml/runs/functions.py
```diff
@@ -222,116 +441,61 @@ def run_flow_on_task(  # noqa: C901, PLR0912, PLR0915, PLR0913
...
-    fields = [*run_environment, time.strftime("%c"), "Created by run_flow_on_task"]
-    generated_description = "\n".join(fields)
-    run = OpenMLRun(
-        task_id=task.task_id,
+    # 7. Create run from results
+    run = _create_run_from_results(
+        task=task,
+        flow=flow,
         flow_id=flow_id,
-        dataset_id=dataset.dataset_id,
```
fkiraly, line 320: why is it fine to delete these args?

fkiraly (changes requested):
Nice! I left some recommendations on how to further simplify the code flow.

geetu040 (changes requested):
please resolve conflicts with main

openml/runs/functions.py
```diff
@@ -222,116 +452,61 @@ def run_flow_on_task(  # noqa: C901, PLR0912, PLR0915, PLR0913
...
-    # execute the run
-    res = _run_task_get_arffcontent(
+    # 6. Execute the run (parallel processing happens here)
+    data_content, trace, fold_evaluations, sample_evaluations = _run_task_get_arffcontent(
         model=flow.model,
         task=task,
         extension=flow.extension,
         add_local_measures=add_local_measures,
         n_jobs=n_jobs,
     )
```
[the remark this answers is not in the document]
geetu040: @Omswastik-11 update the PR description to remove this part.

geetu040 (approved):
@fkiraly, can we merge this now? comments from your last review https://github.com/openml/openml-python/pull/1596#pullrequestreview-3809428642 are resolved.

fkiraly (approved):

Issue #1598, opened 2026-01-06
[MNT] mark tests that require the OpenML test server


geetu040:
Some tests implicitly rely on the OpenML test server, but this dependency is not consistently or explicitly marked.

Tests that require the production server are already labeled with the `production` marker. A small subset of test-server-dependent tests are marked with `server`, but this marker is incomplete and does not cover all such cases. As a result, it is difficult to reliably identify, run, or skip tests that depend on the OpenML test server.

PR #1599, opened 2026-01-06
[MNT] add pytest marker to tests requiring test server

geetu040:
Fixes https://github.com/openml/openml-python/issues/1598

This PR adds the `@pytest.mark.uses_test_server()` marker to tests that depend on the OpenML test server.

Changes
* added `uses_test_server` on the relevant test sets.
* replaced all the `server` markers with `uses_test_server` marker
* removed all the `@pytest.mark.xfail(reason="failures_issue_1544", strict=False)` where the failure was due to race-conditions or server connectivity

### DIFF
--- a/tests/test_runs/test_run_functions.py
+++ b/tests/test_runs/test_run_functions.py
@@ -627,8 +627,8 @@ class TestRun(TestBase):
             sentinel=sentinel,
         )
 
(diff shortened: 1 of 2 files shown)


fkiraly:
great! Can we now skip them, to temporarily disable the failures?

PR #1602, opened 2026-01-07
doc: added test run details for easy lookup

rohansen856:
#### Metadata
* Reference Issue: #1544 
* New Tests Added: NA
* Documentation Updated: NA
* Change Log Entry: 

I am going to incrementally attach my findings about the test failure issues and test results so that everyone can quickly take a look and give their opinions on them.


rohansen856:
@EmanAbdelhaleem i reverted changes from your PR #1587 whch adds the `xfail` tags and then im investigate what are the tests and try to find the cause of the test failures. will raise a PR with the output test results and also add those to our doc.

PR #1576, opened 2025-12-30
[ENH] V1 → V2 API Migration - core structure

geetu040:
Towards #1575

This PR sets up the core folder and file structure along with base scaffolding for the API v1 → v2 migration.

It includes:

* Skeleton for the HTTP client, backend, and API context
* Abstract resource interfaces and versioned stubs (`*V1`, `*V2`)
* Minimal wiring to allow future version switching and fallback support

No functional endpoints are migrated yet. This PR establishes a stable foundation for subsequent migration and refactor work.

Issue #1575, opened 2025-12-30
[ENH] V1 → V2 API Migration


geetu040:
This issue tracks the migration of the `openml-python` SDK from the legacy OpenML API v1 (PHP-based) in [openml/OpenML](https://github.com/openml/OpenML) to the newer API v2 (Python-based) in [openml/server-api](https://github.com/openml/server-api), along with a structural refactor of the codebase to support both APIs during the transition.

The SDK currently relies on the v1 API, which should be progressively replaced by v2 while keeping backward compatibility and fallback where v2 endpoints are incomplete.

### Goals

* Separate API interaction from SDK logic by introducing clear abstraction layers
* Modularize HTTP handling, API resources, and SDK-facing interfaces
* Allow easy switching between API v1 and v2
* Provide a transparent fallback to v1 for missing or unimplemented v2 endpoints

### Proposed design (high level)

* A shared `HTTPClient` responsible for:
  * retries
  * proxy configurations
  * caching
  * exception handling
  * thread safety
* Abstract API "blueprints" (e.g. `DatasetsAPI`, `TasksAPI`) defining SDK-facing methods
* Version-specific implementations (`*V1`, `*V2`) mapping those methods to concrete endpoints
* A fallback proxy that routes calls to v1 if a v2 endpoint is unavailable (unless `strict=True`)
* A central backend/context that controls API version selection and exposes resources internally

<details>

<summary>A draft implementation sketch is included below to illustrate the architecture (not final API).</summary>

```py
# handles HTTP requests, retries, and proxy configuration

class HTTPClient:
    def __init__(self, base_url: str):
        self.base_url = base_url

    def get(self, path, params=None):
        return requests.get(f"{self.base_url}{path}", params=params)

    def post(self, path, params=None, files=None):
        return requests.post(f"{self.base_url}{path}", params=params, files=files)


# abstract blueprints defining SDK-exposed API endpoints and methods

class DatasetsAPI(ABC):
    @abstractmethod
    def get(self, id: int) -> dict: ...

class TasksAPI(ABC):
    @abstractmethod
    def get(self, id: int) -> dict: ...

    @abstractmethod
    def list(self, id: int) -> dict: ...


# version-specific implementations built on top of the API blueprints

class DatasetsV1(DatasetsAPI):
    def __init__(self, http: HTTPClient):
        self._http = http

    def get(self, id: int) -> dict:
        r = self._http.get(f"/api/v1/json/data/{id}")
[...]

PR #1604, opened 2026-01-07
[ENH] V1 → V2 API Migration - estimation procedures

### LINKED ISSUE #1622: [ENH] V1 → V2 API Migration - estimation procedures
V1 → V2 API Migration - estimation procedures. Sub-issue of #1575, for tracking.

EmanAbdelhaleem:
Fixes #1622
Depends on #1576 
Related to: #1575 

#### Details
This PR implements `EstimationProcedures` resource, and refactor its existing functions


geetu040 (commented):
This looks great. The implementation is clean and nicely decoupled. The next step should be adding tests, but let's wait until we have a clearer roadmap for that.
Also, please update the PR description to follow this format:
```
Fixes #...
Depends on #1576
Related to #1575
```

geetu040 (commented):
Can you please take a look at the function `_get_estimation_procedure_list` in `tasks/functions` and see if it can also be replaced via the `list` method or a new method in `EstimationProcedures` class?

EmanAbdelhaleem:
@geetu040 I get this assertion error for the matching test:
```
>       assert output_v1 == output_v2
E       AssertionError: assert ['10-fold Cro...out set', ...] == ['10-fold Cro...out set', ...]
E
E         At index 28 diff: '3-fold Crossvalidation' != '10-fold Crossvalidation'
E         Right contains 3 more items, first extra item: '10% Holdout set'
E         Use -v to get more diff

tests\test_api\test_estimation_procedures.py:107: AssertionError
```

When trying to check the 2 endpoints, there is really a difference in the returned response, the json response has 3 more estimation procedures.

openml/_api/resources/estimation_procedure.py
```diff
@@ -0,0 +1,131 @@
...
+from openml.tasks.task import TaskType
+
+
+class EstimationProceduresV1(ResourceV1, EstimationProceduresAPI):
+    """V1 API implementation for estimation procedures.
+
+    Fetches estimation procedures from the v1 XML API endpoint.
+    """
+
+    def list(self) -> builtins.list[str]:
```
geetu040: this should work
```suggestion
    def list(self) -> list[str]:
```

geetu040 (changes requested):
sdk code look good so far, please take a look at https://github.com/openml/openml-python/issues/1575#issuecomment-3816198783 and make changes accordingly where needed, though I doubt any point affects this PR, but please check.

openml/_api/resources/estimation_procedure.py
EmanAbdelhaleem: It causes pre-commit errors.

EmanAbdelhaleem:
> all tests (existing and new) should pass to make sure we are retaining the original functionality of the sdk

What about the matching test problem?

geetu040:
> What about the matching test problem?

didn't see the comment above, I'll look into it now

geetu040:
> What about the matching test problem?

I check and:
- `localhost - php-api-v1` gives same list
- `localhost - python-api-v2` gives same list
- `https://test.openml.org/`: gives same list
- `https://www.openml.org/`: gives 3 extra values: `{'3-fold Crossvalidation', '5-fold Crossvalidation', '10 times 3-fold Crossvalidation'}`

for now, you should be comparing `https://test.openml.org/` and `localhost - python-api-v2` which should work fine, later we will switch `https://test.openml.org/` with `localhost - php-api-v1` which should also work fine
so can you make sure you are not using the production server which is `https://www.openml.org/` for comparison, since the default config triggers that

EmanAbdelhaleem:
@geetu040 Ready for review

tests/test_api/test_estimation_procedure.py
```diff
@@ -0,0 +1,99 @@
...
+        self.client = self._get_http_client(
+            server="http://localhost:8001/",
+            base_url="",
+            api_key="",
+            timeout_seconds=self.timeout_seconds,
+            retries=self.retries,
+            retry_policy=self.retry_policy,
+            cache=self.cache,
+        )
+        self.resource = EstimationProcedureV2API(self.client)
```
geetu040: you can use `self.http_clients[APIVersion.V2]`

geetu040 (commented):
looks good, let's see if we need to change anything after discussion concludes on design

tests/test_api/test_estimation_procedure.py
EmanAbdelhaleem: updated

geetu040 (changes requested):

geetu040 (changes requested):

geetu040 (approved):
@EmanAbdelhaleem Thanks for the PR. Nicely done!
I have updated the PR to sync with latest changes in base PR.

@PGijsbers please review/merge.

openml/_api/resources/estimation_procedure.py
PGijsbers: Can you try again? It really shouldn't. If it does and you don't understand why or don't know how to best resolve it, please let us know.

PGijsbers (changes requested):
I had a closer look at the old code. Best I can tell `list_estimation_procedures` is never called nor is it actually public (or am I wrong?). The PR also keeps some code duplication that existed in the old style. I would refactor it as follows:

 - remove the current `list` function and rename `list_detailed` to `list`. Going forward, the only way to access the estimation procedures is with all their metadata. It seems that besides the retention of metadata, the only difference is in how to assess that the server response is parseable, and I don't think those differences matter (especially since it looks like `list_estimation_procedures` isn't used).
 - define an EstimationProcedure named tuple which has fields `id`, `name`, `task_type_id`, and `type`. Make the `list` function return that instead.
 - in `_get_estimation_procedure_list` you can convert the named tuples with `._asdict()`
 - in `list_estimation_procedures` just map to the names of the returned values to keep the response identical

This makes sure that at this stage we already reduce duplicate code, and that in the future we easily refactor further to make use of the typed return values instead.

geetu040:
@PGijsbers Thanks for the detailed suggestions. I've applied the changes. This PR is up for review/merge.

openml/estimation_procedures/estimation_procedure.py
```diff
@@ -0,0 +1,60 @@
...
+            "Name",
+            "Type",
+            "Task Type",
+        ]
+
+        _fields = [(key, fields[key]) for key in order if key in fields]
+
+        longest_field_name_length = max(len(name) for name, _ in _fields)
+        field_line_format = f"{{:.<{longest_field_name_length}}}: {{}}"
+        body = "\n".join(field_line_format.format(name, value) for name, value in _fields)
```
PGijsbers:
```suggestion
        fields = {
            "ID": self.id,
            "Name": self.name,
            "Type": self.type,
            "Task Type": self.task_type_id,
        }
        longest_field_name_length = max(len(name) for name in fields)
        field_line_format = f"{{:.<{longest_field_name_length}}}: {{}}"
        body = "\n".join(field_line_format.format(name, value) for name, value in fields.items())
```

Insertion order is preserved since Py3.6 (implementation detail of CPython) and 3.7 (as part of the spec).

PGijsbers (changes requested):
Minor cleanup, I'll commit it myself.

PGijsbers (approved):
will merge if CI looks good

PR #1607, opened 2026-01-08
[ENH] V1 -> V2 migration - Flow (module)

Omswastik-11:


Omswastik-11:
Hi @geetu040 !! 
I left `FlowV2` as not implemented as I don't know how the `V2` endpoints work . Are we going to solve it at a later stage ?
I was having `circular` import issues so I used `lazy` imports + linting suppression . Is it Ok ? or we should do something else?
Review this if I have correctly done the flow module migration , verify the added tests also.

PR #1608, opened 2026-01-08
[ENH] V1 → V2 API Migration - datasets

### LINKED ISSUE #1592: [ENH] V1 → V2 API Migration - datasets
V1 → V2 API Migration - datasets. Sub-issue of #1575, for tracking.

JATAYU000:
#### Metadata
* Fixes #1592
* Depends on:  #1576

* Change Log Entry:This PR implements Datasets resource, and refactor its existing functions


JATAYU000:
FYI @geetu040  Currently the `get_dataset()` function has 3 download requirement
- `download_data` : uses `api_calls._download_minio_bucket()` to download all the files in the bucket if `download_all_files` param was True and `api_calls._download_minio_file()` to download the `dataset.pq` file if it was not found in cache. When download parquet fails it fallback to download `dataset.arff` file with get request
- `download_features` :  if feature_file is passed via init it extracts during initialization else does get request and caches the xml 
- `download_qualities` : if qulities_file is passed via init it extracts during initialization else does get request and caches the xml 

### Issues:
- The data files `.pq` and `.arff` are common for versions and doesn't make sense to be downloaded multiple times
- Path handling for download to return the path especially the data files, As mentioned in the meet I can try the Download specific class which uses the cache mixin and only inherited by dataset resource.
- Current implementation in `OpenMLDataset` has v1 specific parsing which in my opinion should be using the current interface (api_context)
### Example:
current `load_features()` [ref link](https://github.com/openml/openml-python/blob/c5f68bf15e1b18ee8593de6435120ed5c0dd1971/openml/datasets/dataset.py#L797)
This calls a function which downloads and returns a file path and then parse from the file path
This can be changed by changing that function's definition [ref link](https://github.com/openml/openml-python/blob/c5f68bf15e1b18ee8593de6435120ed5c0dd1971/openml/datasets/functions.py#L1252) to get -> parse -> return features instead of file paths

```python
def _get_dataset_features_file(did_cache_dir: str | Path | None, dataset_id: int) -> dict[int, OpenMLDataFeature]:
        return _features
```
Or by updating the Dataset class to use the underlining interface method from api_context directly.

```python
def _load_features(self) -> None:
       ...
        self._features = api_context.backend.datasets.get_features(self.dataset_id)
```
Another option is to add `return_path` to client requests, which in my opinion would be wasteful since adding a param to all the methods of client for just the dataset resource, and that too which could be handled without it as mentioned above.

openml/_api/resources/base/resources.py
```diff
@@ -0,0 +1,83 @@
...
+
+    @abstractmethod
+    def list(
+        self,
+        limit: int,
+        offset: int,
+        *,
+        data_id: list[int] | None = None,  # type: ignore
+        **kwargs: Any,
+    ) -> pd.DataFrame: ...
```
geetu040, line 51: can we not have same signature for all 3 methods: `DatasetsAPI.list`, `DatasetsV1.list`, `DatasetsV2.list`? does it raise pre-commit failures since a few might not be used?

openml/_api/resources/dataset.py
```diff
@@ -0,0 +1,650 @@
...
+        return dataset
+
+    def list(
+        self,
+        limit: int,
+        offset: int,
+        *,
+        data_id: list[int] | None = None,  # type: ignore
+        **kwargs: Any,
+    ) -> pd.DataFrame:
```
geetu040, line 127: you can make this simple using private helper methods

openml/_api/resources/datasets.py
```diff
@@ -0,0 +1,650 @@
...
+        ----------
+        dataset_id : int
+            OpenML id of the dataset
+
+        Returns
+        -------
+        bool
+            True if the deletion was successful. False otherwise.
+        """
+        return openml.utils._delete_entity("data", dataset_id)
```
geetu040: if you implement the delete logic yourself instead of `openml.utils._delete_entity`, how would that look? I think it would be better.

openml/_api/resources/datasets.py
```diff
@@ -0,0 +1,650 @@
...
+            return dataset, response
+
+        return dataset
+
+    def list(
+        self,
+        limit: int,
+        offset: int,
+        **kwargs: Any,
+    ) -> pd.DataFrame:
```
geetu040, line 989: same as above, it can use private helper methods

openml/datasets/functions.py
```diff
@@ -65,17 +64,9 @@ def list_qualities() -> list[str]:
     -------
     list
     """
-    api_call = "data/qualities/list"
-    xml_string = openml._api_calls._perform_api_call(api_call, "get")
-    qualities = xmltodict.parse(xml_string, force_list=("oml:quality"))
-    # Minimalistic check if the XML is useful
-    if "oml:data_qualities_list" not in qualities:
-        raise ValueError('Error in return XML, does not contain "oml:data_qualities_list"')
+    from openml._api import api_context
```
geetu040: can't we have this import at the very top? does it create circular import error? if not, should be moved to top from all functions.

geetu040 (changes requested):
Left an intermediate review. This is solid work and well done overall. Nice job. I'll look into the download part now.

openml/_api/resources/datasets.py
JATAYU000: Makes Sense , It would look like a delete request from client along with exception handling

openml/datasets/functions.py
JATAYU000: It does raise circular import

openml/_api/resources/base/resources.py
JATAYU000: Oh that v2 signature was experimental, idk how pre-commits did not catch that, Will make them same

openml/_api/resources/dataset.py
JATAYU000: @geetu040  I could not find a good common block to make an helper since the filters are passed via url in v1 and via json object in v2 , and both have different parsing, If you have any specific Idea on that please let me know

geetu040: looking at `DatasetsV1` I can think of these helper methods: `_build_list_call`, `_parse_and_validate_xml`, `_parse_dataset`
you can do something similar for `DatasetsV2` though they can be different.

JATAYU000: > you can do something similar for DatasetsV2 though they can be different.

I see, That opens more options.

geetu040:
> FYI @geetu040 Currently the `get_dataset()` function has 3 download requirement

Thanks for a detailed explanation, I now have good understanding of the download mechanism.

> `download_data` : uses `api_calls._download_minio_bucket()` to download all the files in the bucket if `download_all_files` param was True and `api_calls._download_minio_file()` to download the `dataset.pq` file if it was not found in cache. When download parquet fails it fallback to download `dataset.arff` file with get request

minio can be handled easily, we will use a separate client along with `HTTPClient` or implement it's methods in the `HTTPClient`, which work independently of the api version

> * `download_features` :  if feature_file is passed via init it extracts during initialization else does get request and caches the xml
> 
> * `download_qualities` : if qulities_file is passed via init it extracts during initialization else does get request and caches the xml

these are actually different objects in both apis, v1 uses xml and v2 keeps them in json

> The data files `.pq` and `.arff` are common for versions and doesn't make sense to be downloaded multiple times

yes you are right, they are the same files, which are not required to be downloaded again for both versions, but isn't this true for almost all the http objects? they may have different format `xml` or `json`, slightly different structure, but if parsed most are identical, so shouldn't this rule be applied to all the responses?

> Path handling for download to return the path especially the data files, As mentioned in the meet I can try the Download specific class which uses the cache mixin and only inherited by dataset resource.

I don't understand this point

> Current implementation in `OpenMLDataset` has v1 specific parsing which in my opinion should be using the current interface (api_context)

agreed, should be handled by `api_context`

> Another option is to add `return_path` to client requests, which in my opinion would be wasteful since adding a param to all the methods of client for just the dataset resource, and that too which could be handled without it as mentioned above.

agreed, adding `return_path` for just one specific method of one resource is not preffered

---

in conclusion, I may ask, if we ignore the fact that it downloads the `.arff` files for both versions separately, does everything else works out smooth without any blocker? I think ignoring this part is not really bad because conceptually this rule could be applied to almost every other response object

JATAYU000:
@geetu040 making a new client for `minio` handles just the parquet file, we would still need to migrate `download_text_file()` for the arff file (this is also used by tasks and runs) 
So maybe we can have a `DownloadClient` which can contain all of these along with a save method which can save content to a specified path and hence also fixes our issue with features/qualities path ?

FYI the new commit adds better handling of feature and qualites in OpenMLDataset class moving the v1 specific parsing logic to the interface. So only part left is to handle 
- return path of saved file (feature, qualities, arff, pq)
- downloader for arff or implementation of `download_text_file()` which is used for arff download
- minio file and bucket download for the pq file

geetu040:
From the standup discussion and earlier conversations, I think we can agree on a few points:

* Have a separate client for `MinIO` interactions alongside `HTTPClient`. In future if we plan to add more providers like dropbox, google drive, e.t.c, we don't end up with too many changes, rather have their client implemented as a separate class and just pass that down to relevant resource.
* `DownloadClient` doesn't feel like the right abstraction; instead, implement download-specific methods directly in `HTTPClient`.

Consider this a green light to experiment with the client design. Try an approach, use whatever caching strategy you think fits best, and aim for a clean, sensible design. Feel free to ask for suggestions or reviews along the way. I'll review it in code. Just make sure this doesn't significantly impact the base classes or other stacked PRs.

JATAYU000:
The points do make sense to me, I will propose the design along with how it would be used in the resource.

JATAYU000:
@geetu040 I have a design implemented which needs reviews


`MinIOClient` similar to `HTTPClient` is being used by `DatasetAPI` from `self._minio` , It implements 2 methods download file and download bucket, it uses `_get_cache_dir()` for the destination
New method `download` implemented under `HTTPclient` that can be used for features,qualities and arff files, along with specific `v1/v2` interface using handler callback.

Question:
- most methods signature include cache_directory how should that be handled? if the directory is passed use that if not use our cache dir? i am not sure how this would effect the old users
- Also the caching implemented currently suggest the Response() is cached but I remember from a meeting you mentioned the respective files (.xml .json) are cached, I am not sure about it , I have went through the design as if the caching is done on the response.

geetu040:
> @geetu040 I have a design implemented which needs reviews

I have taken a quick look, the design looks really good, though I have some suggestions/questions in the code, which I will leave in a detailed review. But this in general fixes all our blockers without damaging the original design.

> most methods signature include cache_directory how should that be handled? if the directory is passed use that if not use our cache dir? i am not sure how this would effect the old users

Is it provided by the user? I don't think so. In that case, how does it affect the users? From looking at the code, this cache directory is generated programmatically inside the functions, we can completely remove these lines and always rely on the `CacheMixin` class. How does that sound?

> Also the caching implemented currently suggest the Response() is cached but I remember from a meeting you mentioned the respective files (.xml .json) are cached, I am not sure about it , I have went through the design as if the caching is done on the response.

`CacheMixin._set_cache_response` will look at the response object and extract `json` or `xml` content from it and save it respectively in `.json` and `.xml` files.
`CacheMixin._get_cache_response` will read these files (`.json` or `.xml`) from the given path and create a dummy Response object then fill it with status_code and content. Therefore a `Response` object will be returned.

JATAYU000:
> `CacheMixin._set_cache_response` will look at the response object and extract `json` or `xml` content from it and save it respectively in `.json` and `.xml` files.
> `CacheMixin._get_cache_response` will read these files (`.json` or `.xml`) from the given path and create a dummy Response object then fill it with status_code and content. Therefore a `Response` object will be returned.

This makes a sense now. having an independent download method as I have setup is better than updating requests/caching to return path right?

> we can completely remove these lines and always rely on the CacheMixin class. How does that sound?

Yes that would work, but the function definition would be changed i.e. tests etc corresponding to them

geetu040:
> This makes a sense now. having an independent download method as I have setup is better than updating requests/caching to return path right?

I am not sure about that, would require a detailed review

> Yes that would work, but the function definition would be changed i.e. tests etc corresponding to them

yes, that is expected

openml/_api/resources/base/resources.py
satvshr: Is `mypy` supposed to catch that?

openml/_api/resources/dataset.py
```diff
@@ -0,0 +1,1102 @@
...
+                f.write(response.text)
+            return file_path
+
+        return self._http.download(url_ext, __handler, encoding)
+
+    def download_features_file(self, dataset_id: int) -> Path:
+        path = f"data/features/{dataset_id}"
+        return self.download_file(path)
+
+    def download_qualities_file(self, dataset_id: int) -> Path:
```
satvshr: May I know why this function (and similar functions), which are used only in `DatasetsV1` and V2, and not called inside `functions.py`, not private? It becomes hard to distinguish what is actually being called by the SDK and what is being abstracted away.

openml/_api/http/client.py
```diff
@@ -0,0 +1,302 @@
...
+        source : str
+            URL to a MinIO bucket.
+        destination : str | Path
+            Path to a directory to store the bucket content in.
+        """
+        destination = self._get_cache_dir(source, {}) if destination is None else Path(destination)
+        parsed_url = urllib.parse.urlparse(source)
+
+        # expect path format: /BUCKET/path/to/file.ext
+        _, bucket, *prefixes, _file = parsed_url.path.split("/")
```
satvshr: `_file` should be renamed to `_` given it is never called, surprised ruff does not call it out.

openml/datasets/functions.py
```diff
@@ -96,7 +80,7 @@ def list_datasets(
 
     Parameters
     ----------
-    data_id : list, optional
+    dataset_id : list, optional
```
satvshr: Should be `data_id`

openml/datasets/functions.py
```diff
@@ -1114,7 +937,7 @@ def _get_dataset_description(did_cache_dir: Path, dataset_id: int) -> dict[str,
 
 def _get_dataset_parquet(
     description: dict | OpenMLDataset,
-    cache_directory: Path | None = None,
+    cache_directory: Path | None = None,  # noqa: ARG001
```
satvshr: Why the `ARG001` addition, I assume ruff checks were passing before this was added too?

Edit1: Just saw that this is not being used, why not remove it then?

Edit2: This is happening in more than 1 place, I won't mention it in all of them so we can just discuss it here.

openml/datasets/functions.py
```diff
@@ -1144,47 +967,14 @@ def _get_dataset_parquet(
...
-                destination=output_file_path,
-            )
-        except (FileNotFoundError, urllib3.exceptions.MaxRetryError, minio.error.ServerError) as e:
-            logger.warning(f"Could not download file from {url}: {e}")
-            return None
-    return output_file_path
+    return api_context.backend.datasets.download_dataset_parquet(description, download_all_files)
 
 
 def _get_dataset_arff(
```
satvshr, line 911: @geetu040 This function is not being used yet is still present, because it is being used in tests. Should the test involving this be removed, or should this function be kept? (Same scenario exists in tasks migration too)

Edit: I see this happening in multiple places (`_get_dataset_features_file` and `_get_dataset_qualities_file`)

openml/datasets/functions.py
```diff
@@ -1208,47 +998,15 @@ def _get_dataset_arff(
...
-        md5_checksum_fixture = description.md5_checksum
-        assert description.url is not None
-        assert description.dataset_id is not None
-
-        url = description.url
-        did = int(description.dataset_id)
-    else:
-        raise TypeError("`description` should be either OpenMLDataset or Dict.")
-
-    save_cache_directory = (
```
satvshr, line 1239: I assume cache paths for V1 and V2 are different, hence this code is being removed?

openml/datasets/functions.py
JATAYU000: Still under work,This method is now updated to not use the param, working on removing it here and update the corresponding test when starting to write tests for this pr, as mentioned in the meet

openml/_api/resources/dataset.py
JATAYU000: It is called in `functions.py` I haven't updated it yet, was still working on it

openml/_api/resources/base/resources.py
JATAYU000: yes unused parameters are caught under #ARG001 as seen in the `cache_directory` params.

openml/datasets/functions.py
```diff
@@ -1459,4 +1120,6 @@ def delete_dataset(dataset_id: int) -> bool:
     bool
         True if the deletion was successful. False otherwise.
     """
-    return openml.utils._delete_entity("data", dataset_id)
```
satvshr, line 1476: @geetu040 @JATAYU000 I remember being on a call and discussing this, didn't we decide to implement this in the base PR (core) itself and leave it out of our migration implementations? If not, I will need to change tasks.

satvshr (commented):
Left a few comments, will look at this again once it is actually ready for review with all implementations complete.

openml/datasets/functions.py
JATAYU000: Yes it was, since it is not yet added I haven't removed this yet

openml/datasets/functions.py
JATAYU000: @satvshr refering to the conversation [here](https://github.com/openml/openml-python/pull/1608#issuecomment-3727215045)

There are xml specific parsing logic in `OpenMLDataset` which should be implemented under the current interfaces for `v1/v2` support. The load methods for features and qualities are updated to use the `api_context`

satvshr:
Is there an API_KEY that you are using to test the endpoints? I was going to run a test script for your code but I could not given I have no API_KEY for it.

JATAYU000:
> Is there an API_KEY that you are using to test the endpoints? I was going to run a test script for your code but I could not given I have no API_KEY for it.

If you are talking about the invalid api_key regex match in v2 you can set this in the config
```
key="AD000000000000000000000000000000"
```

SubhadityaMukherjee:
> > Is there an API_KEY that you are using to test the endpoints? I was going to run a test script for your code but I could not given I have no API_KEY for it.
> 
> If you are talking about the invalid api_key regex match in v2 you can set this in the config
> 
> ```
> key="AD000000000000000000000000000000"
> ```

Um why are we discussing API keys here. Even if its just for testing.

openml/_api/clients/http.py
```diff
@@ -0,0 +1,480 @@
...
+        )
+
+    def download(
+        self,
+        url: str,
+        handler: Callable[[Response, Path, str], Path] | None = None,
+        encoding: str = "utf-8",
+        file_name: str = "response.txt",
+        md5_checksum: str | None = None,
+    ) -> Path:
```
geetu040, line 806: we should ensure that download method always has `cache` object, raise an exception if `cache` is `None`, then in the following code use path from cache

openml/_api/resources/base/resources.py
```diff
@@ -0,0 +1,143 @@
...
+
+    @abstractmethod
+    def delete(self, dataset_id: int) -> bool: ...
```
geetu040: you can remove it from here as well
see point 5 in https://github.com/openml/openml-python/issues/1575#issuecomment-3816198783

openml/datasets/functions.py
```diff
@@ -475,59 +379,13 @@ def get_dataset(  # noqa: C901, PLR0912
         if did_cache_dir.exists():
             _remove_cache_dir_for_id(DATASETS_CACHE_DIR_NAME, did_cache_dir)
 
-    did_cache_dir = _create_cache_directory_for_id(
-        DATASETS_CACHE_DIR_NAME,
+    return api_context.backend.datasets.get(
```
geetu040: to keep the functionality of `force_refresh_cache`, you can use `reset_cache` for `http.get`
see point 1 in https://github.com/openml/openml-python/issues/1575#issuecomment-3816198783

geetu040 (changes requested):
sdk code look good so far, please take a look at https://github.com/openml/openml-python/issues/1575#issuecomment-3816198783 and make changes accordingly where needed.

SimonBlanke:
@JATAYU000 The recent changes in the base branch fixes the test fail in the windows test job. Please update the base branch.

tests/test_datasets/test_dataset.py
```diff
@@ -469,16 +470,3 @@ def test__read_qualities(static_cache_dir, workdir, mocker):
     assert pickle_mock.dump.call_count == 1
 
 
-
-def test__check_qualities():
```
geetu040, line 473: why remove this test?

openml/_api/resources/dataset.py
```diff
@@ -0,0 +1,1089 @@
...
+        collection_date: str | None = None,
+        language: str | None = None,
+        default_target_attribute: str | None = None,
+        ignore_attribute: str | builtins.list[str] | None = None,
+        citation: str | None = None,
+        row_id_attribute: str | None = None,
+        original_data_url: str | None = None,
+        paper_url: str | None = None,
+    ) -> int:
+        raise NotImplementedError(self._not_supported(method="edit"))
```
EmanAbdelhaleem: You can just use `self._not_supported(method="edit")`
No need for  `raise NotImplementedError()`

tests/test_api/test_datasets.py
```diff
@@ -0,0 +1,340 @@
...
+import pandas as pd
+from openml._api.resources.base.fallback import FallbackProxy
+from openml.testing import TestAPIBase
+from openml._api.resources.dataset import DatasetV1API, DatasetV2API
+
+
+class TestDatasetV1API(TestAPIBase):
+    def setUp(self):
+        super().setUp()
+        self.client = self._get_http_client(
```
EmanAbdelhaleem: since this is V1, using 
`self.client = self.http_client` will do.

tests/test_datasets/test_dataset.py
geetu040: got it, since it's moved to resource api, should be tested at resource class level then

tests/test_api/test_datasets.py
geetu040: since the recent change, it can be used from `self.http_clients[APIVersion.V1]`

JATAYU000: Alright, I have made the change

tests/test_api/test_datasets.py
```diff
@@ -0,0 +1,340 @@
...
+            timeout=self.timeout,
+            retries=self.retries,
+            retry_policy=self.retry_policy,
+            cache=self.cache
+        )
+        self.dataset_v1 = DatasetV1API(self.v1_client,self.minio_client)
+        self.dataset_v2 = DatasetV2API(self.v2_client,self.minio_client)
+        self.dataset_fallback = FallbackProxy(self.dataset_v1,self.dataset_v2)
+    
+    @pytest.mark.uses_test_server()
```
geetu040: if all the methods use `@pytest.mark.uses_test_server()`, you can simply move this decorator above the class

tests/test_datasets/test_dataset.py
```diff
@@ -233,6 +233,7 @@ def test_get_data_corrupt_pickle(self):
         assert isinstance(xy, pd.DataFrame)
         assert xy.shape == (150, 5)
 
+    @pytest.mark.skip("Datasets cache")
```
geetu040: remove this instead of skipping, since we are not going to use it anywhere in future.

tests/test_datasets/test_dataset_functions.py
```diff
@@ -1728,89 +1733,66 @@ def test_delete_dataset(self):
...
     content_file = (
         test_files_directory / "mock_responses" / "datasets" / "data_delete_has_tasks.xml"
     )
-    mock_delete.return_value = create_request_response(
-        status_code=412,
-        content_filepath=content_file,
-    )
+    content_xml = content_file.read_text()
+    requests_mock.delete(ANY, text=content_xml, status_code=412)
+
```
geetu040: why not using `create_request_response`

geetu040 (changes requested):
- can you please look at the failing test, is it related to your code? `tests/test_runs/test_run_functions.py::TestRun::test_format_prediction_non_supervised`
- maybe use design from here if it helps [tests/test_api/test_versions.py](https://github.com/geetu040/openml-python/blob/migration/tests/test_api/test_versions.py)
- your tests involve publishing something, that should be deleted, for this cleanup you can use `TestBase._mark_entity_for_removal`

tests/test_api/test_datasets.py
JATAYU000: Yes, will do

tests/test_datasets/test_dataset_functions.py
JATAYU000: Sorry I missed this comment, I tried could not mock request, but your comment in the base pr fixes this, I have updated it

tests/test_datasets/test_dataset_functions.py
```diff
@@ -1745,12 +1750,13 @@ def test_delete_dataset_not_owned(mock_delete, test_files_directory, test_api_ke
     ):
         openml.datasets.delete_dataset(40_000)
 
-    dataset_url = "https://test.openml.org/api/v1/xml/data/40000"
-    assert dataset_url == mock_delete.call_args.args[0]
+    dataset_url = "https://test.openml.org/api/v1/data/40000"
+    assert dataset_url == mock_delete.call_args.kwargs.get("url")
+    assert 'DELETE' == mock_delete.call_args.kwargs.get("method")
```
geetu040: Failing tests are genuine:

expected url is
```https://test.openml.org/api/v1/xml/data/40000```
not
```https://test.openml.org/api/v1/data/40000```

there was a bug that would ignore `xml` in link, but it's fixed now.

**Review by geetu040 — CHANGES_REQUESTED** [review]

File: tests/test_datasets/test_dataset_functions.py (hunk 22 above)
Comment (JATAYU000) [reply tests/test_datasets/test_dataset_functions.py]: Thank you for pointing it out, it is fixed now

File: openml/datasets/functions.py (hunk 11 above)
Comment (satvshr) [reply openml/datasets/functions.py:1239]: @JATAYU000 I think you missed this

File: tests/test_datasets/test_dataset.py [hunk 23]
```diff
@@ -233,53 +233,6 @@ def test_get_data_corrupt_pickle(self):
...
-            2,
-            download_data=False,
-            download_features_meta_data=False,
-            download_qualities=False,
-        )
-        assert change_time == os.stat(did_cache_dir).st_mtime
-        assert _dataset.features == _compare_dataset.features
-        assert _dataset.qualities == _compare_dataset.qualities
-
-        # -- Test without cache
```
satvshr, line 261: You removed the test without cache too? Shouldn't this still be kept if it tests something that does not require caching?

tests/test_datasets/test_dataset_functions.py
```diff
@@ -438,30 +441,30 @@ def test__getarff_md5_issue(self):
         }
         n = openml.config.connection_n_retries
         openml.config.connection_n_retries = 1
+        openml.config._sync_api_config()
 
         self.assertRaisesRegex(
             OpenMLHashException,
-            "Checksum of downloaded file is unequal to the expected checksum abc when downloading "
-            "https://www.openml.org/data/download/61. Raised when downloading dataset 5.",
+            "Checksum of downloaded file is unequal to the expected checksum abc "
```
satvshr, line 404: Did ruff force this change? Seems unnecessary and does not need to be changed otherwise.

tests/test_datasets/test_dataset_functions.py
```diff
@@ -565,7 +568,7 @@ def test__retrieve_class_labels(self):
         labels = openml.datasets.get_dataset(2).retrieve_class_labels(
             target_name="product-type",
         )
-        assert labels == ["C", "H", "G"]
+        assert labels == ["C", "G", "H"]
```
satvshr: Why this change?

tests/test_datasets/test_dataset_functions.py
```diff
@@ -617,6 +620,7 @@ def test_data_status(self):
         # admin key for test server (only admins can activate datasets.
         # all users can deactivate their own datasets)
         openml.config.apikey = TestBase.admin_key
+        openml.config._sync_api_config()
```
satvshr: I am a bit confused as to why this function is being called everywhere, is it needed in these places?

tests/test_api/test_datasets.py
```diff
@@ -0,0 +1,242 @@
+from __future__ import annotations
+from pathlib import Path
+import time
+
+from openml import OpenMLDataset
+import pytest
+import pandas as pd
+from openml._api.resources.base.fallback import FallbackProxy
```
satvshr: How do you not get an unused import error? Did you forget to use it?

openml/datasets/functions.py
```diff
@@ -3,42 +3,33 @@
...
-    _get_cache_dir_for_id,
-    _remove_cache_dir_for_id,
 )
 
 from .dataset import OpenMLDataset
 
 if TYPE_CHECKING:
     import scipy
 
+
```
satvshr: remove new line

openml/datasets/functions.py
```diff
@@ -1035,13 +843,7 @@ def _topic_add_dataset(data_id: int, topic: str) -> int:
     -------
     Dataset id
     """
-    if not isinstance(data_id, int):
```
satvshr, line 1052: Check being removed

openml/datasets/functions.py
```diff
@@ -1060,15 +862,10 @@ def _topic_delete_dataset(data_id: int, topic: str) -> int:
...
```
satvshr, line 1077: Check being removed

openml/datasets/functions.py
```diff
@@ -1060,15 +862,10 @@ def _topic_delete_dataset(data_id: int, topic: str) -> int:
...
-        raise TypeError(f"`data_id` must be of type `int`, not {type(data_id)}.")
-    form_data = {"data_id": data_id, "topic": topic}  # type: openml._api_calls.DATA_TYPE
-    result_xml = openml._api_calls._perform_api_call("data/topicdelete", "post", data=form_data)
-    result = xmltodict.parse(result_xml)
-    data_id = result["oml:data_topic"]["oml:id"]
-    return int(data_id)
+    return openml._backend.dataset.delete_topic(data_id, topic)
 
 
+# TODO used by tests only
```
satvshr: Just remove the test no?

Same question for all the other "TODO used by test only" comments

openml/datasets/functions.py
```diff
@@ -1269,28 +994,11 @@ def _get_dataset_features_file(did_cache_dir: str | Path | None, dataset_id: int
...
-    return features_file
-
-
-def _get_qualities_xml(dataset_id: int) -> str:
-    url_extension = f"data/qualities/{dataset_id!s}"
-    return openml._api_calls._perform_api_call(url_extension, "get")
+    return openml._backend.dataset.download_features_file(dataset_id)
 
 
+# TODO remove cache dir
```
satvshr: Still a todo?

openml/datasets/functions.py
```diff
@@ -1313,96 +1021,10 @@ def _get_dataset_qualities_file(
...
-                # quality file stays as None
-                logger.warning(f"No qualities found for dataset {dataset_id}")
-                return None
-
-            raise e
-
-    return qualities_file
-
-
-def _create_dataset_from_description(
```
satvshr, line 1357: Was this being used only in tests too?

satvshr (commented):
Questions and minor changes, else looks pretty good!

tests/test_datasets/test_dataset_functions.py
geetu040: This is coming from the core pr, this is to sync the legacy config with the new api config, it's temporary and should be updated in the cleanup

tests/test_api/test_datasets.py
geetu040: I doubt that the test files are being checked by pre-commit, I found many spots being missed by it

geetu040 (changes requested):
I have updated the `HTTPClient` with `download` method, please sync with the base branch

tests/test_api/test_datasets.py
satvshr: That is weird, even the duplicates error I pointed out in core should have been caught by ruff, will have a look

openml/datasets/functions.py
JATAYU000: Caching path is removed from here, since handled in the http client, not due to different paths?

openml/datasets/functions.py
JATAYU000: mb, missed to remove it

openml/datasets/functions.py
JATAYU000: no this was used by get_dataset

tests/test_datasets/test_dataset_functions.py
JATAYU000: The checksum error raised does not point out its was for dataset {id} in the client

tests/test_datasets/test_dataset_functions.py
JATAYU000: Oh,  this was part of fixing the tests which used the test server, before it was mentioned only to change the parts where caching was tested, This change was not reverted

tests/test_datasets/test_dataset.py
JATAYU000: even though did not use caching the paths is checked for, and the path is changed now
did_cache_dir
Also, have restored test_equality_comparison

openml/datasets/functions.py
JATAYU000: THere are a lot of tests relied on it,  same  goes to the other ones removing would result in many tests  being removed

JATAYU000:
@geetu040 These are the two functions under `datatsets/functions.py`
- `_get_online_dataset_format`
- `_get_online_dataset_arff`

They are being used by the following tests alone
- `test_create_dataset_pandas`
- `test_get_online_dataset_format`
- `test_create_dataset_sparse`
- `test_create_dataset_numpy`
- `test_create_dataset_list`
- `test_get_online_dataset_arff`
- `test_create_dataset_row_id_attribute_inference`

geetu040:
> @geetu040 These are the two functions under `datatsets/functions.py`

I would say let's keep them since they are still being used by tests

openml/datasets/functions.py
satvshr: Did you add the checks back?

JATAYU000: yes

geetu040 (changes requested):

JATAYU000:
@geetu040 `_dataset_file_is_downloaded` using cache path to check is `description.xml`, `features.xml`, etc
which is being used by multiple tests via `_dataset_description_is_downloaded(id)` and are failing, What should be done about it?

geetu040:
> @geetu040 `_dataset_file_is_downloaded` using cache path to check is `description.xml`, `features.xml`, etc
> which is being used by multiple tests via `_dataset_description_is_downloaded(id)` and are failing, What should be done about it?

I realized there was a leakage with `openml.config.servers`. I have fixed it now, which should fix these failures.
Please sync with the base PR again and let me know if the tests are still failing.

JATAYU000:
The tests which `mock_delete` was  mocking delete instead of request 
update it to:
`@mock.patch.object(requests.Session, "request")`

Now only v2 tests are failing

openml/_api/resources/dataset.py
```diff
@@ -0,0 +1,1549 @@
...
+        -------
+        Path
+        """
+
+        def __handler(response: Response, path: Path, encoding: str) -> Path:
+            with path.open("w", encoding=encoding) as f:
+                f.write(response.text)
+            return path
+
+        return self._http.download(url_ext, __handler, encoding, file_path)
```
geetu040: how is this different from the default `handler`?

JATAYU000: This is no different from the default, but the same in v2 is different, I thought having them defined here helps better understanding of both v1 and v2?
Would you recommend to use the default handler in v1 dataset?

geetu040: Okay, I think it's fine.

geetu040 (changes requested):

JATAYU000:
@geetu040 There is one issue which is still not address in v2 status update
as we had discussed earlier that the temporary fix (passing api_ke in the url) for passing `api_key` and json through `request` needs to be changed and need to figure out a better way to do it.

geetu040:
> as we had discussed earlier that the temporary fix (passing api_ke in the url) for passing `api_key` and json through `request` needs to be changed and need to figure out a better way to do it.

I though we were going to use that temporary fix for this PR, since this happens only for a very particular case.

JATAYU000:
But you mentioned earlier that this is not an optimal fix and we should think of a better solution?

JATAYU000:
@geetu040 
All the test failures are related to minio unable to download in multiple threads? Do you have any idea why this would be happening?
```text
 DEBUG    urllib3.connectionpool:connectionpool.py:544 http://localhost:8000 "GET /minio/datasets/0000/0002/dataset_2.pq HTTP/1.1" 200 31505
WARNING  openml._api.resources.dataset:dataset.py:758 Could not download file from http://localhost:8000/minio/datasets/0000/0002/dataset_2.pq:
```

Also wwhy are tests on Windows not using Local (OPENML_USE_LOCAL_SERVICES)? 
https://github.com/openml/openml-python/blob/8a5532fe4444da1947143cae7379b6c55554782b/.github/workflows/test.yml#L179

tests/test_api/test_datasets.py
```diff
@@ -0,0 +1,285 @@
...
+        except Exception:
+            pass
+    raise TimeoutError(f"Dataset did not become {status} within given time")
+
+def _status_update_check(dataset, dataset_id, status):
+    dataset.status_update(dataset_id, status)
+    _wait_for_dataset_being_processed(dataset, dataset_id, status)
+
+
+def test_v1_get(dataset_v1):
```
geetu040, line 45: this and the following tests should use the marker `@pytest.mark.test_server()`

geetu040 (changes requested):
> All the test failures are related to minio unable to download in multiple threads? Do you have any idea why this would be happening?

May be the CI can't handle these many services in parallel? I believe so. I will check if reducing the parallel tests fixes this.

> Also wwhy are tests on Windows not using Local (OPENML_USE_LOCAL_SERVICES)?

https://github.com/openml/openml-python/pull/1629#issuecomment-3836181974

If you use the marker correctly it will skip these anyway.

geetu040:
Please add in description `Fixes #1592`

geetu040:
I believe https://github.com/openml/openml-python/pull/1611#pullrequestreview-3991387133 also affects this PR

geetu040 (commented):
Please see https://github.com/openml/openml-python/pull/1619#pullrequestreview-3996391505 regarding changed cache path

geetu040 (changes requested):
I've updated https://github.com/openml/openml-python/pull/1619#pullrequestreview-3996391505. Please sync with main and update cache files accordingly.

openml/datasets/dataset.py
```diff
@@ -954,6 +953,48 @@ def _parse_publish_response(self, xml_response: dict) -> None:
...
+
+        Parameters
+        ----------
+        tag : str
+            Tag to attach to the dataset.
+        """
+        if self.dataset_id is None:
+            raise ValueError(
+                "Dataset does not have an ID. Please publish the dataset before tagging."
+            )
```
geetu040, line 990: use `ObjectNotPublishedError` as in [_tag_openml_base](https://github.com/openml/openml-python/blob/main/openml/utils/_openml.py#L113)

openml/datasets/dataset.py
```diff
@@ -954,6 +953,48 @@ def _parse_publish_response(self, xml_response: dict) -> None:
...
+
+        Parameters
+        ----------
+        tag : str
+            Tag to remove from the dataset.
+        """
+        if self.dataset_id is None:
+            raise ValueError(
+                "Dataset does not have an ID. Please publish the dataset before untagging."
+            )
```
geetu040: use `ObjectNotPublishedError` as in [_tag_openml_base](https://github.com/openml/openml-python/blob/main/openml/utils/_openml.py#L113)

geetu040 (changes requested):

JATAYU000:
@geetu040 
This is the test which fails in ci but passes in local , related to minio path in `_dataset_data_file_is_downloaded()`
`tests/test_datasets/test_dataset_functions.py::test_get_dataset_lazy_behavior` 

I would be testing with some debuggers, here is the failed workflow https://github.com/openml/openml-python/actions/runs/23589352348/job/68690383404?pr=1608

JATAYU000:
@geetu040 I have reverted a class label test which I updated due to improper class fields I recieved in my local testing
`tests/test_datasets/test_dataset_functions.py::TestOpenMLDataset::test__retrieve_class_labels`

I received
```python
['nowin', 'won']
None
["f","t"]
```
for the three respective class label assertion in my local, but ci seems to be the previous so I have reverted the change.

geetu040:
> for the three respective class label assertion in my local, but ci seems to be the previous so I have reverted the change.

it's doing fine now, right? I am looking at the changes and they don't seem to be touched now.

JATAYU000:
> it's doing fine now, right? I am looking at the changes and they don't seem to be touched now.

 Its different only in my local, it works in ci , I have reverted it back to old

openml/_api/clients/minio.py
```diff
@@ -16,13 +25,135 @@ class MinIOClient:
...
+    def headers(self) -> dict[str, str]:
+        return openml.config._HEADERS
+
     @property
     def path(self) -> Path:
         return Path(openml.config.get_cache_directory())
+
+    def _get_path(self, url: str) -> Path:
+        parsed_url = urllib.parse.urlparse(url)
+        return self.path / "minio" / parsed_url.path.lstrip("/")
```
geetu040: this should be made public and probably give a better name? and we could directly use this method to find where the path could look for a particular url
use-case is in tests

openml/_api/clients/minio.py
```diff
@@ -16,13 +25,135 @@ class MinIOClient:
...
+
+    def download_minio_file(
+        self,
+        source: str,
+        destination: str | Path | None = None,
+        exists_ok: bool = True,  # noqa: FBT002
+        proxy: str | None = "auto",
+    ) -> Path:
```
geetu040, line 49: I think we can remove these parameters: `destination` and `exists_ok` altogether since the destination is always dynamically created from a given url using `self._get_path`

openml/_api/clients/minio.py
```diff
@@ -16,13 +25,135 @@ class MinIOClient:
...
+        except minio.error.S3Error as e:
+            if e.message is not None and e.message.startswith("Object does not exist"):
+                raise FileNotFoundError(f"Object at '{source}' does not exist.") from e
+            # e.g. permission error, or a bucket does not exist (which is also interpreted as a
+            # permission error on minio level).
+            raise FileNotFoundError("Bucket does not exist or is private.") from e
+
+        return destination
+
+    def download_minio_bucket(self, source: str, destination: str | Path | None = None) -> None:
```
geetu040, line 110: as suggested above
```suggestion
    def download_minio_bucket(self, source: str,) -> None:
```

openml/testing.py
```diff
@@ -53,6 +55,8 @@ class TestBase(unittest.TestCase):
     logger = logging.getLogger("unit_tests_published_entities")
     logger.setLevel(logging.DEBUG)
 
+    http_client: HTTPClient = HTTPClient(api_version=APIVersion.V1)
```
geetu040, line 58: also add minio_client here, we could use the method `minio_client._get_path` where we need to find the path from a given url

tests/conftest.py
```diff
@@ -237,6 +225,30 @@ def _expected_static_cache_state(root_dir: Path) -> list[Path]:
...
+
+        _c_root_dir / "api" / "v1" / "xml" / "data" / "2",
+        _c_root_dir / "api" / "v1" / "xml" / "data" / "2" / "body.xml",
+        _c_root_dir / "api" / "v1" / "xml" / "data" / "qualities" / "2",
+        _c_root_dir / "api" / "v1" / "xml" / "data" / "qualities" / "2" / "body.xml",
+        _c_root_dir / "api" / "v1" / "xml" / "data" / "features" / "2",
+        _c_root_dir / "api" / "v1" / "xml" / "data" / "features" / "2" / "body.xml",
+        _c_root_dir / "data" / "download" / "1666876" / "phpFsFYVN" / "body.arff",
+
+        _c_root_dir / "datasets" / "30" / "dataset_30.pq",
```
geetu040: the expected path for this minio file in `tests/files` can be found using this script:
```py
import openml

openml.config.set_root_cache_directory("tests/files")
openml.config.use_test_servers()

path = openml._backend.minio_client._get_path("http://data.openml.org/dataset30/dataset_30.pq")
```
therefore, rename file and update it's location in `conftest.py`:
```
tests/files/org/openml/test/datasets/30/dataset_30.pq
->
tests/files/org/openml/test/minio/dataset30/dataset_30.pq
```

geetu040 (changes requested):

openml/_api/clients/minio.py
JATAYU000: @geetu040 Would we then remove these tests which checks if minio download to particular path and directory? eg `test__download_minio_file_to_path`

openml/_api/clients/minio.py
JATAYU000: when `download_minio_file` is called `exists_ok` defaults to True,
Only when `download_minio_bucket` calls the download files passes exists_ok as False, if we remove it wouldn't that functionality removed?

geetu040: I didn't see that, in that case, I think it's fine to keep it

openml/_api/clients/minio.py
geetu040: yes, they can be removed

JATAYU000: https://github.com/openml/openml-python/pull/1608#discussion_r3043466557
Since we are not removing it , its fine to keep it right? should that be removed still?

geetu040: we are keeping `exists_ok` but still removing `destination`, isn't that so?

openml/_api/clients/minio.py
```diff
@@ -16,13 +25,139 @@ class MinIOClient:
...
+            else Path(destination)
+        )
+        parsed_url = urllib.parse.urlparse(source)
+        if destination.suffix:
+            destination = destination.parent
+        # expect path format: /BUCKET/path/to/file.ext
+        _, bucket, *prefixes, _ = parsed_url.path.split("/")
+        prefix = "/".join(prefixes)
+
+        client = minio.Minio(endpoint=parsed_url.netloc, secure=False)
```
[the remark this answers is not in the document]
JATAYU000: out of scope for migration

geetu040 (commented):
@JATAYU000 what's the status of this PR? It has been stale for quite a while now. Are you still working on it, planning to continue it, or waiting for a review?

tests/files/org/openml/test/datasets/30/dataset_30.pq
geetu040, line 1: this file has been removed but not replaced by another file, why is that?

JATAYU000: This file is not used by any tests at the moment, There are some minio tests that needs to be updated that might require this file to be moved. will keep this in mind

openml/_config.py
```diff
@@ -495,6 +495,10 @@ def get_cache_directory(self) -> str:
         reversed_url_suffix = os.sep.join(url_parts)  # noqa: PTH118
         return os.path.join(self._root_cache_directory, reversed_url_suffix)  # noqa: PTH118
 
+    def get_minio_download_path(self, url: str) -> str:
+        parsed_url = urlparse(url)
+        return os.path.join(self.get_cache_directory(), "minio", parsed_url.path.lstrip("/"))  #  noqa: PTH118
```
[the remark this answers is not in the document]
JATAYU000: Have updated the new tests,

openml/_api/resources/dataset.py
```diff
@@ -1,11 +1,1556 @@
...
+        """
+        legal_status = {"active", "deactivated"}
+        if status not in legal_status:
+            raise ValueError(f"Illegal status value. Legal values: {legal_status}")
+
+        data: dict[str, str | int] = {"dataset_id": dataset_id, "status": status}
+        result = self._http.post(
+            f"datasets/status/update/?api_key={self._http.api_key}", json=data, use_api_key=False
+        ).json()
+        server_data_id = result["dataset_id"]
```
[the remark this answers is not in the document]
JATAYU000: This is necessary since for v2 endpoints we require to pass api_key and json payload due to how it is handled in the client one of them is discarded hence we agreed on this method

tests/test_tasks/test_task_functions.py
```diff
@@ -164,7 +165,7 @@ def test_get_task(self):
             os.path.join(openml.config.get_cache_directory(), "tasks", "1", "datasplits.arff")
         )
         assert os.path.exists(
-            os.path.join(openml.config.get_cache_directory(), "datasets", "1", "dataset_1.pq")
+            os.path.join(openml.config.get_cache_directory(), openml.config.get_minio_download_path(dataset._parquet_url))
         )
```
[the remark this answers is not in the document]
JATAYU000: fixed

openml/_api/clients/minio.py
```diff
@@ -16,13 +25,129 @@ class MinIOClient:
...
+            if (file_destination.parent / file_destination.stem).exists():
+                # Marker is missing but archive exists means the server archive changed
+                # force a refresh
+                shutil.rmtree(file_destination.parent / file_destination.stem)
+
+            with contextlib.suppress(FileExistsError):
+                self.download_minio_file(
+                    source=source.rsplit("/", 1)[0]
+                    + "/"
+                    + file_object.object_name.rsplit("/", 1)[1],
```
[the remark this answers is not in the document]
JATAYU000: change of existing implementation, out of scope for refactor

openml/_api/resources/dataset.py
```diff
@@ -1,11 +1,1528 @@
...
+            xml_content = response.text
+            description = xmltodict.parse(xml_content)["oml:data_set_description"]
+
+            features_file = None
+            qualities_file = None
+
+            if download_features_meta_data:
+                features_file = self.download_features_file(dataset_id)
+            if download_qualities:
+                qualities_file = self.download_qualities_file(dataset_id)
```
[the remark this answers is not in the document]
JATAYU000: This here implements get_dataset() method from `dataset_functions.py`

openml/_api/clients/minio.py
```diff
@@ -16,13 +25,125 @@ class MinIOClient:
...
+        destination = destination.expanduser()
+        destination.parent.mkdir(parents=True, exist_ok=True)
+
+        if proxy == "auto":
+            resolved_proxies = requests.utils.get_environ_proxies(parsed_url.geturl())
+            proxy = requests.utils.select_proxy(parsed_url.geturl(), resolved_proxies)  # type: ignore
+
+        proxy_client = ProxyManager(proxy) if proxy else None
+
+        client = minio.Minio(endpoint=parsed_url.netloc, secure=False, http_client=proxy_client)
```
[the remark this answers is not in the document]
JATAYU000: update in minio logic is out of scope for this pr

openml/_api/clients/minio.py
```diff
@@ -16,13 +25,125 @@ class MinIOClient:
...
+        """
+        destination = Path(openml.config.get_minio_download_path(source))
+        parsed_url = urllib.parse.urlparse(source)
+        if destination.suffix:
+            destination = destination.parent
+        # expect path format: /BUCKET/path/to/file.ext
+        _, bucket, *prefixes, _ = parsed_url.path.split("/")
+        prefix = "/".join(prefixes)
+
+        client = minio.Minio(endpoint=parsed_url.netloc, secure=False)
```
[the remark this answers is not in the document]
JATAYU000: update in minio logic is out of scope for this pr

openml/_api/clients/minio.py
```diff
@@ -16,13 +25,125 @@ class MinIOClient:
...
+        proxy: str, optional (default = "auto")
+            The proxy server to use. By default it's "auto" which uses ``requests`` to
+            automatically find the proxy to use. Pass None or the environment variable
+            ``no_proxy="*"`` to disable proxies.
+        """
+        destination = Path(openml.config.get_minio_download_path(source))
+        parsed_url = urllib.parse.urlparse(source)
+
+        # expect path format: /BUCKET/path/to/file.ext
+        bucket, object_name = parsed_url.path[1:].split("/", maxsplit=1)
```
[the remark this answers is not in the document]
JATAYU000: update in minio logic is out of scope for this pr

openml/_api/clients/http.py
```diff
@@ -98,16 +99,32 @@ def _get_body_filename_from_response(self, response: Response) -> str:
...
             return "body.xml"
 
+        if response.content.startswith(b"PK\x03\x04"):
+            return "body.zip"
+
+        try:
+            arff.loads(response.text)
+            return "body.arff"
+        except arff.ArffException:
+            pass
```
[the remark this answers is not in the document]
JATAYU000: @geetu040  What is your opinion on this suggestions?

tests/test_api/test_datasets.py
```diff
@@ -0,0 +1,322 @@
...
+
+@pytest.fixture
+def dataset_v2(http_client_v2, minio_client) -> DatasetV2API:
+    return DatasetV2API(http=http_client_v2, minio=minio_client)
+
+
+def _wait_for_dataset_being_processed(dataset, did, status='active', n_tries=10, wait_time=10):
+    for attempt in range(n_tries):
+        try:
+            result = dataset.list(limit=1, offset=0, data_id=[did], status="all")
```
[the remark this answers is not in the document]
JATAYU000: The default is "active"

openml/_api/clients/http.py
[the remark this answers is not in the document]
geetu040: yes, this has been addressed in https://github.com/openml/openml-python/pull/1616, which should get merged first

openml/_api/resources/dataset.py
```diff
@@ -1,11 +1,1582 @@
...
+logger = logging.getLogger(__name__)
+
+
+NO_ACCESS_GRANTED_ERRCODE = 112
+
 
 class DatasetV1API(ResourceV1API, DatasetAPI):
     """Version 1 API implementation for dataset resources."""
 
+    @openml.utils.thread_safe_if_oslo_installed
```
geetu040: I think this is redundant, we already apply lock on `get_dataset` in `datasets/functions.py`

openml/_api/resources/dataset.py
```diff
@@ -1,11 +1,1582 @@
...
+            xmltodict.parse(dataset_xml)["oml:data_set_description"]
+        )
+        with arff_file.open("r", encoding="utf8") as f:
+            return f.read()
+
 
 class DatasetV2API(ResourceV2API, DatasetAPI):
     """Version 2 API implementation for dataset resources."""
+
+    @openml.utils.thread_safe_if_oslo_installed
```
geetu040: same as above

openml/datasets/functions.py
```diff
@@ -1076,11 +879,8 @@ def _topic_delete_dataset(data_id: int, topic: str) -> int:
...
-    form_data = {"data_id": data_id, "topic": topic}  # type: openml._api_calls.DATA_TYPE
-    result_xml = openml._api_calls._perform_api_call("data/topicdelete", "post", data=form_data)
-    result = xmltodict.parse(result_xml)
-    data_id = result["oml:data_topic"]["oml:id"]
-    return int(data_id)
+
+    return openml._backend.dataset.delete_topic(data_id, topic)
 
 
 def _get_dataset_description(did_cache_dir: Path, dataset_id: int) -> dict[str, Any]:
```
geetu040: this function still uses old api calls, though it's never really used in the code base, except for the tests. I'd suggest to remove this entirely from everyhwere

openml/utils/_openml.py
```diff
@@ -454,6 +455,24 @@ def _create_lockfiles_dir() -> Path:
     return path
 
 
+@contextlib.contextmanager
+def file_lock(lock_path: str) -> Iterator[None]:
```
geetu040, line 459: 1. Did you have to add this because of the failing tests in CI?
2. We already have a lock in the codebase: `thread_safe_if_oslo_installed`, this is different because it applies lock at file levels instead of functions
3. Do you think this `file_lock` could be avoided inside functions and replaced by `thread_safe_if_oslo_installed` at some function level.
4. If we are keeping it, can we align this function more with `thread_safe_if_oslo_installed`, which would mean giving a similar name and if possible keeping `file_lock` (resource/file-scoped) and `thread_safe_if_oslo_installed` (function/id-scoped) as separate wrappers over one shared external-lock primitive.

tests/test_api/test_datasets.py
```diff
@@ -0,0 +1,322 @@
...
+
+    assert isinstance(output, dict)
+    assert len(output.keys()) == 37
+
+@pytest.mark.test_server()
+def test_v1_get_qualities(dataset_v1):
+    output = dataset_v1.get_qualities(2)
+
+    assert isinstance(output, dict)
+    assert len(output.keys()) == 107
```
geetu040: quite hardcoded, should be avoided, may be `>= 100` would be my suggestion, similarly in other places for other v1/v2 and get/list_quantiles.

tests/test_api/test_datasets.py
```diff
@@ -0,0 +1,322 @@
...
+@pytest.mark.test_server()
+def test_v1_feature_add_remove_ontology(dataset_v1):
+    did = 11
+    fid = 0
+    ontology = "https://www.openml.org/unittest/" + str(time.time())
+    output = dataset_v1.feature_add_ontology(did, fid, ontology)
+    assert output
+
+    output = dataset_v1.feature_remove_ontology(did, fid, ontology)
+    assert output
```
geetu040, line 172: doesn't really check if the functions actually work, can you check the output or the ontologies if they have been updated after each call.

tests/test_api/test_datasets.py
```diff
@@ -0,0 +1,322 @@
...
+@pytest.mark.skipif(
+    not os.environ.get(openml.config.OPENML_TEST_SERVER_ADMIN_KEY_ENV_VAR),
+    reason="Test requires admin key. Set OPENML_TEST_SERVER_ADMIN_KEY environment variable.",
+)
+@pytest.mark.test_server()
+def test_v1_add_delete_topic(dataset_v1):
+    openml.config.apikey = TestBase.admin_key
+    topic = f"test_topic_{str(time.time())}"
+    dataset_v1.add_topic(31, topic)
+    dataset_v1.delete_topic(31, topic)
```
geetu040, line 184: same as above

tests/test_api/test_datasets.py
```diff
@@ -0,0 +1,322 @@
...
+
+@pytest.mark.test_server()
+def test_v2_feature_add_remove_ontology(dataset_v2):
+    with pytest.raises(OpenMLNotSupportedError):
+        dataset_v2.feature_add_ontology(2, 0, "https://www.openml.org/unittest/" + str(time.time()))
+
+@pytest.mark.test_server()
+def test_v2_add_delete_topic(dataset_v2):
+    with pytest.raises(OpenMLNotSupportedError):
+        dataset_v2.add_topic(2, 'test_topic_' + str(time.time()))
```
geetu040, line 258: probably update for delete as well

geetu040 (changes requested):
Really nice, we are close to finish this, just left a few minor comments.

PR #1611, opened 2026-01-09
[ENH] V1 → V2 API Migration - Tasks

### LINKED ISSUE #1621: [ENH] V1 → V2 API Migration - tasks
V1 → V2 API Migration - tasks. Sub-issue of #1575, for tracking.

satvshr:
#### Metadata
* Reference Issue: stacks on #1576, fixes #1621
* New Tests Added: No
* Documentation Updated: No


examples/Advanced/fetch_evaluations_tutorial.py
```diff
@@ -178,4 +178,4 @@ def plot_flow_compare(evaluations, top_n=10, metric="predictive_accuracy"):
     function="predictive_accuracy", flows=[6767], size=100, parameters_in_separate_columns=True
 )
 
-print(evals_setups.head(10))
\ No newline at end of file
+print(evals_setups.head(10))
```
geetu040, line 181: Let's keep these changes away from this PR. If there are some ruff errors in the existing code, they should be fixed in another PR which will probably get merged soon.

geetu040 (changes requested):
From a high-level review, I noticed a few points that need adjustment:
- Caching can likely be removed from the SDK, since these concerns should be handled by the base client.
- I don't see the `api_context` being used in `tasks/functions`, so it's not clear to me how the SDK is actually using the new API interface here.
- Instead of moving entire methods out of `tasks/functions.py`, it would be better to stick to the goal of minimal SDK changes while enabling v2 support.
- API calls should be updated at the specific root functions (for example `_get_task_description`, `OpenMLTask._download_split`).
- For listing tasks, please follow the approach discussed in [#1575 comment](https://github.com/openml/openml-python/issues/1575#issuecomment-3744991650).

examples/Advanced/fetch_evaluations_tutorial.py
satvshr: Accidentally had ran `ruff format .` on this branch, ruff PR getting merged solved these issues automatically though.

openml/_api/resources/base.py
```diff
@@ -0,0 +1,169 @@
...
+    def __init__(self, http: HTTPClient):
+        self._http = http
+
+
+class DatasetsAPI(ResourceAPI, ABC):
+    @abstractmethod
+    def get(self, dataset_id: int) -> OpenMLDataset | tuple[OpenMLDataset, Response]: ...
+
+
+class TasksAPI(ResourceAPI, ABC):
```
geetu040: why are the methods commented out?

openml/_api/resources/base.py
```diff
@@ -0,0 +1,169 @@
...
+    # # Task listing (V1 only)
+    # @abstractmethod
+    # def list_tasks(
+    #     self,
+    #     *,
+    #     task_type: TaskType | None = None,
+    #     offset: int | None = None,
+    #     size: int | None = None,
+    #     **filters: Any,
+    # ):
```
geetu040: this method should simply be called `list`

openml/_api/resources/tasks.py
```diff
@@ -0,0 +1,606 @@
...
+    OpenMLSupervisedTask,
+    OpenMLTask,
+    TaskType,
+)
+
+TASKS_CACHE_DIR_NAME = "tasks"
+
+
+class TasksV1(TasksAPI):
+    @openml.utils.thread_safe_if_oslo_installed
```
geetu040: you can remove this, it's not needed, it's related to cache and should be handled at client

openml/_api/resources/tasks.py
```diff
@@ -0,0 +1,606 @@
...
+
+    def _get_task_description(self, task_id: int) -> OpenMLTask:
+        result = self._http.get(f"task/{task_id}", return_response=True)
+
+        if isinstance(result, tuple):
+            task, _response = result
+        else:
+            task = result
+
+        return task
```
geetu040: you should not copy this entirely from `tasks/functions.py`, only the specific part which loads the task object should be here, which would probably be
```
       response = self._http.get(f"task/{task_id}")
       task = self._create_task_from_xml(response.text)
       return task
```

openml/_api/resources/tasks.py
```diff
@@ -0,0 +1,606 @@
...
+        if kwargs is not None:
+            for operator, value in kwargs.items():
+                if value is not None:
+                    if operator == "task_id":
+                        value = ",".join([str(int(i)) for i in value])  # noqa: PLW2901
+                    api_call += f"/{operator}/{value}"
+
+        return self.__list_tasks(api_call=api_call)
+
+    def __list_tasks(self, api_call: str) -> pd.DataFrame:  # noqa: C901, PLR0912
```
geetu040: maybe use better helper functions like `_create_list_url` and `_parse_list_response`?

openml/_api/resources/tasks.py
```diff
@@ -0,0 +1,606 @@
...
+                        stacklevel=2,
+                    )
+                else:
+                    warnings.warn(
+                        f"Could not find key {e} in {task_}!", RuntimeWarning, stacklevel=2
+                    )
+
+        return pd.DataFrame.from_dict(tasks, orient="index")
+
+    def _get_estimation_procedure_list(self) -> list[dict[str, Any]]:
```
geetu040: no need for this, see the other comment on `_get_estimation_procedure_list`

openml/_api/resources/tasks.py
```diff
@@ -0,0 +1,606 @@
...
+                )
+
+        return procs
+
+    def get_tasks(
+        self,
+        task_ids: list[int],
+        download_data: bool | None = None,
+        download_qualities: bool | None = None,
+    ) -> list[OpenMLTask]:
```
geetu040: keep this method in `tasks/functions.py`, because we are sticking to the rule "minimal sdk changes for v1/v2 compatibility"

openml/_api/resources/tasks.py
```diff
@@ -0,0 +1,606 @@
...
+
+    def create_task(
+        self,
+        task_type: TaskType,
+        dataset_id: int,
+        estimation_procedure_id: int,
+        target_name: str | None = None,
+        evaluation_measure: str | None = None,
+        **kwargs: Any,
+    ) -> (
```
geetu040: this should stay in `tasks/fucntions.py`

openml/_api/resources/tasks.py
```diff
@@ -0,0 +1,606 @@
...
+        ----------
+        task_id : int
+            OpenML id of the task
+
+        Returns
+        -------
+        bool
+            True if the deletion was successful. False otherwise.
+        """
+        return openml.utils._delete_entity("task", task_id)
```
geetu040: I'll implement this in the base class that you can replace with later

openml/_api/resources/tasks.py
```diff
@@ -0,0 +1,606 @@
...
+        cls = {
+            TaskType.SUPERVISED_CLASSIFICATION: OpenMLClassificationTask,
+            TaskType.SUPERVISED_REGRESSION: OpenMLRegressionTask,
+            TaskType.CLUSTERING: OpenMLClusteringTask,
+            TaskType.LEARNING_CURVE: OpenMLLearningCurveTask,
+        }[task_type_id]
+
+        return cls(**common_kwargs)
+
+    def list_task_types(self) -> list[dict[str, str | int | None]]:
```
geetu040: is this used anywhere?

openml/tasks/functions.py
```diff
@@ -1,129 +1,23 @@
...
-    task_xml_path = tid_cache_dir / "task.xml"
-    try:
-        with task_xml_path.open(encoding="utf8") as fh:
-            return _create_task_from_xml(fh.read())
-    except OSError as e:
-        openml.utils._remove_cache_dir_for_id(TASKS_CACHE_DIR_NAME, tid_cache_dir)
-        raise OpenMLCacheException(f"Task file for tid {tid} not cached") from e
-
-
-def _get_estimation_procedure_list() -> list[dict[str, Any]]:
```
geetu040, line 73: keep this method here and inside try to use the method `list_estimation_procedures` already implemented in `evaluations/functions.py`

geetu040 (changes requested):
I have left some comments, please take a look and make sure the signature of all methods in `TasksAPI`, `TasksV1` and `TasksV2` stay same.

openml/_api/resources/base.py
satvshr: I was going to remove them, if I add abstract methods they have to be for shared functions right? The only shared function right now is `get`.

openml/_api/resources/base.py
satvshr: Replaced the name in `TasksV1` (where the function actually exists)

openml/_api/resources/tasks.py
satvshr: I am confused as to what youre trying to say here, do you mean I should transfer the functionalities of `list` (previously `list_tasks`) to `_create_list_url`, and rename `__list_tasks` to `_parse_list_response`?

openml/_api/resources/tasks.py
satvshr: Did you mean to highlight the entire `get` function or only `_get_task_description`? Is this:
```
        dataset = get_dataset(task.dataset_id, **get_dataset_kwargs)
        # List of class labels available in dataset description
        # Including class labels as part of task meta data handles
        #   the case where data download was initially disabled
            task.class_labels = dataset.retrieve_class_labels(task.target_name)
        # Clustering tasks do not have class labels
        # and do not offer download_split
        if download_splits and isinstance(task, OpenMLSupervisedTask):
            task.download_split()
```
not useful? Why?

openml/_api/resources/tasks.py
satvshr: `_get_estimation_procedure_list` is being called inside `__list_tasks`, what to do about that? Call `_get_estimation_procedure_list` inside `__list_tasks` by importing it from `tasks/functions.py`? But that does not make sense, so may I know what you are thinking?

openml/_api/resources/tasks.py
satvshr: Shouldn't I do the same for `create_task` and `delete_task` too?

Edit: Just saw your comment below :D

openml/_api/resources/tasks.py
satvshr: Nope, there is an endpoint for it though, same for `get_task_type`.

openml/_api/resources/tasks.py
satvshr: As discussed today during the standup, makes sense

openml/tasks/functions.py
satvshr: `list_estimation_procedures` returns only the `"oml:name"` whereas `_get_estimation_procedure_list` requires more items, they make call the same API, and `list_estimation_procedures` may be somewhat of a subset of `_get_estimation_procedure_list`, but that does not mean it can be used inside `_get_estimation_procedure_list`

openml/_api/resources/base.py
geetu040: > if I add abstract methods they have to be for shared functions right?

1. they create blueprint of this resource, so one can look at the resource class to see which are the public methods and what do their inputs and outputs look like
2. these methods are expected to be implemented in all the child classes, so yes they are used for shared functions

> The only shared function right now is `get`.

`list`, `delete`, ...?

openml/_api/resources/base.py
geetu040: signatures should be same for all 3 classes as of now: https://github.com/openml/openml-python/pull/1611#pullrequestreview-3666316645

openml/_api/resources/tasks.py
geetu040: > Did you mean to highlight the entire `get` function or only `_get_task_description`? Is this:

both

> not useful? Why?

what should I look here? this is dataset related code.

openml/_api/resources/tasks.py
geetu040: this works fine but `__list_tasks` is not a good name for a helper function in this class.
I'd suggest you take a look at datasets PR, it does something similar: https://github.com/openml/openml-python/pull/1608

openml/_api/resources/tasks.py
satvshr: > what should I look here? this is dataset related code.

It is assigning attributes to the task object, don't you think that's useful?

openml/_api/resources/base.py
satvshr: > `list`, `delete`, ...?

Not there for `v2`

openml/_api/resources/tasks.py
geetu040: why it doesn't make sense? estimation-procedure is a separate resource, it should not be coupled with the tasks, also listing this is already a duplicate code seen in both `tasks/functions.py` and `evaluations/functions.py`. Ideally I would prefer that you import this from `evaluations/functions.py` and make it's output work in this code.

openml/_api/resources/tasks.py
geetu040: I'd say just remove it, since it suits a `TaskType` resource, though it's not needed anywhere now

openml/_api/resources/base.py
satvshr: As I said above, in terms of functionality only `get` matches, else i'd have done that.

openml/_api/resources/tasks.py
satvshr: Oh ok only a rename? Will do.

openml/_api/resources/tasks.py
satvshr: I added `tasktype` as part of this PR too, there is only one endpointt for `Tasks` and 2 for `TaskType`, so I just added it into `Tasks` given theyre clubbed together in the docs for endpoints too

openml/_api/resources/tasks.py
satvshr: Discussion solved [below](https://github.com/openml/openml-python/pull/1611#discussion_r2695440254)

geetu040: just saw this comment: https://github.com/openml/openml-python/pull/1611#discussion_r2695440254

if it can't be used from `evaluations/functions.py`, then my suggestion is
1. remove `procs = self._get_estimation_procedure_list()` entirely from `TasksV1`
2. rather add this line inside `tasks/functions.py::list_tasks`
3. pass the `procs` are an argument to `TasksV1.list` through `partial`
4. in `TaskV1` just use this from the newly added parameter

openml/_api/resources/tasks.py
geetu040: ok I see, you are asking if this should live in the sdk or the resource class? if it can stay out of the resource class then it should

openml/_api/resources/base.py
geetu040: still the base class should have these, in the v2 class just raise an exception or maybe skip it and the exception will be raised automatically

openml/_api/resources/base.py
geetu040: I still don't understand

openml/_api/resources/tasks.py
geetu040: > theyre clubbed together in the docs for endpoints too

in v2 if I remember correctly, this is not the case. anyways I still think they should be removed as they are not being used anywhere in the sdk

openml/_api/resources/tasks.py
satvshr: > 3\. pass the `procs` are an argument to `TasksV1.list` through `partial`

To do that it will have to go through 2 functions before call, `list` -> `_fetch_tasks_df` -> `_get_estimation_procedure_list`. If you want to keep code in the sdk and not in resources, I should move `_fetch_tasks_df` to `functions.py` instead.

openml/_api/resources/tasks.py
satvshr: I was talking about v2 only, `tasks` and `tasktype` share the same header in the v2 docs page. `tasktype` endings are not being used anywhere so should `tasktypes` just not exist? Want me to remove it from this PR and park it into a draft PR if it is ever used (which will probably get lost over  time), or should we let the endpoints and functions to de-serialize those endpoints exist for no reason?

satvshr:
@geetu040 I will explain the confusion I am facing to the best of my abilities, and I do feel communicating on the PR threads is yielding no results, hence I will just put everything here:
1. Over [here](https://github.com/openml/openml-python/pull/1611#discussion_r2698007303) you say the base class should have v1 functions like list, delete, and create. This means to me that all V1 related functions should be moved from `functions.py` to `TasksV1` (**from SDK to resources**) and calls inside `functions.py` should be replaced with calls to `api_context.backend.task.method`. Over [here](https://github.com/openml/openml-python/pull/1611#discussion_r2697263064), you say something similar, stating that TasksV1, V2, and TasksAPI should have the same function signatures.
2. Over [here](https://github.com/openml/openml-python/pull/1611#discussion_r2698001272)  and [here](https://github.com/openml/openml-python/pull/1611#discussion_r2699682711) you mention that we should try to **keep most of the code in SDK and not resources** (contradictory to point 1).
Completely contrary to point 1 you mention `get_tasks` along with `create_task` and `delete_task` should stay in sdk over [here](https://github.com/openml/openml-python/pull/1611#discussion_r2694949307).

I seem to be getting 2 contradictory messages from your end which is where the confusion arises.

openml/_api/resources/tasks.py
geetu040: I think you should just completely remove them and I won't recommend to create a separate draft PR for now

openml/_api/resources/tasks.py
geetu040: keep `_fetch_tasks_df` in the resource class, just call `_get_estimation_procedure_list` in `list_tasks`, then pass it as a parameter to `TasksV1.list` through `partial` and then as a parameter to `_fetch_tasks_df`

geetu040:
> I seem to be getting 2 contradictory messages from your end which is where the confusion arises.

I don't feel confused with the picture I have in my mind, I think I need to explain it even better. I'll try to document each method in detail what should be moved and what should stay.

openml/_api/resources/tasks.py
satvshr: > just call `_get_estimation_procedure_list` in `list_tasks`

I will do this given that is what you want (and for this not to serve as a blocker) but I do not agree with this as you will put in a purely V1 function into `tasks/functions.py`, which is supposed to be a module for both V1 and V2.

satvshr:
Kindly resolve the comments which have been solved, it would be easier to navigate!

geetu040:
> Kindly resolve the comments which have been solved, it would be easier to navigate!

It doesn't look like I have the typical "Resolve" button.

Edit: It's visible in the "Files changed" section, I'll resolve them there.

openml/_api/resources/tasks.py
geetu040: yes you are right about that, this is a v1 specific code, but we can remove this later with `api_context.backend.estimation_procedures.list`
and I am not totally against your preference of keeping it in the `TasksV1` but in the case either we'll have to decouple this and apply the above solution eventually, but until we don't have estimation-procedures merged and it cannot be used via `api_context` we can use this solution of yours

openml/tasks/task.py
```diff
@@ -150,7 +150,7 @@ def get_dataset(self, **kwargs: Any) -> datasets.OpenMLDataset:
 
         Accepts the same keyword arguments as the `openml.datasets.get_dataset`.
         """
-        return datasets.get_dataset(self.dataset_id, **kwargs)
+        return datasets.get_dataset(self.dataset_id, **kwargs)  # Shrivaths
```
geetu040: can you undo this?

openml/tasks/task.py
geetu040, line 1: we have an api call in `_download_split`, what do you think about that? also does it relate with the datasets PR?

openml/_api/resources/tasks.py
geetu040, line 1: this looks really good and clean now, just one open thread which needs discussion now: https://github.com/openml/openml-python/pull/1611#discussion_r2707210799

openml/_api/resources/base.py
geetu040, line 1: please see the previous comments on `TasksAPI`

geetu040 (changes requested):
Really nice, the code is in a better shape now. Many lines have been removed in the functions.py which seem fine but I'll take a deeper look. In other comments mostly some cleaning is required.

openml/tasks/task.py
satvshr: Yeah, I was going to do that before it was going to be in a mergeable state.

openml/tasks/task.py
satvshr: > we have an api call in `_download_split`, what do you think about that?

I remember you and Shrivaths discussing cache paths for V1 and V2 and I think the understanding was that the paths would differ, if that is the case, we should change this later on and leave it as it is right now.

> also does it relate with the datasets PR?

Only the `get_dataset` part, hence the Shrivaths comment

openml/_api/resources/tasks.py
satvshr: If you agreed with my approach, can we not close that too?

openml/_api/resources/base.py
satvshr: Solved with the latest commit

openml/_api/resources/tasks.py
```diff
@@ -0,0 +1,384 @@
...
+        return cls(**common_kwargs)  # type: ignore
+
+    def list(
+        self,
+        limit: int,
+        offset: int,
+        task_type: TaskType | int | None = None,
+        **kwargs: Any,
+    ) -> pd.DataFrame:
+        raise NotImplementedError("Task listing is not available in API v2 yet.")
```
geetu040: see point 8 in https://github.com/openml/openml-python/issues/1575#issuecomment-3816198783

openml/tasks/functions.py
```diff
@@ -385,7 +141,6 @@ def get_tasks(
     return tasks
 
 
-@openml.utils.thread_safe_if_oslo_installed
```
geetu040, line 350: see point 3 in https://github.com/openml/openml-python/issues/1575#issuecomment-3816198783

openml/tasks/task.py
```diff
@@ -46,7 +46,7 @@ class _EstimationProcedure(TypedDict):
     data_splits_url: str | None
 
 
-class OpenMLTask(OpenMLBase):
+class OpenMLTask(OpenMLBase, ResourceAPI):
```
geetu040: should not inherit `ResourceAPI`

openml/tasks/task.py
```diff
@@ -172,10 +172,7 @@ def _download_split(self, cache_file: Path) -> None:
                 pass
         except OSError:
             split_url = self.estimation_procedure["data_splits_url"]
-            openml._api_calls._download_text_file(
-                source=str(split_url),
-                output_path=str(cache_file),
-            )
+            self._http.download(url=str(split_url), file_name="datasplits.arff")
```
geetu040: use `api_context.backend.task.[something]` here

openml/tasks/functions.py
```diff
@@ -589,13 +242,7 @@ def create_task(
...
         task_cls = OpenMLRegressionTask  # type: ignore
     else:
-        raise NotImplementedError(
-            f"Task type ID {task_type:d} is not supported. "
-            f"Supported task type IDs: {TaskType.SUPERVISED_CLASSIFICATION.value},"
-            f"{TaskType.SUPERVISED_REGRESSION.value}, "
-            f"{TaskType.CLUSTERING.value}, {TaskType.LEARNING_CURVE.value}. "
-            f"Please refer to the TaskType enum for valid task type identifiers."
-        )
+        raise NotImplementedError(f"Task type {task_type:d} not supported.")
```
geetu040: let's keep it as it was

geetu040 (changes requested):
sdk code look good so far, please take a look at https://github.com/openml/openml-python/issues/1575#issuecomment-3816198783 and make changes accordingly where needed.

openml/tasks/task.py
satvshr: I want to use the `download` method from the `HTTPClient` and not a function from tasks though?

satvshr: How am I supposed to do that? [`download` is a method in the HTTPClient.](https://github.com/JATAYU000/openml-python/blob/5c8791ac68cc613732bd64608c181bc256ab379e/openml/_api/clients/http.py#L452)

geetu040: Still this should be under a method now called `openml._backend.task.[something]` and then we can take this problem to `TaskV1.[something]`.

I see this function depends on `download` of `HTTPClient`, I'll either add that method in the core PR or we can plan to merge datasets PR before this.

satvshr: Will be moving this to `TaskAPI` after discussion.

openml/_api/resources/task.py
```diff
@@ -0,0 +1,383 @@
...
+                        stacklevel=2,
+                    )
+                else:
+                    warnings.warn(
+                        f"Could not find key {e} in {task_}!", RuntimeWarning, stacklevel=2
+                    )
+
+        return pd.DataFrame.from_dict(tasks, orient="index")
+
+    def _get_estimation_procedure_list(self) -> builtins.list[dict[str, Any]]:
```
EmanAbdelhaleem: @geetu040  We already have implemented this function in the `estimation_procedure` resource, so I think we don't need it to be here, right?

openml/_api/resources/task.py
```diff
@@ -0,0 +1,383 @@
...
+        return cls(**common_kwargs)  # type: ignore
+
+    def list(
+        self,
+        limit: int,
+        offset: int,
+        task_type: TaskType | int | None = None,
+        **kwargs: Any,
+    ) -> pd.DataFrame:
+        raise NotImplementedError(self._not_supported(method="list"))
```
EmanAbdelhaleem: You can just use `self._not_supported(method="list")`

no need for `raise NotImplementedError()`

tests/test_api/test_task.py
```diff
@@ -0,0 +1,61 @@
...
+        assert "tid" in tasks_df.columns
+
+    # @pytest.mark.uses_test_server()
+    def test_estimation_procedure_list(self):
+        """Verify that estimation procedure list endpoint works."""
+        procs = self.resource._get_estimation_procedure_list()
+        assert isinstance(procs, list)
+        assert len(procs) > 0
+        assert "id" in procs[0]
+
```
EmanAbdelhaleem: I think you need to add `TestTasksV2` class, and also add v1 and v2 matching tests and fallback tests as mentioned here: https://github.com/openml/openml-python/issues/1575#issuecomment-3810582840

openml/_api/resources/task.py
```diff
@@ -0,0 +1,383 @@
...
+        return cls(**common_kwargs)  # type: ignore
+
+    def list(
+        self,
+        limit: int,
+        offset: int,
+        task_type: TaskType | int | None = None,
+        **kwargs: Any,
+    ) -> pd.DataFrame:
+        """
```
EmanAbdelhaleem, line 157: @geetu040 
Should we all use the same style or not?
for me, I have splitted this this into 3 functions for more readability and less coupling
- list()
- _build_url()
- _parse_list_xml()

tests/test_api/test_task.py
satvshr: I dont need to add `TestTasksV2` given those tests are a subset of `TestTasksV1`. Matching tests are in the shared class.

openml/_api/resources/task.py
satvshr: I don't think we need private functions to match each other in terms of nomenclature or functionality as long as we share the same common function names (which we do).

openml/_api/resources/task.py
satvshr: Just checked, once we merge your PR we can replace it with the `_get_details` function yes.

openml/_api/resources/task.py
geetu040: same nomenclature is preferred
I would say you should do this, will make it better anyways

openml/_api/resources/task.py
geetu040: yes, that's the plan

tests/test_api/test_task.py
geetu040: you still need `TestTasksV2` and check if they raise the not implemented error
i think you should use this design then, since there are many common utilities
[tests/test_api/test_versions.py](https://github.com/geetu040/openml-python/blob/migration/tests/test_api/test_versions.py)

geetu040 (changes requested):
- rename `tasks` to `task`, in `TestTasksV1`, `test_tasks.py just for consistency
- and other comments from @EmanAbdelhaleem's review

openml/_api/resources/task.py
satvshr: Will do, but I don't necessarily agree. Something similar happened in another project where we were working on 2 algorithms and I suggested something similar but Franz mentioned it is not something that can consistently be maintained across the entire codebase hence does not need change. Given we have 8-9 resources it does not feel fruitful to have people share the same nomenclature for private functions across all PRs, I don't have a strong opinion on this though and I will be making the requested change.

geetu040: yeah if your function is lengthy then it's always better to break it into components, otherwise agreed this is not necessary

satvshr: Oh yeah I agree if the function is lengthy, over here (eman's suggestion) we are just removing 15 lines of code and replacing with 1, hence the hesitance. Anyways, I did change it so the pr is ready for review again :D

tests/test_tasks/test_task_functions.py
```diff
@@ -151,20 +113,6 @@ def test__get_task_live(self):
         # https://github.com/openml/openml-python/issues/378
         openml.tasks.get_task(34536)
 
-    @pytest.mark.uses_test_server()
-    def test_get_task(self):
-        task = openml.tasks.get_task(1, download_data=True)  # anneal; crossvalidation
-        assert isinstance(task, OpenMLTask)
```
geetu040, line 159: I would say you keep this test and only remove the lines that check for cache, basically remove all the lines after these 2

Since all tests that were testing get function used cache so all were removed and there was no unit-test left for `get_task` method, therefore, keep this test as said above

geetu040 (changes requested):
\

geetu040 (changes requested):
I have updated the `HTTPClient` with `download` method, please sync with the base branch, and use this method where needed.

tests/test_api/test_task.py
```diff
@@ -0,0 +1,74 @@
...
+import pytest
+import pandas as pd
+from openml._api.resources.task import TaskV1API, TaskV2API
+from openml._api.resources.base.fallback import FallbackProxy
+from openml.exceptions import OpenMLNotSupportedError
+from openml.testing import TestAPIBase
+from openml.enums import APIVersion
+from openml.tasks.task import TaskType
+
+
```
Omswastik-11, line 12: I think here we need to implement `TestResourceAPIBase` and move all common tests there but here it seems not much common tests . may be @geetu040 any thoughts ?

https://discord.com/channels/1439727064231182487/1468994885297246399/1469408217082888369

geetu040 (changes requested):

satvshr:
Below errors, caused by the lack of a v2 API endpoint, I assume:
```
FAILED tests/test_api/test_task.py::test_fallback_get_matches - AttributeError: 'str' object has no attribute 'value'
FAILED tests/test_api/test_task.py::test_fallback_get - AttributeError: 'str' object has no attribute 'value'
```

Following errors caused by the `tests/test_api/test_versions.py` file which I think we will be removing?
```
ERROR tests/test_api/test_versions.py::test_v1_publish - TypeError: Can't instantiate abstract class DummyTaskV1API without an implementation for abstract methods 'get', 'list'
ERROR tests/test_api/test_versions.py::test_v1_delete - TypeError: Can't instantiate abstract class DummyTaskV1API without an implementation for abstract methods 'get', 'list'
ERROR tests/test_api/test_versions.py::test_v2_publish - TypeError: Can't instantiate abstract class DummyTaskV2API without an implementation for abstract methods 'get', 'list'
ERROR tests/test_api/test_versions.py::test_v1_tag - TypeError: Can't instantiate abstract class DummyTaskV1API without an implementation for abstract methods 'get', 'list'
ERROR tests/test_api/test_versions.py::test_v2_delete - TypeError: Can't instantiate abstract class DummyTaskV2API without an implementation for abstract methods 'get', 'list'
ERROR tests/test_api/test_versions.py::test_v1_untag - TypeError: Can't instantiate abstract class DummyTaskV1API without an implementation for abstract methods 'get', 'list'
ERROR tests/test_api/test_versions.py::test_v2_tag - TypeError: Can't instantiate abstract class DummyTaskV2API without an implementation for abstract methods 'get', 'list'
ERROR tests/test_api/test_versions.py::test_v2_untag - TypeError: Can't instantiate abstract class DummyTaskV2API without an implementation for abstract methods 'get', 'list'
ERROR tests/test_api/test_versions.py::test_fallback_publish - TypeError: Can't instantiate abstract class DummyTaskV1API without an implementation for abstract methods 'get', 'list'
ERROR tests/test_api/test_versions.py::test_fallback_delete - TypeError: Can't instantiate abstract class DummyTaskV1API without an implementation for abstract methods 'get', 'list'
ERROR tests/test_api/test_versions.py::test_fallback_tag - TypeError: Can't instantiate abstract class DummyTaskV1API without an implementation for abstract methods 'get', 'list'
ERROR tests/test_api/test_versions.py::test_fallback_untag - TypeError: Can't instantiate abstract class DummyTaskV1API without an implementation for abstract methods 'get', 'list'
```

geetu040:
These are now fixed in the base PR.

tests/test_api/test_task.py
geetu040, line 1: no need for fallback tests

geetu040 (changes requested):

geetu040 (changes requested):

geetu040 (commented):
the tests are failing because the cache path has changed
```
old: tests/files/org/openml/test/tasks/1882
new: tests/files/org/openml/test/api/v1/xml/task/1882
```
Let's discuss this on slack and see how we proceed with this change.

geetu040 (commented):
Please see https://github.com/openml/openml-python/pull/1619#pullrequestreview-3996391505 regarding changed cache path

geetu040 (changes requested):
The base PR is merged now, please sync with main.

geetu040:
@satvshr the tests are failing because the cache is not in the expected format, see the first point in https://github.com/openml/openml-python/pull/1619#pullrequestreview-3996391505 and how it's done in [#1619](https://github.com/openml/openml-python/pull/1619/changes#diff-8d30a21a57cb6230b16ca86586024904aaee21f2aeb00f33a28b8f1239eb250e)
so you need `meta.json` and `headers.json` which you can simply copy from the other pr and rename `task.xml` to `body.bin`

openml/_api/resources/base/resources.py
```diff
@@ -22,6 +28,60 @@ class TaskAPI(ResourceAPI):
...
+        ...
+
+    def download(
+        self,
+        url: str,
+        handler: Callable[[Response, Path, str], None] | None = None,
+        encoding: str = "utf-8",
+        file_name: str = "response.txt",
+        md5_checksum: str | None = None,
+    ) -> Path:
```
geetu040: where is this used really, I cannot find?

openml/_api/resources/task.py
```diff
@@ -1,11 +1,401 @@
...
+        -------
+        task: OpenMLTask
+        """
+        if not isinstance(task_id, int):
+            raise TypeError(f"Task id should be integer, is {type(task_id)}")
+
+        response = self._http.get(f"task/{task_id}", enable_cache=True)
+        return self._create_task_from_xml(response.text)
+
+    def _create_task_from_xml(self, xml: str) -> OpenMLTask:
```
geetu040: this method doesn't use `self`, make this a static method, apply this for other private methods as well.

openml/_api/resources/task.py
```diff
@@ -1,11 +1,401 @@
...
+                        stacklevel=2,
+                    )
+                else:
+                    warnings.warn(
+                        f"Could not find key {e} in {task_}!", RuntimeWarning, stacklevel=2
+                    )
+
+        return pd.DataFrame.from_dict(tasks, orient="index")
+
+    def _get_estimation_procedure_list(self) -> builtins.list[dict[str, Any]]:
```
geetu040: this should stay out of `_api/`

openml/tasks/functions.py
geetu040, line 1: though there are lot's of changes in this file but they are all clean, nicely done.

I see you are importing `TaskV1API` and `TaskV2API`, which is completely unacceptable and kills the purpose of migration. Please find a way to bypass this, move that functionality in the classes and access them through a common `TaskAPI` interface

openml/tasks/functions.py
```diff
@@ -1,129 +1,28 @@
...
-    task_xml_path = tid_cache_dir / "task.xml"
-    try:
-        with task_xml_path.open(encoding="utf8") as fh:
-            return _create_task_from_xml(fh.read())
-    except OSError as e:
-        openml.utils._remove_cache_dir_for_id(TASKS_CACHE_DIR_NAME, tid_cache_dir)
-        raise OpenMLCacheException(f"Task file for tid {tid} not cached") from e
-
-
-def _get_estimation_procedure_list() -> list[dict[str, Any]]:
```
geetu040, line 73: don't remove this function, keep it as it is, it's updated here [#1604 (changes)](https://github.com/openml/openml-python/pull/1604/changes#diff-fdaf60448460bf4c7af496380c2f8967b0cabe577a9153256b8397f9f80e0ecc)
call it separately in `list_tasks` and pass onwards to `openml._backend.task.list` as a parameter

tests/test_api/test_task.py
geetu040, line 1: v1, v2 and equivalent tests are missing for `get`

tests/test_api/test_task.py
```diff
@@ -0,0 +1,200 @@
...
+@pytest.fixture
+def task_v2(http_client_v2, minio_client) -> TaskV2API:
+    return TaskV2API(http=http_client_v2, minio=minio_client)
+
+
+@pytest.fixture
+def task_fallback(task_v1, task_v2) -> FallbackProxy:
+    return FallbackProxy(task_v2, task_v1)
```
geetu040: fallback tests are not needed

tests/test_api/test_task.py
```diff
@@ -0,0 +1,200 @@
...
+
+
+def _get_first_tid(task_api: TaskV1API, task_type: TaskType) -> int:
+    """Helper to find an existing task ID for a given type using the V1 resource."""
+    tasks = task_api.list(limit=1, offset=0, task_type=task_type)
+    if tasks.empty:
+        pytest.skip(f"No tasks of type {task_type} found on test server.")
+    return int(tasks.iloc[0]["tid"])
```
geetu040: is it really needed, can we not just rely on a specific id? it's always the same database created from the fronzen docker image.

tests/test_tasks/test_task_functions.py
```diff
@@ -26,36 +27,6 @@ def setUp(self):
...
-            2,
-        )
-
-    @pytest.mark.test_server()
-    def test__get_estimation_procedure_list(self):
-        estimation_procedures = openml.tasks.functions._get_estimation_procedure_list()
-        assert isinstance(estimation_procedures, list)
-        assert isinstance(estimation_procedures[0], dict)
-        assert estimation_procedures[0]["task_type_id"] == TaskType.SUPERVISED_CLASSIFICATION
-
```
geetu040, line 58: this should be kept as it is given the previous comments

tests/test_tasks/test_task_functions.py
```diff
@@ -138,11 +109,6 @@ def test_list_tasks_per_type_paginate(self):
                     assert j == task["ttid"]
                     self._check_task(task)
 
-    @pytest.mark.test_server()
-    def test__get_task(self):
-        openml.config.set_root_cache_directory(self.static_cache_dir)
-        openml.tasks.get_task(1882)
-
```
geetu040, line 145: apply the suggestion in https://github.com/openml/openml-python/pull/1619#pullrequestreview-3996391505

tests/test_tasks/test_task_functions.py
```diff
@@ -153,20 +119,10 @@ def test__get_task_live(self):
         # https://github.com/openml/openml-python/issues/378
         openml.tasks.get_task(34536)
 
-    @pytest.mark.test_server()
-    def test_get_task(self):
-        task = openml.tasks.get_task(1, download_data=True)  # anneal; crossvalidation
-        assert isinstance(task, OpenMLTask)
```
geetu040, line 159: you can remove the path check below this, but don't remove the test function entirely, instead use mocking as suggested in https://github.com/openml/openml-python/pull/1619#pullrequestreview-3996391505

tests/test_tasks/test_task_functions.py
```diff
@@ -153,20 +119,10 @@ def test__get_task_live(self):
...
-            os.path.join(openml.config.get_cache_directory(), "tasks", "1", "datasplits.arff")
-        )
-        assert os.path.exists(
-            os.path.join(openml.config.get_cache_directory(), "datasets", "1", "dataset_1.pq")
-        )
-
+    @pytest.mark.skipif(
+        os.getenv("OPENML_USE_LOCAL_SERVICES") == "true",
+        reason="Pending resolution of #1657",
+    )
```
geetu040: I don't think we are skipping these anymore

tests/test_tasks/test_task_functions.py
```diff
@@ -153,20 +119,10 @@ def test__get_task_live(self):
...
     @pytest.mark.test_server()
     def test_get_task_lazy(self):
```
geetu040, line 141: do for this as suggested in https://github.com/openml/openml-python/pull/1619#pullrequestreview-3996391505

tests/test_tasks/test_task_functions.py
```diff
@@ -208,12 +170,6 @@ def assert_and_raise(*args, **kwargs):
         # Now the file should no longer exist
         assert not os.path.exists(os.path.join(os.getcwd(), "tasks", "1", "tasks.xml"))
 
-    @pytest.mark.test_server()
-    def test_get_task_with_cache(self):
```
geetu040, line 212: do for this as suggested in https://github.com/openml/openml-python/pull/1619#pullrequestreview-3996391505

tests/test_tasks/test_task_functions.py
```diff
@@ -224,13 +180,19 @@ def test_get_task_different_types(self):
         # Issue 538, get_task failing with clustering task.
         openml.tasks.functions.get_task(126033)
 
+    @pytest.mark.skipif(
+        os.getenv("OPENML_USE_LOCAL_SERVICES") == "true",
+        reason="Pending resolution of #1657",
+    )
```
geetu040: I don't think we should skip these

tests/test_tasks/test_task_functions.py
```diff
@@ -224,13 +180,19 @@ def test_get_task_different_types(self):
...
     @pytest.mark.test_server()
     def test_download_split(self):
```
geetu040, line 211: as suggested in https://github.com/openml/openml-python/pull/1619#pullrequestreview-3996391505

tests/test_tasks/test_task_methods.py
geetu040, line 1: nicely done, do the same for `tests/test_tasks/test_task_functions.py`

tests/test_tasks/test_task_methods.py
```diff
@@ -34,8 +34,14 @@ def test_tagging(self):
 
     @pytest.mark.test_server()
     def test_get_train_and_test_split_indices(self):
+        import unittest.mock
```
geetu040: move import to the top

tests/conftest.py
geetu040, line 1: would fail if you delete the old files in `tests/files`, you can keep it like the suggestion in https://github.com/openml/openml-python/pull/1619#pullrequestreview-3996391505

geetu040 (changes requested):
delete the old files in `tests/files` so git identifies and marks them as renamed files which would make it easier to track

a lot of irrelevant files are touched in tests, probably from your other pr's pre-commit, please revert them

see the comments on `_get_estimation_procedure_list`
don't remove this function, keep it as it is, it's updated here [#1604 (changes)](https://github.com/openml/openml-python/pull/1604/changes#diff-fdaf60448460bf4c7af496380c2f8967b0cabe577a9153256b8397f9f80e0ecc)
call it separately in `list_tasks` and pass onwards to `openml._backend.task.list` as a parameter
it should stay out of `_api/`
keep `test__get_estimation_procedure_list`

remove `TaskV1API` and `TaskV2API` import in `openml/tasks/functions.py`

geetu040 (changes requested):
I've updated https://github.com/openml/openml-python/pull/1619#pullrequestreview-3996391505. Please sync with main and update cache files accordingly.

openml/tasks/task.py
```diff
@@ -241,6 +241,46 @@ def _parse_publish_response(self, xml_response: dict) -> None:
...
+    def push_tag(self, tag: str) -> None:
+        """Annotates this task with a tag on the server.
+
+        Parameters
+        ----------
+        tag : str
+            Tag to attach to the task.
+        """
+        if self.task_id is None:
+            raise ValueError("Task does not have an ID. Please publish the task before tagging.")
```
geetu040: use `ObjectNotPublishedError` as in [_tag_openml_base](https://github.com/openml/openml-python/blob/main/openml/utils/_openml.py#L113)

openml/tasks/task.py
```diff
@@ -241,6 +241,46 @@ def _parse_publish_response(self, xml_response: dict) -> None:
...
+
+        Parameters
+        ----------
+        tag : str
+            Tag to remove from the task.
+        """
+        if self.task_id is None:
+            raise ValueError(
+                "Dataset does not have an ID. Please publish the dataset before untagging."
+            )
```
geetu040: same as above

geetu040 (changes requested):

openml/tasks/functions.py
satvshr: We will then have to alter the signature of `openml._backend.task.list` to accept `_get_estimation_procedure_list` as a  permanent argument, which doesnt make sense?

openml/_api/resources/base/resources.py
satvshr: The place it was being used was removed by another PR, removing this

openml/tasks/functions.py
geetu040: I think that would make sense, why not?

openml/tasks/task.py
geetu040, line 1: There is cache and api-calls which are needed to be updated in this file. I think you'd only need to update `OpenMLTask.download_split`. You should use `_http.download` here. Let me know if you need help.

geetu040 (changes requested):

openml/tasks/functions.py
satvshr: The function is being called in multiple places so I assume it is something that will be used a lot, won't this change just prevent positional arguments in the future?

openml/tasks/task.py
geetu040: as discussed on [discord](https://discord.com/channels/1439727064231182487/1486708381916991528), no need to use `_http.download`, instead use `_http.get` where applicable.

openml/tasks/functions.py
geetu040: discussion moved over to [discord](https://discord.com/channels/1439727064231182487/1468511432713699340)

geetu040 (commented):
@satvshr what's the status of this PR? It has been stale for quite a while now. Are you still working on it, planning to continue it, or waiting for a review?

satvshr:
@geetu040 you had pushed some commits so I was awaiting instructions on how to proceed next, unless you were picking it up, was my thought.

geetu040:
> @geetu040 you had pushed some commits so I was awaiting instructions on how to proceed next, unless you were picking it up, was my thought.

I think we discussed but I've made the changes in both datasets and this pr. You can continue if you are willing to work on it. That would require first to fix the tests in the CI by updating tests suits with the moved cache files.

PR #1619, opened 2026-01-20
[ENH] V1 → V2 API Migration - setups

### LINKED ISSUE #1625: [ENH] V1 → V2 API Migration - setups
V1 → V2 API Migration - setups. Sub-issue of #1575, for tracking.

EmanAbdelhaleem:
Fixes #1625 
Depends on #1576
Related to #1575

Details
This PR implements `Setups` resource, and refactor its existing functions


EmanAbdelhaleem:
@geetu040 Can you please check if there is something wrong in the get method?
For some reason when I remove the caching from the method in `setups/functions` and rely on my implementation for the migrated api call, it doesn't work as I receive an HTML not XML response. Although it works fine on the original `get_setup` method in `setups/functions`

openml/_api/resources/setups.py
```diff
@@ -0,0 +1,230 @@
...
+            - xml_content: The raw XML response from the server
+            - setup: An initialized OpenMLSetup object parsed from the XML
+        """
+        url_suffix = f"/setup/{setup_id}"
+        setup_response = self._http.get(url_suffix)
+        xml_content = setup_response.text
+        result_dict = xmltodict.parse(xml_content)
+
+        setup = self._create_setup(result_dict)
+        return xml_content, setup
```
geetu040: no need to return `xml_content`, its being used for cache which will be removed
see point 1 in https://github.com/openml/openml-python/issues/1575#issuecomment-3816198783

openml/_api/resources/setups.py
```diff
@@ -0,0 +1,230 @@
...
+        raise NotImplementedError("V2 API implementation is not yet available")
+
+    def _create_setup(self, result_dict: dict) -> OpenMLSetup:
+        raise NotImplementedError("V2 API implementation is not yet available")
+
+    def get(self, setup_id: int) -> tuple[str, OpenMLSetup]:
+        raise NotImplementedError("V2 API implementation is not yet available")
+
+    def exists(self, file_elements: dict[str, Any]) -> int:
+        raise NotImplementedError("V2 API implementation is not yet available")
```
geetu040: see point 8 in https://github.com/openml/openml-python/issues/1575#issuecomment-3816198783

openml/setups/functions.py
```diff
@@ -120,13 +116,12 @@ def get_setup(setup_id: int) -> OpenMLSetup:
     try:
         return _get_cached_setup(setup_id)
     except openml.exceptions.OpenMLCacheException:
-        url_suffix = f"/setup/{setup_id}"
-        setup_xml = openml._api_calls._perform_api_call(url_suffix, "get")
+        result: tuple[str, OpenMLSetup] = api_context.backend.setups.get(setup_id=setup_id)
```
geetu040: set `use_cache=True` in `get`
see point 1 in https://github.com/openml/openml-python/issues/1575#issuecomment-3816198783

openml/setups/functions.py
```diff
@@ -90,7 +86,7 @@ def _get_cached_setup(setup_id: int) -> OpenMLSetup:
         setup_file = setup_cache_dir / "description.xml"
         with setup_file.open(encoding="utf8") as fh:
             setup_xml = xmltodict.parse(fh.read())
-            return _create_setup_from_xml(setup_xml)
+            return _create_setup(setup_xml)
```
geetu040: remove all cache related code
see point 1 in https://github.com/openml/openml-python/issues/1575#issuecomment-3816198783

geetu040 (changes requested):
you need to work on cache, see other PRs for tasks/datasets for reference
please take a look at https://github.com/openml/openml-python/issues/1575#issuecomment-3816198783 and make changes accordingly where needed.

EmanAbdelhaleem:
Running the tests for both files:
`tests\test_api\test_setups.py` or `tests\test_setups\test_setup_functions.py` give 
```openml.exceptions.OpenMLServerException: https://www.openml.org/api/v1/xml/setup/exists/ returned code 103: Authen...``` for setup_exists related tests.

To reproduce:
```pytest tests\test_setups\test_setup_functions.py```

**Comment by EmanAbdelhaleem** [comment]
@geetu040 Ready for review

File: tests/test_api/test_setup.py [hunk 5]
```diff
@@ -0,0 +1,170 @@
...
+    
+    @pytest.mark.sklearn()
+    @pytest.mark.uses_test_server()
+    def test_exists_existing_setup(self):
+        """Test exists() returns setup_id when setup exists"""
+        flow = self.extension.model_to_flow(sklearn.naive_bayes.GaussianNB())
+        flow.name = f"{get_sentinel()}{flow.name}"
+        flow.publish()
+        TestBase._mark_entity_for_removal("flow", flow.flow_id, flow.name)
+        TestBase.logger.info(f"collected from {__file__.split('/')[-1]}: {flow.flow_id}")
```
geetu040: it's nice that you have it here, but it won't take affect, I should add `TestBase` as parent for `TestAPIBase` so this works

tests/test_api/test_setup.py
```diff
@@ -0,0 +1,170 @@
...
+        TestBase.logger.info(f"collected from {__file__.split('/')[-1]}: {run.run_id}")
+        # download the run, as it contains the right setup id
+        run = openml.runs.get_run(run.run_id)
+
+        # execute the function we are interested in
+        setup_id = openml.setups.setup_exists(flow)
+        assert setup_id == run.setup_id
+
+
+class TestSetupV2(TestAPIBase):
```
geetu040: this class has no tests, you should check if it not implemented methods raise the correct exception

geetu040 (changes requested):
https clients are already initialized, you copy design from [tests/test_api/test_versions.py](https://github.com/geetu040/openml-python/blob/migration/tests/test_api/test_versions.py)

tests/test_api/test_setup.py
```diff
@@ -0,0 +1,170 @@
...
+    """Tests for V1 XML API implementation of setups."""  
+  
+    _multiprocess_can_split_ = True  
+  
+    def setUp(self) -> None:    
+        super().setUp() 
+        self.resource = SetupV1API(self.http_client)
+        self.extension = SklearnExtension()
+  
+    @pytest.mark.uses_test_server()
```
geetu040: can be moved to class level

tests/test_setups/test_setup_functions.py
geetu040, line 1: good, hopefully other tests are not breaking.

tests/test_api/test_setup.py
EmanAbdelhaleem: You are talking about the extension, right?
I followed the code of the existing tests. and it's initialized in the setUp() function, I am not even sure if moving it would make difference, the setUp() is already called before running the class tests, rigth?

geetu040: I was talking about `@pytest.mark.uses_test_server()`

tests/test_api/test_setup.py
```diff
@@ -0,0 +1,170 @@
...
+        flow = self.extension.model_to_flow(dectree)
+        flow.name = f"{sentinel}{flow.name}"
+        flow.publish()
+        TestBase._mark_entity_for_removal("flow", flow.flow_id, flow.name)
+        TestBase.logger.info(f"collected from {__file__.split('/')[-1]}: {flow.flow_id}")
+
+        # although the flow exists (created as of previous statement),
+        # we can be sure there are no setups (yet) as it was just created
+        # and hasn't been ran
+        setup_id = openml.setups.setup_exists(flow)
```
EmanAbdelhaleem: @geetu040 
I am not sure how to test the `exists` method without mocking as it is dependent on a Flow object and a run Object. These tests I am using here, which I copied from `test_setups_functions.py`, are not actually testing the API cuz they are using the `openml.setups.setup_exists()`

The API methods signature is `def exists(self, file_elements: dict[str, Any]) -> int | bool:`, so, I need to pass `file_elements` if I want to test it, and to get these file elements, the following code is used in the `def setup_exists(flow: OpenMLFlow) -> int:` function in `setups/functions.py`

```
# sadly, this api call relies on a run object
    openml.flows.functions._check_flow_for_server_id(flow)
    if flow.model is None:
        raise ValueError("Flow should have model field set with the actual model.")
    if flow.extension is None:
        raise ValueError("Flow should have model field set with the correct extension.")

    # checks whether the flow exists on the server and flow ids align
    exists = flow_exists(flow.name, flow.external_version)
    if exists != flow.flow_id:
        raise ValueError(
            f"Local flow id ({flow.id}) differs from server id ({exists}). "
            "If this issue persists, please contact the developers.",
        )

    openml_param_settings = flow.extension.obtain_parameter_values(flow)
    description = xmltodict.unparse(_to_dict(flow.flow_id, openml_param_settings), pretty=True)
    file_elements = {
        "description": ("description.arff", description),
    }
```

So, I first need to create and publish the flow as in these tests and then get the `file_elements` as in the above code, However, I think this might be wrong for a unit test, What do you think?

geetu040: yes you are right, this is not ideal for a unit test, but we already have many _not so ideal_ tests, adding one more won't be a problem. but I would suggest if mocking can make it simple for you then sure go ahead and use that, we are already using it in some places where without mocking irrelevant parts of code would need to be executed.

tests/test_api/test_setup.py
```diff
@@ -0,0 +1,176 @@
...
+from openml.testing import TestBase
+
+  
+from openml._api.resources.setup import SetupV1API, SetupV2API
+from openml.setups.setup import OpenMLSetup
+from openml.testing import TestAPIBase
+from openml.enums import APIVersion
+from openml._api.resources.base.fallback import FallbackProxy
+from openml.exceptions import OpenMLNotSupportedError
+
```
Omswastik-11, line 16: I think here we need to implement `TestResourceAPIBase` and move all common tests there . 

https://discord.com/channels/1439727064231182487/1468994885297246399/1469408217082888369

tests/test_api/test_setup.py
EmanAbdelhaleem: This question is outdated actually, I would ask for a review for the recent changes, both for the tests and the resource module.

geetu040 (changes requested):

geetu040 (changes requested):

geetu040 (commented):
To check: Does https://github.com/openml/openml-python/pull/1611#pullrequestreview-3991387133 affect this PR?

geetu040 (commented):
some old tests will fail because the cache path has changed
```
old: tests/files/org/openml/test/tasks/1882
new: tests/files/org/openml/test/api/v1/xml/task/1882
```
regarding this we can do the following, as i've done in this PR
* move the relevant cache test files to new directory - (reference - [setup/1/body.xml](https://github.com/openml/openml-python/pull/1619/changes#diff-3682c8f08293492199786cebc6903a183dc5edc9db91a96fc6aeb82c1eae040c))
    * create new directory `tests/files/org/openml/test/api/v1/xml/[resource]/[id]/`, move the `description.xml` file there and rename it to `body.xml`
    * inshort a file `tests/files/org/openml/test/setups/1/description.xml` becomes `tests/files/org/openml/test/api/v1/xml/setup/1/body.xml`
    * move other downloadable files from `minio` and `HTTPClient.download` accordingly
* if there are any tests that rely on cache, update them by mocking `Session.request` - (reference - [test_setup_functions](https://github.com/openml/openml-python/pull/1619/changes#diff-c90dbab833c610dbaf84eb42dc7d7af823ab937949b4ca20dea818afbb83f373))
* update `conftest.py` with new file paths, for now you can manually hardcode the new paths and remove this resource from the code earlier, we can later clean this up - (reference - [conftest.py](https://github.com/openml/openml-python/pull/1619/changes#diff-e52e4ddd58b7ef887ab03c04116e676f6280b824ab7469d5d3080e5cba4f2128))

FYI: @satvshr @Omswastik-11 @JATAYU000

geetu040 (approved):
@EmanAbdelhaleem Thanks for the PR. Nicely done!
I have updated the PR to sync with latest changes in base PR.

@PGijsbers please review/merge.

openml/_api/resources/setup.py
```diff
@@ -1,11 +1,278 @@
...
+            }
+        elif isinstance(xml_parameters, builtins.list):
+            parameters = {
+                int(xml_parameter["oml:id"]): self._create_setup_parameter_from_xml(xml_parameter)
+                for xml_parameter in xml_parameters
+            }
+        else:
+            raise ValueError(
+                f"Expected None, list or dict, received something else: {type(xml_parameters)!s}",
+            )
```
PGijsbers, line 109:
```suggestion
        if isinstance(xml_parameters, dict):
            xml_parameters = [xml_parameters]
        if not isinstance(xml_parameters, list):
            raise ValueError(
                f"Expected None, list or dict, received something else: {type(xml_parameters)!s}",
            )
            
        parameters = {
            int(xml_parameter["oml:id"]): self._create_setup_parameter_from_xml(xml_parameter)
            for xml_parameter in xml_parameters
        }
```

I realize this is just our code that was moved around, but I had to do a double take to see what's going on so I figured I might as well suggest the change now to make it easier to parse in the future.

openml/_api/resources/setup.py
```diff
@@ -1,11 +1,278 @@
...
+        setup_response = self._http.post(api_call, files=file_elements)
+        xml_content = setup_response.text
+        result_dict = xmltodict.parse(xml_content)
+
+        setup_id = int(result_dict["oml:setup_exists"]["oml:id"])
+        return setup_id if setup_id > 0 else False
+
+    def _to_dict(
+        self, flow_id: int, openml_parameter_settings: builtins.list[dict[str, Any]]
+    ) -> OrderedDict:
```
PGijsbers: Please turn it into a static method since it doesn't use the object's state.

openml/_api/resources/setup.py
```diff
@@ -1,11 +1,278 @@
...
+                int(xml_parameter["oml:id"]): self._create_setup_parameter_from_xml(xml_parameter)
+                for xml_parameter in xml_parameters
+            }
+        else:
+            raise ValueError(
+                f"Expected None, list or dict, received something else: {type(xml_parameters)!s}",
+            )
+        return OpenMLSetup(setup_id, flow_id, parameters)
+
+    def _create_setup_parameter_from_xml(self, result_dict: dict[str, str]) -> OpenMLParameter:
```
PGijsbers: Please turn it into a static method since it doesn't use the object's state.

openml/_api/resources/setup.py
```diff
@@ -1,11 +1,278 @@
...
+        assert isinstance(setups_dict["oml:setups"]["oml:setup"], list), type(
+            setups_dict["oml:setups"]
+        )
+
+        return [
+            self._create_setup({"oml:setup_parameters": setup_})
+            for setup_ in setups_dict["oml:setups"]["oml:setup"]
+        ]
+
+    def _create_setup(self, result_dict: dict) -> OpenMLSetup:
```
PGijsbers: Please turn it into a static method since it doesn't use the object's state.

tests/test_api/test_setup.py
```diff
@@ -0,0 +1,150 @@
...
+def get_sentinel():
+    # Create a unique prefix for the flow. Necessary because the flow is
+    # identified by its name and external version online. Having a unique
+    #  name allows us to publish the same flow in each test run
+    md5 = hashlib.md5()
+    md5.update(str(time.time()).encode("utf-8"))
+    sentinel = md5.hexdigest()[:10]
+    return f"TEST{sentinel}"
```
PGijsbers:
```suggestion
    sentinel = uuid.uuid4().hex[:10]
    return f"TEST{sentinel}"  
```
(and import `uuid`)

The generated string is deterministic based on the current time. That means that if this is used for multiple tests, and they are executed in parallel, there is a non-zero chance the same value is generated (or when the clock is set back). Admittedly, the precision of `time.time()` is enough that this likely doesn't happen, but it's good to avoid the chance. Moreover, generating it with `uuid` clearly signals the intent that we are just looking for a random string.

PGijsbers (changes requested):
I left a few remarks, additionally the ability to add and remove tags was added, but I did not see any tests added that test this new functionality. Those tests should be added.

I believe we had briefly discussed this before on a different PR, but I can't find it easily. Why store the response body as a binary file? The response is always text (either xml or json, potentially also some error text but text regardless). I do not think it is good to store these files as binary because it doesn't play well with git -and- because it makes it hard to review the file content (see for example the recent `xz`/`ssh` exploit). I would prefer these files have an extension that matches their content, and if that needs to be more generic, then we should use `.txt`.

openml/_api/resources/setup.py
geetu040: updated

openml/_api/resources/setup.py
geetu040: updated

openml/_api/resources/setup.py
geetu040: updated

tests/test_api/test_setup.py
geetu040: good catch, updated

geetu040:
> I left a few remarks, additionally the ability to add and remove tags was added, but I did not see any tests added that test this new functionality. Those tests should be added.

I have removed tagging from OpenMLSetup, since it was unrelated to this PR and already tracked in https://github.com/openml/openml-python/issues/1648

> Why store the response body as a binary file?

I stored it in binary since the `requests.Response` object saves the content in binary format, therefore I used the native extension, but you point makes sense as well. I have updated the cache to store content in relevant files: `body.json`, `body.xml` otherwise `body.txt` and the implementation is clean, so I guess we are fine this way.

---

@PGijsbers This PR is up for review/merge.

PGijsbers (approved):

PR #1620, opened 2026-01-20
[BUG] Correct path to cache for cleanup of task download failure

PGijsbers:
Fixes an issue that I believe was introduced in #1584: the function called already generates the directory that is task specific, so there is no need to check for a subdirectory. I believe that this patch achieves the intended behavior change of #1584. The corresponding unit test now passes (`tests/test_tasks/test_task_functions.py::test_removal_upon_download_failure`), where it would previously fail since the subdirectory was never made.


fkiraly (commented):
Have you checked that the test is running on remote?

PGijsbers:
> Have you checked that the test is running on remote?

At the time of this PR it was up-to-date and CI passed. I'll update the PR.

PR #1628, opened 2026-01-22
[BUG] Test Failures caused because of pandas 3

### LINKED ISSUE #1627: [BUG] Test failures
7 tests seem to be failing in all these PRs ([link](https://github.com/openml/openml-python/pull/1577), [link](https://github.com/openml/openml-python/actions/runs/21228880416/job/61082770429), [here](https://github.com/openml/openml-python/actions/runs/21243115804/), and [link](https://github.com/openml/openml-python/actions/runs/21228673827/job/61082120329?pr=1560)). These bugs should be looked into and only be skipped if the failures are erratic.

satvshr:
#### Metadata
* Reference Issue: fixes #1627

This PR fixes the 7 recurring bugs across all current PRs because of pandas 3:
1. `test_get_data_pandas` bug: Solved type error for dataframe columns having `str` datatype for `pandas==3` and `object` for older versions.
2. `test_get_sparse_dataset_dataframe`, `test_get_sparse_dataset_rowid_and_ignore_and_target`, and `test_get_sparse_dataset_dataframe_with_target` bug: typecasting `type_` to a np array.
3. bugs in `test_flow_functions.py`: `ext_version` can now be `nan` because of `pandas 3`.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -83,6 +83,7 @@ jobs:
         python-version: ${{ matrix.python-version }}
 
     - name: Install test dependencies, scikit-learn, and optional pandas
+      shell: bash
       run: |
         python -m pip install --upgrade pip
         pip install -e .[test] scikit-learn==${{ matrix.scikit-learn }}


tests/test_datasets/test_dataset.py
```diff
@@ -102,21 +102,24 @@ def test_get_data_pandas(self):
         assert isinstance(data, pd.DataFrame)
         assert data.shape[1] == len(self.titanic.features)
         assert data.shape[0] == 1309
+        # Dynamically detect what this version of Pandas calls string columns.
+        str_dtype = data["name"].dtype.name
+
```
geetu040, line 107: Isn't that cheating? why would it ever fail for "name" column if we are dynamically extracting the dtype. Is it to make sure it works with `pandas<3.0.0`?

would make sense to replace `"object"` simply with `"str"` in `col_dtype` if we are setting `pandas>=3.0.0` in dependencies

CC: @fkiraly, are we bounding to this pandas version in `pyproject.toml`?

satvshr: > Isn't that cheating?

It is definitely cheating! Thought it was better to get a quick fix out and then fine-tune the fixes later on. It was between this or bounding pandas, and for me this was the better/less intense option between the 2.

PR #1629, opened 2026-01-29
[MNT] Dockerized tests for CI runs using localhost

### LINKED ISSUE #1614: [MNT] Intermediate test plan
We want to move tests from relying on the remote server (`test.openml.org`) to a locally running server setup. This is needed to avoid race conditions and issues caused by a shared, mutable central database (as discussed in https://github.com/openml/openml-python/issues/1586).

At the moment, test cases use both the production server (`www.openml.org`) and the remote test server (`test.openml.org`). The goal is not to remove all remote calls at once, but to gradually replace calls to `test.openml.org` with calls to a local server.

The local setup will be started via Docker containers in the CI workflows. We will need three services:

* `test-database`
* `php-api-v1`
* `python-api-v2`

The SDK does not yet support `python-api-v2`; work on this is ongoing in https://github.com/openml/openml-python/issues/1575. However, we still plan to enable the local `python-api-v2` service now, since tests targeting it will be added as the migration progresses.

I'll follow up with more details on:

* how these Docker services will be started programmatically in the workflows, and
* how the pytest configuration can be adjusted to use the local services instead of remote endpoints.

satvshr:
#### Metadata
* Reference Issue: fixes #1614, stacks on #1630
* New Tests Added: No
* Documentation Updated: No

#### Details 
This PR implements the setting up of the v1 and v2 test servers in CI using docker via `localhost`.

### DIFF
--- a/tests/test_setups/test_setup_functions.py
+++ b/tests/test_setups/test_setup_functions.py
@@ -159,7 +159,7 @@ class TestSetupFunctions(TestBase):
 
         setups = openml.setups.list_setups(flow=flow_id)
 
-        assert len(setups) > 0  # TODO: please adjust 0
+        assert len(setups) >= 2
         for setup_id in setups:
             assert setups[setup_id].flow_id == flow_id
 


satvshr:
Docker fails to build for Windows, given that the image of database are built using Linux binaries. It works locally because Docker Desktop ends up using WSL in the background (as far as I can understand) to run these containers without errors. Currently I am skipping running docker for the windows env during CI runs only, and using this for tests that will end up using the docker servers:
```python
is_windows_ci = (sys.platform == "win32" or os.name == "nt") and os.environ.get("CI") == "true"
skip_on_windows_ci = pytest.mark.skipif(
    is_windows_ci,
    reason="Skipping Docker tests on Windows CI (No WSL2 available)"
)

# Used in the form:
@skip_on_windows_ci
def test_database_connection():
    ...
```

Since we eventually plan on migrating all tests to these local servers, and we would like windows to run them too, we should think of another method to run these tests on windows CI.

geetu040:
> Docker fails to build for Windows, given that the image of database are built using Linux binaries.

makes sense, github's windows runners don't have a wsl layer, we should think about that.

tests/conftest.py
```diff
@@ -296,6 +302,46 @@ def with_test_cache(test_files_directory, request):
...
+    except (requests.exceptions.ConnectionError, requests.exceptions.Timeout):
+        return False
+
+def _start_docker():
+    """Logic to spin up the containers and wait for initialization."""
+    subprocess.run(["docker", "compose", "up", "-d"], check=True, capture_output=True, text=True)
+    subprocess.run(["docker", "wait", "openml-test-setup-ci"], check=True)
+
+@pytest.fixture(scope="session", autouse=True)
+def openml_docker_stack(tmp_path_factory, worker_id):
```
geetu040: please undo these changes. this is a really good implementation but we don't want to manually setup the docker services when pytests are run, that could be problematic for developers setting up pytests (it also clones the other repo inside this which is really bad), though we'll give them an option to setup their docker services themselves and toggle the pytest behaviour to use localhosts via an env variable

docker-compose.yml
geetu040, line 1: should be removed

docker/update.sh
geetu040, line 1: should be removed

.github/workflows/test.yml
geetu040, line 1: there should be following steps to setup the docker services, though these should be ignored for window machines
1. clone the repo, which you are already doing
2. run the docker services
3. verify that localhost endpoints are live

tests/conftest.py
satvshr: > it also clones the other repo inside this which is really bad

I am still reading all the reviews, but this does not happen locally, only in CI.

docker/update.sh
satvshr: It is needed to update the database image which we are using (PR description)

docker-compose.yml
satvshr: It is needed to run docker....why should it be removed?

geetu040: when we clone the server-api repo, it already has the docker-compose file and should be used from there.

docker/update.sh
geetu040: as above, this should come from cloning server-api repo

tests/conftest.py
geetu040: yes, that is correct and in that case it's redundant. because the containers should already be running and verified in a previous setup of work, no need to rerun/validate here then

satvshr:
`openml/rest-api:0.0.1-alpha` image for [the rest (v2) api](https://hub.docker.com/r/openml/rest-api) for my note.

satvshr:
@geetu040 Doesn't #1630 close #1614 ? If services PR [13](https://github.com/openml/services/pull/13) and [15](https://github.com/openml/services/pull/15) add the v1 and v2 apis to the services repo, and #1630 allows them to be used as a `test_server` on changing `TEST_SERVER_URL` to `http://localhost:8000`  what is the point of this PR again?

Edit: I think the only thing I can do in addition to Peter's PR is run the setup during CI, so I close this PR and open a new one stacking on #1630 and adding this additional functionality.

geetu040:
> Doesn't #1630 close #1614 ? If services PR [13](https://github.com/openml/services/pull/13) and [15](https://github.com/openml/services/pull/15) add the v1 and v2 apis to the services repo, and #1630 allows them to be used as a `test_server` on changing `TEST_SERVER_URL` to `http://localhost:8000` what is the point of this PR again?

#1630 adds variable `TEST_SERVER_URL` throughout the sdk to easily configure test-server path.

Point of this PR is to get the docker services running in the CI.

> Edit: I think the only thing I can do in addition to Peter's PR is run the setup during CI, so I close this PR and open a new one stacking on #1630 and adding this additional functionality.

Keep this PR, how it already is pointing to `main`. For now just focus on adding the services in the CI. We need to see how long do the services take to setup in the CI, based on that, we can move forward with replacing remote test-server configured with local test-server in the CI. By then we'll see if we need to stack on top of Pieter's PR if it's not already merged to `main`.

.github/workflows/test.yml
```diff
@@ -101,11 +98,25 @@ jobs:
...
+      run: git clone --depth 1 https://github.com/openml/services.git
+
+    - name: Start Docker Services
+      if: matrix.os == 'ubuntu-latest'
+      working-directory: ./services
+      run: |
+        sudo systemctl stop mysql.service
+        docker compose --profile rest-api --profile minio --profile evaluation-engine up -d
+        docker wait openml-test-database-setup
+
```
geetu040, line 120: It takes around 2:30 minutes to start the services, that's probably acceptable.

geetu040, line 120: In the next step, can we verify that the api is working by doing curl something? maybe `http://localhost:8000/api/v1/xml/data/1`

.github/workflows/test.yml
```diff
@@ -101,11 +98,25 @@ jobs:
         echo "BEFORE=$git_status" >> $GITHUB_ENV
         echo "Repository status before tests: $git_status"
 
+    - name: Clone Services
+      if: matrix.os == 'ubuntu-latest'
+      run: git clone --depth 1 https://github.com/openml/services.git
```
geetu040: please checkout to this branch https://github.com/openml/services/pull/13

.github/workflows/test.yml
satvshr: Was going to mention this during the meeting

satvshr: Good idea!

.github/workflows/test.yml
satvshr: I forgot that, good reminder! Will merge 13 (v1 and image updates) and 15  (v2) into main for now to know the correct execution times.

satvshr:
> where do you set the `TEST_SERVER_URL` from env in tests?

I was going to remove that given it is not needed right now, `TEST_SERVER_URL = "https://test.openml.org"` is harcoded in `config.py`

.github/workflows/test.yml
```diff
@@ -101,6 +101,38 @@ jobs:
...
+
+        git fetch origin pull/15/head:pr-15
+        git merge pr-15 --no-edit
+
+    - name: Start Docker Services
+      if: matrix.os == 'ubuntu-latest'
+      working-directory: ./services
+      run: |
+        sudo systemctl stop mysql.service
+        docker compose --profile rest-api --profile minio --profile evaluation-engine up -d
```
geetu040: why do we need to run `evaluation-engine`?

.github/workflows/test.yml
geetu040, line 1: add env var
```
OPENML_USE_LOCAL_SERVICES: "true"
```

.github/workflows/test.yml
```diff
@@ -101,6 +101,38 @@ jobs:
...
+        docker wait openml-test-database-setup
+
+    - name: Verify API is Reachable
+      if: matrix.os == 'ubuntu-latest'
+      run: |
+        timeout 180s bash -c 'until curl -sSf http://localhost:8000/api/v1/xml/data/1 > /dev/null; do 
+                  echo "Server still booting... retrying in 5s"; 
+                  sleep 5; 
+                done'
+        curl -I http://localhost:8000/api/v1/task/1
```
geetu040:
```suggestion
      run: |
        echo "Waiting for API to become available..."
        timeout 30s bash -c 'until curl -sSf http://localhost:8080/api/v1/task/1 > /dev/null; do sleep 2; done'

        echo "Response:"
        curl http://localhost:8080/api/v1/task/1
```

tests/conftest.py
```diff
@@ -277,7 +277,7 @@ def with_server(request):
         openml.config.apikey = None
         yield
         return
-    openml.config.server = "https://test.openml.org/api/v1/xml"
+    openml.config.server = f"{openml.config.TEST_SERVER_URL}/api/v1/xml"
```
geetu040, line 282: check the env variable `OPENML_USE_LOCAL_SERVICES` and update server accordingly
```py
@pytest.fixture(autouse=True)
def with_server(request):
    if os.getenv("OPENML_USE_LOCAL_SERVICES") == "true":
        openml.config.TEST_SERVER_URL = "http://localhost:8080"

    if "production" in request.keywords:
        openml.config.server = "https://www.openml.org/api/v1/xml"
        openml.config.apikey = None
        yield
        return
    openml.config.server = f"{openml.config.TEST_SERVER_URL}/api/v1/xml"
    openml.config.apikey = TestBase.user_key
    yield
```

.github/workflows/test.yml
satvshr: Similar to what there was earlier, changed it the previous commit and was going to revert it after CIs

satvshr: Similar to what there was earlier, changed it the previous commit and was going to revert it after CIs

PGijsbers:
I didn't look at it too closely but considering it looks like the local evaluations   go wrong its not likely to do with any server connection issues [https://github.com/openml/openml-python/actions/runs/21919012927/job/63293911025?pr=1629#logs](https://github.com/openml/openml-python/actions/runs/21919012927/job/63293911025?pr=1629#logs) . The error message about datatypes immediately makes me think of pandas, and this PR does not contain the fixes from https://github.com/openml/openml-python/pull/1628. I have to assume that is the underlying issue for that error.

The other error you sent _is_ strange [https://github.com/openml/openml-python/actions/runs/21940621497/job/63365126160?pr=1629](https://github.com/openml/openml-python/actions/runs/21940621497/job/63365126160?pr=1629). I'll have a closer look after my next meeting.

update: I can't quickly find a reason for the error. I added it on my list to check later.

tests/conftest.py
geetu040: not needed here, better to use this env variable to change `TEST_SERVER_URL` in `config.py`

.github/workflows/test.yml
```diff
@@ -101,6 +101,33 @@ jobs:
...
+        # # === PATCH 1: Use 'nginx' hostname ===
+        # # This works inside Docker (DNS) and on Host (via /etc/hosts hack above)
+        # sed -i 's/localhost:8000/nginx:8000/g' config/database/update.sh
+        
+        # # === PATCH 2: Fix Path Mismatch ===
+        # # Ensure we use /data/ which Nginx recognizes
+        # sed -i 's|/minio/|/data/|g' config/database/update.sh
+        
+        # echo "=== Patched Update Script ==="
+        # cat config/database/update.sh | grep "nginx"
```
geetu040: why extra work here? locally just running the services is enough

openml/config.py
```diff
@@ -28,6 +29,11 @@
 OPENML_TEST_SERVER_ADMIN_KEY_ENV_VAR = "OPENML_TEST_SERVER_ADMIN_KEY"
 _TEST_SERVER_NORMAL_USER_KEY = "normaluser"
 
+if sys.platform.startswith("win"):
+    TEST_SERVER_URL = "http://localhost"
+else:
+    TEST_SERVER_URL = "http://localhost:8000"
+
```
geetu040: we should actually use an env variable here, please see https://github.com/openml/openml-python/pull/1629/changes#r2797509441
should be controlled by that env variable, which if not set, should default to use https://test.openml.org/

geetu040 (changes requested):

.github/workflows/test.yml
satvshr: Kindly ignore these, the pr isnt ready for review yet as tests are still failing and I was trying to debug tests.

openml/config.py
satvshr: This is not how I plan to resolve this either, just a temporary fix to the windows issue.

geetu040:
The tests are taking too long because `connection_n_retries` is set to `5`, you can set it to `1` for this PR, to avoid delays in CI.

satvshr:
> The tests are taking too long because `connection_n_retries` is set to `5`, you can set it to `1` for this PR, to avoid delays in CI.

Will do that to prevent hold ups for other CIs in the repo, for my branch it is noticeable if a run is going to fail if it has been stuck on a single test for more than a minute.

geetu040:
> Will do that to prevent hold ups for other CIs in the repo, for my branch it is noticeable if a run is going to fail if it has been stuck on a single test for more than a minute.

yeah but each job in this PR still takes full 150 minutes

openml/config.py
```diff
@@ -157,7 +157,7 @@ def _resolve_default_cache_dir() -> Path:
     "cachedir": _resolve_default_cache_dir(),
     "avoid_duplicate_runs": False,
     "retry_policy": "human",
-    "connection_n_retries": 5,
+    "connection_n_retries": 1,
```
geetu040: I don't think this would work, since we change this again in `conftest.py`.
To be completely sure that this works, you can temporarily set `n_retries = 1` in `_api_calls.py::_send_request`

.github/workflows/test.yml
```diff
@@ -101,6 +101,12 @@ jobs:
         echo "BEFORE=$git_status" >> $GITHUB_ENV
         echo "Repository status before tests: $git_status"
 
+    - name: Clone Services & Apply Universal Patch
+      if: matrix.os == 'ubuntu-latest'
+      run: |
+        git clone --depth 1 https://github.com/openml/services.git
+        cd services
+
```
geetu040, line 120: you are not running these services yet.

geetu040 (changes requested):

.github/workflows/test.yml
satvshr: DId not realise I accidentally removed it

tests/test_flows/test_flow.py
```diff
@@ -222,6 +223,7 @@ def test_publish_existing_flow(self, flow_exists_mock):
             f"collected from {__file__.split('/')[-1]}: {flow.flow_id}",
         )
 
+    @pytest.mark.skip(reason="Pending resolution of #1657")
```
geetu040: skip these only if `OPENML_USE_LOCAL_SERVICES` is set to `True`

tests/test_flows/test_flow.py
geetu040, line 1: some tests are skipped though they are not mentioned in https://github.com/openml/openml-python/issues/1657, why is that?

satvshr: I was using the failures from [here](https://github.com/openml/openml-python/actions/runs/22128792999/job/63964351099?pr=1629)

.github/workflows/test.yml
```diff
@@ -145,6 +164,20 @@ jobs:
...
         pytest -n 4 --durations=20 --dist load -sv --reruns 5 --reruns-delay 1 -m "not test_server"
 
+    - name: Upload coverage
+      if: matrix.code-cov && always()
+      uses: codecov/codecov-action@v4
+      with:
+        files: coverage.xml
+        token: ${{ secrets.CODECOV_TOKEN }}
+        fail_ci_if_error: true
+        verbose: true
```
geetu040, line 174: why is this part moved above?

tests/conftest.py
```diff
@@ -272,6 +272,8 @@ def as_robot() -> Iterator[None]:
 
 @pytest.fixture(autouse=True)
 def with_server(request):
+    if os.getenv("OPENML_USE_LOCAL_SERVICES") == "true":
+        openml.config.TEST_SERVER_URL = "http://localhost:8000"
```
geetu040, line 276: I would ask for you suggestion. Do you think it should be here or rather moved to `config.py`, since you had it there before this.
I think, having this in `config.py` would be better, it would also be in favor of global config architecture in https://github.com/openml/openml-python/pull/1577 and migration in https://github.com/openml/openml-python/pull/1576

.github/workflows/test.yml
satvshr: Codecov was giving errors, I do not recall why the errors were occurring.

tests/conftest.py
satvshr: First thing that comes to mind is the fact that something pertaining to the testing env should be kept in `conftest`, though I have no hard opinion on this. If it is more convenient for you I can move it to config too.

geetu040: Let's keep it here then

openml/config.py
geetu040: you can undo this now

openml/_api_calls.py
```diff
@@ -362,7 +362,7 @@ def _send_request(  # noqa: C901, PLR0912
     files: FILE_ELEMENTS_TYPE | None = None,
     md5_checksum: str | None = None,
 ) -> requests.Response:
-    n_retries = max(1, config.connection_n_retries)
+    n_retries = 1
```
geetu040: you can undo this now

openml/config.py
satvshr: Was just waiting to see if tests pass before reverting all changes, apologies!

geetu040:
The PR looks good, tests are also passing. I will approve this as https://github.com/openml/openml-python/pull/1629#pullrequestreview-3858883917 is addressed.

satvshr:
> Is it complete from your side? Looks fine to me.

Yes, it is.

> Also are we sure we have skipped all the right tests?

I did not use your issue as reference but rather the failures which I observed in this PR alone, in the links mentioned [here](https://github.com/openml/openml-python/issues/1657#issuecomment-3952625028) in the comment I left on the issue itself.

> I will also take a look at the skips, match them with reported tests locally and on the issue #1657

I have updated the comment linked above on #1657 but feel free to cross-verify if you want to.

geetu040 (approved):
Great work on this.

Tests that pass on the remote test server but fail on the local test server are currently skipped, and we're tracking those in https://github.com/openml/openml-python/issues/1657

FYI: running the docker servers in CI takes less than 3 minutes and the tests that previously relied on remote-test-server now using the local-test-server are much faster and reliable since the probability for race-conditions is highly reduced.

@fkiraly @PGijsbers please review and merge.

PGijsbers:
I looked at this PR just now and I think we should first investigate the reason for the failure of the tests. Has anyone tried this locally? Because when I try it locally it works (spin up services, point python to `localhost:8000` and run tests). There are a few small exceptions, but the vast majority of marked tests pass. This makes me think it is rather an issue with the deployment on GH CI, perhaps there are issues with write permissions or so.

geetu040:
> Has anyone tried this locally?

@PGijsbers Yes, they fail locally as well.

I have added some details on the failures in https://github.com/openml/openml-python/issues/1657

Quoting from above
> Could it be a problem with the `arff-to-pq` service in [docker-compose](https://github.com/openml/services/blob/main/docker-compose.yaml) of [openml/services](https://github.com/openml/services)? Since the image mentioned `openml-arff-to-pq-converter-cron` is not available as a public frozen docker image on [docker-hub](https://hub.docker.com/u/openml) and therefore the service fallbacks to use the context given in [config/arff-to-pq-converter](https://github.com/openml/services/tree/main/config/arff-to-pq-converter) for normal user (non openml-dev team with access to the private image).
> But this is just my guess, it can be totally irrelevant to the failure.

PR #1630, opened 2026-01-30
[ENH] Allow using a local test server

PGijsbers:
Update the tests to allow connecting to a local test server instead of a remote one (requires https://github.com/openml/services/pull/13).

Running the tests locally:
  - Locally start the services (as defined in https://github.com/openml/services/pull/13) using `docker compose --profile "rest-api" --profile "evaluation-engine" up -d`. Startup can take a few minutes, as currently the PHP container still builds the ES indices from scratch.
    
    I noticed that the `start_period` for some services isn't sufficient on my M1 Mac, possibly due to some containers requiring Rosetta to run, slowing things down. You can recognize this by the services reporting "Error" while the container remains running. To avoid this, you can either increase the `start_period` of the services (mostly elastic search and php api), or you can simply run the command again (the services are then already in healthy state and the services that depended on it can start successfully).
  
    The following containers should run: openml-test-database, openml-php-rest-api, openml-nginx, openml-evaluation-engine, openml-elasticsearch, openml-minio

 - Update the `openml/config.py`'s `TEST_SERVER_URL` variable to `"http://localhost:8000"`.
 - Run the tests (`python -m pytest -m "not production" tests`).

This PR builds off unmerged PR https://github.com/openml/openml-python/pull/1620.

### DIFF
--- a/tests/test_datasets/test_dataset_functions.py
+++ b/tests/test_datasets/test_dataset_functions.py
@@ -107,7 +107,7 @@ class TestOpenMLDataset(TestBase):
         for did in datasets:
             self._check_dataset(datasets[did])
 
-    @pytest.mark.uses_test_server()
+    @pytest.mark.test_server()
     def test_tag_untag_dataset(self):
         tag = "test_tag_%d" % random.randint(1, 1000000)
         all_tags = _tag_entity("data", 1, tag)
@@ -115,12 +115,12 @@ class TestOpenMLDataset(TestBase):
         all_tags = _tag_entity("data", 1, tag, untag=True)
         assert tag not in all_tags
 
-    @pytest.mark.uses_test_server()
+    @pytest.mark.test_server()
     def test_list_datasets_length(self):
         datasets = openml.datasets.list_datasets()
         assert len(datasets) >= 100
 
-    @pytest.mark.uses_test_server()
+    @pytest.mark.test_server()
     def test_list_datasets_paginate(self):
         size = 10
         max = 100
@@ -135,12 +135,12 @@ class TestOpenMLDataset(TestBase):
                 categories=["in_preparation", "active", "deactivated"],
             )
 
-    @pytest.mark.uses_test_server()
+    @pytest.mark.test_server()
     def test_list_datasets_empty(self):
         datasets = openml.datasets.list_datasets(tag="NoOneWouldUseThisTagAnyway")
         assert datasets.empty
 
-    @pytest.mark.production()
+    @pytest.mark.production_server()
     def test_check_datasets_active(self):
         # Have to test on live because there is no deactivated dataset on the test server.
         self.use_production_server()
@@ -159,7 +159,7 @@ class TestOpenMLDataset(TestBase):
         )
         openml.config.server = self.test_server
 
-    @pytest.mark.uses_test_server()
+    @pytest.mark.test_server()
     def test_illegal_character_tag(self):
         dataset = openml.datasets.get_dataset(1)
         tag = "illegal_tag&"
@@ -169,7 +169,7 @@ class TestOpenMLDataset(TestBase):
         except openml.exceptions.OpenMLServerException as e:
             assert e.code == 477
 
-    @pytest.mark.uses_test_server()
+    @pytest.mark.test_server()
     def test_illegal_length_tag(self):
         dataset = openml.datasets.get_dataset(1)
         tag = "a" * 65
@@ -179,7 +179,7 @@ class TestOpenMLDataset(TestBase):
         except openml.exceptions.OpenMLServerException as e:
             assert e.code == 477
 
-    @pytest.mark.production()
+    @pytest.mark.production_server()
     def test__name_to_id_with_deactivated(self):
         """Check that an activated dataset is returned if an earlier deactivated one exists."""
         self.use_production_server()
@@ -187,19 +187,19 @@ class TestOpenMLDataset(TestBase):
         assert openml.datasets.functions._name_to_id("anneal") == 2
         openml.config.server = self.test_server
 
-    @pytest.mark.production()
+    @pytest.mark.production_server()
     def test__name_to_id_with_multiple_active(self):
         """With multiple active datasets, retrieve the least recent active."""
         self.use_production_server()
         assert openml.datasets.functions._name_to_id("iris") == 61
 
-    @pytest.mark.production()
+    @pytest.mark.production_server()
     def test__name_to_id_with_version(self):
         """With multiple active datasets, retrieve the least recent active."""
         self.use_production_server()
         assert openml.datasets.functions._name_to_id("iris", version=3) == 969
 
-    @pytest.mark.production()
+    @pytest.mark.production_server()
     def test__name_to_id_with_multiple_active_error(self):
         """With multiple active datasets, retrieve the least recent active."""
         self.use_production_server()
@@ -211,7 +211,7 @@ class TestOpenMLDataset(TestBase):
             error_if_multiple=True,
         )
 
-    @pytest.mark.uses_test_server()
+    @pytest.mark.test_server()
     def test__name_to_id_name_does_not_exist(self):
         """With multiple active datasets, retrieve the least recent active."""
         self.assertRaisesRegex(
@@ -221,7 +221,7 @@ class TestOpenMLDataset(TestBase):
             dataset_name="does_not_exist",
         )
 
-    @pytest.mark.uses_test_server()
+    @pytest.mark.test_server()
     def test__name_to_id_version_does_not_exist(self):
         """With multiple active datasets, retrieve the least recent active."""
         self.assertRaisesRegex(
@@ -232,7 +232,7 @@ class TestOpenMLDataset(TestBase):
             version=100000,
         )
 
-    @pytest.mark.uses_test_server()
+    @pytest.mark.test_server()
     def test_get_datasets_by_name(self):
         # did 1 and 2 on the test server:
         dids = ["anneal", "kr-vs-kp"]
@@ -240,7 +240,7 @@ class TestOpenMLDataset(TestBase):
         assert len(datasets) == 2
         _assert_datasets_retrieved_successfully([1, 2])
 
-    @pytest.mark.uses_test_server()
+    @pytest.mark.test_server()
     def test_get_datasets_by_mixed(self):
         # did 1 and 2 on the test server:
         dids = ["anneal", 2]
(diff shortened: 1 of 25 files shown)


tests/files/localhost:8080
```diff
@@ -0,0 +1 @@
+/Users/pietergijsbers/repositories/openml-python/tests/files/org/openml/test
```
PGijsbers, line 8080: will make this relative

openml/tasks/functions.py
```diff
@@ -415,9 +415,10 @@ def get_task(
     if not isinstance(task_id, int):
         raise TypeError(f"Task id should be integer, is {type(task_id)}")
 
-    cache_key_dir = openml.utils._create_cache_directory_for_id(TASKS_CACHE_DIR_NAME, task_id)
-    tid_cache_dir = cache_key_dir / str(task_id)
-    tid_cache_dir_existed = tid_cache_dir.exists()
+    task_cache_directory = openml.utils._create_cache_directory_for_id(
+        TASKS_CACHE_DIR_NAME, task_id
+    )
+    task_cache_directory_existed = task_cache_directory.exists()
```
geetu040, line 421: `task_cache_directory` now points to the folder with all the tasks instead of the folder with the specific task ...

openml/tasks/functions.py
```diff
@@ -431,8 +432,8 @@ def get_task(
         if download_splits and isinstance(task, OpenMLSupervisedTask):
             task.download_split()
     except Exception as e:
-        if not tid_cache_dir_existed:
-            openml.utils._remove_cache_dir_for_id(TASKS_CACHE_DIR_NAME, tid_cache_dir)
+        if not task_cache_directory_existed:
+            openml.utils._remove_cache_dir_for_id(TASKS_CACHE_DIR_NAME, task_cache_directory)
```
geetu040, line 436: .. and therefore remove the complete directory, instead of the specific task. is that intentional?

tests/files/localhost_8000
```diff
@@ -0,0 +1 @@
+org/openml/test
```
geetu040, line 1: why this symlink is uploaded? probably a mistake?

or is this used for linking remote-test-server cache with local-test-server cache. if this was intentional, can I ask what's it's purpose and this path can't be tracked on windows because of the `:`

leads to CI failure on windows: [job/61983305389](https://github.com/openml/openml-python/actions/runs/21512661186/job/61983305389?pr=1630)

geetu040 (changes requested):
Thanks for these fixes.
I hope the test-server can be fully replicated locally for the tests now.
Left a few questions, and `tests/files/localhost:8080` file needs to revised.

tests/files/localhost_8000
PGijsbers: > or is this used for linking remote-test-server cache with local-test-server cache.

Yes, it is. There is a cache folder checked in to the repository [here](https://github.com/openml/openml-python/tree/main/tests/files/org/openml/test). Because cache lookup depends on the configured server, they are not found if the server is configured to be `localhost:8000`, this symlink solves that.

I suppose we can update the cache directory resolution to drop the port number (and update the symlink file accordingly), then it should continue to work while windows PCs can still check out the repository.

update: done.

openml/tasks/functions.py
PGijsbers: No, it does not. The `_create_cache_directory_for_id` function, as the name implies, returns the cache directory for the specific identifier, e.g., `[...]/org/openml/test/tasks/1882`

openml/config.py
```diff
@@ -27,6 +27,9 @@
 OPENML_SKIP_PARQUET_ENV_VAR = "OPENML_SKIP_PARQUET"
 _TEST_SERVER_NORMAL_USER_KEY = "normaluser"
 
+TEST_SERVER_URL = "https://test.openml.org"
+TEST_SERVER_URL = "http://localhost:8000"
+
```
geetu040, line 32: unintentional mistake here:
```suggestion
TEST_SERVER_URL = "https://test.openml.org"

```

openml/config.py
```diff
@@ -469,7 +472,9 @@ def get_cache_directory() -> str:
 
     """
     url_suffix = urlparse(server).netloc
-    reversed_url_suffix = os.sep.join(url_suffix.split(".")[::-1])  # noqa: PTH118
+    url_parts = url_suffix.split(".")[::-1]
+    url_parts_no_port = [part.split(":")[0] for part in url_parts]
+    reversed_url_suffix = os.sep.join(url_parts_no_port)  # noqa: PTH118
```
geetu040: this ignores the port completely, as a result apis running on different ports will lead to same path, e.g (php-api on port:8080 and python-api on port:8082) which should not really be the case.
```
~/.cache/openml/localhost
```

tests/files/localhost_8000
geetu040: Thanks for the update.
How safe is the symlink here? I was getting errors like "could not update file in this path", but I'll have a look again.

satvshr:
Recent changes seem to be [breaking my PR](https://github.com/openml/openml-python/actions/runs/21992956358/job/63544915574).

PGijsbers:
A few of the tests now fail with the local output, the PHP API reports that no api key is provided even though there is. Investigating.

openml/config.py
geetu040: resolved

PGijsbers:
Solved with [this commit](github.com/openml/services/pull/13/commits/1a0cf720826b4eb7cc6574cc408114e84465450e). Now all tests should work locally again.

tests/files/localhost_8000
PGijsbers: It should be safe, I am not really sure where those errors come from. What OS are you using? What did you do to set up for the tests?

geetu040: Looks like it coming from the locally running php-test-server, not really related to the path here:

There are a bunch of tests in `tests/test_datasets/` which give this error when publishing in the local services.
```
pytest tests/test_datasets/test_dataset_functions.py::TestOpenMLDataset::test_create_dataset_sparse

FAILED tests/test_datasets/test_dataset_functions.py::TestOpenMLDataset::test_create_dataset_sparse - openml.exceptions.OpenMLServerException: http://localhost:8080/api/v1/xml/data/ returned code 132: Failed to move the files - None
```

Does this test pass for you when using the local php-service?

I am on Ubuntu and I updated the `TEST_SERVER_URL` to use `"http://localhost:8080"` replacing `"https://test.openml.org"`

geetu040 (approved):
We can merge this PR, it works as expected. There is one unresolved comment (https://github.com/openml/openml-python/pull/1630#discussion_r2821385236) but that seems to be unrelated to the changes here.

tests/files/localhost_8000
PGijsbers: Yes, it does work for me locally.

tests/test_runs/test_run_functions.py
```diff
@@ -1813,7 +1813,6 @@ def test_initialize_model_from_run_nonstrict(self):
 
 @mock.patch.object(requests.Session, "delete")
 def test_delete_run_not_owned(mock_delete, test_files_directory, test_api_key):
-    openml.config.start_using_configuration_for_example()
```
mfeurer, line 1816: Why don't we need this anymore?

tests/test_flows/test_flow_functions.py
```diff
@@ -453,7 +453,6 @@ def test_delete_flow(self):
 
 @mock.patch.object(requests.Session, "delete")
 def test_delete_flow_not_owned(mock_delete, test_files_directory, test_api_key):
-    openml.config.start_using_configuration_for_example()
```
mfeurer, line 456: Why don't we need this anymore?

openml/tasks/functions.py
```diff
@@ -415,9 +415,10 @@ def get_task(
     if not isinstance(task_id, int):
         raise TypeError(f"Task id should be integer, is {type(task_id)}")
 
-    cache_key_dir = openml.utils._create_cache_directory_for_id(TASKS_CACHE_DIR_NAME, task_id)
-    tid_cache_dir = cache_key_dir / str(task_id)
-    tid_cache_dir_existed = tid_cache_dir.exists()
+    task_cache_directory = openml.utils._create_cache_directory_for_id(
+        TASKS_CACHE_DIR_NAME, task_id
```
mfeurer, line 419: Do I read this PR correctly in that the previous version was incorrect?

tests/test_datasets/test_dataset_functions.py
```diff
@@ -527,27 +527,20 @@ def test_deletion_of_cache_dir(self):
...
     def test_publish_dataset(self):
-        # lazy loading not possible as we need the arff-file.
-        openml.datasets.get_dataset(3, download_data=True)
-        file_path = os.path.join(
-            openml.config.get_cache_directory(),
-            "datasets",
-            "3",
-            "dataset.arff",
-        )
+        arff_file_path = self.static_cache_dir / "org" / "openml" / "test" / "datasets" / "2" / "dataset.arff"
```
mfeurer, line 535: Why do we change the dataset here?

tests/test_datasets/test_dataset_functions.py
```diff
@@ -1734,7 +1728,6 @@ def test_delete_dataset(self):
 
 @mock.patch.object(requests.Session, "delete")
 def test_delete_dataset_not_owned(mock_delete, test_files_directory, test_api_key):
-    openml.config.start_using_configuration_for_example()
```
mfeurer, line 1737: Why is this removed?

openml/tasks/functions.py
PGijsbers: That's correct.

tests/test_datasets/test_dataset_functions.py
PGijsbers: Because dataset 2 is already in the prepopulated cache and dataset 3 is not. Since the test tries to publish a dataset and does not test the downloading of it, I figure we can make the test more reliable and faster by using a cached file.

tests/test_flows/test_flow_functions.py
PGijsbers: The test configuration is already invoked by the auto-use fixture. It's the default, so we don't need to switch to it.

tests/test_runs/test_run_functions.py
PGijsbers: The test configuration is already invoked by the auto-use fixture. It's the default, so we don't need to switch to it.

tests/test_datasets/test_dataset_functions.py
PGijsbers: The test configuration is already invoked by the auto-use fixture. It's the default, so we don't need to switch to it.

PR #1637, opened 2026-02-16
[MNT] Update ruff and mypy versions

### LINKED ISSUE #1639: [MNT] `pre-commit` updates
We should update the `pre-commit` versions of the repository with the latest versions supported by `pre-commit`. Currently in `pre-commit` we face some problems:
1. [Mypy checks](https://github.com/openml/openml-python/blob/da993f74df36eae7c6f0c08ee0597515df4c7a0a/pyproject.toml#L336) are ignoring the entire tests repository, same with [ruff](https://github.com/openml/openml-python/blob/da993f74df36eae7c6f0c08ee0597515df4c7a0a/pyproject.toml#L152).
2. Upgrading the mypy version from `1.13.0` to the most recent stable version `1.19.1` results in [failures](https://results.pre-commit.ci/run/github/17939040/1771255830.5CDLPfHqR4C4a1n7A4TZIQ) as observed in #1637.

cc: @fkiraly @PGijsbers

satvshr:
fixes #1639.

Only major changes in this PR:
1. `ruff` version update from v0.14.10 to v0.15.1
2. `mypy` version update from v1.13.0 to v1.19.1
   * Had to add a few files to be ignored by `mypy` (instead of using the suggested approach by @fkiraly to only check for files changed and then chip at it brick-by-brick to fix mypy errors, the architecture of those functions would most likely need to be changed, so if someone was adding a simple test they'd have to fix a ton of errors occuring in that file for other functions, which doesnt make sense)
3. Adding a fixed file ending check to prevent unwarranted git diffs
4. Removing trailspaces
5. **Including** the `tests` directory in ruff and mypy checks:
   * Fixed `ruff` wherever possible and added `noqa` tags to other places
   * Had to add a few files to be ignored by `mypy` (for the reason mentioned above in point 2)


geetu040 (changes requested):
I think we should keep skipping the `tests/files` directory. The content there is generated programmatically, and excluding it helps ensure we don't accidentally modify generated artifacts.

Not sure, but after this change it might make sense to adjust `pre-commit-config.yaml` or `pyproject.toml` make the skip rule less restrictive.

Also please sync with `main`, there are some conflicts.

Other than that, the rest of the changes look good to me.

satvshr:
> I think we should keep skipping the `tests/files` directory.

I think Franz was the one who told me to include the `tests/files` directory. We should discuss this today in the meeting.

tests/files/localhost_8000
geetu040, line 1: Is this a manual change or an automatic pre-commit fix? Given the settings, nothing should be touched in `tests/files/`, so I wonder what triggered this.

pyproject.toml
```diff
@@ -328,13 +325,62 @@ check_untyped_defs = true
 
 warn_return_any = true
 
-
 [[tool.mypy.overrides]]
-module = ["tests.*", "openml.extensions.sklearn.*"]
+module = [
```
geetu040, line 330: we should track this in an issue that the following files are being skipped by pre-commit, and should be updated with time.

tests/files/localhost_8000
satvshr: It's a pre-commit file ending fix, it's not being skipped, `addopts` only assured pytrst doesn't touch the directory given it was causing failures.

pyproject.toml
satvshr: I'll open an issue if you'd like me too.

tests/files/localhost_8000
geetu040: but we are skipping pre-commit for `tests/files/` as well, right? then how is this touched?

pyproject.toml
geetu040: yes please that would be nice, also mention how could one reproduce these failures, probably remove this file from the list right?

satvshr: what do you mean by "this" file over here?

tests/files/localhost_8000
satvshr: Can you tell me where do you think this is happening? We are not skipping tests from pre-commit as far as I know?

pyproject.toml
satvshr: Opened an issue [here](https://github.com/openml/openml-python/issues/1696). It could serve as a good first issue for new contributors, epsecially with the next ESoC around.

geetu040: > what do you mean by "this" file over here?

I mean to reproduce these failures for a particular file one would simply remove for-example `tests.test_datasets.test_dataset` from the list `module` in `pyproject.toml`, right?

tests/files/localhost_8000
geetu040: https://github.com/openml/openml-python/pull/1637#pullrequestreview-3859124493

pyproject.toml
satvshr: yes

fkiraly (changes requested):
@PGijsbers should check whether the changes in the precommits are fine with the entire team

.pre-commit-config.yaml
```diff
@@ -1,48 +1,41 @@
...
     hooks:
       - id: check-added-large-files
-        files: ".*"
       - id: check-case-conflict
-        files: ".*"
       - id: check-merge-conflict
-        files: ".*"
       - id: check-yaml
-        files: ".*"
+        exclude: ^mkdocs\.yml$
```
PGijsbers, line 33: Why is it excluded?

PGijsbers (commented):
LGTM; small question though

.pre-commit-config.yaml
satvshr: Removed it earlier because of the errors occurring [here](https://results.pre-commit.ci/run/github/17939040/1773658298.C1Cft0zWSQ-i3nHv9oxSiQ). I thought I should leave it untouched given I did not plan on making any "real" changes with this PR.

PR #1651, opened 2026-02-18
fix: add HTTP request timeouts to prevent hung threads

suhr25:
## Summary

All HTTP traffic in OpenML funnels through `openml/_api_calls.py::_send_request`.
Previously, every `session.get()`, `session.post()`, and `session.delete()` call was executed **without a timeout**, meaning `requests` defaulted to `timeout=None` (infinite wait).

Under slow, stalled, or half-open TCP connections, this caused:

* Threads to block indefinitely in `recv()`
* TCP sockets (file descriptors) to remain open
* Retry logic to never trigger (no exception raised)
* Worker starvation in long-running pipelines

Since `_send_request` is the single HTTP dispatch function used by dataset downloads, task listing, run uploads, and metadata calls, this impacted the entire library surface.

---

### 1. Add Configurable Timeouts

Extended `openml/config.py` with two new fields:

```python
connect_timeout: float
read_timeout: float
```

Defaults:

```python
"connect_timeout": 9.05,
"read_timeout": 300.0,
```

Backwards compatibility is preserved using `.get()` in `_setup()`:

```python
connect_timeout = float(
    config.get("connect_timeout", _defaults["connect_timeout"])
)
read_timeout = float(
    config.get("read_timeout", _defaults["read_timeout"])
)
```

These values are now also exposed via `get_config_as_dict()`.

---

### 2. Apply Timeout to All HTTP Verbs

In `_send_request`, a unified timeout tuple is created:

```python
timeout = (config.connect_timeout, config.read_timeout)
```

And passed to every request:

```python
response = session.get(
    url, params=data, headers=_HEADERS, timeout=timeout
)
```

Same change applied to `delete()` and `post()`.

---

### 3. Ensure Timeouts Trigger Retries

Added `requests.exceptions.Timeout` to the retryable exception block:

```python
except (
    requests.exceptions.Timeout,
    requests.exceptions.ChunkedEncodingError,
    requests.exceptions.ConnectionError,
    requests.exceptions.SSLError,
    ...
):
```

This ensures timeouts integrate with existing retry/backoff logic.

---

### 4. Test Updates

Updated configuration tests to:

* Include the two new timeout fields
* Adjust expected config length
* Validate custom timeout injection via `_setup()`

---

## Verification

* Confirmed all HTTP entry points route through `_send_request`
* Verified timeout tuple is applied to `get`, `post`, and `delete`
* Confirmed backward compatibility with configs missing new keys
* Ensured retry behavior now activates on `Timeout`
* Updated unit tests to reflect expanded config surface

After this change:

* Hung connections terminate predictably
* Threads are released
* Sockets are closed
* Retry logic functions as intended
* Worker processes degrade gracefully under server slowness instead of stalling permanently

This converts a silent, unbounded hang into a controlled, observable failure mode with retry semantics.

### DIFF
--- a/tests/test_openml/test_api_calls.py
+++ b/tests/test_openml/test_api_calls.py
@@ -7,6 +7,7 @@ from unittest import mock
 
 import minio
 import pytest
+import requests
 
 import openml
 from openml.config import ConfigurationForExamples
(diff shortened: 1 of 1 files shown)


suhr25:
Hi @fkiraly , Fixes a critical issue where HTTP calls had no timeout, causing threads to hang forever and sockets to leak under slow or stalled connections.
Adds configurable timeouts and integrates `Timeout` into retry logic so failures are bounded and handled cleanly.

PR #1652, opened 2026-02-18
[MNT] Update setup list test assertion

### LINKED ISSUE #1645: [MNT] Update Setup List Test Assertion to Be More Specific
The test `test_setup_list_filter_flow` in [test_setup_functions.py#L139](https://github.com/openml/openml-python/blob/main/tests/test_setups/test_setup_functions.py#L139) has a placeholder assertion `assert len(setups) > 0`. This was likely added when the test server had no setups, but now it should check for a specific expected count based on real data.

**Steps:**
1. Run the test against the production server (or check the test server data) to see how many setups are returned for flow ID 5873.
2. Replace the assertion with `assert len(setups) == expected_count` (e.g., `assert len(setups) == 5` if that's the actual number).
3. If the count varies, consider making it `assert len(setups) >= expected_minimum` with a comment explaining why.
4. Ensure the test still passes and update any related comments.

The test will be more robust and fail appropriately if the setup count changes unexpectedly, preventing false positives.  

Source: Mentioned TODO in source code (same line)

omkar-334:
Fixes #1645

```
import openml

openml.config.server = "https://www.openml.org/api/v1/xml"
setups = openml.setups.list_setups(flow=5873)
```
Screenshot 2026-02-18 at 7 12 32 PM

Current number of setups is 2. 

> 1. For now, please keep using the production server for this test as the test server is not guaranteed to have setups (yet)
> 2. Make the assertion based on a minimum. It is possible that people upload more runs for the flow (and consequently, more setups). It is also possible they delete them, but not so likely. So I would check a lower bound that is lower than what is currently on the production server.

cc @geetu040 @PGijsbers so should we use the lower bound as 1?  
```
assert len(setups) >= 1
```

### DIFF
--- a/openml/_api_calls.py
+++ b/openml/_api_calls.py
@@ -22,8 +22,8 @@ from urllib3 import ProxyManager
 from . import config
 from .__version__ import __version__
 from .exceptions import (
+    OpenMLAuthenticationError,
     OpenMLHashException,
-    OpenMLNotAuthorizedError,
     OpenMLServerError,
     OpenMLServerException,
     OpenMLServerNoResult,
@@ -515,11 +515,7 @@ def __parse_server_exception(
         400,  # run/42 delete
         460,  # task/42 delete
     ]:
-        msg = (
-            f"The API call {url} requires authentication via an API key.\nPlease configure "
-            "OpenML-Python to use your API as described in this example:"
-            "\nhttps://openml.github.io/openml-python/latest/examples/Basics/introduction_tutorial/#authentication"
-        )
-        return OpenMLNotAuthorizedError(message=msg)
+        msg = f"The API call {url} requires authentication via an API key."
+        return OpenMLAuthenticationError(message=msg)
 
     return OpenMLServerException(code=code, message=full_message, url=url)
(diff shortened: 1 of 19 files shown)


omkar-334:
Screenshot 2026-02-18 at 7 17 57 PM
tests passing

tests/test_setups/test_setup_functions.py
```diff
@@ -143,7 +143,7 @@ def test_setup_list_filter_flow(self):
 
         setups = openml.setups.list_setups(flow=flow_id)
 
-        assert len(setups) > 0  # TODO: please adjust 0
+        assert len(setups) == 2
```
geetu040: As suggested in https://github.com/openml/openml-python/issues/1645#issuecomment-3919532653 and https://github.com/openml/openml-python/pull/1652#discussion_r2822428799
Please update this to check against a lower bound, probably `assert len(setups) >= 2`

omkar-334: will do, thanks

omkar-334: @geetu040 updated this. thanks!

PR #1656, opened 2026-02-18
[MNT] Fix race condition in `OpenMLSplit._from_arff_file`

### LINKED ISSUE #1641: [MNT] Race condition in `OpenMLSplit._from_arff_file` when running tests in parallel

omkar-334:
Fixes #1641

This PR adds separate temp directories per test in test_split.py to avoid race conditions when running with multiple workers. 

cc @geetu040  
I made this PR because I suspect #1643 is made by an OpenClaw bot.

### DIFF
--- a/tests/test_tasks/test_split.py
+++ b/tests/test_tasks/test_split.py
@@ -30,13 +30,19 @@ class OpenMLSplitTest(TestBase):
             / "1882"
             / "datasplits.arff"
         )
-        # Use temporary directory to avoid conflicts with other tests
-        self.temp_dir = tempfile.TemporaryDirectory()
-        self.arff_filepath = Path(self.temp_dir.name) / "datasplits.arff"
+        # Use a unique temp directory for each test to avoid race conditions
+        # when running tests in parallel (see issue #1641)
+        self._temp_dir = tempfile.TemporaryDirectory()
+        self.arff_filepath = Path(self._temp_dir.name) / "datasplits.arff"
         shutil.copy(source_arff, self.arff_filepath)
+        self.pd_filename = self.arff_filepath.with_suffix(".pkl.py3")
 
     def tearDown(self):
-        self.temp_dir.cleanup()
+        # Clean up the entire temp directory
+        try:
+            self._temp_dir.cleanup()
+        except (OSError, FileNotFoundError):
+            pass
 
     def test_eq(self):
         split = OpenMLSplit._from_arff_file(self.arff_filepath)
--- a/openml/_api_calls.py
+++ b/openml/_api_calls.py
@@ -22,8 +22,8 @@ from urllib3 import ProxyManager
 from . import config
 from .__version__ import __version__
 from .exceptions import (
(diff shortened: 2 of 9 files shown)


omkar-334:
Screenshot 2026-02-19 at 12 53 02 AM
All tests pass

omkar-334:
> May I suggest to pick up more challenging issues next instead of eating up all the good first issues?

Sure, I’ll look into more challenging issues next.

fkiraly:
it looks like the bot in https://github.com/openml/openml-python/pull/1643 is not reacting to the review - would you be able to action the review request, and then we merge this PR instead?

omkar-334:
> it looks like the bot in #1643 is not reacting to the review - would you be able to action the review request, and then we merge this PR instead?

Yes, I'll review your comments in that PR and make the changes.

tests/test_tasks/test_split.py
```diff
@@ -29,13 +30,15 @@ def setUp(self):
             / "1882"
             / "datasplits.arff"
         )
-        self.pd_filename = self.arff_filepath.with_suffix(".pkl.py3")
+        # Use temporary directory to avoid conflicts with other tests
+        self.temp_dir = Path(tempfile.mkdtemp())
```
geetu040: copying the comment from https://github.com/openml/openml-python/pull/1643#discussion_r2826021251

`mkdtemp` is fine, but I'd prefer `TemporaryDirectory` since it's safer and reduces cleanup code.

Then in `tearDown` you can simply call `self.temp_dir.cleanup()`

tests/test_tasks/test_split.py
```diff
@@ -29,13 +30,15 @@ def setUp(self):
...
+        self.arff_filepath = self.temp_dir / "datasplits.arff"
+        shutil.copy2(source_arff, self.arff_filepath)
```
geetu040: `shutil.copy` should be enough

omkar-334: resolved

tests/test_tasks/test_split.py
omkar-334: resolved

tests/test_tasks/test_split.py
```diff
@@ -29,14 +30,14 @@ def setUp(self):
...
+        shutil.copy(source_arff, self.arff_filepath)
 
     def tearDown(self):
-        try:
-            os.remove(self.pd_filename)
-        except (OSError, FileNotFoundError):
-            #  Replaced bare except. Not sure why these exceptions are acceptable.
-            pass
+        if hasattr(self, "temp_dir"):
+            self.temp_dir.cleanup()
```
geetu040: why do we need to check this?
```suggestion
        self.temp_dir.cleanup()
```

omkar-334: hey @geetu040, missed this. sorry. I've resolved it now

PR #1676, opened 2026-02-20
[ENH] OpenmMLBenchmarkRunner class

EmanAbdelhaleem:
This is a draft PR for creating a `OpenmMLBenchmarkRunner` class that handles running benchmarks in parallel instead of using for loops

### DIFF
--- a/openml/benchmarks/__init__.py
+++ b/openml/benchmarks/__init__.py
@@ -1,0 +1,1 @@
+# License: BSD 3-Clause
--- a/openml/benchmarks/benchmark_runner.py
+++ b/openml/benchmarks/benchmark_runner.py
@@ -1,0 +1,128 @@
+# License: BSD 3-Clause
+from __future__ import annotations
+
+from concurrent.futures import Future, ThreadPoolExecutor, as_completed
+from typing import TYPE_CHECKING, Any
+
+import sklearn.metrics
+from tqdm import tqdm
+
+import openml
+
+if TYPE_CHECKING:
+    from openml.runs import OpenMLRun
+    from openml.study import OpenMLBenchmarkSuite
+    from openml.tasks import OpenMLTask
+
+
+class OpenMLBenchmarkRunner:
+    """Run an estimator against every task in an OpenML benchmark suite.
+
+    Parameters
+    ----------
+    benchmark_suite : OpenMLBenchmarkSuite | str | int
+        Suite object, alias (e.g. "OpenML-CC18"), or integer suite ID.
+    estimator : sklearn-compatible estimator
+    n_jobs : int, optional (default=1)
+        Number of parallel workers. -1 uses one thread per task.
+    upload_runs : bool, optional (default=True)
+        Whether to publish runs to the OpenML server.
+    """
+
+    def __init__(
+        self,
+        benchmark_suite: OpenMLBenchmarkSuite | str | int,
+        estimator: Any,
+        *,
+        n_jobs: int = 1,
+        upload_runs: bool = True,
+    ) -> None:
+        if isinstance(benchmark_suite, (str, int)):
+            benchmark_suite = openml.study.get_suite(benchmark_suite)
+
+        self.benchmark_suite: OpenMLBenchmarkSuite = benchmark_suite
+        self.estimator: Any = estimator
+        self.n_jobs: int = n_jobs
+        self.upload_runs: bool = upload_runs
+        self.results: dict[int, dict[str, Any]] = {}
+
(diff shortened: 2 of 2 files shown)


EmanAbdelhaleem:
@jgyasu @fkiraly I need a review here. 
I also have 2 questions regarding the pause and resume methods:
1. Should the pause and resume methods get called when the user presses some keyboard key like `P` for pause and `R` for resume? or do you have a different workflow in mind?
2. If the user has paused while some task was in progress, should the task end first before pausing or should the task get interrupted and deprecated?

jgyasu:
I haven't reviewed the code yet but I will answer your question for now,
> Should the pause and resume methods get called when the user presses some keyboard key like P for pause and R for resume? or do you have a different workflow in mind?

The goal is to make the benchmark runner robust to crashes or estimator failures. For example, if the 4th estimator fails during execution, the entire benchmark run currently breaks and previously computed results may not be safely stored (correct me if I am wrong?). Instead, the runner should persist results incrementally after each successful run locally.

This would allow us to:
* Continuously save results as they are produced.
* Avoid losing progress if a failure occurs.
* Restart the benchmark later and automatically skip already completed runs.
* Execute only the remaining combinations.

So, this is more about checkpointing and resumability than about manual pause/resume. Of course you can then think that the user can pause the experiment with an keyboard interrupt too, and in that case the checkpointing and resumability would work too. @fkiraly do you want to add anything?

EmanAbdelhaleem:
> this is more about checkpointing

As far as I know, we deal with classical models currently, right? so, wouldn't checkpointing be much?

Also, what do you exactly mean by storing the results? Is it saving the output that is printed on the terminal or something else?
As far as I know, the current behaviour of benchmarking is as follows:
1. Select a benchmarking Suite
2. Iterate over the tasks
3. Run model on task
4. Store the results by publishing the runs (if you want)

So, storing is done by publishing, and you can publish each single run when its finished, you don't need to wait till the experiment is done. The problem with looping is that if some run failed, all the benchmark experiment is gonna crash, I solved this by using threading instead of looping.

So, I am not sure what do you exactly mean by `save results as they are produced`. You can already do this by publishing the run, right?

jgyasu:
> As far as I know, we deal with classical models currently, right? so, wouldn't checkpointing be much?

What checkpointing has to do with classical models? I did not get you

> So, I am not sure what do you exactly mean by save results as they are produced. You can already do this by publishing the run, right?

What if there is an error in publishing? What happens when the internet connection breaks?

> Also, what do you exactly mean by storing the results? Is it saving the output that is printed on the terminal or something else?

Storing the result of the benchmark runs in local disk, maybe in JSON or CSV format (design question). And then the results should be loadable in-memory as dataframes.

EmanAbdelhaleem:
> What checkpointing has to do with classical models? I did not get you

As far as I know, checkpointing is about saving the model's sate. This is necessary with deep learning as it usually takes time, but with classical models. A single run usually doesn't take much time, So, I think checkpointing might be much. Let me know what you think.

> What if there is an error in publishing? What happens when the internet connection breaks?

I think it is not gonna get published then? You still can save the results of the run locally even with the model pickled if you use the method `run.to_filesystem(path)`. However, it's not saved in a dataframe, it's serialized to be uploaded later using `run.from_filesystem(path)` that can instantiate an `OpenMLRun` object based on the saved files.

I think what I can automate here is the `path` passed to these methods. What do you think?

jgyasu:
> As far as I know, checkpointing is about saving the model's sate.

I think you are misunderstanding the term "checkpointing" here. We are not storing the model's weights, we are checkpointing the results after each run.

> You still can save the results of the run locally even with the model pickled if you use the method run.to_filesystem(path)

Really? I can't find `run.to_filesystem(path)` function, can you share the module you found it in?

EmanAbdelhaleem:
> Really? I can't find `run.to_filesystem(path)` function, can you share the module you found it in?


They are methods in the `OpenMLRun` class, check `runs/run.py`.
They also exist in other several classes, for example, they also exist in `OpenMLRunTrace`.

btw, you have mentioned in a previous mentoring meeting if I could add parallelization for the folds or the different hyperparameters used, when checking I found that it's already implemented in the current code, check `_run_task_get_arffcontent` at `runs/functions.py`.

So, the current code has implementation for saving the results locally already, but it will need automation in defining the 'path' to save to. However, this save implementation is for the whole run, not for each trace (a single hyperparameter combination).
The current code also has parallelization for the different hyperparameter settings. 

What I added in this PR is parallelization for the tasks in the benchmark suite(whole runs).

PR #1688, opened 2026-02-23
[MNT] Remove temporary dummy jobs from CI workflow

### LINKED ISSUE #1573: [MNT] remove no longer existing jobs from required list in CI
There are a few jobs in openml that should be removed from being required, since they are no longer valid:

* jobs that no longer exist or run: `docker`, `(windows-latest, Py, sk0.24.*, sk-only:false)`
* jobs that relate to long-deprecated python versions (3.8 and 3.9)

Without this, not even maintainers can merge pull requests.

@PGijsbers, @joaquinvanschoren, would you kindly be able to fix, this is blocking us from using CI for testing.

To fix this, an admin or maintainer should go (on GitHub GUI) to:

Settings -> branches (left menu) -> `main`, edit -> scroll down to "Require status checks to pass before merging" and click the "X" next to the above.

While you are there, you could also go to the "general" pane and ensure checkmarks are like this:

Image

Otherwise, PR may get merged as individual commits and mess up the `main` commit history. Squashing is an easy way to avoid conflicts and rebase accidents.

moksha-hub:
- Removed 3 temporary dummy jobs from `.github/workflows/test.yml`:
  - `dummy_windows_py_sk024`
  - `dummy_windows_py_sk023`
  - `dummy_docker`

These jobs were added in PR #1572 as a temporary workaround when branch protection rules required jobs that no longer existed. They are no longer needed and should be removed.

Fixes #1573

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -165,30 +165,3 @@ jobs:
         token: ${{ secrets.CODECOV_TOKEN }}
         fail_ci_if_error: true
         verbose: true
-
-  dummy_windows_py_sk024:
-    name: (windows-latest, Py, sk0.24.*, sk-only:false)
-    runs-on: ubuntu-latest
-    steps:
-      - name: Dummy step
-        run: |
-          echo "This is a temporary dummy job."
-          echo "Always succeeds."
-
-  dummy_windows_py_sk023:
-    name: (ubuntu-latest, Py3.8, sk0.23.1, sk-only:false)
-    runs-on: ubuntu-latest
-    steps:
-      - name: Dummy step
-        run: |
-          echo "This is a temporary dummy job."
-          echo "Always succeeds."
-
-  dummy_docker:
-    name: docker
-    runs-on: ubuntu-latest
-    steps:
-      - name: Dummy step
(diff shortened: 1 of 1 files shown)


moksha-hub:
@geetu040 please have a look.

geetu040 (approved):
I think this is fine. We should remove these jobs. But I will leave this for @fkiraly, since he created the dummy jobs and ask if the reason for creating these in the first place is not relevant anymore.

I am still seeing these jobs in the CI with the message "Expected — Waiting for status to be reported". I think these should be removed if approved to run.

@fkiraly please approve the workflows and merge if it looks fine given the above context.

fkiraly (changes requested):
The dummy jobs are still required, because the admin console still has them as required jobs.

@PGijsbers, could you remove these from the admin console first? Otherwise tests will not be able to pass (these are dependent on ancient versions).

PGijsbers:
Removed

PGijsbers (approved):

geetu040 (changes requested):
@moksha-hub please resolve conflicts with main.

PR #1692, opened 2026-02-26
[MNT] Diagnose and address long test runtimes (#1633)

### LINKED ISSUE #1633: [MNT] diagnose and address long test runtimes
Current test runtimes are in the order of 1 hour.

We should diagnose the biggest time sinks and think about strategies to mitigate them, while optimally not reducing coverage.

Abhishek9639:
# [MNT] diagnose and address long test runtimes. Closes #1633

## Changes

Current CI test runs take 1–2+ hours. This PR diagnoses the bottleneck and implements several improvements:

### Root Cause
The `production_server` tests (74 tests) make live API calls to openml.org, taking ~1h 23m in CI even with 4-worker parallelization.

### Improvements

1. **Global per-test timeout** (`pyproject.toml`)
   - Added `timeout = 600` (10 min) to `[tool.pytest.ini_options]`
   - Prevents any single test from hanging indefinitely

2. **CI workflow improvements** (`.github/workflows/test.yml`)
   - Changed `--durations=20` → `--durations=0` to report ALL test durations for diagnosis
   - Added explicit `--timeout=600` to all 3 pytest invocations

3. **Fixture optimization** (`tests/conftest.py`)
   - Changed `verify_cache_state` fixture scope from `function` → `module`
   - Reduces redundant filesystem I/O (was running before/after EVERY test)

4. **Benchmark script** (`scripts/profile_tests.sh`)
   - New script for easy local test duration profiling
   - Configurable marker filters

### Test Distribution Analysis
| Category | Count | CI Time |
|---|---|---|
| All tests | 368 | |
| `production_server` | 74 | ~1h 23m (**bottleneck**) |
| `test_server` | 196 | excluded from CI |
| `sklearn`-only | 6 | ~1 min |
| Non-server | 99 | fast |

### Verification
- All pre-commit checks pass (ruff, ruff-format, mypy)
- All 368 tests still collect correctly


Abhishek9639:
Hii @geetu040 and @fkiraly,
Fixed the code quality checks. All pre-commit checks are now passing.
Please review it.

.github/workflows/test.yml
```diff
@@ -119,7 +119,7 @@ jobs:
           marks="not production_server and not test_server"
         fi
 
-        pytest -n 4 --durations=20 --dist load -sv $codecov -o log_cli=true -m "$marks"
+        pytest -n 4 --durations=0 --timeout=600 --dist load -sv $codecov -o log_cli=true -m "$marks"
```
geetu040: `durations`: should not be set to 0, we could possibly use this in CI
`timeout`: we need to think about this, I would try to avoid setting this explicitly, but let's discuss. what's your reasoning to set a `timeout` here?

scripts/profile_tests.sh
geetu040, line 1: this file looks fine and we probably need this
duration, timeout and output file path should take value from users like markers and should have default value

tests/conftest.py
```diff
@@ -254,7 +255,7 @@ def test_api_key() -> str:
     return TestBase.user_key
 
 
-@pytest.fixture(autouse=True, scope="function")
+@pytest.fixture(autouse=True, scope="module")
```
geetu040: why did you have to change this?

geetu040 (changes requested):
Thanks, `scripts/profile_tests.sh` file will be used, but I not sure about other `duration` and `timeout` related changes. See comments below.

.github/workflows/test.yml
Abhishek9639: Makes sense, I've reverted both changes durations is back to 20 and removed timeout from CI.

scripts/profile_tests.sh
Abhishek9639: Update done The script now supports m (marker), d (durations), t (timeout), and o (output file) as CLI arguments, each with sensible default values. Let me know if you'd like any tweaks.

tests/conftest.py
Abhishek9639: You're right, this change wasn't necessary. I've reverted it back to scope=\"function\". Thanks for pointing it out

geetu040:
you should mention the issue https://github.com/openml/openml-python/issues/1633 without the keyword `Fixes #1633` since it doesn't close it, rather adds script to help debug this.

Abhishek9639:
@geetu040,
I’ve made all the changes you suggested. Could you please review it once?
And if any further changes are needed, please let me know.
Thanks

scripts/profile_tests.sh
```diff
@@ -0,0 +1,51 @@
...
+echo "Timeout per test: ${TIMEOUT}s"
+echo "Output file: $OUTPUT_FILE"
+echo ""
+
+pytest \
+  --durations="$DURATIONS" \
+  --timeout="$TIMEOUT" \
+  -q \
+  -m "$MARKER_FILTER" \
+  2>&1 | tee "$OUTPUT_FILE"
```
geetu040: I would also prefer additional argument `-n` for more workers and remove `-q` to get full pytest output
```diff
pytest \
+ --dist=load \
+ -n=$NUM_WORKERS \
  --durations="$DURATIONS" \
  --timeout="$TIMEOUT" \
- -q \
  -m "$MARKER_FILTER" \
  2>&1 | tee "$OUTPUT_FILE"
```
This would mimic the exact pytest command in CI

geetu040 (changes requested):
see the comment below

scripts/profile_tests.sh
Abhishek9639: Thanks for pointing out the changes.
I’ll make the changes.

Abhishek9639:
@geetu040,
Updated Added -n for workers with --dist=load and removed -q for full output the script now mimics the exact CI pytest command. Please review
If any further changes are needed, please let me know.
Thanks

geetu040 (approved):
LGTM

@fkiraly please merge.

PGijsbers:
Is this ready for review or not?:) I was pinged for this but I see there are still commits being pushed

Abhishek9639:
Hi @PGijsbers,
Yes, it's ready for review now! Sorry about the confusion the force-push was just to clean up the commit history. 
No more changes planned.
Thanks

scripts/profile_tests.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
```
PGijsbers: Can you make an argument for why we need this file?
To me, I think it would make more sense to expand the testing documentation with an example invocation to help people diagnose long runtimes.

This just seems to make the setup _less_ flexible (e.g., it might be sensible to only run tests of specific classes or files, or use other arguments like --setup-only to investigate setup times, each of these cases would require editing this file or just using a `pytest` invocation anyway. It's also brittle to changes in pytest (which to be fair is to me is not so much of a concern, the interface is stable).

tests/conftest.py
```diff
@@ -106,7 +106,8 @@ def delete_remote_files(tracker, flow_names) -> None:
     if "flow" in tracker:
```
PGijsbers, line 106: Unless I am missing something, it looks like there are only formatting changes in this file?

PGijsbers (commented):
I am a bit confused as to the purpose of this PR. I would reject but I assume I am missing some context here.

Abhishek9639:
Hi @PGijsbers, 
Thanks for the review and sorry for the back-and-forth earlier. I've addressed both of your comments:

- **`scripts/profile_tests.sh`**: removed. You're right that a fixed wrapper script makes things less flexible than just invoking `pytest` directly (e.g. `--setup-only`, restricting to a class/file, or any other pytest argument all need either editing the script or bypassing it anyway).
- - **`tests/conftest.py`**: reverted. As you noted, the changes there were purely formatting noise and didn't belong in this PR. The file is now identical to `main`.
In place of the script, I've added a small **"Diagnosing Slow Tests"** subsection to `CONTRIBUTING.md` with example `pytest` invocations (`--durations`, `--timeout`, `--setup-only`, marker filters, and an `xdist` parallel run) so contributors can mix and match the arguments they actually need.

The PR now only touches `CONTRIBUTING.md`. Please take another look when you get a chance.

scripts/profile_tests.sh
Abhishek9639: Done!

tests/conftest.py
Abhishek9639: Done!

geetu040:
> I am a bit confused as to the purpose of this PR. I would reject but I assume I am missing some context here.

@PGijsbers I don't think there is much context here. It is as you see it. There was no definite issue or requirement for this PR. It was a utility work to log long test runtimes. I actually relied on this script and seeing the PR actually completed and though that it could be useful for others as well.

Now it's changed quite a bit. It has moved from utility script to a part in contributing guide.
I would leave it up to you, if you think it would make sense and be useful for developers in future. Otherwise I don't have a strong preference on this.

PGijsbers (approved):
Thanks, I think it's more useful this way :)

PR #1694, opened 2026-03-01
[MNT] Fix race condition in OpenMLSplit tests during parallel execution #1641

### LINKED ISSUE #1641: [MNT] Race condition in `OpenMLSplit._from_arff_file` when running tests in parallel

codervinitjangir:
Metadata

Reference Issue: Fixes #1641

New Tests Added: No (Fixes existing tests)

Documentation Updated: No

Change Log Entry: Fix race condition in OpenMLSplit tests by using unique temporary directories for parallel execution.

Details
What does this PR implement/fix?
This PR fixes a race condition in OpenMLSplitTest that occurs when tests are run in parallel (e.g., using pytest -n). I updated the setUp method to create a unique temporary directory for each test run using tempfile.mkdtemp() and ensured that the input ARFF file is copied into this isolated path. I also added proper cleanup in tearDown using shutil.rmtree().

Why is this change necessary?
Previously, parallel workers were sharing the same file path for the generated .pkl.py3 cache file. This resulted in intermittent EOFError because one worker would delete the file during tearDown while another worker was still trying to read it.

How can I reproduce the issue this PR is solving and its solution?
The issue can be reproduced by running the following command multiple times:
Without this fix, the tests fail intermittently (roughly 1 in 10 runs). With this fix, the tests pass consistently in parallel environments.

Any other comments?
Following the maintainer's suggestion in the issue thread, this approach avoids shared paths entirely by giving each worker its own tempfile directory.


codervinitjangir:
Hi @geetu040 , I have submitted a PR to resolve this race condition using the unique tempfile directory approach we discussed. The CI workflows are currently awaiting approval to run. Could you please trigger them when you have a moment? Thanks!

geetu040 (changes requested):
This issue has already been fixed, but we can still use your PR to further clean this file. Can you please sync with `main` and apply your changes?

codervinitjangir:
Hi @geetu040 , thanks for the update! I'm glad the core issue is resolved. I will sync my branch with main, apply the cleanup changes using the tempfile approach, and update this PR shortly.

codervinitjangir:
Hi @geetu040, I accidentally closed the PR! Reopened it now. I have synced the branch with main and applied the tempfile cleanup changes as requested. Let me know if everything looks good!

tests/test_tasks/test_split.py
```diff
@@ -18,9 +20,12 @@ class OpenMLSplitTest(TestBase):
     # than 5 seconds + rebuilding the test would potentially be costly
 
     def setUp(self):
```
geetu040, line 20: don't change anything here, keep it as it was

tests/test_tasks/test_split.py
```diff
@@ -30,23 +35,21 @@ def setUp(self):
...
     def tearDown(self):
-        # Clean up the entire temp directory
         try:
-            self._temp_dir.cleanup()
-        except (OSError, FileNotFoundError):
+            shutil.rmtree(self.test_dir)
+        except OSError:
+            #  Replaced bare except. Not sure why these exceptions are acceptable.
             pass
+        super().tearDown()
```
geetu040:
```suggestion
    def tearDown(self):
        self._temp_dir.cleanup()
```

tests/test_tasks/test_split.py
```diff
@@ -30,23 +35,21 @@ def setUp(self):
...
 
     def test_eq(self):
         split = OpenMLSplit._from_arff_file(self.arff_filepath)
-        assert split == split
+        assert split == split  # noqa: PLR0124
```
geetu040: this line can be removed
```suggestion
```

geetu040 (changes requested):

tests/test_tasks/test_split.py
geetu040: `tearDown` can simply be written as:
```py
    def tearDown(self):
        self._temp_dir.cleanup()
```

codervinitjangir:
Hi @geetu040 , I have applied all the cleanup changes exactly as suggested. The setUp method is back to its original state, tearDown is updated to use self._temp_dir.cleanup(), and the redundant assert has been removed. Let me know if it's good to go!

tests/test_tasks/test_split.py
geetu040: https://github.com/openml/openml-python/pull/1694#discussion_r2888977053 is still unresolved

tests/test_tasks/test_split.py
```diff
@@ -30,23 +33,13 @@ def setUp(self):
...
-        # Clean up the entire temp directory
-        try:
-            self._temp_dir.cleanup()
-        except (OSError, FileNotFoundError):
-            pass
+        self._temp_dir.cleanup()
 
     def test_eq(self):
         split = OpenMLSplit._from_arff_file(self.arff_filepath)
-        assert split == split
```
geetu040, line 49: sorry about that, I think this should be kept, since it tests eq

geetu040 (changes requested):

geetu040 (changes requested):
You haven’t reverted the changes in the `setUp` function, it should remain as it was.

The previous review comment about this is still unresolved. Please address it and then request another review once it's fixed.

codervinitjangir:
Hi @geetu040 , I finally realized my local main branch was out of sync with the upstream repository, which caused those extra modified lines in setUp and test_eq. I have now fetched directly from upstream, synced everything perfectly, and the ONLY change in this PR is the tearDown simplification exactly as you requested. Thanks for your patience!

geetu040 (approved):
@fkiraly There were 3 PRs fixing https://github.com/openml/openml-python/issues/1641
You merged both https://github.com/openml/openml-python/pull/1643 and https://github.com/openml/openml-python/pull/1656, one to close the issue and other to credit the other author.
One review was not addressed on the previous PRs when merging, which has been addressed in this PR.
So, please approve the workflows and merge.

PR #1740, opened 2026-09-17
TEST Add sparse dataset row ID and ignore coverage

### LINKED ISSUE #1644: [MNT] Complete Test for Sparse Dataset Row ID and Ignore Attributes Inclusion
The test `test_get_sparse_dataset_rowid_and_ignore_and_target` in [test_dataset.py#L382](https://github.com/openml/openml-python/blob/main/tests/test_datasets/test_dataset.py#L382) currently sets `ignore_attribute` and `row_id_attribute` on a sparse dataset but calls `get_data()` with `include_row_id=False` and `include_ignore_attribute=False`. This means the test doesn't verify behavior when these attributes are actually included in the output.

**Steps:**
1. Add a new test case or extend the existing one to call `get_data(target="class", include_row_id=True, include_ignore_attribute=True)`.
2. Assert the correct output shape (should exclude the ignored columns and include row IDs if applicable).
3. Verify that categorical indicators are updated accordingly.
4. Run the test to ensure it passes with the sparse dataset.

The test should validate that sparse datasets handle row ID and ignore attributes correctly when included, improving test coverage for edge cases.

Source: Mentioned TODO in source code (same line)

ChiruChirag007:
## Description

This PR completes the sparse dataset test for row ID and ignore attribute inclusion.

The existing test only covered the case where row ID and ignore attributes were excluded. This change adds coverage for retrieving the dataset with both `include_row_id=True` and `include_ignore_attribute=True`.

## Changes

- Added a test case for including row ID and ignore attributes.
- Verified the expected feature shape `(600, 20000)`.
- Verified the categorical indicator length.
- Verified that all categorical indicators are `False`.
- Verified the target shape `(600,)`.

## Testing

The following test passes:

`pytest tests/test_datasets/test_dataset.py::OpenMLDatasetTestSparse::test_get_sparse_dataset_rowid_and_ignore_and_target`

Pre-commit checks also pass.

Closes #1644


ChiruChirag007:
Hi maintainers,
I have completed the test coverage requested in issue #1644.
The test now covers the case where both `include_row_id=True` and `include_ignore_attribute=True`.
The targeted test passes locally, and all local pre-commit checks pass.
Thank you for reviewing the contribution!

